PR #689, opened 2020-11-17
feat: Update favorite chips to use Button

cbrevik:
Also fix issues with FavoriteIcons fill color, and allow Button to control icon fill (shouldn't use ThemeIcon in that case).

Added FadeBetween to fade between two children. Also removed explicit marginBottom on Buttons.

### DIFF
--- a/src/components/search-button/index.tsx
+++ b/src/components/search-button/index.tsx
@@ -76,8 +76,7 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
-    paddingHorizontal: theme.spacings.medium,
-    paddingVertical: theme.spacings.small,
+    padding: theme.spacings.medium,
   },
   buttonContainer: {
     flexDirection: 'row',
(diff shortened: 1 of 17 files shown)


src/components/search-button/index.tsx
```diff
@@ -76,8 +76,7 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
-    paddingHorizontal: theme.spacings.medium,
-    paddingVertical: theme.spacings.small,
+    padding: theme.spacings.medium,
```
rikkekantega, line 79: Just commenting on this: Changed to minHeight on the other branch, but both increased padding and defined minHeight might as well live together :)

rikkekantega (commented):
image
The two rows of favoritechips should be provided some spacing between :thinking:

rikkekantega:
image

Favorites are now being capped on the right end side, not sure if intentional

cbrevik:
Oops, will fix both issues! My last commit messed it up a bit

cbrevik:
Should be fixed 👍 @rikkekantega Good catch 😅

rikkekantega (approved):
I love this! :grinning:  LGTM+ :star:

Issue #487, opened 2020-09-30
[Feature] Increase element size and/or spacing on journey search


hildeor:
## Origin

[Usability test with Blindeforbundet, 23.09.2020, finding 1.](https://github.com/AtB-AS/innsikt/blob/master/brukertester/2020-09-23_Blindeforbundet.md#1)

> Det er for trangt nederst i app mellom “Fra” “Til” “Favoritter” og Tabs. Vanskelig å treffe ønsket element.

[Usability test with Blindeforbundet, 29.10.2020.](https://github.com/AtB-AS/innsikt/blob/master/brukertester/2020-10-29_Blindeforbundet.md#15)

> Knapp for å bytte innhold mellom fra/til-felt bør komme etter til-feltet

## Motivation
_A short description of what user needs or business goals this feature will solve._

## Hypotheses and assumptions
_A list of hypotheses and assumptions we have made about the user or the proposed solution._

## Proposed solution
_A coarse description of a proposed solution, that may include wireframes and graphic design._

- Adjusting spacing, sizing, margins on the input elements to increase the touch area according to design in Figma
- Move the switch direction button according to design i Figma
Figma: https://www.figma.com/file/P624oig7YgUQIbz7uFwdHw/App-0.9?node-id=495%3A0

![Front --No favorites](https://user-images.githubusercontent.com/8980523/97967198-2e6f4280-1dbd-11eb-96bd-960ae84eb68f.png)
NB: Check Figma to see latest.

PR #684, opened 2020-11-16
feat: Adds update profile screen

mikaelbr:
I was so inspired by @cbrevik's PR that I wanted to make my own with bloated scope 😬 :trollface:  This mainly rewrites Profile screen. But with that, a few new components were introduced from the design system. Mainly item lists and input groups (from design system in Figma). These are general components that can be used anywhere. So I rewrote some of the different places to test them out.

After some discussions with @cbrevik and some soul searching, I think I'll restructure the item group and list group concept and smash them together as one big interoperable sections with different standardized children. But as this PR is large enough in scope already, I'll make a separate PR and we can discuss some of the architectural decisions there.

No longer marked as a draft, but should not be merged until all items on the todo list below is checked.

Note: There are some places there are still hardcoded. They are marked as TODO in the code. We don't _need_ to fix them now but would be nice to be able to.

## TODO

- [x] Accessibility check for the profile.
- [x] Check for regressions of accessibility on assistant view.
- [x] Check various sizes

## Features

- Adds item groups components
- Adds "full size" header with proper padding for color behind the status bar.
- Rewrites profile screen and flow.
- Adds the ability to reorder favorites.

## Other changes
- Reduced header height (was too large padding now)
- Changes "faded" color to gray 500 (seemed more in line with the design) 
- Adds the ability to reorder favorites
- Changes add/edit favorite screen to be a modal.
- Introduces input group components (NOTE: Subject to change structure in separate PR)
- Rewrites Assistant inputs to new input group types (same as in Profile)

## Out of scope

- Rewrite input group for departures view, or location search modal.

Screenshot 2020-11-17 at 15 46 29
Screenshot 2020-11-17 at 15 46 27
image
image

image

PR #691, opened 2020-11-18
fix: Increase journey search input size

rikkekantega:
#487

Adjusted input field size and did other changes according to design (move swap button, border color change, font size etc)

Changes on margins for timepicker field (height on Fade) not included, as Christian is currently working on the Fade component.

image

### DIFF
--- a/src/components/search-button/index.tsx
+++ b/src/components/search-button/index.tsx
@@ -78,7 +78,7 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
     alignItems: 'center',
     paddingHorizontal: theme.spacings.medium,
     paddingVertical: theme.spacings.small,
-    height: theme.sizes.touchable,
+    minHeight: theme.sizes.touchable,
   },
   buttonContainer: {
     flexDirection: 'row',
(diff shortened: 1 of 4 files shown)


cbrevik:
Should this be solved with explicit height? Ref. when font sizes scale? Might `paddingVertical` not the same, while allowing to grow?

rikkekantega:
> Should this be solved with explicit height? Ref. when font sizes scale? Might `paddingVertical` not solve the same, while allowing to grow?

Of course, you are entirely right about that. Perhaps just use minHeight instead, since we are defining this (44px) as a "minimum" touchable size. Don't want it to shrink and be harder to hit if font size is decreased.

mikaelbr:
I think I might have accidentally touched this in #684

rikkekantega:
> I think I might have accidentally touched this in #684

Have you solved the #487 issue entirely, or just that I'll need to consider your changes? :)

mikaelbr:
It's equal to the height of the design in Figma which is 44. So I think so

rikkekantega:
> It's equal to the height of the design in Figma which is 44. So I think so

Oh, shoot that I didn't know you were working on these things as well as profile. I pulled your branch now and see the new design is implemented there, I'll just close this PR. :)

mikaelbr:
Sorry, my fault. Didn't intend to, nor is it strictly in the scope. I just started implementing the general components as a part of the new input form for adding favorites and saw that they were generic. One thing lead to another..

PR #695, opened 2020-11-19
fix: FadeBetween height requirement, listen to parent (like it should)

cbrevik:


src/screens/Assistant/FadeBetween.tsx
```diff
@@ -21,16 +21,19 @@ type FadeProps = {
   style?: ViewStyle;
   children: [ReactElement, ReactElement];
   duration?: number;
+  isParentAnimating?: boolean;
```
mikaelbr, line 24: Should this just not be optional? Or at the very least defaulted to `false`

mikaelbr (approved):
Nice! Haven't checked it out locally bug figure you have tested it. I don't quite follow how the animation should be relative with the different isanimating/isvisible permutations, though.

src/screens/Assistant/FadeBetween.tsx
cbrevik: Sure, can definitely default it to false 👍

cbrevik:
At the moment, the logic means: if parent is animating, keep the "old" child height until animation is finished, then switch to new visible child height.

This only works for two elements though, so if we are going to support more I need some way to cache keys, and lookup which elements we are transitioning between. For this use case, it works out okay though.

mikaelbr:
There are some issues it seems with favorites timing and empty list of favorites when animating from "result view" to "no result view". But that is not related to this. Just noting it down while I remember

PR #696, opened 2020-11-20
fix: Delay DisappearingHeader recalculation till after animation

cbrevik:
Think this is a bit better than the fix from last evening, since it forces DisappearingHeader to wait with content height recalculation till after animation is done.

Also fixes glitch in FadeBetween so it doesn't enable animation till after component mount


rikkekantega (approved):
LGTM :+1:

PR #712, opened 2020-11-23
feat: adds user preference to select start screen

mikaelbr:
Adds user preferences context to app. Also adds two settings:

- Setting preferred startup screen
- Overriding dark mode locally (not listening to phone settings)

Note: From the design I've added an option "override phone". Logically I have a hard time managing if it should show dark mode or not when not explicitly saying override. Imagine scenario: We toggle to dark mode in-app, but switch back to non-dark mode. And then we go back to phone settings and activate darkmode, but the app is still light. Now there is no option to ever let the phone decide darkmode or not. Another option is to only ever override darkmode, so setting darkmode option to true sets option but deactivating it removes the option. But this would cause you not to be able to override to have light mode if the phone it self is in dark mode. So having a flag saying to override I think is the most pragmatic choice.

image

image

image

### DIFF
--- a/src/components/sections/index.tsx
+++ b/src/components/sections/index.tsx
@@ -1,4 +1,8 @@
+// Sections
 export {default as Section} from './section';
+export {default as RadioSection} from './radio-section';
+
+// Links and actions
 export {default as LinkItem} from './link-item';
 export {default as HeaderItem} from './header-item';
 export {default as FavoriteItem} from './favorite-item';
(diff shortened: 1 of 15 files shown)


pettersuul:
Looks good! Should the override setting be "inverted", is it more understandable if it says "use phone settings" or something similar?

mikaelbr:
Yes, that sounds better.

cbrevik (approved):
Love it, these are really cool features! 🚀

Issue #666, opened 2020-11-10
[Designsync] Update tab-bar


pettersuul:
Use same color for text and icon on active tab.
Figma / Components 1.0 https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components-1.0?node-id=92%3A0

![image](https://user-images.githubusercontent.com/56530785/98677617-8aa60980-235d-11eb-971c-b60966a98002.png)

Issue #649, opened 2020-11-05
[Feature] Adjust navbar height to contain complete label with text enhancement


hildeor:
## Origin
_Links to received feedback, user research or other findings._

As observed in 2 usability tests with Blindeforbundet, the navbar is difficult to use in combination with text enlargement because the labels are truncated. 

![IMG_0150](https://user-images.githubusercontent.com/8980523/98247409-e5152380-1f73-11eb-9022-06e1ed084ca1.PNG)

## Proposed solution
_A coarse description of a proposed solution, that may include wireframes and graphic design._

- Adjust navbar height so that the complete labels are readable. 
- Hide icons in navbar if text is enlarged
- Change background color of active tab (?)

mikaelbr:
Should be solved by https://github.com/AtB-AS/mittatb-app/pull/717

PR #717, opened 2020-11-24
fix: tab bar icons, sizing and coloring

mikaelbr:
Updates Icons to the latest icons (small SVG changes), fixes coloring when active and text sizing.

Also sets max font size increase. Unsure if this is the best approach and most accessible. But feels it might be the most pragmatic (and also following WCAG guidelines). Open for input on this. See screenshots. Might solve #649 

Solves #666 

image

image

image


rikkekantega:
We've actually already added 2 as max font multiplier for theme text. This is indeed WCAG compliant. I think further font scaling isn't what's going to improve accessibility the most. We should perhaps rather explore options like implementing pinch zoom and contrast theme in the future.

src/navigation/TabNavigator.tsx
```diff
@@ -49,37 +52,55 @@ const NavigationRoot = () => {
...
+function tabSettings(
+  tabBarLabel: string,
+  Icon: (svg: SvgProps) => JSX.Element,
+): TabSettings {
+  return {
+    tabBarLabel: ({color}) => (
+      <Text
+        style={{color, fontSize: 14, lineHeight: 14}}
+        maxFontSizeMultiplier={2}
+      >
```
rikkekantega, line 100: Should we perhaps consider using ThemeText component and type here instead?

mikaelbr: You're right. I did this initially but found that the default lineHeight didn't work in this case, and ended up doing it manually. But at this point it's easier to use ThemeText and override line height manually. Updated

rikkekantega (approved):
Code LGTM :rocket:

hildeor (approved):

PR #727, opened 2020-11-25
fix: Add button role for all buttons + favorite name for a11ylabel

rikkekantega:


src/components/sections/favorite-item.tsx
```diff
@@ -28,10 +28,14 @@ export default function FavoriteItem(props: FavoriteItemProps) {
   if (!withOnPress(props)) {
     return <FavoriteItemContent {...props} />;
   }
+  const a11yLabel =
+    props.favorite.name === props.favorite.location.name
```
mikaelbr, line 32: Hm. I don't think this does what you want? `props.favorite.name` is optional and is undefined if it should use location name. So a better test would be:
```suggestion
    props.favorite.name
```

mikaelbr (changes requested):
🚀 Just one fix with the favorite name check.

src/components/sections/favorite-item.tsx
rikkekantega: Still early in the morning here :sweat_smile:

rikkekantega: Hmm no, actually not. We do save the address as favorite name if a user defined name isn't given. That was why I ended up with checking for equality here actually, to read out both name and address if they are different. But I see that the typing on favorites indeed has name as optional, so I still should consider possibility of it being undefined.
I do need more coffee. :coffee:

mikaelbr: Hm. I disagree with myself here. I actually think the name should be undefined yeah, as it's then easier to change the address of a favorite later. I guess I initially set it as defined to avoid null behavior, but the type says otherwise. I think we should at some point change this behavior.

rikkekantega: I agree :point_up:

PR #730, opened 2020-11-25
fix: Misc QA fixes

rikkekantega:
- Use button component on "Utvid kart"
- Update message component to obey design system on contrast bg/text for error messages 
- Minor text adjustments ticket view
- Accessible ticket count controls
- Logo click label and "navigate to start" action consistant with user preference
- Make tab bar height adjust to font scaling, and disappearingheader content height calculation use responsive tab bar height
- Make screen readers ignore icon picker, as it's currently not operatable by using screen reader navigation 

2x font resizing tab bar:
image


src/utils/navigation.tsx
```diff
@@ -1,13 +1,40 @@
...
+  Preference_ScreenAlternatives,
+} from '../preferences';
+import {TabNavigatorParams} from '../navigation/TabNavigator';
+import {useWindowDimensions} from 'react-native';
 
-const HOME_TAB_NAME = 'Assistant';
-
-export const useNavigateHome = () => {
+export const useBottomNavigationStyles = (): {height: number} => {
+  const {fontScale} = useWindowDimensions();
```
mikaelbr, line 12: Hey! I didn't know about that! I used `DeviceInfo` which is way more work! Thank you!

mikaelbr (approved):
Looks good! There's been some feedback about the "click on the logo" functionality, where clicking it expects to "reset the view" (we've also discussed this previously). But that is probably a different discussion than this QA fix.

cbrevik (approved):
Nice! Lots of good fixes here 🚀

rikkekantega:
> Looks good! There's been some feedback about the "click on the logo" functionality, where clicking it expects to "reset the view" (we've also discussed this previously). But that is probably a different discussion than this QA fix.

I agree, we need to work out a better logo look and function.

PR #731, opened 2020-11-25
fix: disables map when on iOS with voice over to prevent brutal crash

mikaelbr:
- Disable map for now when Voice Over is active on iOS. This is to prevent brutal app crash which is out of our control. Should be reverted as soon as possible.


rikkekantega (approved):
Gr8 :rocket:

PR #733, opened 2020-11-25
feat: Update payment flow to capture in backend

cbrevik:
# Changes
* Remove credit card capture from frontend, backend handles this in redirect now
* Add "active reservation", which is shown in tickets list, and keeps polling for updated payment status until it is either `CAPTURE` or a "handled" status like `CANCEL`. In the future we can probably build upon this to enbable retrying, etc.
* After payment status reaches `CAPTURE`, we poll for fare contracts
* This means NETS and Vipps payment flows are pretty similar. It also makes both flows more robust, since less has to be explicitly handled by frontend.
* Rewrite `TicketContext` to use a reducer

# Screenshot
The active reservation shown in the ticket list isn't great yet. But I'd like that to be left outside the scope of this PR. Want to land functionality for now:

image


src/TicketContext.tsx
```diff
@@ -3,59 +3,207 @@ import React, {
...
+        const {fare_contracts: fareContracts} = await listFareContracts();
+        dispatch({type: 'UPDATE_FARE_CONTRACTS', fareContracts});
+      } catch (err) {
+        console.warn(err);
+      }
+    },
+    [dispatch],
+  );
+
+  useInterval(
```
mikaelbr, line 164: This should still poll even if inactive?

mikaelbr (approved):
Didn't see any obvious issues. Think we should do proper QA-ing for this, but it's still closed beta, so we've got time for that also

src/TicketContext.tsx
cbrevik: Oops good catch, should be disabled if there are no active reservations, or if there are, and every one is CAPTURE'd

PR #734, opened 2020-11-25
refactor: use useWindowDimensions for font scale instead of DeviceInfo

mikaelbr:


cbrevik (approved):
https://github.com/AtB-AS/mittatb-app/blob/master/src/chat/user.ts#L26

Third way to check font scale 🤪

cbrevik:
It uses `Dimensions` in the end too though

rikkekantega (approved):
:ship: :it:

PR #736, opened 2020-11-26
fix: tab height issue on iOS and max scale

mikaelbr:
My Android Emulator crashed hard now, need to do a full reset. Could you check, @rikkekantega ?

Max size for iOS:

image


rikkekantega (approved):
Looks good! Nice to get this sorted out on iOS! :+1: :woozy_face:

PR #737, opened 2020-11-26
fix: A11ylabels for ticket support icons

rikkekantega:
Added a11ylabels for ticket support icon (upper left corner), so that screen reader users also can use this to determine if a ticket is valid or not.

Screenshot_1606378677


cbrevik (approved):
Awesome, thanks :D

PR #738, opened 2020-11-26
fix: hack/workaround for offset issues in header animation. NOT proper fix

mikaelbr:
This is ridiculous, I know. Some of this doesn't make any sense and is brittle as a wood glued vase. However, it seems to workaround with our current issues with animated header for now. I think we should rethink this component and look into shared elements to do this instead, but that would require more work now and postpone the wanted release. But we should do proper QA-ing of this.

The issues seem to be caused by `scrollY` animated value being triggered by RefreshControl and when animating back to the initial state after refreshing, it's no longer the correct value. Not an issue on Android as the animation is positioned absolute and elevated. This doesn't seem to be possible on iOS with the current implementation.

Should properly check on Android also as the implementation has changed some. Initial tests seem OK


cbrevik (approved):
LGTM, bit difficult to review changes here. So I'd rather QA in master :D

PR #739, opened 2020-11-26
fix: less buggy workaround animating header (but still not perfect)

mikaelbr:
Less errored fix (and less code), reverting the old fix. Still an issue, but I don't think it is blocking, even though it is annoying:

image


cbrevik (approved):
Yeah, LGTM. Annoying as you say, but probably lesser evil until we do a more thorough rewrite of this

Issue #635, opened 2020-11-02
[Feature] Communicate upcoming features in beta to engage users


rikkekantega:
## Origin
_Links to received feedback, user research or other findings._

Many user seem to quit using the app.
## Motivation
_A short description of what user needs or business goals this feature will solve._

Improve user retention
Engage users

## Hypotheses and assumptions
_A list of hypotheses and assumptions we have made about the user or the proposed solution._

## Proposed solution
_A coarse description of a proposed solution, that may include wireframes and graphic design._

Communicate upcoming / in work features, perhaps through the open space on app front page
Figma / App 1.0 / Assistant: https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=2172%3A0
Figma / Components 1.0 / Infobox: https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components-1.0?node-id=838%3A20

![image](https://user-images.githubusercontent.com/56530785/100353515-e5db2b80-2fee-11eb-8cb0-80f71079489d.png)

pettersuul:
![Screenshot 2020-11-24 at 11 54 18](https://user-images.githubusercontent.com/56530785/100085000-ca83eb00-2e4b-11eb-9993-8ec3c1bb466f.png)

pettersuul:
If we are using illustration with AtB-colors, we may need to have both dark- and light-mode versions available.

pettersuul:
![Uploading image.png…]()

gorandalum:
Some conclusions after a talk at the office:
- First version only text, no image.
- Use Firebase as source.
- Provide Norwegian and English version of the text.
- Allow optional link, and Norwegian and English link text. Not support inline links, but a "Read more" link at the bottom.

Thing mentioned, but not implemented for the first version:
- Image
- Possibility for the user to cross out the message box
- Different styling for more important/critical messages

PR #741, opened 2020-11-26
feat: Add component for highlighting news (#635)

gorandalum:
Add an additional parameter 'highlightComponent' to the
DisapperingHeader, which may be used for showing highlighted text and/or
images.

This new component is used on the search assistant screen where a news
banner is shown. The text and link for this news banner is fetched from
the remote config in Firebase.


gorandalum:
Screenshot 2020-11-26 at 16 46 04

rikkekantega:
Screenshot_20201126-170710_atb

The newsbanner component (or wrapper) could perhaps benefit from having a background color here, so we can control text/bg contrast.😊

src/components/disappearing-header/index.tsx
```diff
@@ -284,6 +292,13 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
     right: 0,
   },
 
+  highlightComponent: {
```
rikkekantega, line 295: Is there a possibility that font scaling up to 200% and/or a little more text content to the higlightComponent could possibly render off screen and become inaccessible? 
Just wanted to mention this, as we had those issues on the ticketing splash screen and onboarding views. We solved it by making the contents scrollable

rikkekantega (changes requested):
This is going to be great! And whow nice to see that you're up and running, contributing to the app already! :star_struck: 
I agree on keeping images etc out of scope. Lots of value to this feature with text only for the  first edition. I left two comments to consider, though - Just recognized a couple of possible accessibility issues.

gorandalum:
> The newsbanner component (or wrapper) could perhaps benefit from having a background color here, so we can control text/bg contrast.😊

I talked to Petter about this now. The sketches used the new colors for dark mode, which is not yet changed in the app. I'll create a new issue for changing the colors, if it doesn't already exist, and it is a requirement that the new issue is fixed before merging this issue. This was preferred over creating a temporary workaround for dark mode text in this highlight component.

src/components/disappearing-header/index.tsx
gorandalum: I'll look into it 👍

gorandalum: Yeah, lots of text and font scaling did not look too good. Talked with Petter about this, and he was not a big fan of scrolling.

Proposed solution:
- We set a description in Firebase Remote Config that specifies the maximum length for the text field
Screenshot 2020-11-27 at 10 41 11
- A text length of up to 120 characters and link text length of up to 40 characters will fit with font size set to max on both iOS and Android, also on small screens. This is tested locally.
- Overflowing text will be hidden. This should not happen unless someone do not follow the descriptions in Remote Config.

Does this sound as a good enough solution?

rikkekantega (approved):
LGTM :rocket:

gorandalum:
> > The newsbanner component (or wrapper) could perhaps benefit from having a background color here, so we can control text/bg contrast.😊
> 
> I talked to Petter about this now. The sketches used the new colors for dark mode, which is not yet changed in the app. I'll create a new issue for changing the colors, if it doesn't already exist, and it is a requirement that the new issue is fixed before merging this issue. This was preferred over creating a temporary workaround for dark mode text in this highlight component.

Change of plans. In addition to changing colors as mentions above, we also had to add support for light/dark-mode of the svg banner. Instead of starting to fix both these things we decided to change the svg banner to white and a opacity of `0.2`. This made it usable on both light and dark mode, without being to light compared to the white text in dark mode.

src/components/disappearing-header/index.tsx
gorandalum: After testing with the maximum accessibility text size in iPhone settings we saw that 120 characters was actually too many to be presented without overflowing. We ended up adding ScrollView, for the cases where both the number of characters and accessibility text size is maxed out. We preserve the text length descriptions in Firebase Remote Config so scrolling rarely have to be used.

PR #745, opened 2020-11-30
feat: i18n

rikkekantega:
i18n

Using Lobo-T for translations, and RNLocalize for localization

We decided on structuring translations as closely related to the visual representation as possible. They are typed, so usages (and even existence) is easily verified.

Per now all translate files are structured in a separate folder. Project component/screen folder structure somehow seemingly re-created. This is done to make the files easier to find for a non-delevoper doing only translations. We should consider if this works, or if we should move translations even closer to the code.

i18n is as of now added as a remote config setting. Default turned off. 

Some strings are still awaiting transformation to being translatable, but considering the major changes, we should probably merge with master sooner rather than later.

TODOS: 
Handling locale in date utils
Some translateabling


src/LanguageContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <lobot.LanguageProvider value={currentLanguage}>
+      {children}
+    </lobot.LanguageProvider>
+  );
+};
+function useLanguage() {
+  const [currentLanguage, setCurrentLanguage] = useState(DEFAULT_LANGUAGE);
+  const [locale, setLocale] = useState(preferredLocale);
+
+  const onChange = () => {
```
mikaelbr, line 23: Nitpick mode: I think this can be moved inside useEffect where it is used. That way we don't recreate it every time

src/LanguageContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+      ? getAsAppLanguage(locale.languageCode) ?? DEFAULT_LANGUAGE
+      : DEFAULT_LANGUAGE;
+    setCurrentLanguage(language);
+  }, [locale]);
+
+  return currentLanguage;
+}
+function preferredLocale(): RNLocalize.Locale | undefined {
+  const preferredSystemLocales = RNLocalize.getLocales();
+  const locale = preferredSystemLocales.find(
```
mikaelbr, line 44: What happens if nothing is found? Is that possible?

src/LanguageContext.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
mikaelbr, line 1: Discussion: I wonder if we could have this context in `src/translation` folder and move the `useTranslation` hook as a part of this module. With this we would co-locate everything relevant for translations to that particular module. Also we could do `import { useTranslation } from 'src/translation'` at some point.

src/LanguageContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+import * as RNLocalize from 'react-native-localize';
+import {useState, useEffect} from 'react';
+import {Language, lobot, DEFAULT_LANGUAGE} from './utils/language';
```
mikaelbr, line 4: I think maybe also co-locate these utils in same translation module.

src/location-search/LocationSearch.tsx
```diff
@@ -27,6 +27,9 @@ import {LocationSearchNavigationProp} from './';
 import LocationResults from './LocationResults';
 import {LocationSearchResult} from './types';
 import useDebounce from './useDebounce';
+import {useTranslation} from '../utils/language';
+import {LocationSearchTexts} from '../translations';
+import {TranslateFunction} from '../translations/utils';
```
mikaelbr, line 32: To give more context to the discussion from previous comments: I think this is semantically the same module and should be retrievable from the same file as "public API".

src/screens/Nearby/NearbyResults.tsx
```diff
@@ -52,10 +55,16 @@ const NearbyResults: React.FC<NearbyResultsProps> = ({
...
   const onPress = (departure: EstimatedCall) => {
     const {publicCode, name} = getLineNameFromEstimatedCall(departure);
+    const title = publicCode
+      ? `${publicCode} ${name}`
+      : name
+      ? name
+      : t(dictionary.travel.line.defaultName);
     navigation.navigate('DepartureDetailsModal', {
-      title: publicCode ? `${publicCode} ${name}` : name,
+      title: title,
```
mikaelbr, line 67: suuuuuuuuuper nitpicky stuff
```suggestion
      title,
```

src/screens/Profile/FavoriteList/index.tsx
```diff
@@ -54,23 +56,21 @@ export default function FavoriteList({navigation}: ProfileScreenProps) {
...
-                accessibilityHint: 'Aktivér for å redigere',
+                accessibilityHint: t(FavoriteListTexts.favoriteItem.a11yHint),
               }}
               onPress={navigateToEdit}
             />
           ))}
         </Sections.Section>
 
         <Sections.Section withPadding>
-          {!!items.length && (
```
mikaelbr, line 65: Here we introduce a regression error I think.

src/utils/date.ts
```diff
@@ -91,7 +103,15 @@ export function formatToClockOrRelativeMinutes(
 
   return secondsToMinutesShort(diff);
 }
-
+export function formatLocaleTime(date: Date, language?: Language) {
```
mikaelbr, line 106: Does it exist something in date-fn that can handle this?

src/utils/date.ts
```diff
@@ -10,46 +10,57 @@ import {
 import nb from 'date-fns/locale/nb';
```
mikaelbr, line 10: Should we have more locales here now?

mikaelbr (changes requested):
Some minor comments. Looks good! Only thing I'm unsure of is the structure of the different files regarding translate/language. See comments for context. Marking as request change but it's mostly to mark it as a discussion point.

android/settings.gradle
```diff
@@ -1,3 +1,5 @@
 rootProject.name = 'atb'
+include ':react-native-localize'
```
cbrevik, line 2: This shouldn't be necessary since packages will autolink 0.60+

android/app/src/main/java/no/mittatb/MainApplication.java
```diff
@@ -11,6 +11,7 @@
 import com.bugsnag.android.Configuration;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
+import com.reactcommunity.rnlocalize.RNLocalizePackage;
```
cbrevik, line 14: Remove this too, because of auto-link.

ios/Podfile
```diff
@@ -15,6 +15,8 @@ target 'atb' do
   permissions_path = '../node_modules/react-native-permissions/ios'
   pod 'Permission-LocationWhenInUse', :path => "#{permissions_path}/LocationWhenInUse.podspec"
 
+  pod 'RNLocalize', :path => '../node_modules/react-native-localize'
```
cbrevik, line 18: Because of auto-link this shouldn't be necessary

cbrevik (changes requested):
Going to do a more thorough read-through. But requesting changes on these so you're aware.

src/utils/date.ts
rikkekantega: Something.. Perhaps! Haven't looked into it yet, the date functions are still on the todo list

src/remote-config.ts
```diff
@@ -3,6 +3,7 @@ import remoteConfig from '@react-native-firebase/remote-config';
 export type RemoteConfig = {
   enable_ticketing: boolean;
   enable_intercom: boolean;
+  enable_i18n: boolean;
```
cbrevik, line 6: Good idea! We should do this more often, that is to say feature toggling big features in case of regressions or extended testing. Love it! 🙌

src/utils/date.ts
rikkekantega: Probably

android/settings.gradle
rikkekantega: Ah, wasn't aware of that. :thinking:

android/app/src/main/java/no/mittatb/MainApplication.java
rikkekantega: :+1:

ios/Podfile
rikkekantega: :+1:

src/remote-config.ts
rikkekantega: wee :smile:

src/screens/Profile/FavoriteList/index.tsx
rikkekantega: Somewhat faulty merge with master, probably. Not intentional. I'll look through and compare.

src/screens/Nearby/NearbyResults.tsx
rikkekantega: Yes, but really non-pleasing to the eye :grimacing: Some automatic refactoring gone wrong

src/location-search/LocationSearch.tsx
rikkekantega: :+1:

src/LanguageContext.tsx
rikkekantega: :+1:

src/LanguageContext.tsx
rikkekantega: will do

src/LanguageContext.tsx
cbrevik: That's what happened in Mapbox with Voice Over (hence the crash) :D We should definitely default with Norwegian here

src/components/map/BackArrow.tsx
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
+import {TouchableOpacity, View} from 'react-native';
```
cbrevik, line 2: Are these used?

src/LanguageContext.tsx
rikkekantega: Nothing found as nothing returned by RNLocalize.getLocales()?

cbrevik: Yeah it is possible.

rikkekantega: Yep. Then locale will be undefined, as given by return type. This method is actually just intended to return a locale IF a match on configured App languages is found from users preferenced system locales. As of now, we only use locale to decide on language, with a fallback to default language. 
But we might consider having a "fallback locale" if locale is to be used for anything else :thinking:

rikkekantega: @cbrevik what, that's just creepy

cbrevik: Sorry I should have been more specific, you will get some `preferredSystemLocales` back, but you aren't guaranteed those will be either Norwegian or English. In other words, `preferredLocale()` might return `undefined`.

rikkekantega: Ah okay, but as my comment above, I'm aware of that. 😊
As of now, we only use locale to decide on language, with a fallback to default language.
But we might consider having a "fallback locale" if locale is to be used for anything else 🤔 What do you think?

src/LanguageContext.tsx
rikkekantega: Agreed, done :tada:

ios/Podfile
```diff
@@ -15,6 +15,7 @@ target 'atb' do
   permissions_path = '../node_modules/react-native-permissions/ios'
   pod 'Permission-LocationWhenInUse', :path => "#{permissions_path}/LocationWhenInUse.podspec"
 
+
```
cbrevik, line 18:
```suggestion
```

cbrevik (approved):
When cloning this branch locally, my yarn.lock is modified on dep install. Is it updated?

rikkekantega:
> When cloning this branch locally, my yarn.lock is modified on dep install. Is it updated?

Yes, I've added to packages, react-native-localize and lobo-t. Aren't we supposed to track the lockfile?

cbrevik:
Yep, we are. But when I run `yarn install` on your branch, yarn.lock is modified. So I was wondering if you've checked in the latest modifications?

rikkekantega:
That's strange. I don't have an uncommitted change for the yarn.lock file :thinking:

cbrevik:
> That's strange. I don't have an uncommitted change for the yarn.lock file 🤔 Running yarn install here doesn't modify the lockfile either.

After deleting the branch, and re-fetching, I no longer experience this. Weird stuff! So false alarm 🙈

PR #751, opened 2020-11-30
fix: support for arbitrary falsey items in section

mikaelbr:
There were some issues with handling falsey values in section wrapper. This checks for actual first and last index of valid child to set proper radius.

Point is now to get a list indicating what is valid children and not. With that list we find both first and last index that is valid. These should get radiuses set. Technically valid children can be strings, but I think that is kind of outside the scope to patch that kind of usage.


cbrevik (approved):
👍 👍 👍

PR #753, opened 2020-11-30
fix: Section hotfixes

cbrevik:


mikaelbr (approved):
Verified fix for empty and non-empty favorite lists.

PR #760, opened 2020-12-02
fix: Terminal payment redirect load needs time to finish

cbrevik:
Move payment redirect check to `loadEnd` so race condition where callback is cancelled if we navigate away from the WebView too fast.


rikkekantega (approved):
Seems legit

PR #764, opened 2020-12-04
feat: Passenger selection in ticket buy process

gorandalum:


gorandalum:
Contains a lot of changes. May be better to view each commit separately when looking at the changes.

src/TicketContext.tsx
```diff
@@ -170,22 +217,28 @@ const TicketContextProvider: React.FC = ({children}) => {
...
     [],
     !activeReservations.some((res) => res.paymentStatus === 'CAPTURE'),
   );
 
   useEffect(() => {
-    updateFareContracts();
+    updateFareContractTickets();
+  }, []);
+
+  useInterval(updateFareContractTypes, 10000);
```
gorandalum, line 230: Refreshing every ten minute. Or does it need to be refreshed at a set interval at all?

mikaelbr: Haven't read/seen any of the code but just a heads up that you can look into: I think there are limitations on timeout time for android. Around 2 minutes iirc

src/screens/Ticketing/Purchase/Travellers/traveller-types.tsx
```diff
@@ -0,0 +1,35 @@
+import {TravellerType} from '../../../../api/fareContracts';
+
+export type Traveller = {
+  type: TravellerType;
+  text: string;
+};
+
+export type TravellerWithCount = Traveller & {count: number};
+
+export const travellerTypes: Traveller[] = [
```
gorandalum, line 10: Using this temporary as the source for the traveller type options. Will use `userTypes` service or json when the model is updated to include the traveller types which are used when buying tickets.

src/screens/Ticketing/Purchase/index.tsx
```diff
@@ -1,28 +1,47 @@
...
 import {ActiveTicketsScreenName} from '../Tickets';
-import {ReserveOffer} from '../../../api/fareContracts';
+import {FareContractType, ReserveOffer} from '../../../api/fareContracts';
+import {DetailScreenRouteProp} from '../../TripDetailsModal/Details';
+import {RouteProp} from '@react-navigation/core';
 
-type PaymentParams = {offers: ReserveOffer[]};
+type TravellersParams = {
+  refreshOffer?: boolean;
+  fareContractType: FareContractType;
```
gorandalum, line 12: Added the `fareContractType` as a param in navigation. Thought of setting it in the `TicketContext` instead, making it not necessary to pass it around between navigation stack screens. What is preferred?

src/screens/Ticketing/Tickets/ReceiptModal.tsx
```diff
@@ -1,11 +1,11 @@
 import React, {useState} from 'react';
 import {View, TouchableOpacity, Modal, TextInput} from 'react-native';
-import {FareContract, sendReceipt} from '../../../api/fareContracts';
+import {FareContractTicket, sendReceipt} from '../../../api/fareContracts';
 import {StyleSheet} from '../../../theme';
 import Text from '../../../components/text';
 
 interface ModalProps {
-  fareContract?: FareContract;
+  fareContract?: FareContractTicket;
```
gorandalum, line 8: Note to self: rename this to fareContractTicket

src/screens/Ticketing/Tickets/Ticket.tsx
```diff
@@ -67,8 +67,8 @@ const Ticket: React.FC<Props> = ({fareContract: fc, now}) => {
         <View style={styles.ticketInfoContainer}>
           <ThemeText style={styles.travellersText}>
             {fc.user_profiles.length > 1
-              ? `${fc.user_profiles.length} voksne`
-              : `1 voksen`}
+              ? `${fc.user_profiles.length} billetter`
```
gorandalum, line 70: As of now we do not get the traveller types in the `FareContractTicket'. Talked with Torbjørn about this, and it will be changed to returning a `travellers` array which has reference to both the ticket type and traveller type.

src/TicketContext.tsx
gorandalum: I see now that I wrote 10 minutes, it is actually set to 10 seconds, which probably is unnecessary often.

cbrevik: We should probably retrieve this on app startup instead. This shouldn't update that often?

src/screens/Ticketing/index.tsx
```diff
@@ -6,5 +6,5 @@ import {useRemoteConfig} from '../../RemoteConfigContext';
 export default function Ticketing() {
   const {enable_ticketing} = useRemoteConfig();
 
-  return !enable_ticketing ? <Splash /> : <Tickets />;
+  return <Tickets />;
```
cbrevik, line 9: I'd recommend enrolling to ticketing in simulator instead of commenting out this. Bit risky if it accidentally hits production.

src/screens/Ticketing/Tickets/ReceiptModal.tsx
cbrevik: I'm not sure if we should go for that naming standard. Unless I've missed something, I think `FareContract` contains tickets, so `FareContractTicket` might be misleading.

src/api/fareContracts.ts
```diff
@@ -11,9 +12,16 @@ export async function list(): Promise<ListTicketsResponse> {
   return response.data;
 }
 
+export async function listTypes(): Promise<FareContractType[]> {
```
cbrevik, line 15: Should we call these `ProductTypes` instead? As far as I know, a `FareContract` can contain multiple products? This might be a good discussion point wrt domain modeling, of course.

src/screens/Ticketing/Purchase/Travellers/index.tsx
```diff
@@ -77,116 +92,114 @@ const Travellers: React.FC<Props> = ({navigation, route: {params}}) => {
...
+            onPress={payWithCard}
+            viewContainerStyle={[
+              styles.paymentButton,
+              styles.cardPaymentButton,
+            ]}
+            textContainerStyle={{marginLeft: 20}}
+          />
+        </View>
       </View>
+      <View style={styles.bottomColor} />
```
gorandalum, line 196: This feels a little hackish, but couldn't find a better solution on the fly. 

Originally the footer was using `{ position: absolute; bottom: 0 }` making it stick to the bottom, but this made the `<ScrollView>` for the travellers not work optimally as it did not know how much space it had available. Then I changed the footer to not be using absolute positioning, and instead use `justifyContent: 'space-between'` on the parent. But this made a gap under the footer with a different color because of the `<SafeAreaView>`, and this `<View style={styles.bottomColor} />` with absolute positioning was added just for filling the color at the bottom of the screen.

I would love to learn a better solution, if anyone has.

src/screens/Ticketing/index.tsx
gorandalum: Enrolled now. Reverting this.

src/api/fareContracts.ts
gorandalum: The name was based on the json file, which was named `fare-contract-types.json`. The actual endpoint we now use is named `preassigned-fare-products`. Is naming the type `PreassignedFareProduct` ok?

src/screens/Ticketing/Tickets/ReceiptModal.tsx
gorandalum: Aight, reverting these changes. The change was made for better distinction between ` FareContract(Ticket)` and `FareContractType`, but now when the latter is renamed to ` PreassignedFareProduct` there is less need for better distinction.

src/api/fareContracts.ts
gorandalum: Renamed to `PreassignedFareProduct` for now.

src/api/fareContracts.ts
```diff
@@ -1,19 +1,29 @@
...
 
   const url = 'ticket/v1/ticket/' + customerId;
   const response = await client.get<ListTicketsResponse>(url);
 
+  return response.data.fare_contracts;
+}
+
+export async function listPreassignedFareProducts() {
+  const url =
+    'https://gateway-5pwmu3xz7a-ew.a.run.app/api/v1/reference-data/ATB/preassigned-fare-products';
```
gorandalum, line 17: Now using temporary endpoints, as is it not yet deployed in the staging environment. Should not release with these endpoints.

gorandalum: Updated to the mittatb-endpoints.

src/TicketContext.tsx
```diff
@@ -37,15 +47,18 @@ type TicketReducer = (
   action: TicketReducerAction,
 ) => TicketReducerState;
 
-const ticketReducer: TicketReducer = (prevState, action) => {
+const ticketReducer: TicketReducer = (
+  prevState,
+  action,
+): TicketReducerState => {
```
cbrevik, line 53: Is typing `TicketReducerState` necessary here when the variable is already typed to `TicketReducer` (which in turn is already typed to return `TicketReducerState`)?

src/TicketContext.tsx
```diff
@@ -56,6 +69,19 @@ const ticketReducer: TicketReducer = (prevState, action) => {
         isRefreshingTickets: false,
       };
     }
+    case 'SET_IS_REFRESHING_PREASSIGNED_FARE_PRODUCTS': {
+      return {
+        ...prevState,
+        isRefreshingPreassignedFareProducts: true,
+      };
+    }
```
cbrevik, line 77: Do you need `SET_IS_REFRESHING_PREASSIGNED_FARE_PRODUCTS` as its own action type? Since it only happens on startup, couldn't `isRefreshingPreassignedFareProducts` be set as `true` in `initialReducerState` further down?

src/TicketContext.tsx
```diff
@@ -151,9 +183,25 @@ const TicketContextProvider: React.FC = ({children}) => {
...
+    async function () {
+      try {
+        dispatch({type: 'SET_IS_REFRESHING_PREASSIGNED_FARE_PRODUCTS'});
+        const preassignedFareProducts = await listPreassignedFareProducts();
+        dispatch({
+          type: 'UPDATE_PREASSIGNED_FARE_PRODUCTS',
+          preassignedFareProducts,
+        });
       } catch (err) {
         console.warn(err);
```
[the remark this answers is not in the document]
cbrevik: Yes definitely, if this fails then you have no better way out of this. I'm actually sort of leaning towards using Firebase RemoteConfig here instead, and so we can bundle a default set of products with the app. And update/cache via RemoteConfig.

src/screens/Ticketing/Purchase/Travellers/index.tsx
```diff
@@ -210,40 +226,47 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
...
+    marginTop: theme.spacings.medium,
+    flex: 1,
+    minHeight: 64,
+  },
+  vippsPaymentButton: {
+    marginRight: 6,
+  },
+  cardPaymentButton: {
+    marginLeft: 6,
+  },
```
cbrevik, line 262: I think we have to look into the use of these styles.  My thinking is that flexbox could probably solve some of these problems for us. Also should rather use `theme.spacings` here rather than explicit numbers.

src/screens/Ticketing/Purchase/Travellers/index.tsx
cbrevik: We can look at this together if you want

cbrevik (commented):
I've looked over briefly, and commented on a few things. I'll spend some time this morning looking at reducer/hooks-related changes locally

cbrevik (approved):
LGTM 💯 Awesome job on this. Since the size of this is as big as it is, it should probably be additionally heavily tested in QA

PR #767, opened 2020-12-08
feat: Basic order details

cbrevik:
# Changes
* Refactor `Ticket` into own module, use `Section`-groups for content.
* Add Detail-modal for `Ticket` (which has Order-id)
* Add `Sections.GenericItem`, for when you need to use arbitrary/custom content inside a `Section`-item.
* Fix so `Section` items uses available space outside `ScrollView`
* Add disabled prop for `Sections.LinkItem`
* Add `withBottomPadding` on `Section`-group, when you need padding on the bottom of items

# Screenshot
image


src/assets/svg/icons/other/Banner.tsx
```diff
@@ -5,8 +5,7 @@ function SvgBanner(props: SvgProps) {
   return (
     <Svg width={20} height={20} fill="black" viewBox="0 0 320 160" {...props}>
       <Path
-        fill="white"
-        opacity={0.2}
+        fill="#AAE6EC"
```
gorandalum, line 8: Was this change intended?

gorandalum (approved):
Looks very good, except for what may be a regression in `Banner.tsx`.

src/assets/svg/icons/other/Banner.tsx
cbrevik: Sort of. The SVG-components under `src/assets` are generated by `yarn generate-svgs`, which uses the SVGs from `assets` under the root-folder as source. In other words, these are auto-generated and shouldn't really be modified directly. I've added a commit which adjusts Banner-opacity from DisappearingHeader instead

PR #772, opened 2020-12-11
fix: Use closest location from reverse geo lookup

gorandalum:
Fixed a bug (three places) where the second closest location from the
hook useReverseGeocoder was used as the current location.

As every usage of useReverseGeocoder was mainly interested in only the
closest location, this was added as an own field in the returned state.

### DIFF
--- a/src/location-search/map-selection/index.tsx
+++ b/src/location-search/map-selection/index.tsx
@@ -64,13 +64,11 @@ const MapSelection: React.FC<Props> = ({
     ],
   );
 
-  const {locations, isSearching, error} = useReverseGeocoder(
+  const {closestLocation: location, isSearching, error} = useReverseGeocoder(
     centeredCoordinates,
   );
 
   const {location: geolocation} = useGeolocationState();
-
-  const location = locations?.[0];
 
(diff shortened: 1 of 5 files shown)


cbrevik:
This is one of those 🙈 bugs. Awesome catch!

src/location-search/map-selection/index.tsx
```diff
@@ -64,14 +64,12 @@ const MapSelection: React.FC<Props> = ({
...
   );
 
-  const {locations, isSearching, error} = useReverseGeocoder(
+  const {closestLocation: location, isSearching, error} = useReverseGeocoder(
     centeredCoordinates,
   );
 
   const {location: geolocation} = useGeolocationState();
 
-  const location = locations?.[0];
```
cbrevik, line 73: Weirdly enough here it's correct. Might be why it hasn't been discovered, seeing as "Map"-selection is more visual and precise.

cbrevik (approved):
Tested locally, and seems to fix the issue! Great catch 🙌🏻

mikaelbr (approved):

PR #773, opened 2020-12-11
feat: Simple tariff zone selection

gorandalum:
Added simple tariff zone selection functionality, just listing the names of the
available tariff zones. There is not yet any functionality for searching by
location, adresses, bus stops, or using maps.


gorandalum:
Not yet ready for review. Have to do a cleanup job regarding base commit for this branch.

src/screens/Ticketing/Purchase/Travellers/index.tsx
```diff
@@ -35,14 +36,25 @@ const Travellers: React.FC<TravellersProps> = ({
 
   const {userProfilesWithCount, addCount, removeCount} = useUserCountState();
 
+  const {tariffZones} = useTicketState();
+  const {
```
gorandalum, line 40: For now just using `tariffZones[0]` as the default, which conveniently is _Sone A_. We probably want a more reliable method for selecting the default zone. Should the default zone already be flagged when it is returned from backend?

src/screens/Ticketing/Purchase/Travellers/index.tsx
```diff
@@ -35,14 +36,25 @@ const Travellers: React.FC<TravellersProps> = ({
...
+    fromTariffZone = tariffZones[0],
+    toTariffZone = tariffZones[0],
```
gorandalum, line 42: For now just using tariffZones[0] as the default, which conveniently is Sone A. We probably want a more reliable method for selecting the default zone. Should the default zone already be flagged when it is returned from backend?

cbrevik: When trying this locally now, my default zone was "Sone B3". This should probably be part of the "data washing" which happens in Remote Config.

cbrevik:
zone-search

Seems like there's some problems with the text input? Unsure if this is just me. I have to type letters several times before they register

src/screens/Ticketing/Purchase/Travellers/index.tsx
gorandalum: > Seems like there's some problems with the text input? Unsure if this is just me. I have to type letters several times before they register

Ahh. I've used `debouncedText` instead of `text` as the InputField value. Fixing.

gorandalum:
> Seems like there's some problems with the text input? Unsure if this is just me. I have to type letters several times before they register

Ahh. I've used debouncedText instead of text as the InputField value. Fixing.

gorandalum:
> Ahh. I've used debouncedText instead of text as the InputField value. Fixing.

@cbrevik Now fixed this.

Note: This branch was force pushed as i rebased on master.

src/screens/Ticketing/Purchase/Travellers/index.tsx
gorandalum: Yes. The ordering of the tariff zones from the reference data endpoint has changed now. 

We can probably change this later, this is a "problem" we have with all the reference data now. More ticket types than only "Enkeltbillett" are shown, user profiles are not ordered, and no (good) way to specify default tariff zone. This will be improved when it is fully decided where to put the "data washing".

cbrevik (approved):
Great job on this, looks good to me! We probably have to look a bit at the behaviour of saving "zone" choices. And there's currently a bug in the backend which only gives us offers on "Sone A". But lets get this into master 💯

PR #774, opened 2020-12-11
feat: Add receipt screen, in ticket details modal

cbrevik:
# Changes
* Rename `TicketDetails` to `TicketModal`
* Add `StackNavigator` in `TicketModal`-root.
* Re-design `Receipt`-screen.
* Add `success`-type in `MessageBox`. CC @pettersuul hope it looks okay according to theme? See screenshot with success message.

# Screenshots
Now there is an option to "Be om kvittering" under Ticket-details:
image
Initial screen with info:
image
Success message (added the new MessageBox-style for success here):
image
Error message:
image


gorandalum (approved):
Looks good!

One question: In the design for ticket purchase flow there was the text "Avbryt" in the upper left corner, but in this design there is a **X**-icon in the upper left corner. Is there a conscious choice where to use one or the other?

cbrevik:
>One question: In the design for ticket purchase flow there was the text "Avbryt" in the upper left corner, but in this design there is a X-icon in the upper left corner. Is there a conscious choice where to use one or the other?

Good point! I wonder if it would make more sense to use "Lukk" though, since you're not aborting a process.

Issue #754, opened 2020-12-01
[Feature] Group departures by line


mikaelbr:
## Origin
_Links to received feedback, user research or other findings._

See issue #420 

## Acceptance criteria

- [ ] Should be able to search for stop places and addresses and see departures nearby.
- [ ] Should show multiple departures for each line
- [ ] Should group by distinct lines e.g. `24 City Syd` and `24 City Syd v/Kroppanbrua` would be different groupings even if they are technically the same line.
- [ ] Should keep screen up to date with realtime dates
- [ ] Should be able to expand/contract stop places. 
- [ ] Should be able to show more lines on quays.
- [ ] Should be able to navigate to departure details
- [ ] Should hide old/outdated departures.
- [ ] Should show all types of transportation modes.
- [ ] Should be accessible with screen reader with a flow that makes sense and able to skip between lines and quays.

## Motivation
_A short description of what user needs or business goals this feature will solve._

Allow the user to more easily see relevant information.

## Hypotheses and assumptions
_A list of hypotheses and assumptions we have made about the user or the proposed solution._

Many users use the same line every day from the same stop, and want to easily find these departures.

## Proposed solution
_A coarse description of a proposed solution, that may include wireframes and graphic design._

Using:

![image](https://user-images.githubusercontent.com/56530785/98776574-7874ab80-23ef-11eb-9de8-2b128dfa7454.png)

Figma / App 1.0 / Favorite https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=8%3A1529
Figma / App 1.0 / Departure https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=4%3A982

mikaelbr:
Should we actually group by line or by named line? There can be some variations on lines it seems.

Description of a line:

> A group of routes which is generally known to the public by a similar name or number

We have examples like:

```
Line ID: ATB:Line:200109141544126_24_200109141539610
Fronttext on bus: 24 City Syd via Kroppanmarka

---

Line ID: ATB:Line:200109141544126_24_200109141539610
Fronttext on bus: Sjetnmarka via Kroppanm.-Okstad
```

There can be quite the difference. So currently thinking we also should group on fronttext.

PR #775, opened 2020-12-12
feat: Adds grouped departure view

mikaelbr:
Todo:
- [x] Verify accessibility/screen readers
- [x] Verify Dark Mode
- [x] Change color of button to be lighter. This is dependent on changing buttons to have accents.
- [x] Verify language strings. 

Dependent on service:
- [x] https://github.com/AtB-AS/atb-bff/pull/57

This is a big change on both the BFF and App. For Code Review I think it would be helpful to run it locally. Could also see acceptance criteria on #754 

Note:
- Haven't verified `HiddenHeader` component in screen reader. Difficult to do with iOS in Simulator so need to verify on device.
- Now the logic for "left transportations" is as follows: If under 1 minute until arrival, show "Now". If estimated arrival is passed within 1 minutes show "now" but disabled button. After that, remove. No re-fetching new departures. Meaning, if users have the view open for a long time it will get fewer and fewer departures in the view.
- Currently, we have some out of sync color naming/structure. Currently just working around that but should be fixed at some point.

Screenshot 2020-12-15 at 16 37 47
Screenshot 2020-12-15 at 16 37 40
Screenshot 2020-12-15 at 16 37 24


rikkekantega:
This seems to be implemented exactly as designed, so this is more of an idea for discussion... I know it would cause more work, but I feel the need to throw it out here.

When navigating "departures" view, my first impression is that there is **a lot** of data. Some of it hidden away by collapsable stops and within "load more" for the individual quay, but I'm not sure if the right datas are hidden at first :thinking: 

image
:point_up:  I can't help but feel that my first goal as a traveller here, would be to be able to navigate to my desired departures from the quay. Which would be "70 - Stjørdal", "70D Hommelvik" and "79 - Vikhammeråsen". But they are hidden within "Load more" right now, because of other sooner departures. I guess I would never know how many "load more"'s I'd have to do to get the full overview of possible lines I'd like to compare... :thinking: I know of the upcoming favorite function that will solve some of this, but I still feel like this will be an obstacle whenever I want to get an overview for a new quay I don't usually travel from. E.g if I suddenly decided to pop by a shop on my way home, and suddenly was going to check departures from "Dronningens gate" instead.

Based on my first impression, I just got this feeling. What if all possible lines from a quay are visible on the first load, and _they_ are collapsable? Then you would remove some of the "noise" made by all the time boxes, you get a quick overview  And you could toggle the departures you'd like to display times for.. And you would easily just skip past lines you're not interested in with a screen reader.

I know this would mean added work, and expanding/changing scope at this point would cost a lot of resources. So I'm not saying that we must consider this right now. But I'm curious what you and the others think

rikkekantega:
Have run this locally, and the initial impression is that the grouped departure view seem to work as intended. Haven't looked into code yet. 

I noticed just a few things while running fast through by using screen reader. First of all, so many reachable elements, it's really pleasing to test through this! :clap: 

I know you haven't checked off screen reader verification, not sure if you meant to do that yourself, but perhaps these "findings" are somewhat helpful:

image

- Perhaps the toggleable stop name header could be considered a **switch** with an on/off state? (https://reactnative.dev/docs/accessibility#accessibilityrole)
Right now it reads "Prinsens gate - Skjul". Without the visual support of the chevron and the contained quays, the function isn't very easy to understand.

image

- The distance label on quay is inaccessible, due to the parent having an accessibilityLabel set (screen reader then ignores child content, unless they're explicitly marked as accessbile). Info on distance could possibly be added to the platform header accessibility label.

image
- Last of all, these guys. :point_up:  Perhaps the times could be a11y-prefixed, like "1. Avgang", "2. Avgang" and so on? 
I really like the way greyed out "Nå" actually reads that the departure has probably passed :+1: :D 

I'm also really tempted to create a reusable component to enable horizontal scroll views to be easily skippable for screen readers. We already have a possbile issue of many favorites that may be annoying to step through, and now these departure times. But I don't want to expand the scope here, I'm going to describe that as a separate issue.

mikaelbr:
I've investigated the ability to skip content. It's kind of finicky as I do it now as you imperatively have to select what component and can only jump to a component that is accessible. So it is manual, verbose, and kind of awkward. But according to my tests it works and I think it adds good value for screen readers.

I think it is possible to make it a reusable component that is more generic, but I think this can be refactored later 😅

rikkekantega:
> I've investigated the ability to skip content. It's kind of finicky as I do it now as you imperatively have to select what component and can only jump to a component that is accessible. So it is manual, verbose, and kind of awkward. But according to my tests it works and I think it adds good value for screen readers.
> 
> I think it is possible to make it a reusable component that is more generic, but I think this can be refactored later 😅

mikaelbr:
FYI all: I've verified (at least tried to) usage of this in regards to accessibility, themes and language now. Awaiting any feedback for CR. I've started a new branch with favorites.

src/screens/Nearby/section-items/line.tsx
```diff
@@ -0,0 +1,225 @@
...
+          <View style={styles.transportationMode}>
+            <ThemedTransportationIcon mode={group.lineInfo?.transportMode} />
+          </View>
+          <ThemeText>{title}</ThemeText>
+        </TouchableOpacity>
+      </View>
+      <ScrollView
+        horizontal
+        contentContainerStyle={styles.scrollContainer}
+        accessibilityElementsHidden
```
rikkekantega, line 97: Also need to provide the Android variant, `importantForAccessibility="no"

mikaelbr: Ah. Thanks 😍  (one would think this could be patched internally...)

mikaelbr: Should it be `importantForAccessibility="no-hide-descendants"` ?

rikkekantega (approved):
Really great work! :raised_hands: :tada: 

Just a minor comment about adding the `importantForAccessibility` value to support Android screen readers as well.

I've read through code and tested locally, and everything just look really nice as far as I can tell! :raised_hands: :star: :clap:

src/screens/Nearby/section-items/line.tsx
rikkekantega: Yes, doesn't make sense that they are two different properties, but perhaps to support the no/no-hide-descendants difference on Android . "no" hides only the current element, "no-hide-descendats" hides all children. The latter is an easy "go to", I think that's almost always what we'd really want to do.

mikaelbr:
@rikkekantega but 59b438a2064d7ef40c0f4311a9774d76ecb1516c should fix that comment, right?

ios/Podfile.lock
```diff
@@ -692,7 +692,7 @@ SPEC CHECKSUMS:
   Intercom: 5d8058e3476ab6b0350dea165338164364eb1acb
   lottie-ios: 48fac6be217c76937e36e340e2d09cf7b10b7f5f
   lottie-react-native: 1fb4ce21d6ad37dab8343eaff8719df76035bd93
-  Mapbox-iOS-SDK: 6a67397df2bcafe5d2851604192aa8de6f8d6c89
+  Mapbox-iOS-SDK: 2563ed87ead6ec08f1c2090873977b8a7be335a8
```
mikaelbr, line 695: My lock file keep on insisting on this, but you didn't get the same, @cbrevik ?

cbrevik: No, thats quite weird

src/components/button/index.tsx
```diff
@@ -48,6 +52,17 @@ const Button: React.FC<ButtonProps> = ({
...
+    .current;
+
+  React.useEffect(() => {
+    Animated.timing(fadeAnim, {
+      toValue: disabled ? DISABLED_OPACITY : 1,
+      duration: 200,
+      easing: Easing.inOut(Easing.ease),
+      useNativeDriver: true,
+    }).start();
+  }, [disabled, fadeAnim]);
```
cbrevik, line 65: 😍

src/components/sections/section-utils/use-section-item.ts
```diff
@@ -33,6 +35,7 @@ export default function useSectionItem({
     padding: spacing,
     alignSelf: isInline ? 'flex-start' : undefined,
     ...mapToBorderRadius(theme, radiusSize, radius),
+    backgroundColor: transparent ? undefined : theme.background.level0,
```
cbrevik, line 38: `backgroundColor` also accepts `'transparent'`, not a big deal tho, just an FYI

src/components/sections/section.tsx
```diff
@@ -33,6 +33,7 @@ export default function SectionGroup({
     <View style={containerStyle}>
       {React.Children.map(children, (child, index) => {
         if (!React.isValidElement(child)) return child;
+        if (child == null) return child;
```
cbrevik, line 36: Trying to catch `undefined` as well?

src/index.tsx
```diff
@@ -13,15 +13,34 @@ import SearchHistoryContextProvider from './search-history';
...
+      values2?: unknown,
+    ) => {
+      if (typeof values === 'string') {
+        metadata[values] = values2;
+      } else {
+        metadata = {
+          ...metadata,
+          ...values,
+        };
+      }
```
cbrevik, line 39: :o what's going on here?

src/index.tsx
```diff
@@ -43,6 +63,14 @@ async function setupConfig() {
 trackAppState();
 enableScreens();
 
+// Adds ability to automatically animate between layout paint changes.
+if (
+  Platform.OS === 'android' &&
+  UIManager.setLayoutAnimationEnabledExperimental
+) {
+  UIManager.setLayoutAnimationEnabledExperimental(true);
+}
```
cbrevik, line 72: https://github.com/facebook/react-native/issues/25832

Bit unsure about this, maybe we should feature toggle this?

cbrevik (approved):
Awesome work! Looked through, didn't find anything blocking. Only thing I'm wondering is if we should feature toggle `setLayoutAnimationEnabledExperimental` in case of issues on Android devices.

src/components/sections/section-utils/use-section-item.ts
mikaelbr: Good point! But I'll just leave it and fix it later 😬

src/components/sections/section.tsx
mikaelbr: Yeah, can do that. It was a specific use case where returning `null` from component is a common use case. But as you say, undefined can also be a thing.

src/index.tsx
mikaelbr: Oh yeah. It's to support overloading functions in the simplest way 😬

src/components/sections/section.tsx
cbrevik: Was a reference to it being a falsy check `==`. Was wondering myself if `undefined` is a case.

src/index.tsx
mikaelbr: Yeah, good point. I'll fix

mikaelbr: Just removing it entierly. Triggering some errors

src/components/sections/section.tsx
mikaelbr: Yes. That's right. Everything is as it should

PR #778, opened 2020-12-15
chore: Upgrade Entur SKD

gorandalum:
Necessary for retrieving tariff zones on geolocation features venues.


mikaelbr:
Yeah. At some point we should move away from these types and use our own. But until then this looks good to me.

mikaelbr (approved):

cbrevik (approved):
LGTM 👌🏻

PR #779, opened 2020-12-15
feat: Redesign journey details

### LINKED ISSUE #551: [Feature] Improvements to voiceover on journey details
## Origin
_Links to received feedback, user research or other findings.

Discovered while QA-testing for beta in voiceover mode:
Nearby -> Departure details:
- "Mellomstopp" is missing info about button. Now it reads "10 mellomstopp". 
- Each line with time and leg name is now separate focus frames. Suggestion that each line (for instance time + leg is read together. Makes it easier to move between the different legs of the journey as you only have to move the finger up/down, instead of up, down, left, right.

## Motivation

## Hypotheses and assumptions

## Proposed solution
- "Mellomstopp" must be a touchable element, so it's accessible to the screen reader, and the correct instruction is read out ("10 mellomstopp, knapp, dobbelklikk for å aktivere")
- Each leg should indeed be one accessible element. Consider doing this change along with implementing the new design (as shown below, please link this issue to the correct issue if and when it's made)
Journey

### LINKED ISSUE #748: [Designsync] Update journey details
## Acceptance criteria

- [ ] Details contents can be perceived by using a screen reader
- [ ] Details contents can be operated using a sceeen reader
- [ ] Details contents and functions are accessible for font size up to 200%
- [ ] Changes to journey details and trip details match design (see exceptions below)
- [ ] Elements displayed properly when dark theme is switched on

Exceptions to design:
- Map is left out of the trip details view, as this is described in a separate issue 
- Pagination is left out of the trip details view, as currently not decided what to paginate from the different views (trip details is a shared component opened from both nearby and journey details)
- Date in journey details: Was not shown in figma design at the time of development. Could possibly be kept out for now and described as a separate issue (perhaps even as a "quick win")

## Origin

## Motivation

## Hypotheses and assumptions

## Proposed solution

Figma / App 1.0 / Assistant: https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=963%3A0
Figma / App 1.0 / Departures: https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=963%3A1990
FIgma / Components 1.0 / List-items: https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components-1.0?node-id=547%3A6037
FIgma / Components 1.0 / List-groups: https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components-1.0?node-id=547%3A6287

image

rikkekantega:
___

Closes #748 Designsync journey details
Closes #551 Improved voiceover for journey details

____

#551 New voiceover for journey details depended upon design sync for journey details. 
***Voiceover screen recorder video TBP***

Issue #748 describing design sync for journey details also included design for departure details.

I have excluded the map from departure details (because it is described in separate issue). I also left out pagination, since we possibly need to decide on when and what to paginate. The pagination component with the correct look is created and ready to use.

**Sneak peek** 

Journey details
image

Departure details (map and pagination left out as mentioned above)
image
image


mikaelbr:
I can investigate the negative margin bug after submitting #775 for review.

rikkekantega:
> I can investigate the negative margin bug after submitting #775 for review.

Thanks! Although I just figured it out. :D

src/navigation/index.tsx
```diff
@@ -88,21 +88,23 @@ const NavigationRoot = () => {
               <>
                 <Stack.Screen name="TabNavigator" component={TabNavigator} />
                 <Stack.Screen
-                  name="TripDetailsModal"
-                  component={TripDetailsModal}
+                  name="TripDetailsRoot"
```
cbrevik, line 91: I'm wondering if we could re-think a bit how these are used?

My thinking here is that if these two (`TripDetailsRoot`, `DepartureDetails`) are going to be sub-screens and not modals, they should be navigate-able within the context they are used in (`Assistant` and `Nearby`). In that case, `Nearby` and `Assistant` should probably define their own `Stack`-navigators which these two are used within. 

At the moment they are defined globally in the `NavigationRoot`, and we break certain patterns. They look like full-screen modals, but behave like sub-screens. 

In other words, I think we should show the tab-bar, and not lock the user into a full-screen context which is hard(er) to get out of.

cbrevik (changes requested):

src/navigation/index.tsx
rikkekantega: Sounds very reasonable! :raised_hands:

mikaelbr: In this way would we have to add specific stack.screen entries using the root here several places (e.g. departure, assistant, etc)

src/components/transportation-icon/index.tsx
```diff
@@ -1,40 +1,47 @@
...
   children,
 }) => {
   const styles = useStyle();
-  const {fill, icon} = transportationColor(mode, publicCode);
+  const {t} = useTranslation();
+  const {color} = transportationColor(mode, publicCode);
 
   return (
-    <View style={[styles.circle, {backgroundColor: fill}, circleStyle]}>
+    <View style={styles.iconContainer}>
```
mikaelbr, line 35: Note to self: I've also had to do some changes here, but seems as I can just throw them away. (I made an additional component to avoid having to rewrite and break journey details)

src/components/transportation-icon/index.tsx
```diff
@@ -60,25 +67,17 @@ function InnerIcon({
 
   switch (mode) {
     case 'bus':
-      return (
-        <BusSide accessibilityLabel="Buss" key="bus" {...innerIconProps} />
-      );
+      return <BusSide key="bus" {...innerIconProps} />;
```
mikaelbr, line 70: Does this handle dark mode now?

src/utils/transportation-color.ts
```diff
@@ -17,38 +18,38 @@ export default function transportionColor(
       const code = Number(publicCode);
       if (code < 100)
```
mikaelbr, line 19: Btw. There is a transportation submode on service journey which I belive can say if it is a regional bus or not.

src/theme/colors.ts
```diff
@@ -3,9 +3,9 @@ import {StatusBarProps, TextStyle} from 'react-native';
 
 const backgrounds = {
   light__level0: '#FFFFFF',
-  light__level1: '#F5F5F6',
-  light__level2: '#EBECED',
-  light__level3: '#E1E3E4',
+  light__level1: '#E7E8E9',
```
mikaelbr, line 6: Ah gray_50 I see

mikaelbr:
There's an issue with padding on iOS (image), but I figure this will be fixed when moving away from being a modal

image

src/components/transportation-icon/index.tsx
mikaelbr: Guessing the colors just work. So ignore comment

src/screens/Assistant/index.tsx
```diff
@@ -328,12 +328,10 @@ const Assistant: React.FC<Props> = ({
...
   const onPressed = useCallback(
-    (tripPattern) =>
-      navigation.navigate('TripDetailsModal', {
-        from: from!,
-        to: to!,
-        tripPatternId: tripPattern.id!,
-        tripPattern: tripPattern,
+    (tripPatterns, index) =>
+      navigation.navigate('TripDetailsRoot', {
+        initialTripPatterns: tripPatterns,
```
mikaelbr, line 333: This makes it harder to do deep linking at some point as you would require details view to get passed entire search set. Might be a different way to handle this by having two different types of details showing depending on deep linking.

mikaelbr (approved):
I don't really have a comment except for the thing @cbrevik pointed out and good work! There are some comments in the review but they are mostly ramblings and me thinking out loud.

I've tested it locally and clicked around and can't find any obvious issues.

src/theme/colors.ts
rikkekantega: Ah, haha, I "solved" the problem :flushed: :joy:

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <View style={[styles.container, style]}>
+      <Button
+        type="compact"
+        mode="tertiary"
+        disabled={!hasPrevious}
+        iconPosition="left"
+        icon={ArrowLeft}
+        onPress={() => onNavigate(--page)}
```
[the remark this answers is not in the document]
rikkekantega: Good point!

src/screens/Assistant/index.tsx
rikkekantega: Actually didn't consider that, glad you're pointing that out. Don't want to make changes that may hinder future deep linking. Would it be sufficient to alter the details component to accept a singular trip pattern as an alternative to a set of trip patterns? I'm assuming that the only change (more or less) would be skipping pagination

src/theme/colors.ts
rikkekantega: I'll revert this one :point_up_2: Actually not intentional, I think I may have misinterpreted this color as being the "new" gray_100 value at the time the change happened..

src/screens/Assistant/index.tsx
mikaelbr: I think taking a `tripPatternId` in addition and having `initialTripPatterns` and `tripPatternIndex` optional. So using `tripPatternId` to get data but prefill with initialTripPatterns if they exist. Also probably not showing pagination when tripPatternIndex is not defined.

rikkekantega: Sounds very reasonable!

src/utils/transportation-color.ts
rikkekantega: Ah - Will try and figure out if there's any good reason to why the code check was used instead of that initially  :monocle_face:

mikaelbr: I believe we didn't know better. 😬  cc @cbrevik

cbrevik: As far as I know, it actually didn't correctly specify correctly if it was a regional bus or not. If it does now, great 🎉. I haven't checked in a while.

mikaelbr: To be fair: I haven't actually checked the data. Let's just make another issue for this

rikkekantega: How would one know for sure? Do you know of some specific lines that would appear as local when regional, or the other way around?

src/navigation/index.tsx
cbrevik: My thinking was that Assistant gets its own Stack-Navigator. With two screen: the main `Assistant`-screen, and `TripDetailsRoot` as a screen. And Nearby gets two, the main one and  `DepartureDetails`. So specific screens per newly created `Navigator` yeah.

src/utils/transportation-color.ts
cbrevik: Don't remember any specific lines no, just checked trips like Trondheim - Stjørdal. A quick search now on the same trip seems like it has the correct `transportSubmode` under service-journey, so seems like they've fixed it 🎉 Earlier all of them said `localBus`.

```
		"serviceJourney": {
				"id": "ATB:ServiceJourney:310_200610136946289_30_200127114109271",
				"journeyPattern": {
					"line": {
						"id": "ATB:Line:200610136922955_310_200127114109271",
						"name": "Stjørdal - Trondheim - Orkanger",
						"notices": [],
						"publicCode": "310",
						"transportMode": "bus",
						"transportSubmode": "regionalBus"
					},
					"notices": []
				},
				"notices": [],
				"publicCode": "",
				"privateCode": "",
				"transportSubmode": "regionalBus"
			},
			"situations": []
		},
```

rikkekantega: Nice :+1:

src/navigation/index.tsx
rikkekantega: Assistant has been provided its own stack navigator, and tripdetailsroot screen is moved to within Assistant navigator. :+1: 
DepartureDetails already exist as a screen in TripDetailsRoot stack navigator, and will remain there.
I am unsure whether to alter "Nearby" right now, or to do it in a separate branch after both mine Mikaels changes have been merged.

rikkekantega:
> There's an issue with padding on iOS (image), but I figure this will be fixed when moving away from being a modal
> 
> image

Would you mind to check if moving the screen has resolved this? =)

src/utils/transportation-color.ts
rikkekantega: I've changed so transportSubmode is now used instead of publicCode. Updated all usages as well.

cbrevik (approved):
Awesome! Great job, this is gonna be awesome to use 👩🏻‍🎤 🎸

PR #783, opened 2020-12-17
feat: Adds favorite departures (closes #680)

### LINKED ISSUE #680: [Feature] Add favorite departures
## Origin

See issue #420 

## Motivation

Allow the user to save a frequently used departure.

## Hypotheses and assumptions

Many users use the same line every day from the same stop, and want to easily find these departures.

## Proposed solution

## Acceptance Criteria
- [ ] Able to add favorite.
- [ ] Favorites are persisted after app restart.
- [ ] I can remove favorite by toggling star.
- [ ] Able to filter on favorites nearby, by address search and from different quays.
- [ ] Able to add favorites from all types of transportation modes and quays.
- [ ] As a user of screen reader I can navigate and easily understand favorites and how to toggle between.
- [ ] Works as expected in dark mode.
- [ ] (Is translatable to English)
- [ ] Does not drain battery
- [ ] Works in up to 200% font size

Adding:
- "Star" from departures view
image

Using:
- Filter toggle on departures screen
- Toggle only shown if current results contain a favorite
image

Figma / App 1.0 / Favorite https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=8%3A1529
Figma / App 1.0 / Departure https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=4%3A982

mikaelbr:
**This is dependent on production deployment of the BFF. It is also dependent on #775 which should be merged first**

This adds favorite departures from the departure screen. Only scoped to #680 where we can add favorites directly from results. Currently, this is the only way to add favorite departures.

For Code Review, see Acceptance Criteria on #680.

## Examples

image

image

### DIFF
--- a/src/components/sections/action-item.tsx
+++ b/src/components/sections/action-item.tsx
@@ -28,11 +28,15 @@ export default function ActionItem({
 
   if (mode === 'toggle') {
     return (
-      <View style={[style.spaceBetween, topContainer]} {...accessibility}>
+      <View style={[style.spaceBetween, topContainer]}>
         <ThemeText type="body" style={contentContainer}>
           {text}
         </ThemeText>
-        <Switch value={checked} onValueChange={(v) => onPress?.(v)} />
+        <Switch
+          value={checked}
+          onValueChange={(v) => onPress?.(v)}
+          {...accessibility}
+        />
       </View>
     );
   }
--- a/src/components/sections/action-item.tsx
+++ b/src/components/sections/action-item.tsx
@@ -29,12 +29,13 @@ export default function ActionItem({
   if (mode === 'toggle') {
     return (
       <View style={[style.spaceBetween, topContainer]}>
-        <ThemeText type="body" style={contentContainer}>
+        <ThemeText accessible={false} type="body" style={contentContainer}>
           {text}
         </ThemeText>
         <Switch
           value={checked}
           onValueChange={(v) => onPress?.(v)}
+          accessibilityLabel={text}
           {...accessibility}
         />
       </View>
--- a/src/screens/Nearby/NearbyResults.tsx
+++ b/src/screens/Nearby/NearbyResults.tsx
@@ -3,6 +3,7 @@ import sortBy from 'lodash.sortby';
 import React, {useEffect, useState} from 'react';
 import {ActivityIndicator, View} from 'react-native';
 import {QuayGroup, StopPlaceGroup} from '../../api/departures/types';
+import ScreenReaderAnnouncement from '../../components/screen-reader-announcement';
 import {ActionItem} from '../../components/sections';
 import ThemeText from '../../components/text';
 import {Location} from '../../favorites/types';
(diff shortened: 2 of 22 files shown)


rikkekantega:
Seems like using the toggler back and forth to fast (before result has loaded) triggers an error
image

I also managed to get an error when removing favorite. Trying to reproduce to describe steps.

rikkekantega:
Okay, so I'm triggering quite a few errors when clicking around. I'm trying to find consistent reproduction steps. But I expect you would be able to trigger some by combining toggling favorite and adding/removing favorites back and forth.

mikaelbr:
Hm. Maybe there are some regressions after my git management also 😬  I'm checking it out

src/screens/Nearby/NearbyResults.tsx
```diff
@@ -1,93 +1,69 @@
...
+              ? t(NearbyTexts.results.messages.initial)
+              : t(NearbyTexts.results.messages.initial)}
+          </ThemeText>
         </MessageBox>
       </View>
     );
   }
 
-  if (hasNoQuays(departures)) {
+  if (hasNoQuaysWithDepartures(departures)) {
```
rikkekantega, line 53: Wonder if loading state should be considered here?

mikaelbr: Sure!

src/components/sections/action-item.tsx
```diff
@@ -13,20 +13,22 @@ export type ActionItemProps = SectionItem<{
...
+  accessibility,
   ...props
 }: ActionItemProps) {
   const {contentContainer, topContainer} = useSectionItem(props);
   const style = useSectionStyle();
 
   if (mode === 'toggle') {
     return (
-      <View style={[style.baseItem, style.spaceBetween, topContainer]}>
+      <View style={[style.spaceBetween, topContainer]} {...accessibility}>
```
rikkekantega, line 31: Generally, corresponding label should respond to click action by toggling the switch, 
and I wonder if we will make the function more understandable by doing so as well. If the entire containing element has role switch, and an onPress action. 
Right now, for the usage in favorites toggler, talkback reads "Vis favoritter" (no action provided), and then by navigating to the next item, the switch is read out so fast that I cant interpret what is said.  (only the following hint about how to interact, which I think may be Android specific?) Probably default behaviour, but I'm not sure if it's understandable enough in our case

src/translations/screens/Nearby.ts
```diff
@@ -39,20 +42,63 @@ const NearbyTexts = {
...
+    lines: {
+      a11y: {
+        line: _(`Linje:`),
+      },
+      lineNameAccessibilityHint: _(
+        'Aktiver for detaljer om avgang og oversikt over kommende avganger.',
+      ),
+      favorite: {
+        addFavorite: (name: string, place: string) =>
+          _(`Legg til favorittavgang: ${name} fra ${place}`),
```
rikkekantega, line 83: Really nice!!! :raised_hands:

src/components/sections/action-item.tsx
mikaelbr: > Generally, corresponding label should respond to click action by toggling the switch,

Maybe there are different practices on Android / iOS here? I know checkboxes and labels are a thing on the Web but typically iOS pattern is to _not_ have the label (entire line) be pressable to toggle Switches.

mikaelbr: Should we possibly do text as inaccessible and adding an accessible label equal to the text on the Switch itself?

mikaelbr:
Errors you (@rikkekantega ) saw was caused by layout animations 😢  Deactivated on Android for now. Too bad, but can look into how we can get it to work later.

src/components/sections/action-item.tsx
rikkekantega: Okay, seems like clickable label for switch is an Android-only pattern then.

rikkekantega: Your suggestion on labeling the switch: Yes, I think that would make more sense =)

mikaelbr: Hm. Then we should probably do so on android. I can look into it

mikaelbr: I did some accessibility improvements. But not the clickable thing. I think we should work out a good way to do both patterns on the platforms, but tempted to say we should do that as a separate issue as it affects many other things aswell

rikkekantega:
image
Label/star will at least need some space between to be accessible for people with reduced precision (and to people with clubbed thumbs, like me :grimacing: )

mikaelbr:
Have you checked the latest version, @rikkekantega ? I bumped the hitslop some commits ago

rikkekantega:
> Have you checked the latest version, @rikkekantega ? I bumped the hitslop some commits ago

Yep, I just recently pulled, hitslop is set to 14, 8

mikaelbr:
Added some additional spacing.

rikkekantega:
200%
Long line names push favorite icon out of view
image

Switch is hard to hit because it's pushed or too close to text. Label text should probably wrap sooner / have some more margin
image

Quay name should also wrap to avoid colliding/pushing distance
image

mikaelbr:
Good point! Lets do it as a part of the qa flow

PR #784, opened 2020-12-17
feat: Adds ability to remove favorite departure

mikaelbr:
**This is dependent on production deployment of the BFF. It is also dependent on #783 which should be merged first**

This could be merged into #783 as it is a small change

I think this should be a separate change from adding favorite through profile screen. Adding through wizard is a larger change which technically is nice to have (though very nice, don't get me wrong). But I think we can try to QA this and the other dependant changes and deploy that before implementing new add favorite flow.

## Examples

image

image

image


gorandalum (approved):
LGTM :+1:

PR #786, opened 2020-12-18
fix: Add mode "coach" to transport color/icon

rikkekantega:
Noticed when running through favorite branch locally that we weren't handling transport mode "coach". Seems like coach mode is a type of bus. Entur also display it as a bus. I've added this to existing icon/color functions.

image

### DIFF
--- a/src/components/transportation-icon/index.tsx
+++ b/src/components/transportation-icon/index.tsx
@@ -67,6 +67,7 @@ function InnerIcon({
 
   switch (mode) {
     case 'bus':
+    case 'coach':
       return <BusSide key="bus" {...innerIconProps} />;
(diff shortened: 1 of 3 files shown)


mikaelbr (approved):

mikaelbr:
Are the switch cases here now exhaustive covering all cases in the type?

rikkekantega:
> Are the switch cases here now exhaustive covering all cases in the type?

By checking the possible type values, the answer is no.
We are missing bicycle, car, cable, lift, metro, and funicular. All of which does not seem likely for our usage right now, but we sure should handle some of them in the future. Should we perhaps create an issue on adding icons and names for these?

PR #789, opened 2020-12-18
fix: Counter-input accessibility bug

gorandalum:
Even though the count was 1, the accessibility elements for the decrease
button was hidden.


rikkekantega (approved):
:+1:

PR #790, opened 2020-12-18
fix: Remove chat icon since this will crash the ErrorView

cbrevik:
Crashes because `useChatIcon` uses remote-config, which is not available in the `ErrorBoundary`

### DIFF
--- a/src/error-boundary/ErrorView.tsx
+++ b/src/error-boundary/ErrorView.tsx
@@ -18,13 +18,12 @@ type ErrorProps = {
 
 const ErrorView: React.FC<ErrorProps> = ({onRestartApp, errorCode}) => {
   const styles = useStyles();
-  const chatIcon = useChatIcon();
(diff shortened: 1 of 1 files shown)


src/error-boundary/ErrorView.tsx
```diff
@@ -37,7 +36,7 @@ const ErrorView: React.FC<ErrorProps> = ({onRestartApp, errorCode}) => {
             Appen er i læringsmodus, og slike krasj er akkurat det vi trenger
             for å gjøre den enda mer robust.
             {'\n'}
-            Bruk gjerne chat-funskjonen vår til å fortelle oss hva som gikk
+            Bruk gjerne chat-funksjonen vår til å fortelle oss hva som gikk
```
mikaelbr, line 39: 😆

mikaelbr (approved):

PR #791, opened 2020-12-18
chore: Sort tariff zones for easier selection

gorandalum:
This will also make A the default tariff zone as it is the first tariff
zone in the array.


cbrevik (approved):
Awesome 🙌🏻

PR #792, opened 2020-12-18
feat: Search for venues when selecting tariff zone

gorandalum:
Screenshot 2020-12-18 at 13 42 44
Screenshot 2020-12-18 at 13 43 09


cbrevik (approved):
LGTM 💯 I think as discussed earlier that at some point we should look into how these screens relate to `location-search`. There's a lot of the same functionality and components, with some important differences. Anyway, not blocking for this PR 😇

PR #794, opened 2020-12-18
feat: Improve accessibility for Travellers screen

gorandalum:
Improve accessibility and use internationalization for Travellers
screen.

### DIFF
--- a/src/translations/components/Section.ts
+++ b/src/translations/components/Section.ts
@@ -7,6 +7,18 @@ const SectionTexts = {
     placeholder: _('Søk etter adresse eller sted', 'Search for a place'),
     a11yValue: (currentLocation: string) => _(`${currentLocation} er valgt.`),
   },
+  counterInput: {
+    decreaseButton: {
+      a11yLabel: _('Minsk antall'),
+      a11yHint: (text: string, count: number) =>
+        _(`Aktivér for å minske antall ${text} til ${count - 1}`),
+    },
+    increaseButton: {
+      a11yLabel: _('Øk antall'),
+      a11yHint: (text: string, count: number) =>
+        _(`Aktivér for å øke antall ${text} til ${count + 1}`),
+    },
+  },
   favoriteDeparture: {
     from: _('Fra'),
   },
--- a/src/translations/screens/subscreens/Travellers.ts
+++ b/src/translations/screens/subscreens/Travellers.ts
@@ -1,0 +1,36 @@
+import {translation as _} from '../../commons';
+import {UserProfileWithCount} from '../../../screens/Ticketing/Purchase/Travellers/use-user-count-state';
+
+const TravellersTexts = {
+  header: {
+    title: _('Velg sone'),
+    leftButton: {
+      text: _('Avbryt'),
+      a11yLabel: _('Avbryt kjøpsprosessen'),
+    },
+  },
+  errorMessageBox: {
+    title: _('Det oppstod en feil'),
+    failedOfferSearch: _('Klarte ikke å søke opp pris'),
+    failedReservation: _('Klarte ikke å reservere billett'),
+  },
+  tariffZones: {
+    a11yHint: _('Aktivér for å velge soner'),
+  },
+  travellerCounter: {
+    text: (u: UserProfileWithCount) => _(u.name.value),
+  },
+  totalCost: {
+    title: _('Totalt'),
+    label: _('Inkl. 6% mva'),
+  },
+  paymentButtonVipps: {
+    text: _('Betal med Vipps'),
+    a11yHint: _('Aktivér for å betale billett med Vipps'),
+  },
+  paymentButtonCard: {
+    text: _('Betal med bankkort'),
+    a11yHint: _('Aktivér for å betale billett med bankkort'),
+  },
+};
+export default TravellersTexts;
--- a/src/components/sections/counter-input.tsx
+++ b/src/components/sections/counter-input.tsx
@@ -6,6 +6,7 @@ import insets from '../../utils/insets';
 import ThemeIcon from '../theme-icon';
 import {Add, Remove} from '../../assets/svg/icons/actions';
 import {StyleSheet} from '../../theme';
+import {SectionTexts, useTranslation} from '../../translations';
 
 export type CounterInputProps = SectionItem<{
   text: string;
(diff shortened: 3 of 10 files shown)


mikaelbr (approved):
Looks good. With labels and hints, it's often important with a period (`.`) as without it the reading is very unnatural. Doesn't seem to be necessary here now, but just experienced this previously and thought I'd share my experience. 😅

src/translations/components/Section.ts
```diff
@@ -7,6 +7,18 @@ const SectionTexts = {
     placeholder: _('Søk etter adresse eller sted', 'Search for a place'),
     a11yValue: (currentLocation: string) => _(`${currentLocation} er valgt.`),
   },
+  counterInput: {
+    decreaseButton: {
+      a11yLabel: _('Minsk antall'),
+      a11yHint: (text: string, count: number) =>
+        _(`Aktivér for å minske antall ${text} til ${count - 1}`),
```
rikkekantega, line 14: Should we perhaps consider decreasing the count value from where this is called, instead of in the translation text? :thinking:  The reason I'm asking is that we're going to have non-developers working in these files. We could make their job easier by using "preprocessed" values within translations

src/translations/screens/subscreens/Travellers.ts
```diff
@@ -0,0 +1,36 @@
...
+  errorMessageBox: {
+    title: _('Det oppstod en feil'),
+    failedOfferSearch: _('Klarte ikke å søke opp pris'),
+    failedReservation: _('Klarte ikke å reservere billett'),
+  },
+  tariffZones: {
+    a11yHint: _('Aktivér for å velge soner'),
+  },
+  travellerCounter: {
+    text: (u: UserProfileWithCount) => _(u.name.value),
```
rikkekantega, line 21: How would this get translated? Is `u.name` already a translated value? I might not entirely understand how the `LanguageAndText` type is being used, but by the name it would seem like translations might already be defined elsewhere for these properties?

rikkekantega (commented):
Great work on improving accessibility :star_struck: 

What Mikael says is true, but just to complicate that, I have to mention that adding a period isn't always sufficient. :woozy_face:  Sometimes we need to manually append newline to force natural screen reading pauses. Often needed in between label and hint, and within longer descriptions in a singular accessible. Possible that voiceover on iOS and TalkBack on Android read/obey period a little different. Newline seem to work for both. But this adds to several other reading issues, sometimes rephrasing is even needed to make contents perceivable. Best way to know is to check each case using a screen reader. :grimacing: 

I just had some minor code comments/questions

src/translations/components/Section.ts
gorandalum: Yeah this can be done. I contemplated both alternatives when implementing and went for the general rule that the component should just send in it's current state, without having knowledge about the text being presented, and it is the texts-file responsibility to create the translation text, and the values for it, based on the current state. When deciding this I did not know that non-developers were going to work on these files. It is totally ok for me that we have as a general rule that all values come preprocessed to the texts-files :+1:

Generally I put quote a bit of logic in the texts-files. Feel free to comment other places where you think the logic would be better if it was outside of the texts-file.

src/translations/screens/subscreens/Travellers.ts
gorandalum: This is NeTeX `MultilingualString`. It is an object { lang: string, value: string } where lang is the language code. Generally the `name` field has the value for the default language, and another field `alternativeNames` has the additional languages.

So for `UserProfile` it would look something like this:
```
{ 
  ...
  name: { lang: "nob", value: "Voksen" },
  additionalNames: [
    { lang: "nno", value: "Vaksen" },
    { lang: "eng", value: "Adult" }
  ]
```  

So for english translation we would have to use something like `u.additionalNames.filter(n => n.lang === "eng")?.value`. Probably in a helper method as this has to be done several places.

gorandalum:
Regarding the discussion on period/newline, is there some general guidelines we can state? Like:

- Always use `screenReaderPause` instead of a period.
- Always use `screenReaderPause` at the end of `accessabilityLabel`.
- Other?

This does not limit the usage of `screenReaderPause` to these situasjon, since there are situations that demand tailored solutions. Like within longer descriptions, as mentioned.

src/translations/components/Section.ts
rikkekantega: I would argue keeping logic out of translation strings as a general rule, to make translation work easier without the need for developer support.

src/translations/components/Section.ts
```diff
@@ -7,6 +7,18 @@ const SectionTexts = {
...
+    },
+    increaseButton: {
+      a11yLabel: _('Øk antall'),
+      a11yHint: (text: string, count: number) =>
+        _(`Aktivér for å øke antall ${text} til ${count + 1}`),
+    },
```
rikkekantega, line 20: Same here

src/translations/screens/subscreens/Travellers.ts
rikkekantega: Yes, I see. I suspected something like that. Perhaps this kind of logic also could be placed elsewhere? I'm just asking, though, I think we should agree on a common practice on this. My main argument is keeping the text files as clean as possible for translators

rikkekantega:
> Regarding the discussion on period/newline, is there some general guidelines we can state? Like:
> 
> * Always use `screenReaderPause` instead of a period.
> * Always use `screenReaderPause` at the end of `accessabilityLabel`.
> * Other?
> 
> This does not limit the usage of `screenReaderPause` to these situasjon, since there are situations that demand tailored solutions. Like within longer descriptions, as mentioned.



> Regarding the discussion on period/newline, is there some general guidelines we can state? Like:
> 
> * Always use `screenReaderPause` instead of a period.
> * Always use `screenReaderPause` at the end of `accessabilityLabel`.
> * Other?
> 
> This does not limit the usage of `screenReaderPause` to these situasjon, since there are situations that demand tailored solutions. Like within longer descriptions, as mentioned.

Yes, I think we should probably try to make som general guidelines here. As mentioned, in some cases we might need to do some manual adjustments.
I'd like to check how usage of both period and newline affect screen reading on both platforms. Generally I'd argue always using punctuation marks where appropriate within text. Newline/screen reader pause at the end of an accessibility label. And manually adjust for special cases.

src/translations/components/Section.ts
gorandalum: I'll rewrite a couple places to keep the logic in the container instead of in the texts file, and then we can compare. It's almost like an additional layer is needed. One layer for keeping the containers separated from text-logic, and one layer for keeping the text-logic separated from the actual texts. Of course this is probably not worth the investment, just a thought :)

rikkekantega: Might be worth consideration, seems none of the solutions are ideal here.

rikkekantega (approved):
LGTM :rocket: :+1:

PR #796, opened 2020-12-21
feat: Language preferences in profile settings (behind feature flag)

### LINKED ISSUE #780: [Feature] Language preference selection in "Mitt AtB"
## Origin

Default app criteria, English and Norwegian language support

## Motivation

We only support English and Norwegian, and decide on which to use based on system language settings (phone settings). In cases where system language is set to something else than the supported languages, we choose a default on the users behalf. The user should be able to override this default language, so he/she doesn't need to change the language of their entire system to change preference for the app. 

## Hypotheses and assumptions

- The user may be more familiar with the one or other of the supported languages, without using it as a system language on his/her phone
- Easily accessed language switch in app may enable the possibility for foreigners to practice their norwegian skills
- Learning mobility related terminology may be easier having an easily accessed translation in app

## Proposed solution

A switch for language in "Mitt AtB" screen, that explicitly sets system language settings. Option to use system settings if desired.

rikkekantega:
Defaults to "default language", currently set to "Norsk bokmål".

Using phone settings is optional - App then gets system locale, and decides on what will be considered the best app language.

User can choose app language explicitly from "Mitt AtB".

image

closes #780

### DIFF
--- a/src/translations/LanguageContext.tsx
+++ b/src/translations/LanguageContext.tsx
@@ -41,7 +41,7 @@ function useLanguage() {
         RNLocalize.removeEventListener('change', onChange);
       };
     }
-  }, [shouldUseUserPreferenced]);
+  }, [useSystemLanguage, userPreferencedLanguage]);
 
   useEffect(() => {
     if (!shouldUseUserPreferenced) {
--- a/src/translations/LanguageContext.tsx
+++ b/src/translations/LanguageContext.tsx
@@ -28,16 +28,10 @@ function useLanguage() {
   const shouldUseUserPreferenced =
     !useSystemLanguage && !!userPreferencedLanguage;
 
-  const getAndSetPreferencedLanguage = () => {
+  useEffect(() => {
     if (shouldUseUserPreferenced) {
       const newLanguage = getAsAppLanguage(userPreferencedLanguage);
       setCurrentLanguage(newLanguage);
-    }
-  };
-
-  useEffect(() => {
-    if (shouldUseUserPreferenced) {
-      getAndSetPreferencedLanguage();
     } else {
       const onChange = () => {
         setLocale(preferredLocale);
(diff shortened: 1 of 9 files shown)


src/translations/LanguageContext.tsx
```diff
@@ -19,35 +20,62 @@ const AppLanguageProvider: React.FC = ({children}) => {
...
+    } else {
+      const onChange = () => {
+        setLocale(preferredLocale);
+      };
+      RNLocalize.addEventListener('change', onChange);
+      return () => {
+        RNLocalize.removeEventListener('change', onChange);
+      };
+    }
   }, []);
```
gorandalum, line 50: Shouldn't `shouldUseUserPreferenced` be in the dependency list here? Probably a corner case, but from my understanding the localization change listener is now only added if `shouldUseUserPreferenced` is false at initial mounting, and not if the user changes their setting to using the system language.

Adding `shouldUseUserPreferenced` will also make the `useEffect` on line 61 redundant.

gorandalum (approved):
Looks very good! And it feels good to finally review a change that is not 50 files, like most of the PRs the last two weeks (including mine) =)

src/translations/LanguageContext.tsx
rikkekantega: Of course, not sure what I was even thinking. Good catch!

mikaelbr: I think maybe it's time to include dependency list eslint rule. I've also had some issues remembering all deps 😅

mikaelbr:
A question that's maybe not directly related to this. We now feature toggle language selection, but the language is also selected by phone settings. Will people with `en` as default now see some English translations?

rikkekantega:
> A question that's maybe not directly related to this. We now feature toggle language selection, but the language is also selected by phone settings. Will people with `en` as default now see some English translations?

The AppLanguageProvider considers the same feature toggle. :)

PR #799, opened 2020-12-22
refactor: Internalization text in text-file facade

gorandalum:
A commit showing how the logic would be distributed with text in a facade in the texts-files. The main texts in the text-files are still cleanly written, while the text logic is added to the Facade. This is only a PoC of setup where the text-logic is contained somewhere "between" the React component and texts-file.


gorandalum:
Closes this now. May later look at solutions like this if we experience that React Components gets to bloated with text logic.

PR #816, opened 2021-01-05
fix: bottom padding in favorite lists

mikaelbr:
There's a bug when showing many favorites where the bottom padding on safe area sets bottom margin when we have bottom tabs. This deactivates bottom edge from safe area to avoid double padding (both in tab bar and content view).


rikkekantega (approved):
LGTM

PR #818, opened 2021-01-05
feat: Fetch reference data from RemoteConfig

gorandalum:
Instead of retrieving reference data by fetching them with http requests
from the referene data services, the reference data is now fetched from
the RemoteConfig. In addition default reference data is bundled with the
app, as backup for situations where the app is unable to retrieve it
from RemoteConfig. Using RemoteConfig also gives out-of-the-box caching
for the user.


torfjor:
I've populated the cache from the export module for testing now.

gorandalum:
@torfjor Is it an easy task to add the `userTypeString` on `user_profiles` also in RemoteConfig? If not we need to do some rewriting in the app code.

torfjor:
> @torfjor Is it an easy task to add the `userTypeString` on `user_profiles` also in RemoteConfig? If not we need to do some rewriting in the app code.

Should be fixed now.

cbrevik (approved):
LGTM! This will definitely make the ticket buy process more robust 👍🏻

PR #826, opened 2021-01-08
fix: Use new offers endpoint which allow two zones

gorandalum:
The old search offers endpoint was not working properly when the travel
was through more than one zone.

### DIFF
--- a/src/api/fareContracts.ts
+++ b/src/api/fareContracts.ts
@@ -21,7 +21,7 @@ export async function search(
   params: OfferSearchParams,
   opts?: AxiosRequestConfig,
 ): Promise<Offer[]> {
-  const url = 'ticket/v1/search';
+  const url = 'ticket/v1/search/zones';
   const response = await client.post<Offer[]>(url, params, opts);
 
   return response.data;


cbrevik (approved):
💯

mikaelbr (approved):
The changeset is maybe on the larger size for one PR and should be split into to, but OK for now.

src/api/fareContracts.ts
```diff
@@ -21,7 +21,7 @@ export async function search(
   params: OfferSearchParams,
   opts?: AxiosRequestConfig,
 ): Promise<Offer[]> {
-  const url = 'ticket/v1/search';
+  const url = 'ticket/v1/search/zones';
```
cbrevik, line 24: I would have preferred:
```suggestion
  const url = 'ticket/v1/search/' + 'z' + 'o' + 'n' + 'e' + 's';
```

mikaelbr: That is what I call _adding value_!

gorandalum:
Feedback noted, will take it to heart and improve future PRs.

mikaelbr:
But did you see my pun? https://github.com/AtB-AS/mittatb-app/pull/826#discussion_r553904079

gorandalum:
Can't say I did. Is the pun regarding adding additions? 😅

mikaelbr:
Strictly speaking, it should be `adding values`, but come on. I see that having to explain it removes some of the punch but it was a play on irony (suggested change didn't add any value at all) and that it is literally `adding values`.

gorandalum:
It was great @mikaelbr! 🥇

mikaelbr:
yeaah

PR #830, opened 2021-01-12
feat: Simple tariff zone map selection

gorandalum:
Simple implementation, with not-so-pretty zone labels and non-optimal selection method. Can be merged mainly for enabling @pettersuul to get a feel for it while improving the design.

### DIFF
--- a/src/components/sections/section.tsx
+++ b/src/components/sections/section.tsx
@@ -5,6 +5,7 @@ import {ContainerSizingType, SectionItemProps} from './section-utils';
 
 export type SectionProps = PropsWithChildren<{
   withPadding?: boolean;
+  withFullPadding?: boolean;
   withTopPadding?: boolean;
   withBottomPadding?: boolean;
   type?: ContainerSizingType;
@@ -13,6 +14,7 @@ export type SectionProps = PropsWithChildren<{
 export default function SectionGroup({
   children,
   withPadding = false,
+  withFullPadding = false,
   withTopPadding = false,
   withBottomPadding = false,
   type = 'block',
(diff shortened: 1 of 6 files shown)


src/components/sections/section.tsx
```diff
@@ -5,6 +5,7 @@ import {ContainerSizingType, SectionItemProps} from './section-utils';
 
 export type SectionProps = PropsWithChildren<{
   withPadding?: boolean;
+  withFullPadding?: boolean;
```
mikaelbr, line 8: I think we should rewrite this to be `padded: 'full' | 'horizontal' | 'horizontal-bottom' | 'bottom' | 'top'` (etc. matching all cases)

src/components/sections/section.tsx
```diff
@@ -78,6 +81,9 @@ const useInputGroupStyle = StyleSheet.createThemeHook((theme: Theme) => ({
     marginHorizontal: theme.spacings.medium,
     marginBottom: theme.spacings.small,
   },
+  container__fullPadded: {
+    margin: theme.spacings.medium,
```
cbrevik, line 85: Will this be correct? topPadded + bottomPadded are `theme.spacings.large`. Also instead of creating its own padding-style, could maybe re-use the other two?

cbrevik (approved):
LGTM! Just one question. Also I think we need to look take a session to look into how these components are split up. I get a feeling we're recreating even more of existing functionality in LocationSearch/MapSelection.

mikaelbr (changes requested):

mikaelbr:
Regarding splitting up etc. Check #832 where I separate SearchLocation from the content to make it reusable without a screen

src/components/sections/section.tsx
gorandalum: I couldn't find a perfect solution here. Using padded + topPadded made the top padding too large as it was `theme.spacings.large`. I can make the change @mikaelbr suggested below. But maybe in it's own PR as it will probably touch lots of files?

Issue #675, opened 2020-11-10
[Feature] New datetime view


pettersuul:
## Origin
_Links to received feedback, user research or other findings._

## Motivation
_A short description of what user needs or business goals this feature will solve._

Make selecting a time and date more accessible and consistent across platforms.

## Hypotheses and assumptions
_A list of hypotheses and assumptions we have made about the user or the proposed solution._

Current method of datetime selection is not accessible on android (also not great on iOS). By placing this on its own page it will be simpler to use native pickers.

## Proposed solution
_A coarse description of a proposed solution, that may include wireframes and graphic design._

During a journey-search, the user may press a button (default: "Avreise nå (##:##)") to open a view/subview which contains controls to select a departure or arrival time.

Figma / App 1.0 / Action / Datetime https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=1818%3A491

![image](https://user-images.githubusercontent.com/56530785/98686911-6d773800-2369-11eb-84dd-49a639fed4ea.png)

rikkekantega:
Waiting on button design sync to complete this

PR #831, opened 2021-01-12
feat: New datetime view

rikkekantega:
This PR rewrites the date input screen from Assistant to be subview instead of bottom sheet. This improves accessibility and gives more room to explore better ways to set date inputs.

The PR introduces new generic section inputs: Date and Time. These are implemented as native components and works differently on Android, iOS post 13.4, and iOS pre 13.4. See screenshots further down. To support different native modules here the dark mode needed to be set on the UIWindow on iOS.

### Changed to native dark mode on iOS 
This prompted a rewrite of theming and dark-mode preferences. I've tested different OS-es and permutations of override/non-override mode. Seemingly got it working but there might be some edge-cases with opening/closing window, changing state etc. There are two states here (native vs JS) that need to be synchronized. Can be some issues it's hard to "synchronize upstream" (to native) as JS is dependent on values provided from native etc.

Closing #675

Regrettably large change, but it includes changes:

- Rewrites button input to not have empty icon container when no icon is set. (fix some layout bugs on button)
- Rewrites Date Input from bottom sheet to sub-screen of Assistant Screen.
- iOS: Adds native module for setting Dark Mode for native components (UIDateInput).
- Adds Section Items Date and Time. Also adding dependency `RNDateTimePicker`/`@react-native-community/datetimepicker` for wrapping native components.

### Known issues

- There is an issue with the native component UIDateInput for iOS where auto-layout (flex) causes the width to collapse. The current workaround is to set hard-coded width. Not ideal, but works for now.

### Todo

- [x] Verify everything with screen readers. 

FYI: Work started by rikkekantega but finished by @mikaelbr . Questions should be asked to @mikaelbr  

## Examples

Screenshot 2021-02-23 at 13 36 05
Screenshot 2021-02-23 at 13 35 56

Screenshot 2021-02-23 at 15 30 20
Screenshot 2021-02-23 at 13 36 14
Screenshot 2021-02-25 at 10 49 55


src/components/sections/date-input/index.android.tsx
```diff
@@ -0,0 +1,39 @@
...
+      />
+      {show && (
+        <RNDateTimePicker
+          value={parseISO(value)}
+          mode="date"
+          locale={locale}
+          textColor={theme.text.colors.primary}
+          onChange={(_, date) => {
+            onChange(dateToDateString(date));
+            setShow(false);
```
gorandalum, line 33: I feel this removal of the DateTimePicker on change is not optimal. When trying to select my wanted date/time I need to open the DateTimePicker multiple times.

src/screens/Assistant/journey-date-picker/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+import {AssistantParams} from '..';
+
+export type DateTimePickerParams = {
+  searchTime: SearchTime;
+  callerRouteName: string;
+  callerRouteParam: string;
+};
+
+export type DateTimeNavigationProp = NavigationProp<AssistantParams>;
+export type DateTimeRouteProp = RouteProp<AssistantParams, 'DateTimePicker'>;
```
gorandalum, line 30: I get this in the console, even though I can not see by just reviewing the code what is causing it. 

>[Wed Feb 24 2021 11:10:57.479]  WARN     Non-serializable values were found in the navigation state. Check:
>
>TabNavigator > Assistant > DateTimePicker > params.onChange (Function)
>
>This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you can use 'navigation.setOptions' instead. See https://reactnavigation.org/docs/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state for more details.

ios/ChangeNativeBridge.m
```diff
@@ -0,0 +1,35 @@
...
+      
+      AppDelegate *appDelegate = (AppDelegate *)[UIApplication sharedApplication].delegate;
+      
+      if ([mode isEqualToString:@"dark"]) {
+        appDelegate.window.overrideUserInterfaceStyle = UIUserInterfaceStyleDark;
+      } else if ([mode isEqualToString:@"light"]) {
+        appDelegate.window.overrideUserInterfaceStyle = UIUserInterfaceStyleLight;
+      } else {
+        appDelegate.window.overrideUserInterfaceStyle = UIUserInterfaceStyleUnspecified;
+      }
```
gorandalum, line 30: Something is not working as intended with this theme-logic. The text color on iOS is not correct when the system setting is dark mode, and the preference in the app is set to light mode.
Screenshot 2021-02-24 at 11 28 42

gorandalum (changes requested):

gorandalum:
Great work! I had some comments that should be addressed, but overall it looks really good :+1:

ios/ChangeNativeBridge.m
mikaelbr: This looks like old code? 😱 Have you pulled the latest changes?

src/screens/Assistant/journey-date-picker/index.tsx
mikaelbr: I think you are running old code here. This should have been fixed by c6a3c5a

src/components/sections/date-input/index.android.tsx
mikaelbr: This doesn't happen to me. Is it Android version thing or is it due to running "old code" as addressed in other comments.

mikaelbr:
Navigating through this using screen reader on iOS. It works fine. There are some issues I think with the native component itself, but I think it's the best we can do for now. Hopefully, iOS/Apple will improve navigation on the date input form.

I made some small improvements to the Section Action Input setting them as radiogroup, radio items and switches.

gorandalum (approved):
LGTM :+1:

PR #832, opened 2021-01-12
fix: go directly to search location screen when adding favorite

### LINKED ISSUE #674: [Feature] Go straight to search when adding favorite
## Origin

## Motivation

Simplify process of adding a favorite

## Hypotheses and assumptions

The only required field for a favorite place is a place. By sending the user directly to the place-search we may avoid misunderstandings about which inputs to use.

## Proposed solution

When the user adds a new favorite, the first step of the process is to search for a place/stop

Figma / App 1.0 / Action / Favorite https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=8%3A1529

image

mikaelbr:
Fixes #674

TODO:

- [x] Make modal dismissable to close entire modal when opened.

# Examples

image

image

image

https://user-images.githubusercontent.com/606374/104298216-00e0fc80-54c4-11eb-913d-aad4287b9cfc.mov


mikaelbr:
@pettersuul Check out video for flow. We can also do a run-through demo to discuss different flow and behavior.

pettersuul (approved):

src/favorite-chips/index.tsx
```diff
@@ -108,7 +108,9 @@ const FavoriteChips: React.FC<Props> = ({
             text={t(FavoriteTexts.chips.addFavorite)}
             accessibilityRole="button"
             icon={Add}
-            onPress={() => navigation.navigate('AddEditFavorite', {})}
+            onPress={() =>
+              navigation.navigate('AddEditFavorite', {screen: 'SearchLocation'})
```
gorandalum, line 112: Educate me, what makes the 'screen' property mandatory when navigating to AddEditFavorite?

gorandalum (approved):
LGTM!

src/favorite-chips/index.tsx
mikaelbr: It's a sub screen. So two nested navigation stacks. `AddEditFavorite` is the root screen which is a stack itself and `SearchLocation` is a screen within that stack.

rikkekantega (approved):
LGTM :+1:

src/favorite-chips/index.tsx
gorandalum: What threw me off is that it was a mandatory parameter, but we have several other sub screens in the app which is navigated to without specifying the screen parameter. But I think I found it now. If I understand correctly it is the ´NavigatorScreenParams` which wraps the route param type that enables, and makes mandatory, the `screen` property.

PR #838, opened 2021-01-13
feat: Updated view for travel suggestions

### LINKED ISSUE #250: [Feature] Updated view of travel suggestions (phase 2)
## Origin

## Motivation

This issue is a continuation of the following feature: #234 

## Hypotheses and assumptions
The user needs to see several travel suggestions on the screen at the same time, to be able to compare them to each other and to choose the best suggestion.
The user needs to know the estimated arrival, in case he needs to get to an appointment in time.
This user needs to know the total travel time, because he wants to be traveling for the shortest time possible.
The user needs to know about the number of transfers, because he thinks it is comfortable to avoid unnecessary transfers.
The user needs to have detailed information about the first step in the journey, because this is something the user must relate to and plan for immediately.

## Proposed solution

image
image

Figma / Components 1.0 / List-groups https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components-1.0?node-id=547%3A6287
Figma / App 1.0 / Results https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=1786%3A295

rikkekantega:
Closes #250 

Update to result item view in accordance to new design
Also add support for icon sizing
And screen reader hidden utility (because of high potential of reuse)

image

### DIFF
--- a/src/screens/Assistant/ResultItem.tsx
+++ b/src/screens/Assistant/ResultItem.tsx
@@ -181,9 +181,9 @@ const ResultItemFooter: React.FC<{legs: Leg[]; onDetailsPressed(): void}> = ({
           AssistantTexts.results.resultItem.footer.detailsHint,
         )}
         icon={ArrowRight}
-        iconPosition={'right'}
-        mode={'tertiary'}
-        type={'compact'}
+        iconPosition="right"
+        mode="tertiary"
+        type="compact"
         onPress={onDetailsPressed}
         textStyle={styles.detailsButtonText}
       />
--- a/src/screens/Assistant/ResultItem.tsx
+++ b/src/screens/Assistant/ResultItem.tsx
@@ -26,6 +26,7 @@ import ThemeText from '../../components/text';
 import ThemeIcon from '../../components/theme-icon';
 import {
   AssistantTexts,
+  dictionary,
   TranslateFunction,
   useTranslation,
 } from '../../translations/';
@@ -53,7 +54,7 @@ function getFromLeg(legs: Leg[]) {
   const found = legs.find(legWithQuay);
   const fromQuay = (found?.fromEstimatedCall ?? found?.fromPlace)?.quay;
   if (!fromQuay) {
-    return legs[0].fromPlace.name ?? '';
+    return legs[0].fromPlace.name;
   }
   const publicCodeOutput = fromQuay.publicCode ? ' ' + fromQuay.publicCode : '';
   return fromQuay.name + publicCodeOutput;
@@ -158,7 +159,7 @@ const ResultItemFooter: React.FC<{legs: Leg[]; onDetailsPressed(): void}> = ({
 }) => {
   const styles = useThemeStyles();
   const {t} = useTranslation();
-  const quayName = getFromLeg(legs);
+  const quayName = getFromLeg(legs) ?? t(dictionary.travel.quay.defaultName);
   const quayLeg = legs.find(legWithQuay);
   const timePrefix =
     !!quayLeg && !quayLeg.realtime ? missingRealtimePrefix : '';
--- a/src/screens/Assistant/ResultItem.tsx
+++ b/src/screens/Assistant/ResultItem.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {View, TouchableOpacity, Text, AccessibilityProps} from 'react-native';
+import {View, AccessibilityProps, ViewStyle} from 'react-native';
 import {Leg, TripPattern} from '../../sdk';
 import {StyleSheet} from '../../theme';
 import {
@@ -7,7 +7,6 @@ import {
   secondsToDurationShort,
   secondsBetween,
   secondsToMinutesShort,
-  formatToClockOrRelativeMinutes,
   missingRealtimePrefix,
   formatToClock,
 } from '../../utils/date';
@@ -28,9 +27,12 @@ import ThemeIcon from '../../components/theme-icon';
 import {
   AssistantTexts,
   TranslateFunction,
-  dictionary,
   useTranslation,
 } from '../../translations/';
+import {ScrollView} from 'react-native-gesture-handler';
+import {ArrowRight, ChevronRight} from '../../assets/svg/icons/navigation';
+import Button from '../../components/button';
+import {screenReaderHidden} from '../../utils/accessibility';
 
 type ResultItemProps = {
   tripPattern: TripPattern;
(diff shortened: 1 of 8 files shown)


src/translations/screens/Assistant.ts
```diff
@@ -75,13 +76,18 @@ const AssistantTexts = {
...
+      destination: {
+        a11yLabel: _('Destinasjon'),
+      },
       waitRow: {
         label: _('Vent'),
       },
+      footer: {
+        fromLabel: (place: string, time: string) => _(`Fra ${place} ${time}`),
+        detailsLabel: _('Detaljer'),
+        detailsHint: _('Aktivér for å  flere detaljer om reisen'),
```
gorandalum, line 88: Mangler et ord?

rikkekantega: Nice catch, woops

src/theme/colors.ts
```diff
@@ -210,7 +217,7 @@ export const themes: Themes = {
       ...textTypes,
     },
     border: {
-      primary: colors.primary.gray_100,
+      primary: colors.primary.gray_50,
```
mikaelbr, line 220: Is this true for the entire design system?

src/translations/screens/Assistant.ts
```diff
@@ -75,13 +76,18 @@ const AssistantTexts = {
...
+      },
       journeySummary: {
-        situationsWarning: _('Driftsmeldinger gjelder for dette forslaget.'),
```
mikaelbr, line 82: Is it now no longer indications that there are some situations relevant for the result item?

mikaelbr: Ah. Probably catched by `hasSituationsTip`

src/screens/Assistant/ResultItem.tsx
```diff
@@ -51,37 +54,40 @@ function getFromLeg(legs: Leg[]) {
...
 }
 const ResultItemHeader: React.FC<{
   tripPattern: TripPattern;
 }> = ({tripPattern}) => {
-  const quayName = getFromLeg(tripPattern.legs);
   const styles = useThemeStyles();
   const durationText = secondsToDurationShort(tripPattern.duration);
   const {t} = useTranslation();
+  const startTime = tripPattern.legs[0].expectedStartTime;
+  const endTime = tripPattern.legs[tripPattern.legs.length - 1].expectedEndTime;
```
mikaelbr, line 69: Potential undefined? Can't remember now exactly, but maybe there's always at least a walk leg. So this should be fine.

src/screens/Assistant/ResultItem.tsx
```diff
@@ -109,63 +115,120 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
+              ) : (
+                <TransportationLeg leg={leg} />
               );
-            }
-            return <TransportationLeg key={leg.serviceJourney.id} leg={leg} />;
+            return (
+              <View
+                style={styles.legOutput}
+                key={leg.serviceJourney?.id ?? leg.fromPlace.latitude}
+                {...screenReaderHidden}
```
mikaelbr, line 139: Note to self/thinking out loud. Wonder if we can make this as a component at some point

src/screens/Assistant/ResultItem.tsx
```diff
@@ -109,63 +115,120 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
+      </ThemeText>
+      <Button
+        text={t(AssistantTexts.results.resultItem.footer.detailsLabel)}
+        accessibilityHint={t(
+          AssistantTexts.results.resultItem.footer.detailsHint,
+        )}
+        icon={ArrowRight}
+        iconPosition={'right'}
+        mode={'tertiary'}
+        type={'compact'}
```
mikaelbr, line 186:
```suggestion
        iconPosition="right"
        mode="tertiary"
        type="compact"
```

src/screens/Assistant/ResultItem.tsx
```diff
@@ -211,24 +274,10 @@ const FootLeg = ({leg, nextLeg}: {leg: Leg; nextLeg?: Leg}) => {
...
-    <View style={styles.legContainer}>
-      <ThemeText
-        type="label"
-        color="faded"
-        style={[styles.textDeprioritized, styles.time]}
-      >
-        {formatToClockOrRelativeMinutes(leg.expectedStartTime)}
-      </ThemeText>
+    <View style={styles.legContainer} accessibilityLabel={text}>
       <View style={styles.iconContainer}>
```
mikaelbr, line 278: Is this needed?

mikaelbr (approved):
🍰

gorandalum (approved):

gorandalum:
On the ResultItem the main focus point is the line number and line destination name, but when using the screen reader this info is not read at all. Is this by design since it would be to much to read for each ResultItem?

Screenshot 2021-01-14 at 11 29 41

src/screens/Assistant/ResultItem.tsx
rikkekantega: Yeah think there always is, but might stilll consider the possibility, thanks for pointing that out

src/screens/Assistant/ResultItem.tsx
rikkekantega: Sure, a wrapper component might feel more seemless

src/screens/Assistant/ResultItem.tsx
rikkekantega: If you're thinking of the extra outer view: Can't see that it is, probably just the result of incomplete cleaning

rikkekantega:
> On the ResultItem the main focus point is the line number and line destination name, but when using the screen reader this info is not read at all. Is this by design since it would be to much to read for each ResultItem?
> 
> Screenshot 2021-01-14 at 11 29 41

Script from designers said to read (example) "2 bytter. Buss nummer 70, buss nummer 1, og trikk nummer 9, totalt 230 meter å gå" instead. Found that there was too much info to read through on larger journeys in an earlier user test with Blidneforbundet

src/screens/Assistant/ResultItem.tsx
mikaelbr: I think my main thing is that it will be more explicit and easier to document in a potential component library

src/screens/Assistant/ResultItem.tsx
mikaelbr: Yeah, either or. Just grouped styles together I think?

src/theme/colors.ts
rikkekantega: Yep, in every case I managed to find in app 1.0. Though with a different name, but same color value. Going over the new theming and new naming now in a separate update.

src/screens/Assistant/ResultItem.tsx
rikkekantega: Yep, glad you pointed that out, had a label before as well. Unnecessary with the nesting here now

PR #839, opened 2021-01-14
feat: Update ticket screens for i18n use

gorandalum:


gorandalum:
Some things left:
- Not all of the functions in the date-utils file is fully i18n compatible. They some places defaults to Norwegian or the default language. Changing this will affect lots of files, so should probably be done in own PR.
- Several places the names for ticket types, user profiles, etc. are presented in Norwegian ("Enkeltbillett", "Voksen", o.l.). The English texts are returned through the API, but using them is not yet implemented. Did not update it now as it is not necessary to do it before the translation of the texts-files.
- And probably several more places.

rikkekantega:
> Some things left:
> 
> * Not all of the functions in the date-utils file is fully i18n compatible. They some places defaults to Norwegian or the default language. Changing this will affect lots of files, so should probably be done in own PR.
> * Several places the names for ticket types, user profiles, etc. are presented in Norwegian ("Enkeltbillett", "Voksen", o.l.). The English texts are returned through the API, but using them is not yet implemented. Did not update it now as it is not necessary to do it before the translation of the texts-files.
> * And probably several more places.

I'll create a separate issue for the first bullet point.:)

mikaelbr (approved):
LGTM

PR #842, opened 2021-01-14
feat: Add animation to the ticket validity line

gorandalum:


src/screens/Ticketing/Ticket/ValidityLine.tsx
```diff
@@ -57,4 +57,104 @@ const ValidityLine: React.FC<{
...
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    marginVertical: theme.spacings.medium,
+    marginHorizontal: -theme.spacings.medium,
+    flexDirection: 'row',
+  },
+  progressBar: {
+    height: 8,
+    backgroundColor: colors.primary.green_500,
```
rikkekantega, line 149: Changes to theming in app code will be making use of themed color identifiers, as they are defined in Figma. I think green mostly just stay the same green, but many colors will change between light and dark (and potential future themes). Not something you can solve now, just a note that we should remember to change these color references later (I imagine that we'll need to refactor many places due to the new design system, so it'll probably be solved then).

rikkekantega (approved):
LGTM :rocket:

src/screens/Ticketing/Ticket/ValidityLine.tsx
gorandalum: Yeah, was working on this when I asked about the colors on Slack. The colors in dark mode is ok-ish, but probably not exactly how @pettersuul envisioned.

src/screens/Ticketing/Ticket/ValidityLine.tsx
```diff
@@ -57,4 +57,104 @@ const ValidityLine: React.FC<{
...
+/**
+ * Starts an animation for the vertical line offset which loops continuously.
+ * The offset is animated from 0 to 1, which must be interpolated to the actual
+ * pixel offset value in the vertical line component.
+ * @returns the current animated offset value
+ */
+const useAnimatedVerticalLineOffset = () => {
+  const animatedOffset = useRef(new Animated.Value(0)).current;
+  useEffect(() => {
+    Animated.loop(
```
cbrevik, line 69: Should return in useEffect from Animated.loop.start(), since the return value from start() is a function to stop the animation. Potentially it could loop forever even if the screen is unmounted/ticket is no longer valid.
```suggestion
    return Animated.loop(
```

cbrevik (changes requested):
LGTM, only thing is I would return the callback from `start` so animation is ended on component unmount.

Also, we might want to discuss how well this would work as a visual inspection when there is not much time left on your ticket:

https://user-images.githubusercontent.com/4932625/104726871-ae5a4700-5734-11eb-88d9-bed9916821b8.mov

src/screens/Ticketing/Ticket/ValidityLine.tsx
gorandalum: Fixed

gorandalum:
> Also, we might want to discuss how well this would work as a visual inspection when there is not much time left on your ticket (not a blocker for this PR) CC @pettersuul:

@pettersuul proposed a minimum wideness of 5%. Since there is a text right above the progress bar stating minutes left this should probably not be too misleading for the user. I've implemented this fix now.

cbrevik (approved):
🎉

PR #844, opened 2021-01-15
fix: Broken logic on choosing language initial app install

### LINKED ISSUE #843: [BUG] Mismatch between used language and default preferences in app
**Describe the bug**
When no preferred language is set in the preferences, it uses the device language. It ignores that the default language selection in the app is Norwegian.

**To Reproduce**
Steps to reproduce the behaviour (Android emulator):
1. Close the app
2. Ensure device language is English
3. Delete all app data: Long click AtB-app icon -> App Info -> Storage & cache -> Clear storage
4. Start the app
5. See english language in app

**Expected behaviour**
Either:
- The startup language is always Norwegian, as it is selected as the default language in the app preferences. 
- The startup language is the device language, and in the app preferences the switch for whether to use device language is by default set to true.

**Screenshots**
Screenshot 2021-01-15 at 09 08 43

rikkekantega:
Fixed broken langugage logic on initial app install, choosing locale based language instead of default preference selected language.

closes #843

### DIFF
--- a/src/translations/LanguageContext.tsx
+++ b/src/translations/LanguageContext.tsx
@@ -25,21 +25,14 @@ function useLanguage() {
   } = usePreferences();
 
   useEffect(() => {
-    if (!useSystemLanguage) {
-      const newLanguage = getAsAppLanguage(
-        userPreferencedLanguage ?? DEFAULT_LANGUAGE,
-      );
-      setCurrentLanguage(newLanguage);
-    } else {
-      const onChange = () => {
-        setLocale(preferredLocale);
-      };
-      RNLocalize.addEventListener('change', onChange);
-      return () => {
-        RNLocalize.removeEventListener('change', onChange);
-      };
-    }
-  }, [useSystemLanguage, userPreferencedLanguage]);
+    const onChange = () => {
+      setLocale(preferredLocale);
+    };
+    RNLocalize.addEventListener('change', onChange);
+    return () => {
+      RNLocalize.removeEventListener('change', onChange);
+    };
+  }, []);
 
   useEffect(() => {
     if (useSystemLanguage) {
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Looks good! :+1:

With that said it is a mental exercise to wrap the head around this logic, so there is no surprise that a small bug got through.

Could it be easier if the `useEffect` for listening to locale changes was its own method, and maybe it is not strictly necessary to be turned on and off based on whether `useSystemLanguage` is true or not.

Something like this: 
```
useEffect(() => {
   ... adding locale event listener
}, []);

useEffect(() => {
  if (useSystemLanguage) {
    ... set current language from locale
  } else {
    ... set current language from preferences
  }
}, [everything necessary])
```
Just a thought, but maybe not worth the effort since this we now (hopefully) have e fully working implementation.

rikkekantega:
> Looks good! 
> 
> With that said it is a mental exercise to wrap the head around this logic, so there is no surprise that a small bug got through.
> 
> Could it be easier if the `useEffect` for listening to locale changes was its own method, and maybe it is not strictly necessary to be turned on and off based on whether `useSystemLanguage` is true or not.
> 
> Something like this:
> 
> ```
> useEffect(() => {
>    ... adding locale event listener
> }, []);
> 
> useEffect(() => {
>   if (useSystemLanguage) {
>     ... set current language from locale
>   } else {
>     ... set current language from preferences
>   }
> }, [everything necessary])
> ```
> 
> Just a thought, but maybe not worth the effort since this we now (hopefully) have e fully working implementation.

No big effort, and makes sense to do that. I think I just lost some perspective after re-thinking/writing this a few times. So thanks for pointing that out :)

src/translations/LanguageContext.tsx
```diff
@@ -24,33 +24,29 @@ function useLanguage() {
...
         : DEFAULT_LANGUAGE;
       setCurrentLanguage(language);
+    } else {
+      const newLanguage = getAsAppLanguage(
+        userPreferencedLanguage ?? DEFAULT_LANGUAGE,
+      );
+      setCurrentLanguage(newLanguage);
     }
-  }, [locale, useSystemLanguage]);
+  }, [userPreferencedLanguage, useSystemLanguage]);
```
gorandalum, line 49: The `locale` field probably need to be in the dependency list, to ensure that the current language is changed when the locale event listener is fired?

PR #852, opened 2021-01-21
feat: Update to new ticketing flow

gorandalum:
After the user feedback the ticketing flow design was repolished. This
commit updates the app ticketing flow to be more inline with the new
design. One notable difference from the design is that subscreens are
still used where the design is not changed to using action screens.


src/navigation/createDismissableStackNavigator.tsx
```diff
@@ -32,10 +32,10 @@ function DismissableStackRouter(
     ...router,
     actionCreators: {
       ...router.actionCreators,
-      dismiss(params?: object) {
+      dismiss(params?: object & {screen: string}) {
         return {
           type: 'NAVIGATE',
-          payload: {name: options.dismissToScreen, params},
+          payload: {name: params?.screen || options.dismissToScreen, params},
```
cbrevik, line 38: Bit unsure about this change. If you want to navigate to a specific screen, you should use `.navigate()` instead of `.dismiss()`

cbrevik (changes requested):

src/navigation/createDismissableStackNavigator.tsx
gorandalum: I could not get `navigate()` to work out-of-the-box, as only the screens from the current stack and root stack were available, not the sub screens of the tab navigator for tickets. I can look into what the possibilities are. Maybe navigating to the TicketsScreen, and adding a parameter specifying which tab to open?

gorandalum: Added fix for this now.

PR #853, opened 2021-01-21
feat: Show refunded tickets under expired tickets

gorandalum:


cbrevik (approved):
LGTM 🚀

gorandalum:
Oops, ValidityLine still green with animation even though refunded, as long as the original valid to timestamp is not passed. Will fix.

PR #855, opened 2021-01-22
feat: More robust fetching of fare contracts

gorandalum:
When fetching fare contracts failed:
- It did not automatically retry
- Manual retry did not work, since the refresh control was already in
  the refreshing state


cbrevik (approved):
LGTM 👌🏻

PR #856, opened 2021-01-22
feat: Fix product/zone/user names on ticket

gorandalum:


src/screens/Ticketing/Ticket/TicketInfo.tsx
```diff
@@ -0,0 +1,98 @@
...
+  );
+
+  const fromTariffZone = findById(
+    tariffZones,
+    fc.travellers[0].tariff_zone_refs[0],
+  );
+  const toTariffZone = findById(
+    tariffZones,
+    fc.travellers[0].tariff_zone_refs.slice(-1)[0],
+  );
```
cbrevik, line 29: Suggested refactor for readability:

```suggestion
  const {fare_product_ref: productRef, tariff_zone_refs} = fc.travellers[0];
  const [firstZone] = tariff_zone_refs;
  const [lastZone] = tariff_zone_refs.slice(-1);

  const preassignedFareProduct = findById(preassignedFareProducts, productRef);
  const fromTariffZone = findById(tariffZones, firstZone);
  const toTariffZone = findById(tariffZones, lastZone);
```

cbrevik (approved):
Suggested a change in the body of the `TicketInfo`-component.

I thought about looking into `groupByUserProfileNames` as well, since it's doing a "lot". But can't think of any better way of doing it at the moment.

LGTM otherwise 👍🏻

PR #860, opened 2021-01-25
feat: workflow: update reference data & remote config

torfjor:
Syncs src/reference-data/*.json every hour and opens a PR if there are any diffs.

On push: uses an external tool: github.com/atb-as/remoteconfig/cmd/rc that lets you publish key=value pairs as default values in remote config. Need to revisit later when we need to differentiate on environments.

### DIFF
--- a/.github/workflows/publish_remote_config.yml
+++ b/.github/workflows/publish_remote_config.yml
@@ -1,0 +1,30 @@
+name: Publish remote config
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+      - name: Publish remote config
+        env:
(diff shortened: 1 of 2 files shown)


mikaelbr:
@gorandalum or @cbrevik Could any of you who know the ref data/remote config knowledge look into reviewing this?

.github/workflows/publish_remote_config.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish remote config
+
+on:
+  push:
+    branches:
+      - master
```
cbrevik, line 6: Maybe add a path here so it only runs if `reference-data` is updated?

Haven't tested it out, but something like:

```suggestion
  push:
    branches:
      - master
    paths:
      - 'src/reference-data/*.json'
```

cbrevik (approved):
LGTM, just a suggested change

.github/workflows/publish_remote_config.yml
torfjor: Looks like the workflow docs use double star for wildcards. Will do some local testing.

PR #864, opened 2021-01-25
feat: Show ref data names in correct langauge

gorandalum:
The reference data (UserProfiles, TariffZones, PreassignedFareProducts)
format support multiple languages by providing names and other texts in
a type which has a field 'lang' specifying language and a field 'value'
containing the text in that language. In our code this is typed as a
'LanguageAndText'.

This commit add utility functions for selecting the correct text based
on the current language.

### DIFF
--- a/src/api/utils.ts
+++ b/src/api/utils.ts
@@ -50,36 +50,3 @@ export const getAxiosErrorMetadata = (error: AxiosError): ErrorMetadata => ({
 });
 
 export const stringifyUrl = (url: string, query: string) => `${url}?${query}`;
-
-/**
- * Get the text of a field in a NeTeX entity in the correct language. If English
- * is requested, it will fallback to Norwegian if no English text is found. If
- * neither English nor Norwegian is found the first text in the provided texts
- * array is returned.
- */
-export const getTextInLanguage = (
-  texts: LanguageAndText[],
-  language: Language,
-) => {
-  if (language === Language.English) {
-    const englishText = texts.find((t) => t.lang === 'eng');
-    if (englishText) return englishText.value;
-  }
-  const norwegianText = texts.find((t) => t.lang === 'nor' || t.lang === 'nob');
-  if (norwegianText) return norwegianText.value;
-
-  return texts[0].value;
-};
-
-/**
- * Wrapper for getting the name of a NeTeX entity in the given language.
- */
-export const getNameInLanguage = <
-  T extends {
-    name: LanguageAndText;
-    alternativeNames?: LanguageAndText[];
-  }
->(
-  {name, alternativeNames}: T,
-  language: Language,
-) => getTextInLanguage([name, ...(alternativeNames || [])], language);
--- a/src/api/utils.ts
+++ b/src/api/utils.ts
@@ -1,7 +1,5 @@
 import axios, {AxiosError} from 'axios';
 import {RequestIdHeaderName} from './headers';
-import {LanguageAndText} from '../reference-data/types';
-import {Language} from '../translations';
 
 export type ErrorType =
   | 'unknown'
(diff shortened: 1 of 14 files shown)


src/api/utils.ts
```diff
@@ -48,3 +50,36 @@ export const getAxiosErrorMetadata = (error: AxiosError): ErrorMetadata => ({
...
+  const norwegianText = texts.find((t) => t.lang === 'nor' || t.lang === 'nob');
+  if (norwegianText) return norwegianText.value;
+
+  return texts[0].value;
+};
+
+/**
+ * Wrapper for getting the name of a NeTeX entity in the given language.
+ */
+export const getNameInLanguage = <
```
gorandalum, line 77: Kinda feel that the function name `getNameInLanguage` is not the best. I am happy for better sounding alternatives.

cbrevik:
Should the two functions live in `src/api/utils.ts`? Isn't the `src/translation/commons.ts` a more natural place for these? Possibly a new `util.ts` there?

src/api/utils.ts
cbrevik: They are reference-data related. So possibly `getReferenceDataName`?

src/api/utils.ts
```diff
@@ -48,3 +50,36 @@ export const getAxiosErrorMetadata = (error: AxiosError): ErrorMetadata => ({
...
 
 export const stringifyUrl = (url: string, query: string) => `${url}?${query}`;
+
+/**
+ * Get the text of a field in a NeTeX entity in the correct language. If English
+ * is requested, it will fallback to Norwegian if no English text is found. If
+ * neither English nor Norwegian is found the first text in the provided texts
+ * array is returned.
+ */
+export const getTextInLanguage = (
```
cbrevik, line 60: `getReferenceDataText`?

gorandalum:
> Should the two functions live in `src/api/utils.ts`? Isn't the `src/translation/commons.ts` a more natural place for these? Possibly a new `util.ts` there?

Maybe. I kinda thought they belonged in the `src/api/utils.ts` as they are helper methods for the NeTeX multilingual string format, and not generic for other use cases.

I also see some other comments about "rebranding" the methods to be more reference data specific. With this in mind maybe `src/reference-data/utils.ts` is a better location for now?

cbrevik:
Yep, I updated my comment earlier to suggest the same. `src/reference-data/utils.ts` seems fine 👌🏻

src/api/utils.ts
```diff
@@ -1,5 +1,7 @@
 import axios, {AxiosError} from 'axios';
 import {RequestIdHeaderName} from './headers';
+import {LanguageAndText} from '../reference-data/types';
+import {Language} from '../translations';
```
cbrevik, line 4:
```suggestion
```

Probably don't need these imports?

src/reference-data/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+const getReferenceDataText = (texts: LanguageAndText[], language: Language) => {
+  if (language === Language.English) {
+    const englishText = texts.find((t) => t.lang === 'eng');
+    if (englishText) return englishText.value;
+  }
+  const norwegianText = texts.find((t) => t.lang === 'nor' || t.lang === 'nob');
+  if (norwegianText) return norwegianText.value;
+
+  return texts[0].value;
```
cbrevik, line 18: Extra safety measure
```suggestion
  return texts[0]?.value;
```

cbrevik (approved):
LGTⓂ️

src/api/utils.ts
gorandalum: They are needed. Used for typing the function parameters.

cbrevik: Didn't you move them?

gorandalum: Aha. Just saw a comment on `utils.txt`, and didn't think it through 😅

PR #870, opened 2021-01-26
fix: Fix language prioritizing and default to no if enable_i18n is false

mikaelbr:
There is an issue with the language selection when feature toggle is disabled where it defaults to English. That kind of defeat the purpose of the feature toggle as a safety mechanism to enforce "old" way to show things. This PR fixes that.

This PR also fixes what seems to be a small error in handling default system languages where it now looks to just select the _most_ preferred language or English.

### DIFF
--- a/src/translations/LanguageContext.tsx
+++ b/src/translations/LanguageContext.tsx
@@ -36,9 +36,13 @@ function useLanguage() {
 
   useEffect(() => {
     if (useSystemLanguage) {
+      // Get preferred locale from system settings.
+      // If no system settings match our languages we select
+      // English as it is more likely they know english than Norwegian.
+      // If locale is undefined we don't support any of the system languages.
       const language = locale
         ? getAsAppLanguage(locale.languageCode)
-        : DEFAULT_LANGUAGE;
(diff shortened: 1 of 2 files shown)


cbrevik (approved):
LGTM 👌🏻

src/translations/LanguageContext.tsx
```diff
@@ -36,9 +36,13 @@ function useLanguage() {
 
   useEffect(() => {
     if (useSystemLanguage) {
+      // Get preferred locale from system settings.
+      // If no system settings match our languages we select
+      // English as it is more likely they know english than Norwegian.
+      // If locale is undefined we don't support any of the system languages.
       const language = locale
         ? getAsAppLanguage(locale.languageCode)
```
gorandalum, line 44: I kinda feel some of the logic in this file would be easier to follow if `locale` was more strongly typed. As far as I understand it will always be either Norwegian or English, and never undefined.

gorandalum: My bad. I see now that it can be undefined.

PR #873, opened 2021-01-26
chore: updates pod lockfile

mikaelbr:


cbrevik:
We are just switching back and forth. No idea why this is: https://github.com/AtB-AS/mittatb-app/pull/862/files

mikaelbr:
Odd. Maybe I should try to perge something

PR #875, opened 2021-01-26
fix: Allow user to try again when offerSearch fail

gorandalum:


cbrevik (approved):
LGTM 👌🏻 As discussed we should look into the extra offer search later

PR #882, opened 2021-01-27
feat: Ticket inspection with QR Code from API

cbrevik:
## Description
Simplified ticket inspection with QR code from API. This uses the `qr_code` (base64 string) from the returned `fare_contract`, and constructs an SVG-image from it.

## A couple of issues with this solution
* QR-code is generated asynchronously, so might not be available right after a ticket is bought and fare contracts are loaded (must be tested extensively)
* Might negatively impact security level of account since we're generating paper tokens (will have to test this as well)

Still, I'd love to get this in QA though so we can test it a bit further with inspection app as well.

## Screenshots
QR inspection screen:
image
Can navigate to screen from ticket details modal:
image
Or ticket overview:
image


torfjor (approved):
Did some testing, notes:

- Purchased 10 single tickets
- The first purchase had a noticable delay before the `qr_code` property was present and I could click "Show for inspection". The cloud function that triggers the distribution of the qr paper token has some cold start ramp-up time (mainly spent exchanging client credentials for oauth2 tokens).
- For the remaining purchases the "Show for inspection" button was clickable right away without any refreshes.

Code looks okay for testing paper qr tokens, but obviously needs to handle creating and signing mobile tokens when we get there!

👍

src/screens/Ticketing/Ticket/Details/DetailsScreen.tsx
```diff
@@ -41,6 +41,12 @@ export default function DetailsScreen({navigation, route}: Props) {
       orderVersion: fc.order_version,
     });
 
+  const onInspectionNavigate = () =>
+    fc &&
+    navigation.push('TicketInspection', {
+      orderId: fc.order_id,
```
gorandalum, line 47: Couldn't we just pass the entire `fc` object here? and same for navigating to 'TicketReceipt'?

gorandalum (approved):
LGTM!

src/screens/Ticketing/Ticket/Details/DetailsScreen.tsx
cbrevik: We could, but that would make it more difficult to construct deep-links later to screens in app. We prefer to pass id's or simple serializable values when navigating. In this concrete example I'd also imagine that opening this screen when getting a push notification about a Bluetooth-challenge from an inspector is a plausible functionality.

Issue #879, opened 2021-01-27
[Feature] Add paging and date for departure details


mikaelbr:
## Origin

_Links to received feedback, user research or other findings._

From design files and previous discussions. Service journey details/departure details now lacks paging and date. See image from Figma: 

<img width="175" alt="Screenshot 2021-01-27 at 10 56 18" src="https://user-images.githubusercontent.com/606374/105974535-4b4ea580-608e-11eb-822b-7c9c0ce3a733.png">

See https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=4%3A982

## Acceptance Criteria

_List of relevant acceptance criteria as a part of a QA flow_

- [] Can navigate through different departure suggestions
- [] Can see date of specified departure/service journey in details
- [] Renders colors as expected in dark and light mode.
- [] Is translatable to English and Norwegian.
- [] Does not drain battery
- [] Works in up to 200% font size
- [] Is operable and perceivable using screen reader

mikaelbr:
QA: Is operable with screen reader but can be improved. Non blocking issue for now. Added separate bug: https://github.com/AtB-AS/mittatb-app/issues/1037

Issue #746, opened 2020-11-30
[Bug] Sikre at de øverste inputfeltene på reisesøkbildet vises riktig på mobiler med liten skjermstørrelse


mikaelbr:
There are a lot of animation timing issues causing different rendering bugs for both assistant view and departure view as it is now. One problem is that we always have to use the Y translated header container and also do scrolling-handling. To solve this we should separate the concepts.

One way to do this is to use shared elements and add new navigation point. This will also allow for pressing "back" on Android.

monopoint:
This needs to be detailed a bit more before we can work on it. 

- What are the specific problems we are trying to solve?
- what are the acceptance criteria for this issue?
- what functional requirements of the animated header to we need to preserve?

tormoseng:
https://mittatb.slack.com/archives/C018UPET2Q2/p1622640525117800

monopoint:
> https://mittatb.slack.com/archives/C018UPET2Q2/p1622640525117800

denne er fikset på #1359 

@cbrevik jeg synes denne saken er litt uklar, men den berører i hvert fall en god del av det som ble fikset i #1359. skal vi kaste denne saken og si at resten ligger i bøtta som heter "skrive om disappearing header"?

PR #883, opened 2021-01-27
feat: Adds pagination and date to service journey/departure details

mikaelbr:
This adds pagination and date metadata to departure details view.

- Unifies behavior/design for trip pattern details and service journey/departure details.
- Adds disappearing map on scroll.
- Adds pagination for service journey/departure details
- Delays loading indicator 400 ms to avoid it showing too often.

## Examples

Screenshot 2021-01-27 at 15 23 19
Screenshot 2021-01-27 at 15 23 49

Screenshot 2021-01-27 at 15 23 32

## Demo new scroll behavior for details view

https://user-images.githubusercontent.com/606374/106045112-deb0c680-60e0-11eb-915d-af69434e2257.mov

Closing #879

### DIFF
--- a/src/components/disappearing-header/content.tsx
+++ b/src/components/disappearing-header/content.tsx
@@ -1,0 +1,159 @@
+import {useScrollToTop} from '@react-navigation/native';
+import React, {PropsWithChildren, useCallback, useEffect, useRef} from 'react';
+import {
+  Animated,
+  NativeScrollEvent,
+  NativeSyntheticEvent,
+  Platform,
+  RefreshControl,
+  ScrollView,
+  View,
+} from 'react-native';
+import {StyleSheet, useTheme} from '../../theme';
+import throttle from '../../utils/throttle';
+import {useLayout} from '../../utils/use-layout';
+
+type Props = PropsWithChildren<{
+  header: React.ReactNode;
+  onRefresh?(): void;
+  isRefreshing?: boolean;
+
+  onEndReached?(e: NativeScrollEvent): void;
+  onEndReachedThreshold?: number;
+}>;
+
+type Scrollable = {
+  scrollTo(opts: {y: number}): void;
+};
+
+export default function ContentWithDisappearingHeader({
+  header,
+  children,
+  isRefreshing = false,
+  onRefresh,
+
+  onEndReached,
+  onEndReachedThreshold = 10,
+}: Props) {
+  const {
+    contentHeight,
+    onHeaderContentLayout,
+  } = useCalculateHeaderContentHeight();
+  const contentHeightRef = React.useRef(contentHeight);
+  const scrollableContentRef = React.useRef<ScrollView>(null);
+  useScrollToTop(
+    React.useRef<Scrollable>({
+      scrollTo: () =>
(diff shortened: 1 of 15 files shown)


src/components/disappearing-header/content.tsx
```diff
@@ -0,0 +1,159 @@
...
+  },
+}));
+
+function useCalculateHeaderContentHeight() {
+  const {onLayout: onHeaderContentLayout, height: contentHeight} = useLayout();
+  return {
+    contentHeight,
+    onHeaderContentLayout,
+  };
+}
```
cbrevik, line 159: Maybe an unnecessary abstraction?

src/screens/Nearby/section-items/line.tsx
```diff
@@ -72,29 +72,32 @@ export default function LineItem({
...
+  const onPress = (activeItemIndex: number) => {
     navigation.navigate('TripDetails', {
       screen: 'DepartureDetails',
       params: {
-        title,
-        serviceJourneyId: departure.serviceJourneyId!,
-        date: departure.aimedTime,
-        fromQuayId: group.lineInfo?.quayId,
+        activeItemIndex,
+        items,
```
cbrevik, line 86: Was thinking about how we'd link to "trips" with deep-links when we get as far as providing that functionality. But we probably need some other way of looking up trips when we get that far anyway. Not that sending in `items` breaks anything, still serializable, just thinking out loud.

cbrevik (commented):
Tested the branch locally, and dark mode doesn't seem quite right. Unsure if related to this PR or not:
image

src/screens/Nearby/section-items/line.tsx
mikaelbr: I think this still supports deep linking in the same way as before. Items is just references to service journeys which can be URLs

cbrevik: Yeah I think so too, just thinking out loud if this screen is maybe evolving past the scope of "linking to a single trip". If we ever get there that support can deep-link trips for users, or users can save trips. Not important wrt to this PR 👯

src/components/disappearing-header/content.tsx
mikaelbr: Agreed 😬 I initially planned to use this component inside the existing disappearing-header, but decided against it due to #746

mikaelbr:
> Tested the branch locally, and dark mode doesn't seem quite right. Unsure if related to this PR or not:

Definitely related to my changes! Thanks 😬  Will look into it.

src/screens/Nearby/section-items/line.tsx
mikaelbr: Note, this is for service journeys, not trips.

cbrevik:
RefreshControl overlays header. But that's probably a good thing so navigation arrows don't move when you're trying to page through many trips?
image

cbrevik (approved):
LGTM, great job landing this!

mikaelbr:
> RefreshControl overlays header. But that's probably a good thing so navigation arrows don't move when you're trying to page through many trips?

Yes, this is to avoid jumping content. Not perfect, and we can get a better showing of this, but it is an existing problem also so we can solve it in a separate issue

PR #891, opened 2021-01-29
feat: Tariff zone selection improvements

gorandalum:
Main changes:
- The initial tariff zones selection screen now has a map
- The map can be clicked for selecting both from and to zone,
  alternating which is chosen
- Labels are a little better now as they were changed to using a Point
  as the source in stead of a Polygon. For enabling this the library
turfCentroid was added.

Remaining improvements:
- The labels are not as pretty as in the design. Maybe use icons?
- The zone edges in the map is a little "rough" when compared to the
  zone edges in Mobillett. Look into tweaks on the shape source.
- Should give a hint about which zone is up for selection when clicking
  the map next time.
- I feel it is not intuitive enough that the sections which display
  selected zones is clickable and go to zone selection by address.
- I feel like zone selection by current location could be improved.


mikaelbr (approved):
I'll be honest! I haven't fine-combed the code or the initial design. But looking through the code it seems like no obvious issues or architectural/structural issues. And running the code looks to work fine with larger text and the flow seems to work great. So good job! I think this is an excellent improvement for the zone selector – but I still look forward to when we can give some description to the zones 😅

mikaelbr:
The icon in zone list is maybe odd?

src/screens/Ticketing/Purchase/TariffZones/index.tsx
```diff
@@ -138,128 +169,311 @@ const destinationPickerValue = (
...
+  });
 
-  const {t, language} = useTranslation();
+  useEffect(() => {
+    setSelectedZones({
+      ...selectedZones,
+      from: fromTariffZone,
+      to: toTariffZone,
+    });
+  }, [fromTariffZone, toTariffZone]);
```
cbrevik, line 188: Is this effect necessary? Wouldn't the zones be set in the default parameters going into the `selectedZones` `useState`?

cbrevik (approved):
LGTM! Nice implementation :D

src/screens/Ticketing/Purchase/TariffZones/index.tsx
gorandalum: I think it is necessary. When selecting a zone from the venue search (TariffZoneSearch screen) it navigates back to the TariffZones screen which is already in the stack, therefore not initializing the `selectedZones` from scratch.

cbrevik: Alright! In that case it definitely is necessary 👍🏻

gorandalum:
> The icon in zone list is maybe odd?

It is the design from the design sketches. @pettersuul can take this comment into consideration when doing the finishing touches.

PR #894, opened 2021-02-01
feat: Show cyan validity line while processing

gorandalum:


src/screens/Ticketing/Ticket/ValidityLine.tsx
```diff
@@ -8,39 +8,40 @@ import colors from '../../../theme/colors';
...
+type Props =
+  | {
+      status: 'valid';
+      nowSeconds: number;
+      validFrom: number;
+      validTo: number;
+    }
+  | {
+      status: 'reserving';
+      nowSeconds?: never;
```
mikaelbr, line 20: Why `?: never` here and not just 

```ts
type Props =
  | {
      status: 'valid';
      nowSeconds: number;
      validFrom: number;
      validTo: number;
    }
  | {
      status: 'reserving';
     }
  | {
      status: 'expired';
     };
```

src/screens/Ticketing/Ticket/ValidityLine.tsx
```diff
@@ -57,6 +58,46 @@ const ValidityLine: React.FC<{
...
+  const animatedVerticalLineOffset = useAnimatedVerticalLineOffset();
+  const numberOfVerticalLines = getNumberOfVerticalLines();
+  return (
+    <View style={styles.container}>
+      <View
+        style={[
+          styles.progressBar,
+          {width: `${validityPercent}%`},
+          {
+            backgroundColor:
```
mikaelbr, line 79:
```suggestion
          {
            width: `${validityPercent}%`,
            backgroundColor:
```

mikaelbr (approved):

src/screens/Ticketing/Ticket/ValidityLine.tsx
gorandalum: Yeah, I'll remove it. I initially had `{ status, nowSeconds, validFrom, validTo}: Props`, which was not okay if omitting 'never' as it said _"... does not exist on type 'Props'"_.

src/screens/Ticketing/Ticket/ValidityLine.tsx
```diff
@@ -57,6 +58,46 @@ const ValidityLine: React.FC<{
   }
 };
 
+const LineWithVerticalBars = ({
+  status,
+  validityPercent = 100,
+}: {
+  status: 'valid' | 'reserving';
```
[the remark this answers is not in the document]
gorandalum: Hmm. You mean like creating an enum, and using `Exclude<Status, Status.Expired>` to reuse the type.

Maybe it is better just having a `backgroundColor` or `progressBarStyle` parameter for setting the line background color, instead of passing the status?

PR #895, opened 2021-02-02
feat: Add network logging for Bugsnag

cbrevik:
## Changes

* Upgrade to latest Bugsnag (several improvements apparently)
* Add `axios-better-stacktrace` (https://github.com/svsool/axios-better-stacktrace) since we have thousands of entries for `Error node_modules/axios/lib/core/createError.js:16`. Problem is `axios` creates a new Error at that line, which is the one Bugsnag reports. We're losing a lot of context around the error.
* Enable logging of `network-error`s and `timeout`s. But gate it behind a feature flag if we're finding it spams us.

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -5,6 +5,8 @@ import {getAxiosErrorType, getAxiosErrorMetadata} from './utils';
 import Bugsnag from '@bugsnag/react-native';
 import {InstallIdHeaderName, RequestIdHeaderName} from './headers';
 import axiosRetry, {isIdempotentRequestError} from 'axios-retry';
+import axiosBetterStacktrace from 'axios-better-stacktrace';
+import {getBooleanConfigValue} from '../remote-config';
 
 export default createClient(API_BASE_URL);
 
@@ -27,6 +29,7 @@ export function createClient(baseUrl: string) {
   const client = axios.create({
     baseURL: baseUrl,
   });
+  axiosBetterStacktrace(client);
   client.interceptors.request.use(requestHandler, undefined);
   client.interceptors.response.use(undefined, responseErrorHandler);
   axiosRetry(client, {retries: 3, retryCondition});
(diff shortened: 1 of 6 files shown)


src/api/client.ts
```diff
@@ -62,19 +65,14 @@ function responseErrorHandler(error: AxiosError) {
...
       }
       break;
   }
 
   return Promise.reject(error);
 }
-
-function warnConfig(config: AxiosRequestConfig) {
-  if (!config) return;
-  console.warn(`URL: ${config.baseURL}${config.url}`);
```
mikaelbr, line 79: URL is still warned in local development?

mikaelbr (approved):
Good change. Also `axios-better-stacktrace` 👍

src/api/client.ts
cbrevik: Yeah that's handled by the override for `Bugsnag.notify` when `__DEV__` since it also includes metadata

cbrevik:
```bash
[Wed Feb 03 2021 14:41:00.931]  ERROR    [BUGSNAG] [Error: Request failed with status code 404] {
  "requestId": "5f3b7691-b720-449b-99d2-d882d5f33bd5",
  "requestUrl": "ticket/v1/search/zones2",
  "requestMessage": "Request failed with status code 404",
  "responseStatus": 404,
  "responseData": "{\"code\":404,\"message\":\"The current request is not defined by this API.\"}"
}
```

cbrevik:
After further testing `axios-better-stacktrace` doesn't completely solve the problem for us. I tested initially with just catching the error individually. But noticed later that Bugsnag still didn't get the complete stack. The problem seems to be that the patched code which modifies the stack trace is run AFTER the logging-interceptor runs. I've opened an issue about this (https://github.com/svsool/axios-better-stacktrace/issues/1), but don't really see it as a blocker for merging this.

PR #898, opened 2021-02-03
feat: Add recently purchased tickets on buy screen

gorandalum:
Screenshot 2021-02-03 at 12 56 10


src/screens/Ticketing/Tickets/use-recent-tickets.ts
```diff
@@ -0,0 +1,200 @@
...
+const mapUsers = (
+  users: {[user_profile: string]: string},
+  userProfiles: UserProfile[],
+) =>
+  Object.entries(users)
+    .map(([profileId, count]) => ({
+      ...findReferenceDataById(userProfiles, profileId),
+      count: parseInt(count),
+    }))
+    .filter((u): u is UserProfileWithCount => 'id' in u)
```
mikaelbr, line 77: Is this just for casting? Could it be done by explicitly setting generic in `map`? Or is it also a filter, and could potentially be reduce instead of map + filter. Might be micro-optimization though.

src/screens/Ticketing/Tickets/use-recent-tickets.ts
```diff
@@ -0,0 +1,200 @@
...
+    .sort(
+      // Sort by user profile order in remote config
+      (u1, u2) =>
```
mikaelbr, line 80: We also have `sortBy` dependency from lodash.

src/screens/Ticketing/Tickets/use-recent-tickets.ts
```diff
@@ -0,0 +1,200 @@
...
+ * - Remove similar tickets, which means tickets with same product, same zones
+ *   and same travellers
+ * - Return only the first three tickets, which will be the three most recent
+ *   tickets
+ */
+const mapToLastThreeUniqueTickets = (
+  recentFareContracts: RecentFareContract[],
+  remoteConfigState: RemoteConfigContextState,
+): RecentTicket[] => {
+  return recentFareContracts
```
mikaelbr, line 152: What is typically the size of this?

src/screens/Ticketing/Ticket/TicketInfo.tsx
```diff
@@ -92,17 +112,27 @@ const groupByUserProfileNames = (
...
+      );
       return {
-        name,
+        ...userProfile,
         count: idAndCount.count,
       };
     })
     .filter(
-      (nameAndCount): nameAndCount is NameAndCount => nameAndCount.name != null,
+      (userProfileWithCount): userProfileWithCount is UserProfileWithCount =>
```
mikaelbr, line 125: is this mainly to get typing correct or am I not reading this correct? are there cases where mapped data from above does not contain ID?

src/reference-data/utils.ts
```diff
@@ -30,3 +35,10 @@ export const getReferenceDataName = <
   {name, alternativeNames}: T,
   language: Language,
 ) => getReferenceDataText([name, ...(alternativeNames || [])], language);
+
+export const findReferenceDataById = <
+  T extends UserProfile | PreassignedFareProduct | TariffZone
+>(
+  elements: T[],
+  id: string,
```
mikaelbr, line 43: Technically this could be undefined. Although it won't change the implementation as is

mikaelbr (approved):

src/reference-data/utils.ts
gorandalum: You mean the return value? Or the `id` parameter?

mikaelbr: ID parameter (I think). At some point, the passed argument is `someArray[0]` and `someOtherArray.slice(-1)`. Unless we actually know the arrays have the "correct" amount of items.

gorandalum: My thought was that the tariff zones array always was non-empty, as it does not make sense to have a ticket with no tariff zones, and if so both `someArray[0]` and `someOtherArray.slice(-1)` will never be undefined. But as seen in the ongoing chat on Slack these assumptions should be confirmed. And the type should probably then be `[string, ...string[]]` instead of `string[]`.

src/screens/Ticketing/Tickets/use-recent-tickets.ts
gorandalum: Aight. Will use that instead.

src/screens/Ticketing/Tickets/use-recent-tickets.ts
gorandalum: It is not just casting. If the ticket for some reason references a product/tariff zone/user profile that is not in our reference data, then the response from `findReferenceDataById` will be undefined. This `filter` is for filtering out these tickets, and also get the correct typing after the filter.

With that said `reduce` may just as well be used, and could probably make it more readable? My last team had these `filter`-methods with type guards several places, so I may be a little affected by that 😅

src/screens/Ticketing/Tickets/use-recent-tickets.ts
gorandalum: I'll look into it.

gorandalum: All recent orders from Entur is returned. Martin is not sure if they expire or not at some point. But, as of now we hope to get Firestore in place before next production release. When Firestore is used we will not continuously retrieve all elements, but just the diff that has changed. I will also argue for that the data should already be sorted when stored in Firestore.

src/screens/Ticketing/Tickets/use-recent-tickets.ts
gorandalum: I did look into using sortBy. I think for these situations using sortBy did not give more readable nor more compact code, as it could not be used in a chain, and temp variables had to be introduced.

cbrevik (approved):
LGTM 🎉

PR #900, opened 2021-02-04
fix: Fix annoying unmounted race condition on startup

cbrevik:
Not sure if this is the best way. But got annoyed at getting this warning all the time 😬


gorandalum (approved):
LGTM! These warnings have been annoying me too!

PR #901, opened 2021-02-04
feat: Add product selection screen

gorandalum:
If more than one preassigned fare product is defined in RemoteConfig,
the user will be able to go to the product selection screen for
selecting product. In the overview screen the first product will still
be selected as default product.

The old app code also selected the first product in RemoteConfig as
default, and never allowed the user to change the product. To keep the
old app functioning as is, the 'Enkeltbillett buss/trikk' product must
be kept as the first product defined in the list from RemoteConfig.


cbrevik (approved):
LGTM, good job 👍🏻

PR #903, opened 2021-02-04
feat: Use the reference data json files as default

gorandalum:
Use the reference data json file, created/updated by the new workflow for
syncing reference data with RemoteConfig, as defaults for RemoteConfig
in the app.


src/reference-data/types.ts
```diff
@@ -1,9 +1,14 @@
 import {Polygon} from 'geojson';
 
-export type Language = 'nob' | 'nno' | 'nor' | 'eng';
+export enum Language {
```
mikaelbr, line 3: Should this be named something like `ReferenceDataLanguage` instead? To avoid importing wrong language

mikaelbr (approved):

src/reference-data/types.ts
gorandalum: Yeah. Also this type is only used in the utility method `getReferenceDataText`. I'll just define it in that utils file without export, so it doesn't show up as importable other places in the app.

PR #905, opened 2021-02-05
refactor: Standardize header button types

gorandalum:
The screen header mostly had the same functionality for its buttons. The
left buttons were either 'Back', 'Close', 'Cancel' or 'Home', and the
right button was always 'Chat'.

These button types are now standardized as a part of the ScreenHeader
props. Meaning the icon/text, onPress and accessibility is defined in
the ScreenHeader, making it not necessary to duplicate this all over the
app.

Also we were not consistent with usage of text or icons for these header
buttons. It will be easier to be consistent with this new
standardization. After this change mainly text will be used (except for
'Home' and 'Chat), since this is more inline with design guidelines, and
it will give a better differentiation between 'Close' and 'Cancel'.

A custom type may be added later for specifying buttons that is not
included in these standardized types, but as of now I found it not
necessary to add.


gorandalum:
Created as a draft now. If accepted some cleanup has to be done like removing redundant imports and translations, adding translations for the `ScreenHeader` and deciding what the default accessibility should be for these buttons.

gorandalum:
The changes are mainly in `src/ScreenHeader/index.tsx` and `src/ScreenHeader/HeaderButton.tsx`. Other changes are usage of these new props.

gorandalum:
Ment to be reviewed, even though it is still draft @cbrevik @mikaelbr

gorandalum:
The commit in this PR was wrongfully pushed to the `recent_tickets` branch. It was cleaned up from that branch by force pushing, but it seems that it is not possible to "reopen" Github PRs when they have been marked as merged. Even though, we can just use this PR for initial feedback, and I'll make a new one if/when this should get merged to master.

src/ScreenHeader/HeaderButton.tsx
```diff
@@ -1,21 +1,78 @@
 import {AccessibilityProps, TouchableOpacity} from 'react-native';
 import React from 'react';
 import insets from '../utils/insets';
+import {LeftButtonProps, RightButtonProps} from './index';
+import {useNavigation} from '@react-navigation/core';
```
cbrevik, line 5: Always import from `'@react-navigation/native'`, `core` should not be used directly

src/ScreenHeader/HeaderButton.tsx
```diff
@@ -1,21 +1,78 @@
...
+const useIconButton = (
+  buttonProps: LeftButtonProps | RightButtonProps,
+): IconButton | undefined => {
+  const navigation = useNavigation<NavigationProp<RootStackParamList>>();
+  const navigateHome = useNavigateToStartScreen();
+  const chatIcon = useChatIcon();
+  switch (buttonProps.type) {
+    case 'back':
+      return {
+        icon: <ThemeText>Tilbake</ThemeText>,
```
cbrevik, line 50: Translations for these + a11y

PR #906, opened 2021-02-05
fix: Make confirmation screen scrollable

gorandalum:
Users were unable to pay if they used enlarged text and selected
multiple traveller types, as the payment button came below the screen
view, and they were not able to scroll.

I won't be surprised if this also is the case on other screens. Maybe we
could create a View-component which enforced this layout somehow:
-- Header (with safe area top)
-- ScrollView
-- Footer (optional, with safe area bottom)


gorandalum:
The bug was that this view was not scrollable:
Screenshot 2021-02-05 at 13 49 26

cbrevik (approved):

PR #908, opened 2021-02-08
refactor: Standardize header button types

gorandalum:
The screen header mostly had the same functionality for its buttons. The
left buttons were either 'Back', 'Close', 'Cancel' or 'Home', and the
right button was always 'Chat'.

These button types are now standardized as a part of the ScreenHeader
props. Meaning the icon/text, onPress and accessibility is defined in
the ScreenHeader, making it not necessary to duplicate this all over the
app.

Also we were not consistent with usage of text or icons for these header
buttons. It will be easier to be consistent with this new
standardization. After this change mainly text will be used (except for
'Home' and 'Chat), since this is more inline with design guidelines, and
it will give a better differentiation between 'Close' and 'Cancel'.

A custom type may be added later for specifying buttons that is not
included in these standardized types, but as of now I found it not
necessary to add.


mikaelbr:
I wonder now if we should move ScreenHeader here to `components` to over time get a more structured component library

src/screens/Ticketing/Tickets/Tabs.tsx
```diff
@@ -23,18 +23,14 @@ type Props = {
...
   if (must_upgrade_ticketing) return <UpgradeSplash />;
 
   return (
     <View style={styles.container}>
-      {enable_recent_tickets ? (
-        <RecentTicketsScrollView />
-      ) : (
-        <View style={{flex: 1}} />
-      )}
+      <RecentTicketsScrollView />
```
mikaelbr, line 33: No longer need for feature toggle or unintended change?

gorandalum:
> I wonder now if we should move ScreenHeader here to `components` to over time get a more structured component library

Yeah. I can do that. Is there some kind of "rule" whether something should be placed under `components` or not? Why are folders like favorite-chips and location-search not under `components`?

src/ScreenHeader/HeaderButton.tsx
```diff
@@ -1,21 +1,75 @@
 import {AccessibilityProps, TouchableOpacity} from 'react-native';
 import React from 'react';
 import insets from '../utils/insets';
+import {LeftButtonProps, RightButtonProps} from './index';
```
mikaelbr, line 4: Unsure of the proper way to handle types here. Now we have a circular dependency (which works, but is kind of iffy) and it is kind of odd with proper ownership. But I also see the challenge of right vs left button. But my suggestion is to have there be a `HeaderButtonProps` like so in `HeaderButton.tsx`

```
export type ButtonModes = 'back' | 'cancel' | 'close' | 'home' | 'chat'l;
export type HeaderButtonProps = {
  type: ButtonModes;
  onPress?: () => void;
} & AccessibilityProps;
```

and in `ScreenHeader/index.tsx`

```
export type LeftButtonProps = HeaderButtonProps & {
  type: Exclude<ButtonModes, 'chat'>;
};

export type RightButtonProps = Omit<HeaderButtonProps, 'onPress'> & {
  type: 'chat';
};
```

Not perfect, but I think maybe it's better

mikaelbr (approved):

mikaelbr:
> Yeah. I can do that. Is there some kind of "rule" whether something should be placed under `components` or not? Why are folders like favorite-chips and location-search not under `components`?

Good questions. It's kind of a change in progress, but the initial thought was domain-specific and stateful things is separate modules. So things interacting directly with stores, APIs, handling data transformation, etc, would be specific modules/features, but visual reusable components ("dumber") components would be components. That is the distinction currently can be summarized as "modules" vs "components" if that makes sense?

src/screens/Ticketing/Tickets/Tabs.tsx
gorandalum: Both! I'll move it to its own PR.

gorandalum:
> Good questions. It's kind of a change in progress, but the initial thought was domain-specific and stateful things is separate modules. So things interacting directly with stores, APIs, handling data transformation, etc, would be specific modules/features, but visual reusable components ("dumber") components would be components. That is the distinction currently can be summarized as "modules" vs "components" if that makes sense?

Yeah, it made sense. And it will probably not always be black or white whether something is a component or a module. I also feel some other distinction could be made between the "smarter" React components, and the helpers/hooks for data transformation and integration.

gorandalum:
Fixed after feedback:
- Move screen header under components
- Suggested typing change
- Removed unintended removal of feature toggle

PR #913, opened 2021-02-10
fix: Fix header text overlap bug

gorandalum:
The combination of a long header title and large accessibility font
size made the title text overlap with the left button text. With this
change, the left button and right button will be rendered below the
header title if they are to wide to be on the same line.

Tried adding the same logic for the animated header, but decided not to
invest the necessary time for making it work. The animated header, and
its usage in the disappearing header, is quite complex, and changes
there can be error-prone. Instead used a simpler solution with
fontscale, but the buttons will as of now not render below the title on
the animated header.

### DIFF
--- a/src/components/screen-header/animated-header.tsx
+++ b/src/components/screen-header/animated-header.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
-import {Animated, View, ViewProps} from 'react-native';
+import {Animated, useWindowDimensions, View, ViewProps} from 'react-native';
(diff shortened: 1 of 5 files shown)


gorandalum:
Etter fiksing:
Screenshot 2021-02-09 at 15 27 40

mikaelbr:
>  animated header, and its usage in the disappearing header, is quite complex, and changes there can be error-prone

Yes. 😅

mikaelbr (approved):

PR #915, opened 2021-02-10
feat: Firebase (anonymous) Auth + Firestore fare contracts

cbrevik:
## Added:
* `@react-native-firebase/auth`
* `@react-native-firebase/firestore` 
* `@react-native-firebase/messaging` (needed for phone login)

## Changes
* Login automatically as anonymous user in Firebase
* Move nearly all ticket-related functionality to `src/tickets`
* Rewrite app to fetch tickets from Firestore
* Rewrite model for `FareContracts`, introduced concept of `TravelRight`s in the model. Also added a `type`-field with each travel right which will allow us to type guard and differentiate between the different travel right types.
* Also support "unknown" Tickets, but at the moment removed retrieval of them from Firestore

## TODOs
- [x] Differentiate between active and expired fare contracts

## Future stuff, stricken out of scope from this PR
- [ ] ~Better filter queries towards Firebase~
- [ ] ~Paginate expired tickets~
- [ ] ~Support forced refresh (in case for some reason Firebase hiccups)~


src/auth/AuthContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const signOut = useCallback(async function () {
+    await auth().signOut();
+  }, []);
+
+  useEffect(() => {
+    if (state.isInitialized && !state.user) {
+      signInAnonymously();
+    }
+  }, [state.isInitialized]);
```
gorandalum, line 129: I've ran and tested this locally now and all seem good! Really nice including the log in/out buttons for local testing.

With that said I could not get a grasp on the flow of the login process just by reading this file. I had to start debugging and printing text to understand what was really happening. I think a couple comments and some small rearrangements could make the flow a little more clear for future maintainers.

Something like this?
```javascript
  // Subscribe to user changes. Will fire a onChangeEvent immediately on subscription.
  useEffect(
    () => auth().onUserChanged((user) => dispatch({type: 'SET_USER', user})),
    [],
  );

  // Sign in if the onChangeEvent fired immediately on subscription did not include user data.
  useEffect(() => {
    if (state.isInitialized && !state.user) {
      signInAnonymously();
    }
  }, [state.isInitialized]);
  ```

File: src/tickets/types.ts [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+  startDateTime: Timestamp;
+  endDateTime: Timestamp;
+  usageValidityPeriodRef: string;
+  userProfileRef: string;
+  authorityRefs: string[];
+  tariffZoneRefs: string[];
+};
+
+export type PreactivatedSingleTicket = PreactivatedTicket & {
+  type: 'PreActivatedSingleTicket';
```
gorandalum, line 40: Doesn't it need to be `Omit<PreactivatedTicket, 'type'> & { type: ... }`?

src/tickets/types.ts
```diff
@@ -0,0 +1,138 @@
...
+  OTHER = 9,
+  CANCELLED = 20,
+  REFUNDED = 21,
+}
+
+export type TravelRight = {
+  id: string;
+  status: TravelRightStatus;
+  type:
+    | 'PreActivatedSingleTicket'
```
gorandalum, line 22: Hmm, these types, are they actually necessary to store in Firestore? 

I'm thinking especially of the distinction between single ticket, period ticket and unknown ticket. Whether the ticket is activated or not must be in Firestore.

My thinking is that other product information is not included on the fare contract, but fetched from reference data by using the `fareProductRef`. And information whether it is a single ticket or period ticket product will probably also be present there (either directly or as metadata). Also if the `fareProductRef` does not point to a product we have in our reference data it will be an unknown ticket. My thinking is just that we should have a consistent pattern for what the source of information is, but of course there may be reasons making it necessary to store this in Firestore which I am unaware of now.

src/screens/Ticketing/Tickets/Tabs.tsx
```diff
@@ -35,19 +38,21 @@ export const BuyTickets: React.FC<Props> = ({navigation}) => {
...
-          accessibilityHint={t(TicketsTexts.buyTicketsTab.button.a11yHint)}
-          onPress={() =>
-            navigation.navigate('TicketPurchase', {
-              screen: 'PurchaseOverview',
-              params: {},
-            })
-          }
-        />
-      </View>
+      {isInitialized && !!abtCustomerId && (
```
gorandalum, line 41: Could `useAuthState` return a field `isSignedIn` (or something like it) which could be used instead? When I am reading this I am not sure if `isInitialized && !!abtCustomerId` actually is the equivalent as "isSignedIn" or if it actually is something else. Doing this could also prevent exposing fields like `isInitialized` from the auth hook, hiding more of the auth complexity from the other components.

gorandalum (approved):
I have added some comments, but approving as everything works great! 👍

src/auth/AuthContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+const AuthContext = createContext<AuthContextState | undefined>(undefined);
+
+export default function AuthContextProvider({children}: PropsWithChildren<{}>) {
+  const [state, dispatch] = useReducer(authReducer, initialReducerState);
+  const {
+    preferences: {language},
+  } = usePreferences();
+
+  useEffect(() => {
+    if (language) auth().setLanguageCode(language);
```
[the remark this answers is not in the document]
cbrevik: Hm, I'm a bit unsure, seems like an indirect way to get at the language since the language returned from `useTranslation` is changed based on what is the set language in the `PreferencesContext`?

src/auth/AuthContext.tsx
cbrevik: Good idea, can add those!

src/tickets/types.ts
cbrevik: No the `type` will be forced to `PreActivatedSingleTicket` here. If I try to set anything else it will not compile.
image

package.json
```diff
@@ -30,9 +30,12 @@
     "@react-native-community/async-storage": "^2.0.0-rc.1",
     "@react-native-community/async-storage-backend-legacy": "^2.0.0-rc.1",
     "@react-native-community/masked-view": "^0.1.10",
-    "@react-native-firebase/analytics": "^7.6.7",
-    "@react-native-firebase/app": "^8.4.5",
-    "@react-native-firebase/remote-config": "^9.0.10",
+    "@react-native-firebase/analytics": "^10.5.1",
```
mikaelbr, line 33: These require cocoapods 1.10.0, no? Which is dependent on #925 ?

cbrevik: I do not get an error regarding 1.10.0 when using these, so would be cool if you could pull branch local and test :o

src/screens/Ticketing/Tickets/Tabs.tsx
cbrevik: We could probably talk about naming, but there is a difference between the two. Right now we're signing in automatically, but in the future we probably will not. So the `isInitialized` boolean signifies that we've initialized the auth-connection and attempted to fetch any existing users. In other words, when `isInitialized` is true, and there is no users signed in, you're free to attempt a sign in if you want.

The `abtCustomerId` is an id which is generated in the id-based backend after a user is signed in Firebase. This happens asynchronously, and might not be available right after a sign in. But we're unable to purchase or fetch any tickets without it, so we need to check its availability. Also we need to be able to retrieve the id itself. 

So, I'm not sure if a general `isSignedIn` from context will solve the issue in full. But I could distill it down further up in this component?

src/auth/AuthContext.tsx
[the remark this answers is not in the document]
gorandalum: My thinking was that we have logic in LanguageContext for determining the app language. Some weeks ago the logic was that Norwegian was default if no preferences set, but it was later changed to that the system language was default if no preferences set. Maybe later it will go back to being Norwegian as default if we include an easy way to switch to English in the onboarding dialog, as have been talked about. By using `language` from `useTranslation` we will piggyback that logic also here, if we want the auth dialog to always be in the same language as the app language.

src/tickets/types.ts
cbrevik: The generic nature of the travel rights in the id-based backend. As of now, you're right we could probably deduce from the data if it is a single ticket or period ticket. But later when we get things like travel cards, and "klippekort", not to mention other modes of transport or travel rights from other providers (than AtB), the deducing and complexity becomes intractable. So our thinking long-term is that if we let the backend deduce the ticket type, it allows us to make assumption about which data is included in the payload depending on the type.

src/tickets/types.ts
cbrevik: But might be another reason to use `Omit` that I'm not seeing instantly?

src/screens/Ticketing/Tickets/Tabs.tsx
cbrevik: I could maybe return a `isSignedIn` boolean depending on if `abtCustomerId` is set. But technically you are "signed in" to Firebase auth before `abtCustomerId` is set. And if generating that id is slow or fails, then `isSignedIn` would be false, even if it technically should be true. You do have a Firebase auth user which can be signed out, or do other things with.

src/tickets/types.ts
gorandalum: Ok. It may just be Intellij which does not handle this optimally then. Without the `Omit` it resolves usage of `type` to have two possible declarations, both the one from `PreactivatedTicket` and the one from `PeriodTicket`. Although it I see that the typescript compiler gives no error.

src/auth/AuthContext.tsx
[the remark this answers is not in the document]
cbrevik: Alright, I can see that argument, preferences does not have logic itself for deciding language. I'll change it 👍🏻

package.json
mikaelbr: Seems as `@react-native-firebase/app@10.8.0` depends on `Firebase/CoreOnly ~> 7.6.0` which in turn depends on pods 1.10.0. But `10.5.0` as used here is good. Though there are some inconsistencies when merging with the main branch now.

src/screens/Ticketing/Tickets/Tabs.tsx
gorandalum: The suggestion `isSignedIn` was based on my interpretation of `isInitialized && !!abtCustomerId`, which I now understand has more nuances that just being signed in. But I think my misinterpration gives even more incentive to find a better named field for the logic represented by `isInitialized && !!abtCustomerId`, and maybe with some jsdoc.

src/tickets/types.ts
gorandalum: Sounds good, as long as it is a deliberate decision I'm happy!

src/screens/Ticketing/Tickets/Tabs.tsx
cbrevik: Sure, added some more comments and better naming for `isInitialized`

src/tickets/types.ts
cbrevik: It doesn't give an error for you? What's the code you're trying? For the above code I'm getting the following when trying the compiler:
```bash
∟ yarn tsc
yarn run v1.22.10
$ /Users/christianbrevik/Dev/AtB/mittatb-app/node_modules/.bin/tsc
src/tickets/types.ts:48:3 - error TS2322: Type '"UnknownTicket"' is not assignable to type '"PreActivatedPeriodTicket"'.

48   type: 'UnknownTicket',
     ~~~~

  src/tickets/types.ts:21:3
    21   type:
         ~~~~
    The expected type comes from property 'type' which is declared here on type 'PeriodTicket'
```

gorandalum: Yeah, I get the error, so no problem there. It is when reading the `type` it is not handled optimally in Intellij, as it interprets that `type` has two definitions. But no biggie, the typescript compiler is not confused!

Screenshot 2021-02-15 at 09 58 05
Screenshot 2021-02-15 at 09 58 24

mikaelbr (approved):

PR #917, opened 2021-02-11
feat: Default tariff zone from location

gorandalum:
If the user is sharing his location, and the location in within a
tariff zone, use this tariff zone as the default tariff zone when
purchasing ticket. Else the default tariff zone will be the first tariff
zone in RemoteConfig.

There was talks about letting the user select tariff zone based on his
nearest stops, to reduce the risk of selecting a initial tariff zone
that does not include the actual stop the user will travel from. This
may be implemented later if necessary.


mikaelbr (approved):
LGTM

cbrevik (approved):

PR #918, opened 2021-02-11
fix: Long text in LinkItem should not move icon

gorandalum:
Before:
Screenshot 2021-02-11 at 12 29 17
After: 
Screenshot 2021-02-11 at 13 15 28

### DIFF
--- a/src/components/sections/link-item.tsx
+++ b/src/components/sections/link-item.tsx
@@ -11,6 +11,7 @@ import NavigationIcon, {
   NavigationIconTypes,
 } from '../theme-icon/navigation-icon';
 import {SectionItem, useSectionItem, useSectionStyle} from './section-utils';
+import {StyleSheet} from '../../theme';
 
 export type LinkItemProps = SectionItem<{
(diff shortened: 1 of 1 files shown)


src/components/sections/link-item.tsx
```diff
@@ -56,3 +58,8 @@ export default function LinkItem({
     </TouchableOpacity>
   );
 }
+
+const useStyles = StyleSheet.createThemeHook(() => ({
```
mikaelbr, line 62: could this be part of `useSectionStyle`? Or should it just be specific of this particular section item?

gorandalum: You tell me! It is probably not only this section item which may overflow, but moving `flex: 1` as a field of `sectionStyle` does not fix other section items automatically. 

Should I expand this PR to also fix other sections items?

mikaelbr: Hm. I actually think it might be extended. But let's not handle this right now.

PR #924, opened 2021-02-12
refactor: adds TS/Babel paths and @atb/ scope for module resolution

mikaelbr:
Simple enough, changes from:

```ts
import {StyleSheet} from '../../../../../theme';
```

to 

```ts
```

...but for all modules and packages for `src/` root.

This serves multiple purposes:

- Makes it easier to read and navigate through the built in modules.
- Makes it easier to refactor/move modules.
- Makes it easier to scale packages to external packages. E.g. `@atb/theme`, `@atb/sdk` and `@atb/components`.

Could do this as two separate PRs if wanted (one proposing the change the other one implementing on all files).


src/components/sections/favorite-departure-item.tsx
```diff
@@ -6,10 +6,10 @@ import {
...
-import SvgDelete from '../../assets/svg/icons/actions/Delete';
-import {StoredFavoriteDeparture} from '../../favorites/types';
-import {useTheme} from '../../theme';
-import {SectionTexts, useTranslation} from '../../translations';
+import SvgDelete from '@atb/assets/svg/icons/actions/Delete';
+import {StoredFavoriteDeparture} from '@atb/favorites/types';
+import {useTheme} from '@atb/theme';
+import {SectionTexts, useTranslation} from '@atb/translations';
 import {screenReaderPause} from '../accessible-text';
 import ThemeText from '../text';
```
gorandalum, line 14: The logic here is that if the import is just one level down then `@atb/` is not used?

gorandalum: down = up

mikaelbr: Not explicitly there might be somewhere it's missing (especially with `../` as I didn't manually go through every file). But the logic should be:

- If it is part of module/component: use relative
- If it imports something from another module/component: use absolute.

gorandalum (approved):

src/components/sections/favorite-departure-item.tsx
mikaelbr: FWIW, I'm updating for `components/` now

cbrevik (approved):

PR #927, opened 2021-02-15
chore: fixes version consistencies.

mikaelbr:
This was... not pleasant. But the solution itself is simple: After 10.x firebase packages finally match with peerdeps to /app on the same version number. Seems 10.1.0 doesn't require any native libs that require cocoapods 1.10.0. So this seems good enough.


mikaelbr:
Also ran `npm fix audit` (which changes the lock files as yarn doesn't support audit fix itself we have to go through `package-lock.json`) to try to patch node-fetch warning. But this is the reason for why the lock files are totally changed.

cbrevik (approved):
This is what makes developing fun! 🤪

gorandalum (approved):

PR #928, opened 2021-02-16
feat: New purchase flow for period ticket

gorandalum:
Updated the purchase flow in the app to support the design for period
tickets.

Main changes:
- On the "Buy"-tab there is now one button for buying single ticket and
  one button for buying period ticket.
- If buying period ticket only adult, child and senior are available
  user profiles.
- If buying period ticket only one traveller is allowed. This is
  enforced by making the traveller selection a radio section.

Some other notes:
- The filtering of allowed user profiles are now done in the app. This
  should later come from backend (and preferebly Entur) modelled in the
reference data.
- Tried to implement this with as little changes as possible to the
  existing implementatio, meaning even though the visual representation
  of traveller selection only allows a single traveller for period
  tickets, it is backed by the same underlying model which supports
  multiple travellers.
- Not yet implemented selection of future start time. This will be done in 
  its own PR.


gorandalum:
Setting this as draft for now, until the distinction between single ticket and period ticket is added to RemoteConfig.

gorandalum:
Some pictures:
<p float="left">
Screenshot 2021-02-16 at 10 41 19
Screenshot 2021-02-16 at 10 41 27
Screenshot 2021-02-16 at 10 41 41
</p>

gorandalum:
@monopoint Even though @cbrevik and I mainly handles the ticketing tasks we still add you and Mikael as reviewers so you can follow what is happening. Whether or not to do a thorough review is optional.

cbrevik (approved):
LGTM, clean change for introducing single/multiple selection of users. Was wondering if we should do a feature toggle on the "buy"-button for period or not. Depending on if we want it out in production right away next week? We'll have to land the phone login and bring that with in that case?

gorandalum:
Good point. I am merging now, and if a feature toggle is necessary I'll add it in its own PR.

PR #934, opened 2021-02-18
feat: Logging in with phone

gorandalum:


ios/atb/Info.plist
```diff
@@ -32,6 +32,14 @@
...
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>com.googleusercontent.apps.827196677776-fft65c0v4jp8u7fdlqo03ep1u0n66fhf</string>
+			</array>
+		</dict>
```
cbrevik, line 42: Where does this come in?

gorandalum: It's for setting up recaptcha verification on iOS. Described here:
https://firebase.google.com/docs/auth/ios/phone-auth#set-up-recaptcha-verification

cbrevik: Aha. That value seems to be unique per bundle identifier (`debug`, `staging`, `store`), so we'll probably need to override it per environment then too: https://github.com/AtB-AS/mittatb-app/blob/8b249184bdf8c14ec15b1a4b92559dba77e83af7/scripts/override-config-files.sh#L16

cbrevik: On iOS, we have a unique `GoogleService-Info.plist` per env (see `ios`-folder, as well as each folder under `env`). So we can probably extract `REVERSED_CLIENT_ID` from those, then set it in our `Info.plist` as with the other API-keys, using `PlistBuddy`

cbrevik:
image

Got this first time logging in on Simulator, do you notice it too? Seems to be user access issues when accessing Firestore-snapshot of tickets. Haven't found out what causes it yet, but probably a race condition between users (anonymous and phone user), and which has access to what?

src/screens/Profile/Login/ConfirmCode.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <View style={styles.container}>
+      <FullScreenHeader title={'Logg inn'} leftButton={{type: 'back'}} />
+
+      <View style={styles.mainView}>
+        <Sections.Section>
+          <Sections.TextInput
+            label={'Kode'}
+            placeholder="Bekreftelseskode fra SMS"
+            onChangeText={setCode}
```
cbrevik, line 35: Should also use phone pad here, as well as hint to the OS that this input expects an OTP

```suggestion
            onChangeText={setCode}
            keyboardType="phone-pad"
            textContentType="oneTimeCode"
```

cbrevik (changes requested):
A couple of suggested changes + something we should discuss

src/screens/Profile/Login/PhoneInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export default function PhoneInput({navigation}: PhoneInputProps) {
+  const {t} = useTranslation();
+  const styles = useThemeStyles();
+  const {signInWithPhoneNumber} = useAuthState();
+  const [phoneNumber, setPhoneNumber] = useState('');
+
+  const onNext = async () => {
+    await signInWithPhoneNumber(phoneNumber);
```
cbrevik, line 25: Might want to trigger a loading event/indicator before this. It takes a bit of time to navigate to the next screen.

cbrevik:
Testing on Android and works great with the test number! 🌮 🎉 

I don't have a SIM for my test-device, so unable to test a real OTP. But I'm sure one of the testers in QA when we get there can verify if it works.

src/screens/Profile/Login/ConfirmCode.tsx
gorandalum: Fixed

src/screens/Profile/Login/PhoneInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <View style={styles.container}>
+      <FullScreenHeader title={'Logg Inn'} leftButton={{type: 'cancel'}} />
+
+      <View style={styles.mainView}>
+        <Sections.Section>
+          <Sections.GenericItem>
+            <ThemeText>Telefonnummer</ThemeText>
+          </Sections.GenericItem>
+          <Sections.TextInput
+            label={'+ 47'}
```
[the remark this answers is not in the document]
gorandalum: Petter suggested using a dropdown with country codes, but hardcoding it to +47 is good enough at this time.

src/screens/Profile/Login/PhoneInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+            value={phoneNumber}
+            onChangeText={setPhoneNumber}
+            showClear={true}
```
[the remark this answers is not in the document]
gorandalum: Fixed

ios/atb/Info.plist
gorandalum: Fixed

src/screens/Profile/Login/PhoneInput.tsx
gorandalum: Fixed

gorandalum:
> Got this first time logging in on Simulator, do you notice it too? Seems to be user access issues when accessing Firestore-snapshot of tickets. Haven't found out what causes it yet, but probably a race condition between users (anonymous and phone user), and which has access to what?

Yeah. I saw this too. Seems to me as a user access issue right at the time were the user changes from logged in as anonymous too logged in with phone or vica versa. What should be done? Resetting the `abtCustomerId` when changing login method?

cbrevik (approved):
As noted on Slack, translations + `Info.plist` should be done before merging.  Other than that it LGTM 💯. Good job!

src/screens/Profile/Login/PhoneInput.tsx
mikaelbr: Maybe check out `useIsLoading` (in `@atb/utils/use-is-loading.ts`)

src/translations/screens/subscreens/Login.ts
```diff
@@ -0,0 +1,54 @@
+import {translation as _} from '../../commons';
+const LoginTexts = {
```
gorandalum, line 2: @magnusolderoesaether @pettersuul 
You can take a look at the texts and the translations in this file.

PR #935, opened 2021-02-19
fix: Attempt to only dynamically link MapboxEvents

cbrevik:


cbrevik:
This seems to do the trick, tested archiving locally and it worked fine! Tested the build on device and sim, and can't seem to find any fault with the map yet. I suggest merge and test it in QA env.

cbrevik:
This means you have to update your CocoaPods to 1.10.1 cc @gorandalum @mikaelbr

gorandalum (approved):

PR #943, opened 2021-02-25
fix: adds custom lint rules to avoid simple mistakes

mikaelbr:
This adds 3 custom linting rules to make some implicit project knowledge explicit. Namely:

1. We should always use `@react-navigation/native` instead of `@react-navigation/core`. Both work but native is more performant and is best practice. VSCode often suggests `/core`. This lint rule gives a warning when not using `/native` and provides autofix.
1. We should always use `SafeAreaView` from `react-native-safe-area-context` instead of `react-native`. SafeAreaView from react-native calculates padding wrong currently (at least for our use case). 
1. Texts from `/translate` modules should always be wrapped in `t` function provided from `@leile/lobo-t`. This is a naive implementation that doesn't actually check if given identifier `t` is from the correct function. Also the AST traversing is simple and does not account for storing result in variables. This could be done, but ROI etc.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,6 @@
+const rulesDirPlugin = require('eslint-plugin-rulesdir');
+rulesDirPlugin.RULES_DIR = 'eslint-rules/';
+
 module.exports = {
   env: {
     es6: true,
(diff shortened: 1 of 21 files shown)


mikaelbr:
Marked som draft. Want to quickly add checkf or accessibility touchables.

gorandalum (approved):
Great! LGTM :+1:

PR #946, opened 2021-02-26
fix: modal transitions for android

monopoint:
Modal transitions default to scale animation on android and are not consistent with design intentions.
By explicitly using slidefromBottom transitions from IOS on the Assistant screen and slidefromRight transitions on Profile screen we get consistent transitions that are in line with the design.

The disappearing header has its own problems and are handled elsewhere.
Some modals still use scale on Android when they are not out of line with the design (tickets, maps etc)

### DIFF
--- a/src/navigation/index.tsx
+++ b/src/navigation/index.tsx
@@ -21,7 +21,7 @@ import {
   NavigatorScreenParams,
   useLinking,
 } from '@react-navigation/native';
-import {createStackNavigator} from '@react-navigation/stack';
+import {createStackNavigator, TransitionPresets} from '@react-navigation/stack';
 import React, {useEffect, useRef} from 'react';
 import {StatusBar} from 'react-native';
 import {Host} from 'react-native-portalize';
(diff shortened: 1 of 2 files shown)


src/navigation/index.tsx
```diff
@@ -83,43 +83,21 @@ const NavigationRoot = () => {
                 <Stack.Screen
                   name="LocationSearch"
                   component={LocationSearch}
-                  options={{
-                    transitionSpec: {
-                      open: transitionSpec,
-                      close: transitionSpec,
-                    },
-                  }}
+                  options={{...TransitionPresets.ModalSlideFromBottomIOS}}
```
mikaelbr, line 86:
```suggestion
                  options={TransitionPresets.ModalSlideFromBottomIOS}
```

Here you can just use the object directly instead of spreading it.

src/navigation/index.tsx
```diff
@@ -83,43 +83,21 @@ const NavigationRoot = () => {
...
                 <Stack.Screen
                   name="AddEditFavorite"
                   component={AddEditFavorite}
-                  options={{
-                    transitionSpec: {
-                      open: transitionSpec,
-                      close: transitionSpec,
-                    },
-                  }}
+                  options={{...TransitionPresets.ModalSlideFromBottomIOS}}
```
mikaelbr, line 100:
```suggestion
                  options={TransitionPresets.ModalSlideFromBottomIOS}
```

Same as previously

src/navigation/index.tsx
```diff
@@ -83,43 +83,21 @@ const NavigationRoot = () => {
...
                 />
                 <Stack.Screen
                   name="TicketModal"
                   component={TicketModalScreen}
-                  options={{
-                    transitionSpec: {
-                      open: transitionSpec,
-                      close: transitionSpec,
-                    },
-                  }}
```
mikaelbr, line 111: Should this have different behavior?

mikaelbr (changes requested):

mikaelbr:
Added some suggestions. Nitpicky and strictly not needed, but as you've marked this as a draft I thought it might as well a request for change.

mikaelbr:
@monopoint Is this still a draft?

mikaelbr:
Bump 😬

monopoint:
@mikaelbr sorry for the wonky workflow, still getting used to it.
I agree to the suggestions and have merged them.  
Object spread for transitions were used elsewhere in the code base, and I agree that we should use objects directly.
reverted transition specs for login screen. It looked like the settings were identical as defaults, but I might be wrong. Anyway we can keep them for consitency.

mikaelbr (approved):

mikaelbr:
Typically the workflow here is with 1 approved the PR Author can squash and merge the changes.

PR #948, opened 2021-02-26
feat: Use v2 endpoints of ticket

gorandalum:


gorandalum:
The original v1 endpoints are reverted to work as before, without passing bearer token, to work with old versions of the app. The current version of the app has to use the new v2 endpoints for recent tickets and reserving an offer since it is passing bearer token for identifying the user.

gorandalum:
The original v1 endpoints are reverted to work as before, without passing bearer token, to work with old versions of the app. The current version of the app has to use the new v2 endpoints for recent tickets and reserving an offer since it is passing bearer token for identifying the user.

cbrevik (approved):

PR #949, opened 2021-02-26
fix: makes entire result item in Assistant pressable

mikaelbr:
This PR fixes issue with only "Details" being pressable, after several feedbacks and discussions. The latest feedback can be seen at https://app.intercom.com/a/apps/vdemedo2/inbox/inbox/4207849/conversations/159245400004398.

Also, the primary border in dark mode was way too contrasty. Although this is another issue I fixed this in the same go. I can separate this into two different PR-s if necessary, but I was lazy. Sorry!

## Examples
(Attached image of the Map-screen as it also uses border primary style.)

Screenshot 2021-02-26 at 16 18 03 

Screenshot 2021-02-26 at 16 26 05

### DIFF
--- a/src/components/map/MapControls.tsx
+++ b/src/components/map/MapControls.tsx
@@ -14,48 +14,45 @@ const MapControls: React.FC<Props> = ({zoomIn, zoomOut}) => {
   const styles = useStyles();
   const {t} = useTranslation();
   return (
-    <View>
-      <View style={styles.zoomContainer}>
-        <TouchableOpacity
-          onPress={zoomIn}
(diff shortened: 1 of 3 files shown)


mikaelbr:
@pettersuul I've here chosen the border to be `gray_600` for dark mode. I couldn't find any definitions in Figma, but `#fff` as it is now is way too contrasty I think. What do you think about the attached screenshots?

pettersuul (approved):
Looks good to me!

gorandalum (approved):

PR #953, opened 2021-03-01
feat: Fix upcoming tickets visual representation

gorandalum:


gorandalum:
Note: The jury is still out on "upcoming ticket", "future ticket" or another term should be used for these tickets with a future validity. Will not merge until this is decided.

cbrevik (approved):

PR #954, opened 2021-03-01
feat: Fix validity elements

gorandalum:
The validity icon, text and line were not fully synced with the design.
In addition there was a bug where the expired timestamp were wrong
because of a seconds/milliseconds error.


gorandalum:
After this fix, from top to bottom: `reserving`, `upcoming`, `valid`, `expired (less than 1h since expiration)`, `expired (more than 1h since expiration)`, `refunded` and `unknown`.

Screenshot 2021-03-01 at 16 21 43

mikaelbr:
#955 should simplify the coloring of the icon here.

mikaelbr (approved):

mikaelbr (approved):

PR #964, opened 2021-03-02
feat: Use popup/action-sheet for selecting product

gorandalum:
The is done to be inline with the design in Figma.

Some notes:
- Only implemented for product selection for now
- Needed some refactoring of the HeaderButton, making it not use
  'useNavigation' when an onPress handler is provided. The reason for
this is that 'useNavigation' fails in the popup component at it is
detached from the current stack.
- Grouped both the SectionLink used in PurchaseOverview and the actual
  popup in the same Component. This is a simplification done to minimize
  the necessary passing of refs between different Components.


gorandalum:
Screenshot 2021-03-02 at 12 50 50

mikaelbr:
Lazy me for not checking myself, but have you done any screen reader tests here? We used portal/modal previously for date input but I think there were some issues with accessibility with the window itself. Maybe I'm misremembering but I don't think it was the date component that was the entire problem, but modals and screen reader flow also. Do you remember @hildeor ? Can also search through the issues

gorandalum:
I've talked with @cbrevik about this, and yes there is screen reader issues. Working on a solution using https://gorhom.github.io/react-native-bottom-sheet/.

mikaelbr:
Yes, I've been doing a lot of research for emoji screen and landed on that implementation also. 👍

gorandalum:
I've updated this to use [react-native-bottom-sheet](https://gorhom.github.io/react-native-bottom-sheet) now. You can try it out when you have time @mikaelbr @cbrevik.

Some findings:

- The elements of the bottom-sheet are now accessible by screen reader. However, they are not in the "read-flow" of the screen reader when reading elements from top to bottom. They have to be clicked for the screen reader to announce them.
- The "Accessibility support" of the library is some hardcoded accessibility announcements, like [this](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L562). We have to find a way to disable them. Do you know any trick to disable screen reader announcements from libraries? Other alternative can be creating a PR to the library for adding a parameter which can disable accessibility announcements.

gorandalum:
I did some more testing and also tried:
[react-native-scroll-bottom-sheet](https://github.com/rgommezz/react-native-scroll-bottom-sheet/)
[react-native-raw-bottom-sheet](https://github.com/nysamnang/react-native-raw-bottom-sheet/):
[react-native-reanimated-bottom-sheet](https://github.com/osdnk/react-native-reanimated-bottom-sheet/)

My conclusion after trying them all is that neither works perfectly. It seems [react-native-bottom-sheet](https://gorhom.github.io/react-native-bottom-sheet) is still the best alternative, we just need to suppress its internal accessibility announcements.

And also we should maybe consider that the best solution is that we just keep using subscreens for the different selections during ticket purchase. We can live with an emoji picker having substandard screen reader, but for these ticket purchase selections the screen reader should be on point.

cbrevik:
Tested some now, and seems performant and animates in nicely. I didn't get the chance to test w/ Screen Reader yet though. 

Is the backdrop color supposed to be blue btw? Can't remember if this is standard or not.

image

gorandalum:
No, the backdrop color is just not changed from the example code from the library docs. Will fix this before eventual merge.

I've tried a total of five libraries now, and in all of them I pretty easily got the design and animation I wanted. It is just that the screen reader does not perform optimally in neither of them. As of now I am leaning towards keeping the subscreens; we save time and get good accessibility support.

mikaelbr:
I think it's also a possibility to "grap focus" to fix the issue with the flow. Look into `AccessibilityInfo.setAccessibilityFocus`. I can't remember at the moment if we've used it with success or not.

gorandalum:
Pushed some more fixes:
 
- Patched the react-native-bottom-sheet for removing default announcements. This worked great.
- `AccessibilityInfo.setAccessibilityFocus` worked good for setting accessibility on a specific element after bottom sheet opened
- Created a reusable component for BottomSheet, abstracting away most of the ref usage etc.
- Still could not get bottom sheet to have a "read-flow" for VoiceOver. Swiping with two fingers up (read-from-top) or two fingers down (read from cursor to bottom) does not work.

All in all, I will say this was a 90% success.

Then I tried Android:
- When opening purchase overview there is some flickering related to the BottomSheet component. Probably because of the code making the BottomSheet height fit its contents.
- `AccessibilityInfo.setAccessibilityFocus` not working. App crashing probably because of some timing issue as it says the component to set focus on is not found
- Clicks on the BottomSheet are most often not working. The onPress-animation is displayed on TouchableOpacity, but it has no effect. If clicking furiously the click will eventually be registered.
- Some other small bugs

I tried to start to fix some of these Android issues, but this reintroduced other bugs to appear on iOS. I've now probably spent hours totaling a work-week on this and still feel that there is much work left for a good enough solution.

PR #976, opened 2021-03-05
feat: Ticket design with inspection symbol

gorandalum:


gorandalum:
<div float="right">
Screenshot 2021-03-05 at 11 10 58
Screenshot 2021-03-05 at 11 10 47
Screenshot 2021-03-05 at 11 10 36
</div>

src/components/theme-icon/theme-icon.tsx
```diff
@@ -23,6 +23,7 @@ const ThemeIcon = ({
   const settings = {
     fill,
     height: theme.icon.size[size],
+    width: theme.icon.size[size],
```
gorandalum, line 26: Changing just the height had no effect when increasing size, although it does have effect when decreasing size. I couldn't figure out why, and had to also change the width to enable the size increase. From what i can see this should be safe, as all icons have a size of 20x20.

mikaelbr: Could it be some trickery with `preserveAspectRatio` could do the trick?

gorandalum: I played around a little with `preserveAspectRatio`, but could not find a setting which fixed this. With `meet` the icon did not increase in size, with `slice` it did actually increase, but was sliced at width 20. I don't fully understand this, it behaves like the viewBox is 20x20, but the svg file has `viewBox="0 0 40 40"`.

cbrevik: I think this is fine. I think the reason only setting one had no effect is because we explicitly set both the SVG-icon height/width to 20 in the component?

cbrevik (approved):
This looks good to me 👍🏻 Gonna be fun to see in the app

PR #977, opened 2021-03-05
feat: Use RemoteConfig as source for allowed users

gorandalum:
The allowed user profiles was hardcoded in the app, but with this change
it is retrieved from RemoteConfig. The values in RemoteConfig is
populated from Entur.


cbrevik (approved):
LGTM 👌🏻

PR #981, opened 2021-03-08
feat: Feature toggle for login and period ticket

gorandalum:
Created two feature toggles, one for login functionality, and one for
enabling purchase of period tickets.

- If both toggles are enabled there is login functionality under
  "Mitt AtB", and the user will be prompted to login if he tries to buy
  period tickets.
- If only login toggle is enabled the login functionality is present
  under "Mitt AtB", but there is no buy button for period tickets.
- If only toggle for period ticket is enabled then no login
  functionality will be present under "Mitt AtB". The buy button for
  period tickets will be available, and will not open a login prompt
  before ticket purchase.

### DIFF
--- a/src/RemoteConfigContext.tsx
+++ b/src/RemoteConfigContext.tsx
@@ -20,6 +20,8 @@ export type RemoteConfigContextState = Pick<
   | 'enable_i18n'
   | 'enable_creditcard'
   | 'enable_recent_tickets'
+  | 'enable_login'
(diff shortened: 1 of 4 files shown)


cbrevik (approved):
Ah clever, didn't think that we might want to do the third scenario with no login, but period ticket enabled. Good job 👍🏻

cbrevik:
Yep 👌🏻

PR #988, opened 2021-03-09
feat: VAT (mva) defined in RemoteConfig

gorandalum:
Because of corona the VAT is temporary reduced from 12% to 6%. After
this PR we can change the VAT percent in RemoteConfig instead of having
an app release.


gorandalum:
Tested by changing default config locally. Also with decimals:
Screenshot 2021-03-09 at 13 31 04
Screenshot 2021-03-09 at 13 30 29
Screenshot 2021-03-09 at 13 30 09

PR #994, opened 2021-03-10
fix: flex layout supporting shorter screens

monopoint:
### Observation
The onboarding screens do not behave well on shorter screens, spesifically androids, eg. nexus 4.
The screens implement a scroll view for text info and step navigation, resulting in step navigation being pushed out of the screen with no visible cue (e.g. scroll bar) to guide the user forward.
### fix
re-implent views with working flex-layout, shrinking the illustration to make space for text and navigation when vertical space is limited.

### DIFF
--- a/src/screens/Onboarding/components/Illustration.tsx
+++ b/src/screens/Onboarding/components/Illustration.tsx
@@ -15,9 +15,10 @@ const Illustration: React.FC<IllustrationProps> = ({Svg}) => {
 };
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   svgContainer: {
-    backgroundColor: theme.background.level2,
+    backgroundColor: theme.background.level1,
     width: '100%',
     maxHeight: '50%',
+    flexShrink: 1,
   },
 }));
 export default Illustration;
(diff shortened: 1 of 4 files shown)


mikaelbr:
FYI: renamed this to [match conventional commits](https://github.com/AtB-AS/mittatb-app/blob/master/CONTRIBUTING.md#conventional-commits). We typically squash and merge to the main branch and PR titles represent the change to the code base and should as such adhere to commit requirements. The commits them self doesn't always need to match the specified format as long as the actual changes to the main branch (merge) does.

monopoint:
did another refactor because the layout hierarchy wasn't really helpful, and resulted in cropping of the illustration at the top even when the flex layout was respected. layout is now much cleaner.
also changed the background colors to match the design.

monopoint:
bug-nexus4-largest_font
bug-nexus4-normal_font
bug-pixel3a-normal_font
bug-pixel3a-largest_font

Before pictures. larger fonts would push content out of the view and background color is too dark.

monopoint:
**After pix**
fix-nexus4-normal_font
Nexus 4 with normal font settings

fix-nexus4-largest_font
Nexus 4 with largest font settings

fix-pixel3a-normal_font
Pixel 3a with normal font settings

fix-pixel3a-largest_font
Pixel 3a with largest font settings

mikaelbr (approved):
LGTM for now. We have to rewrite this flow at a later point for redesigned onboarding in any case but this is an improvement to the existing layout where the button can be outside of view and/or with scrollbar.

PR #1001, opened 2021-03-16
feat: App in dev/staging uses new staging GCP

gorandalum:


src/api/client.ts
```diff
@@ -54,6 +64,17 @@ function requestHandler(config: AxiosRequestConfig): AxiosRequestConfig {
   return config;
 }
 
+async function requestIdTokenHandler(config: AxiosRequestConfig) {
+  if (config.authWithIdToken) {
+    const retryCount = (config['axios-retry'] as any)?.retryCount;
```
gorandalum, line 69: Feels a little 'hackish' to use this not-exposed variable, but it is what I found was suggested from the library maintainer in several issues on their Github, like here:
https://github.com/softonic/axios-retry/issues/75#issuecomment-502151719

cbrevik: You could extend the type like we do with the retry-boolean: https://github.com/AtB-AS/mittatb-app/blob/a43ff967ecc5babec866e9b8f8f15b6810160abb/src/api/client.ts#L14

Like:
```ts
declare module 'axios' {
  export interface AxiosRequestConfig {
    // Should retry if network error or 5xx idempotent request
    retry?: boolean;
    ['axios-retry']?: { retryCount: number };
  }
}
```

Would perhaps be a little better?

gorandalum: I did try something like that, but could not get it to work as `axios-retry` is already defined on `AxiosRequestConfig`. I tried for example:
```
['axios-retry']: (IAxiosRetryConfig & {retryCount: number}) | undefined;
```
But got "TS2687: All declarations of '['axios-retry']' must have identical modifiers."

Didn't invest too much time into it as I figured it was no significant difference in doing it inline or changing the config.

cbrevik: Alright! I think this is fine, only issue might be if you get an 500 error on a idempotent route and Firebase Auth tries to re-auth 3 times for every user. We might get throttled real quick

cbrevik (approved):
Looks good to me! Let's get it into QA for further testing

src/api/client.ts
gorandalum: Good point @cbrevik! I did look into this now and think I found a solution which both fixes this and removes the "hackish" access of `retryCount`. Take a look at the latest commit.

cbrevik (approved):
Awesome! This is good to go!

PR #1004, opened 2021-03-17
feat: adds recently searched journeys to location search and ability to clear history. Closes #673 

### LINKED ISSUE #673: [Feature] Add last searched journeys
## Origin

See issue #420

## Motivation

Make it easier to search for frequently used journeys.

## Hypotheses and assumptions

The user may want to reuse a previous journey search which is not relative to the current location.

## Proposed solution

When using the assistant to search for the departure-point, show a list of recently used complete (origin to destination) searches.

Figma / App 1.0 / Action / Search https://www.figma.com/file/WsvD8b5PnUwvvRJiyRe6I9/App-1.0?node-id=1839%3A7190

image

mikaelbr:
Closes #673

This PR adds the ability to re-use previous journey searches.

With this, I also added the ability to clear search history as a part of privacy. I feel this is more important now with actual journey searches, even though it is still only stored locally. (I also needed it for testing).

### Example screen shots

Screenshot 2021-03-17 at 22 47 10
Screenshot 2021-03-17 at 22 47 06t of
Screenshot 2021-03-17 at 22 46 46

### DIFF
--- a/src/components/sections/header-item.tsx
+++ b/src/components/sections/header-item.tsx
@@ -1,7 +1,7 @@
+import ThemeText from '@atb/components/text';
 import React from 'react';
 import {View} from 'react-native';
-import ThemeText from '@atb/components/text';
-import {useSectionItem, SectionItem, useSectionStyle} from './section-utils';
+import {SectionItem, useSectionItem} from './section-utils';
 
 export type HeaderItemProps = SectionItem<{
   text: string;
@@ -15,7 +15,6 @@ export default function HeaderItem({
   ...props
 }: HeaderItemProps) {
   const {contentContainer, topContainer} = useSectionItem(props);
-  const style = useSectionStyle();
 
   return (
     <View style={topContainer}>
--- a/src/location-search/JourneyHistory.tsx
+++ b/src/location-search/JourneyHistory.tsx
@@ -13,6 +13,9 @@ type JourneyHistoryProps = {
   onSelect: (journey: JourneySearchHistoryEntry) => void;
 };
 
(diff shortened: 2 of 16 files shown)


mikaelbr:
Limits the current latest journey suggestions to 3 for now. Do we think this needs to be configurable through remote config so we can update that on the fly? cc @pettersuul @hildeor

gorandalum:
Nice. I'll take a look at this now.

Would it make sense to have these "Most recent journeys" on the Assistant screen also? I feel they kinda belong there more than the favourite places which are there now.

Screenshot 2021-03-19 at 08 25 36

gorandalum (changes requested):
Great, this will be a nice new functionality for the users.

I feel that the code is kind of hard to follow, but maybe it is no way around it when we give LocationSearch this added functionality. One other hint that the code is complex is that it all looked ok when reviewing, but when testing I found one or more bugs (I don't know if it is the same one).

Bug 1:
I first selected a recent journey between Kvennhusveien 8 (My position) and Ila. After that I tried changing the from-location, but after selecting Solsiden the from-location was still Kvennhusveien 8:


Bug 2:
I first selected a recent journey between Ila and Solsiden. After that I changed the to-location to Hallset, and then the from-location was unintentionally changed to my position (Kvennhusveien 8):

mikaelbr:
@gorandalum Simplified logic now and seems to fix the issue also. Check if the logic is easier to follow now.

gorandalum:
LGTM! :+1:
And no insects found when testing.

gorandalum (approved):

PR #1005, opened 2021-03-18
fix: Handle possible automatic login on Android

havwig:


src/screens/Profile/Login/ConfirmCode.tsx
```diff
@@ -64,6 +69,20 @@ export default function ConfirmCode({navigation, route}: ConfirmCodeProps) {
     }
   };
 
+  // User might be automatically logged in with Firebase auth, but only on Android
+  // Check user from state and see if it is updated while we wait
+  useEffect(() => {
+    if (user?.phoneNumber) {
```
gorandalum, line 75: I would prefer to let `authenticationType` instead of `user` to decide this. User is a FirebaseAuth internal type that I think we should use sparingly in other parts of the code. I have been thinking of removing if from the returned auth state, just haven't got time yet.

src/screens/Profile/Login/ConfirmCode.tsx
```diff
@@ -64,6 +69,20 @@ export default function ConfirmCode({navigation, route}: ConfirmCodeProps) {
...
+      // OK To use toast here, as this is Android only for now
+      ToastAndroid.showWithGravity(
```
gorandalum, line 77: I am unfamiliar with this ToastAndroid. Will look into this, and it also should be run by the designers. Do you have a screenshot?

src/screens/Profile/Login/ConfirmCode.tsx
```diff
@@ -64,6 +69,20 @@ export default function ConfirmCode({navigation, route}: ConfirmCodeProps) {
...
+        t(LoginTexts.confirmCode.autoVerifyToast),
+        ToastAndroid.SHORT,
+        ToastAndroid.CENTER,
+      );
+      navigation.navigate(afterLogin.routeName as any, afterLogin.routeParams);
+    }
+  });
```
gorandalum, line 84: You should provided a dependency list to make this useEffect not run on every render. With the current code the dependency list should contain `user?.phoneNumber`, or `authenticationType` if the code is changed by the above proposal.

gorandalum (changes requested):

src/screens/Profile/Login/ConfirmCode.tsx
havwig: Fixed.

src/screens/Profile/Login/ConfirmCode.tsx
havwig: Fixed, thanks for the tip!

src/screens/Profile/Login/ConfirmCode.tsx
havwig: Toasts are an OS-level feature on Android (and thus styled by the OS), they are non-interactive and disappear after a short while. See the first paragraph here: https://developer.android.com/guide/topics/ui/notifiers/toasts

Screen shot from the app:
Screenshot_20210318-164504_atb

cbrevik: We do not rely on Toasts anywhere else, so I'd like to run this by design CC @pettersuul. Is there some other way we could better notify the user that the login was successful if the SMS-code was retrieved automatically?

gorandalum: @pettersuul did you have a look at this?

pettersuul: I'm not sure a confirmation is necessary, as the user is informed if the login is unsuccessful, and will immediately see the logged in phone nummer if successful

havwig: OK, I'll remove the toast.

cbrevik (approved):

PR #1006, opened 2021-03-18
feat: Show alerts in app

gorandalum:
Alerts may be configured in FireStore and will immidiately be shown in
the app. The alert are are configured with a context field, which
specifies if the context is travel, ticketing or general. If the context
is travel the alert is shown on the Assistant and Nearby screen, if
ticketing the alert is shown on Ticket screen (including Overview and
Confirmation screen), and if general the alert will be shown on all
mentioned screens.

Also made some changes to the MessageBox representation. Now the message
box background is a solid background color, to make it not be affected
by the underlying background color.


gorandalum:
Some screenshots with different type and context:
Screenshot 2021-03-18 at 15 03 50
Screenshot 2021-03-18 at 15 02 43
Screenshot 2021-03-18 at 15 01 56
Screenshot 2021-03-18 at 15 01 46
Screenshot 2021-03-18 at 15 35 20

src/components/disappearing-header/index.tsx
```diff
@@ -281,6 +295,13 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
     backgroundColor: theme.background.level1,
     flexGrow: 1,
   },
+  alertBoxContainer: {
+    backgroundColor: theme.background.header,
+  },
+  alertBox: {
+    marginHorizontal: 12,
```
mikaelbr, line 302: should this be `theme.spacings.medium`?

src/components/screen-header/index.tsx
```diff
@@ -8,6 +8,8 @@ import React, {useMemo, useState} from 'react';
 import {StyleSheet, useTheme} from '@atb/theme';
 import HeaderButton, {ButtonModes, HeaderButtonProps} from './HeaderButton';
 import ThemeText from '@atb/components/text';
+import {AlertContext} from '@atb/alerts/AlertsContext';
+import AlertBox from '@atb/components/alert-box';
```
mikaelbr, line 12: Should alert-box be part of alerts instead?

src/screens/Assistant/Results.tsx
```diff
@@ -79,7 +81,9 @@ const Results: React.FC<Props> = ({
     return (
       <View style={styles.container}>
         <MessageBox>
-          <ThemeText style={styles.infoBoxText}>
+          <ThemeText
+            style={{...styles.infoBoxText, color: theme.status.info.main.color}}
```
mikaelbr, line 85: I think maybe text color actually gets cascaded? Could wrap ThemeText on top? cc @cbrevik

mikaelbr (approved):
I think collecting alerts within the same folder structure can be better and show clearer intent/responsibility. Also, check out if it's possible to cascade color within the message box using a ThemeText on top. I think this should actually be possible, but not sure.

src/screens/Assistant/Results.tsx
gorandalum: I'll look into this

gorandalum: It is not cascading from one `ThemeText` to the next.

This code which is the default style in `ThemeText` sets the text color if it is not overridden by specifying color in the `style` parameter.
```
const typeStyle = {
  ...theme.text[fontType],
  color: theme.text.colors[color],
};
```

src/components/screen-header/index.tsx
gorandalum: Moved

src/screens/Assistant/Results.tsx
gorandalum: I experimented a little, but could not find a satisfying solution. It may be possible if investing a little more time, but I figured it was not worth it now. This means that if using `children` instead of `message` the `ThemeText` needs to set the color style.

PR #1017, opened 2021-03-22
fix: Refetch recent tickets

gorandalum:
By using fare contracts as dependency for the hook fetching recent
tickets they are updated when logged in user is changed and when new
tickets are bought.


src/screens/Ticketing/Tickets/use-recent-tickets.ts
```diff
@@ -207,7 +212,7 @@ export default function useRecentTickets() {
 
   useEffect(() => {
     fetchRecentFareContracts();
-  }, []);
+  }, [fareContracts]);
```
cbrevik, line 215: Suggest changing to array-length, so we don't get unecessary updates

```suggestion
  }, [fareContracts.length]);
```

cbrevik (changes requested):
I'm a bit unsure under what circumstances this will change. Will the array change if for example a ticket gets its QR-code updated? To avoid unnecessary API-calls, maybe we want to only update if the number of fare contracts changes?

src/screens/Ticketing/Tickets/use-recent-tickets.ts
gorandalum: Good point. Maybe we could use `[fareContracts[0]?.id]` to also eliminate the (very few) cases where logging in/out to a user with equal ticket count?

cbrevik: Hmm, good point! 🤔 That's probably a good idea. Another way (so the way the array sort would not matter) would be to evaluate the latest `created` time. Something like `Math.max(0, ...fareContracts.map((a) => a.created.seconds));` But not sure if we're over-analyzing the whole thing then?

PR #1020, opened 2021-03-23
feat: Navigate to quay departures

monopoint:


src/screens/TripDetails/QuayDepartures/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <View style={styles.screen}>
+      <View style={[styles.header, {paddingTop}]}>
+        <ScreenHeader
+          leftButton={{type: 'back'}}
+          title={route.params.location.name}
+        />
+      </View>
+      <Animated.ScrollView
```
mikaelbr, line 46: Is there a reason for this being animated scroll view?

src/components/departure-list/section-items/line.tsx
```diff
@@ -44,7 +44,7 @@ import {
   View,
 } from 'react-native';
 import {TouchableOpacity} from 'react-native-gesture-handler';
-import {NearbyScreenNavigationProp} from '../Nearby';
+import {NearbyScreenNavigationProp} from '../../../screens/Nearby/Nearby';
```
mikaelbr, line 47: This can be `@atb/screens/Nearby/Nearby` maybe should it be exported from index so it could be `@atb/screens/Nearby`

mikaelbr:
Suggestion for feature at some point: Add map of quays on quay departure list.

src/components/departure-list/DeparturesList.tsx
```diff
@@ -81,6 +81,7 @@ export default function NearbyResults({
               currentLocation={currentLocation}
               lastUpdated={lastUpdated}
               defaultExpanded={i === 0}
+              disableCollapsing={disableCollapsing}
```
mikaelbr, line 84: This also seems to hide headers altogether. Should it be named something more accurate to reflect that?

mikaelbr (changes requested):
This will be a great feature! Good job! Just a couple of things you could consider

src/screens/Nearby/Nearby.tsx
```diff
@@ -127,7 +127,7 @@ const NearbyOverview: React.FC<Props> = ({
     navigation.navigate('LocationSearch', {
       label: t(NearbyTexts.search.label),
       callerRouteName: NearbyRouteNameStatic,
-      callerRouteParam: 'location',
+      callerRouteParam: 'test123',
```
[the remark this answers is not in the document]
monopoint: I dont even

src/components/departure-list/DeparturesList.tsx
```diff
@@ -12,30 +12,30 @@ import {ActivityIndicator, View} from 'react-native';
 import QuaySection from './section-items/quay-section';
 import {hasNoGroupsWithDepartures, hasNoQuaysWithDepartures} from './utils';
```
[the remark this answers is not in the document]
monopoint: :+1:

src/screens/TripDetails/QuayDepartures/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    zIndex: 2,
+  },
+  screen: {
+    backgroundColor: theme.background.level1,
+    flexGrow: 1,
+    flex: 1,
+  },
+  scrollContainer: {
+    paddingTop: theme.spacings.large,
+    paddingBottom: theme.spacings.large,
```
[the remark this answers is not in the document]
monopoint: :+1:

src/screens/TripDetails/QuayDepartures/index.tsx
monopoint: yes, the Animated.scrollView has a refreshControl option that lets us use the pull-down-to-refresh feature

mikaelbr: Regular [`<ScrollView>` should also have that](https://reactnative.dev/docs/scrollview#refreshcontrol)

monopoint: You are indeed correct

src/components/departure-list/DeparturesList.tsx
monopoint: It hides the collapsible tool header. 

scrn223

it seemed unnecessary when collapsing was disabled. 
Do you think we should still show it?

monopoint: @pettersuul maybe?

mikaelbr: I was thinking more like renaming `disableCollapsing` to be `hideHeader` for instance (or something similar ofc)

mikaelbr: On second thought, this will show several quays right? Not one specific quay, but one stop place? In that case we must show the header as it can be named platforms like `Lerkendal 1`, `Lerkendal 2` etc.

monopoint: Ah, yes, In the specific view where disable collapsing is actually used (quay departures), we already show the stop name in the main header.
So this feature was introduced to remove some of the elements that would be perceived as unnecessary clutter.

The option to disable collapsing is used in this view because we know we are only showing platforms for one stop, and the stop name is already displayed in the main header. In this specific case there is no need to collapse platforms to save space, nor show the extra header with the stop name.

If we are showing a view where it makes sense to group platforms together in a stop, we should definitely not be using the  disable collapsing option. 

In the case when we are showing departures for a specific stop, there are two headers showing the stop name and 1 showing the platform name before the actual departures are listed, this removes one of the stop name headers.



Of course, this is my hypothesis. user testing might well prove me wrong.

monopoint: image
vs
image

monopoint: collapses to
image

pettersuul: If only one stop is listed (e.g. searched for å stop in departures), the collapsible stop-header is unneccesary.

image

mikaelbr (approved):

PR #1021, opened 2021-03-23
fix: bug with excessive padding for inline buttons

mikaelbr:
See #1016. This also fixes padding on inline buttons.

Screenshot 2021-03-23 at 10 40 55


cbrevik (approved):
Good catch 🚀

PR #1024, opened 2021-03-23
refactor: rewrites color definitions to use @atb-as/theme

mikaelbr:
Uses shared styles from `@atb-as/theme` as defined in https://github.com/AtB-AS/design-system.

See the documentation of use for colors and typography: https://github.com/AtB-AS/design-system/tree/main/packages/theme

This PR rewrites ThemeText and Colors to match the new distributed theme module which can be shared by other products.

## TODO

- [x] Rewrite colors
- [x] Rewrite typography names to match new design system names

### DIFF
--- a/src/components/message-box/index.tsx
+++ b/src/components/message-box/index.tsx
@@ -59,7 +59,7 @@ const MessageBox: React.FC<MessageBoxProps> = ({
       {retryFunction && (
         <ThemeText
           style={styles.retryText}
-          type="body__link"
+          type="body__primary--underline"
           onPress={retryFunction}
         >
           {t(MessageBoxTexts.tryAgainButton)}
@@ -131,7 +131,7 @@ const useBoxStyle = StyleSheet.createThemeHook((theme) => ({
   content: {
     flex: 1,
   },
-  text: theme.typography.body,
+  text: theme.typography.body__primary,
   retryText: {
     marginTop: theme.spacings.medium,
   },
--- a/src/components/message-box/index.tsx
+++ b/src/components/message-box/index.tsx
@@ -131,7 +131,7 @@ const useBoxStyle = StyleSheet.createThemeHook((theme) => ({
   content: {
     flex: 1,
   },
-  text: theme.text.body,
+  text: theme.typography.body,
   retryText: {
     marginTop: theme.spacings.medium,
   },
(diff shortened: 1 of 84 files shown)


gorandalum (approved):
LGTM! Good work!

src/screens/Profile/DesignSystem.tsx
```diff
@@ -133,12 +133,12 @@ export default function DesignSystem() {
 
         <Sections.Section withPadding withTopPadding>
           <Sections.LocationInput
-            label="Label"
+            label="body__tertiary"
```
cbrevik, line 136: Wrongly replaced text here, should be "Label"? Same with next two?

src/components/message-box/index.tsx
```diff
@@ -59,7 +59,7 @@ const MessageBox: React.FC<MessageBoxProps> = ({
       {retryFunction && (
         <ThemeText
           style={styles.retryText}
-          type="body__link"
+          type="body__primary--underline"
```
cbrevik, line 62: Shouldn't try again text have same color as main body text? Contrast issues?
image

cbrevik (commented):
Is background_2 supposed to change to current shade? Seems like there's not enough contrast with white section-boxes here. Or is it fine @pettersuul?
image

src/screens/Profile/DesignSystem.tsx
mikaelbr: Hah. Fantastic catch! I should have done a case-sensitive replacement AND actually verified. Lazy

pettersuul:
@cbrevik Unsure what values are used here, but background_2 should be darker than background_1 (more contrast)

src/components/message-box/index.tsx
mikaelbr: Saw this also. Think it's an issue from a previous PR (@gorandalum ) but can fix it as part of this aswell.

mikaelbr:
Yep. There were duplicated values and I chose the wrong ones. Fixing up now.

src/screens/Profile/DesignSystem.tsx
mikaelbr: Fixed

src/components/message-box/index.tsx
mikaelbr: Should be solved now

mikaelbr:
@cbrevik Fixed comments now. Would be great if you could just verify quickly if you still have it locally.

cbrevik (approved):
Verified fixes, seems good to go! Awesome job :D

PR #1026, opened 2021-03-23
feat: Group texts for screen reader

gorandalum:
Setting accessibile true on parent views that contain multiple texts,
making them be read by the screen reader in one go, instead of each text
being selectable.


cbrevik (approved):
LGTM 👌🏻

PR #1030, opened 2021-03-24
feat: Display installid in Profile-tab

cbrevik:
Display install-id below furthest down in the Profile-tab. Make it clickable to copy the text, and display alert when you do. Alert text fades out after a bit.

https://user-images.githubusercontent.com/4932625/112430537-96a7cc00-8d3e-11eb-8637-34804ffdd12d.mov

image

### DIFF
--- a/src/screens/Profile/Home/index.tsx
+++ b/src/screens/Profile/Home/index.tsx
@@ -192,29 +192,27 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
           </Sections.Section>
         )}
 
-        <View style={style.debugInfoContainer}>
-          {config?.installId && (
-            <View>
-              {fadeIsAnimating && (
-                <ClipboardFadeContainer>
-                  <ScreenReaderAnnouncement
-                    message={t(ProfileTexts.installId.wasCopiedAlert)}
-                  />
-                  <ThemeText>
-                    ✅ {t(ProfileTexts.installId.wasCopiedAlert)}
-                  </ThemeText>
-                </ClipboardFadeContainer>
-              )}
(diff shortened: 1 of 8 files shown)


src/screens/Profile/Home/index.tsx
```diff
@@ -175,17 +191,48 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
...
   scrollView: {
     marginTop: theme.spacings.medium,
   },
+  debugInfoContainer: {
+    alignItems: 'center',
+    marginVertical: theme.spacings.medium,
+  },
+  debugInfo: {
+    color: theme.colors.background_1.color,
+  },
```
mikaelbr, line 237: I think this is a workaround due to an issue with dark mode. Should not be necessary if merging in master I think.

src/screens/Profile/Home/index.tsx
```diff
@@ -175,17 +191,48 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
...
+
+        <View style={style.debugInfoContainer}>
+          {config?.installId && (
+            <View>
+              {fadeIsAnimating && (
+                <ClipboardFadeContainer>
+                  <ScreenReaderAnnouncement
+                    message={t(ProfileTexts.installId.wasCopiedAlert)}
+                  />
+                  <ThemeText>
```
mikaelbr, line 203: Copy theme text here is wrong color but I think this should be solved with merging master in

mikaelbr (changes requested):
Check out merging master in to see if colors are correct.

mikaelbr (approved):

PR #1033, opened 2021-03-24
feat: Log Firebase Auth Id in Intercom, Bugsnag and API

cbrevik:
Logging out the Firebase uid so we can better debug for specific users.


mjenssen (approved):
LGTM

mikaelbr (approved):

PR #1036, opened 2021-03-25
feat: Refactor enrollment to work for multiple groups

cbrevik:
Moving to setting `group_name`: `"true"` for a certain beta-group on analytics user property. This will allow us to enroll into multiple groups. Also add the enrollment screen under Profile-tab.

image


mikaelbr (approved):
Cool! Could this at some point be extended to support different toggles for different beta features also?

cbrevik:
This PR sooort of solves that, but not completely. It allows you to enroll into several different beta features. But there is no good way to "toggle" yet, as we'd need to "know" about the user-property names beforehand.

PR #1041, opened 2021-03-30
feat: No tariff zone map if screen reader enabled

gorandalum:
Disable the tariff zone map selection if screen reader is enabled, as
MapBox crashes with screen reader. Instead show a list of the available
tariff zones.


src/screens/Ticketing/Purchase/TariffZones/search/TariffZoneResults.tsx
```diff
@@ -48,10 +50,15 @@ const TariffZoneResults: React.FC<Props> = ({tariffZones, onSelect}) => {
                   <ThemeIcon svg={MapPointPin} width={20} />
                 </View>
                 <View style={styles.nameContainer}>
-                  <ThemeText style={styles.nameText}>
+                  <ThemeText type={'body__primary--bold'}>
```
mikaelbr, line 53:
```suggestion
                  <ThemeText type="body__primary--bold">
```

mikaelbr (approved):
Good stuff!

PR #1044, opened 2021-04-06
feat: Set default user profile in preferences

gorandalum:


src/preferences/types.ts
```diff
@@ -16,6 +16,7 @@ export type UserPreferences = {
   overrideColorScheme?: boolean;
   language?: Preference_Language;
   useSystemLanguage?: boolean;
+  defaultUserTypeString?: string;
```
mikaelbr, line 19: Having a hard time understanding what user type and profile is until reading the translation files. Are these names too ambigues?

src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -242,6 +239,49 @@ export const createTravelDateText = (
...
+) => {
+  const {
+    preferences: {defaultUserTypeString},
+  } = usePreferences();
+
+  const isDefaultProfile = useCallback(
+    (u: UserProfile, index: number, filteredProfiles: UserProfile[]) => {
+      if (
+        defaultUserTypeString &&
+        filteredProfiles.some((u) => u.userTypeString === defaultUserTypeString)
```
mikaelbr, line 261: Shadowed variable. Technically not wrong and refers to the same type here, but could be confusing?

mikaelbr (approved):
LGTM! Nice feature for many I think, although somewhat hidden/hard to find.

src/preferences/types.ts
gorandalum: Yeah, we thought about using our own domain names for these things. There even is an iteration of the code where the name `TravellerType` is used instead of `UserProfile`. But after some contemplation we decided on reusing the domain types and names from Entur, which is `UserProfile`. And `UserProfile` has a field `userTypeString` which we now store as the default value in the preferences.

PR #1066, opened 2021-04-13
fix: fixes inaccessible presses

mikaelbr:
An issue with using TouchableOpacity from `react-native-gesture-handler` as it is inaccessible. Add avoid import to eslint rule and fix instances of this occurring today.

This should fix issue with screen reader

### DIFF
--- a/src/screens/Assistant/ResultItem.tsx
+++ b/src/screens/Assistant/ResultItem.tsx
@@ -4,7 +4,6 @@ import {Duration, WalkingPerson} from '@atb/assets/svg/icons/transportation';
 import AccessibleText, {
   screenReaderPause,
 } from '@atb/components/accessible-text';
-import Button from '@atb/components/button';
 import ThemeText from '@atb/components/text';
 import ThemeIcon from '@atb/components/theme-icon';
 import TransportationIcon from '@atb/components/transportation-icon';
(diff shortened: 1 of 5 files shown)


src/screens/Assistant/ResultItem.tsx
```diff
@@ -30,8 +29,13 @@ import insets from '@atb/utils/insets';
...
 import React from 'react';
-import {AccessibilityProps, View, ViewStyle} from 'react-native';
-import {ScrollView, TouchableOpacity} from 'react-native-gesture-handler';
+import {
+  AccessibilityProps,
+  View,
+  ViewStyle,
+  TouchableOpacity,
+} from 'react-native';
+import {ScrollView} from 'react-native-gesture-handler';
```
gorandalum, line 38: Is it important where `ScrollView` is imported from? A quick search in the code now showed 19 imports from `react-native-gesture-handler` and 3 imports fra `react-native`.

gorandalum (approved):

src/screens/Assistant/ResultItem.tsx
mikaelbr: react-native-gesture-handler typically has better performance AFAIK, but in the case of Touchable*s it lacks accessibility. But for ScrollView it should be fine.

PR #1077, opened 2021-04-19
fix: transitions on android infered from design

monopoint:
Screen transition from Assistant to Trip Details should follow design and behavior from ios.


gorandalum:
While doing this, could you also check if Android transitions is correct on the ticketing part of the app? My guess is that it is missing several places.

gorandalum (approved):
LGTM! Kompileringsfeilen vil forsvinne når den kommer på master.

tormoseng:
Ser greit ut på Android.

PR #1120, opened 2021-04-29
feat: Retried requests errors logged only once

gorandalum:
Previously if all three retries failed the error would be logged four
times. Once for the initial request and once for each of the three
retries. With this change only the error from the last retry will be
logged. Also if one of the retries is successful, no error will be
logged.


cbrevik (approved):

gorandalum:
Akseptansekriterier:
Behøver ikke testes

PR #1121, opened 2021-04-29
feat: Lower severity for Network Errors

gorandalum:
The idea is that Bugsnag Errors with severity 'info' does not need
Github Issues to be created.


cbrevik (approved):

gorandalum:
Akseptansekriterier: 
Behøver ikke testes

PR #1202, opened 2021-05-27
feat: New ticket informational screen

gorandalum:
This ticket informational screen appears first time the user navigates
to the tickets screen. If the user accepts the presented information,
the informational screen is removed and the user may use the ticketing
functionality.


gorandalum:
<div>


</div>

monopoint (approved):
ser bra ut

PR #1210, opened 2021-05-31
fix: Emoji popup crashing on some Android versions

gorandalum:
The .map function throws error since the emojis are undefined. I have
not used time on looking into why the emojis are undefined, just quick
fixed this to make the app not crash.


monopoint (approved):
godkjent som quick fix

PR #1242, opened 2021-06-07
feat: Set background color for React navigation

gorandalum:
During transitions the generic white background from React naviation
some times "leaked" through, giving a noticable flash to the user,
especially in dark mode. We got report from a user in Intercom that
this was unpleasant.

The best fix would be to remove these "leaks" of the background color
all together, but I could not find a good solution for this. As such
this was fixed by setting the background color for React navigation to
be 'background_1' from the AtB theme. This makes the background light
gray in light mode and dark gray in dark mode, making the background
leek less unpleasant.

### DIFF
--- a/src/navigation/index.tsx
+++ b/src/navigation/index.tsx
@@ -20,6 +20,7 @@ import {
   NavigationContainerRef,
   NavigatorScreenParams,
   useLinking,
+  DefaultTheme,
 } from '@react-navigation/native';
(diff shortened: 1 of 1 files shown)


gorandalum:
Relevant Intercom-melding:
https://app.intercom.com/a/apps/vdemedo2/inbox/inbox/4537645/conversations/159245400007510

monopoint (approved):
ser ikke så mye forskjell, men det blir def ikke verre

PR #1261, opened 2021-06-14
feat: Re-enabled ticketing toggle and removed beta code button on ticketing splash screen

mortlun:
- Re-enabled ticketing  remote config toggle, to cater for other organisations using the app, where ticketing is not yet implemented in backend.
-  Also removed beta code button on ticketing splash screen since it is not in use anymore.


gorandalum (approved):
LGTM!

PR #1636, opened 2021-10-07
fix: Fixed height being 0 for modal, maxheight for modal is windowheight

brorhb:


brorhb:
Duplicate of https://github.com/AtB-AS/mittatb-app/pull/1635

PR #1822, opened 2021-11-26
chore: Cache path must be string and not env var

cbrevik:


.github/workflows/build-staging-android.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-      - cbrevik/upgrade-react-native-0.66.3
+      - cbrevik/debug-android-build
```
cbrevik, line 6:
```suggestion
```

PR #1905, opened 2021-12-17
feat: Update mobiletoken sdk

gorandalum:
The new version of the mobile-token sdk includes retry logic, and as
such it is no longer necessary to handle it in the app.

Added the possibility for manual force restart when showing ticket
details, and the mobile token creation has failed.


monopoint (approved):
lgtm

cbrevik (approved):

tormoseng:
Har herjet en del med å avvise kall, gi feil responskode, etc med en iOS 15. Ikke sett noen problemer.

PR #2036, opened 2022-01-19
feat: Use correct theme color on line number text

gorandalum:
Før / Etter
<div>
Screenshot 2022-01-19 at 10 03 14
Screenshot 2022-01-19 at 10 37 23
</div>


monopoint (approved):
:+1:

PR #2071, opened 2022-01-26
chore: update ticket prices

monopoint:
https://github.com/AtB-AS/mittatb-app/wiki/Billettinformasjon-i-appen


cbrevik (approved):
Ser bra ut 👍🏻

PR #2222, opened 2022-02-27
feat: List tokens as source for inspectable token

gorandalum:
Mainly use list tokens as the source for which token is inspectable.
This gives less possibility for the firestore-tcard-value, local token
state, and remote tokens, to become out of sync.

Some places the local firestore-tcard-value must still be used to
support users that are not yet enrolled with mobile token.

Some texts are still missing, and are marked with TODO.


rosvik (approved):
Er jo ikke helt stødig på mobile token delen, men ser jo stort sett fint ut! Så tar vi oversettelser og "token toggle toggle" som egne PRs?

PR #2299, opened 2022-03-09
fix: add missing `)` to Fastfile

rosvik:
Ser at bygget feiler pga. en syntaksfeil i Fastfile. Legger til en manglende parantes.

f.eks. https://github.com/AtB-AS/mittatb-app/runs/5478910974?check_suite_focus=true


monopoint (approved):
:+1:

PR #2397, opened 2022-03-25
chore: Enable deploying on Android device

gorandalum:
- The certificate signature in .env for debug is now the signature for
  the debug keystore in the repo. The previous certificate signature was
  the staging keystore.
- Some changes to network security config for debug to enable deploying
  to Android device
- Set exported-field on all activites in AndroidManifest, as it is
  required in next sdk version.
- Updated README.


monopoint (approved):
:+1:

PR #2400, opened 2022-03-25
feat: Reattestation support

gorandalum:
- Use latest version of mobiletoken-sdk which has support for
reattestation.
- Provide logger functions to mobiletoken-sdk instead of handling
  logging in the app.
- Update targetSdkVersion as it is necessary for the latest Entur
  Android Traveller library.


monopoint (approved):
:+1:

Issue #2475, opened 2022-04-05
Use configuration from FireStore


gorandalum:
### Origin
We have a collection called `configuration` in `Firestore`. The values stored here should be made available in the `configuration context` in the app. The app as of now uses `remoteConfig` for some of these values and should be refactored to use Firestore instead.

The buttons for payment should be using values from firestore

in issue #2430 we moved and refactored to use the values in `configuration / referenceData`. This issue will move and refactor in the values in configuration/other and configuration/paymentTypes.

- [x] fetch firestore documents from firestore
   - [x] Update the default json-files with fresh data from production values in firestore 
   - [x] Json-parse the data and cast them to the correct types
- [x] make the data available in the firestoreConfigurationContext
- [x] Refactor to use values from firestore instead of remoteconfig
  - [x] delete corresponding remote config values and code
  - [x] make use of the firestore values
- [x] Ensure the Payment buttons use values from paymentTypes from Firestore
- 
 

![Screenshot 2022-04-05 at 12 26 29](https://user-images.githubusercontent.com/675421/161735274-d527f1e8-de49-49a3-b1d2-da1ba1869093.png)

jorelosorio:
As a track note: The build failed the first time when the PR was merged, with a possible issue fetching Mapbox libraries for Android, but after a second try the next day it worked. We think that the root cause could be that the server was not reached due to a down during that moment. We are aware of this and will keep monitoring for further issues related.

jorelosorio:
@tormoseng Good catch, I new PR is made fixing this at https://github.com/AtB-AS/mittatb-app/pull/2534

tormoseng:
Re-testet OK

PR #2494, opened 2022-04-07
chore: Use configuration from FireStore

jorelosorio:
Lukker https://github.com/AtB-AS/mittatb-app/issues/2475

This issue will move and refactor the values in `configuration/other` and `configuration/paymentTypes`.


src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -150,3 +181,43 @@ function getUserProfilesFromSnapshot(
...
+      case 'vipps':
+        paymentTypes.push(PaymentType.Vipps);
+        break;
+      case 'visa':
+        paymentTypes.push(PaymentType.VISA);
+        break;
+      case 'mastercard':
+        paymentTypes.push(PaymentType.MasterCard);
+        break;
+    }
```
jorelosorio, line 220: There is another way to match those payment methods but for example `MasterCard` doesn't match `mastercard` key in the `Firestore`. If it is possible to match seamless what we have in the Firestore vs the enum keys we could avoid that `switch` block. @gorandalum @monopoint

gorandalum: Good suggestion. Feel free to change the naming of variables in the code the handle this mapping better.

src/screens/Ticketing/Purchase/Payment/SelectPaymentMethod/index.tsx
```diff
@@ -103,6 +89,16 @@ const SelectPaymentMethod: React.FC<Props> = ({
 
   const {user} = useAuthState();
   const {theme} = useTheme();
+  const {paymentTypes} = useFirestoreConfiguration();
+
+  const defaultPaymentOptions: SavedPaymentOption[] = paymentTypes.map(
```
gorandalum, line 94: This seems OK, but just to be sure check how it is handled when the `getSelectedPaymentMethod(previousPaymentMethod)` below has a `previousPaymentMethod` which is not in the firestore config. If "vipps" is removed from Firestore it should not be still usable for those who has it stored as `previousPaymentMethod`.

jorelosorio: Good catch, I was expecting the same but because that saved payment option comes from the server maybe it is filtered. But I will do that as well, thanks.

gorandalum: Just note that there is a difference between `recurringPayments` which are saved payment cards which comes from the server, and the `previousPaymentMethod` which does not come from the server, but is stored locally on the phone storage. However for implementation purposes it is not a big difference whether it comes from a server or phone storage.

jorelosorio: I have covered both cases now, please review it ;)

src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -78,8 +102,15 @@ export const FirestoreConfigurationContextProvider: React.FC = ({children}) => {
...
+      modesWeSellTicketsFor,
+      paymentTypes,
     };
-  }, [preassignedFareproducts, tariffZones, userProfiles]);
+  }, [
+    preassignedFareproducts,
+    tariffZones,
+    userProfiles,
+    modesWeSellTicketsFor,
+  ]);
```
monopoint, line 113: missing paymentTypes as dependancy

monopoint (changes requested):

src/configuration/FirestoreConfigurationContext.tsx
jorelosorio: Fixed

monopoint (approved):
looks good

PR #2534, opened 2022-04-20
chore: Use of VAT Percent attribute from Firestore

jorelosorio:
Lukker https://github.com/AtB-AS/mittatb-app/issues/2475

This issue will refactor the values in configuration/other/vatPercent

### DIFF
--- a/src/configuration/FirestoreConfigurationContext.tsx
+++ b/src/configuration/FirestoreConfigurationContext.tsx
@@ -20,6 +20,7 @@ import {
   defaultUserProfiles,
   defaultModesWeSellTicketsFor,
   defaultPaymentTypes,
+  defaultVatPercent,
 } from '@atb/reference-data/defaults';
 import {PaymentType} from '@atb/tickets';
 
@@ -29,6 +30,7 @@ type ConfigurationContextState = {
   userProfiles: UserProfile[];
   modesWeSellTicketsFor: string[];
   paymentTypes: PaymentType[];
+  vatPercent: number;
 };
 
 const defaultConfigurationContextState: ConfigurationContextState = {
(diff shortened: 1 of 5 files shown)


monopoint (approved):
:+1:

src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -37,6 +39,7 @@ const defaultConfigurationContextState: ConfigurationContextState = {
   userProfiles: [],
   modesWeSellTicketsFor: [],
   paymentTypes: [],
+  vatPercent: 0,
```
gorandalum, line 42: The defaults (like `defaultVatPercent`) should/could probably also be used here? If so then also update the other variables.

src/reference-data/defaults.ts
```diff
@@ -17,3 +17,4 @@ export const defaultModesWeSellTicketsFor: string[] = [
   'shuttleBus',
 ];
 export const defaultPaymentTypes: string[] = ['vipps', 'visa', 'mastercard'];
+export const defaultVatPercent: number = 12;
```
gorandalum, line 20: The vat percent, payment types and modes we sell tickets for are not actually reference data, and the defaults does not belong here. Rather move them closer to the `FirestoreConfigurationContext`, either in that file or in a `defaults.ts` in the same folder.

gorandalum (changes requested):
Some minor changes. The changes is not directly related to your PR, as you were just following the pattern already implemented, but still would be nice to fix now.

src/reference-data/defaults.ts
jorelosorio: Fixed, moved to a separated file in the same folder than `FirestoreConfigurationContext `

jorelosorio:
@gorandalum Changes were made as requested

gorandalum (approved):

PR #2541, opened 2022-04-21
fix: Client number should be read as number digit by digit, not with thousands

jorelosorio:
Closes https://github.com/AtB-AS/mittatb-app/issues/2126

The solution works well for both platforms


gorandalum (approved):
LGTM :+1:

You can change the semantic naming from `chore` to `fix` for this. Generally we have had a habit of using `chore` too broadly. It is intended for changes not affecting production code, and in most situations `refactor` or `fix` is more appropriate. Going forward we are going to base our release notes more on the PR-titles, and therefore we will focus more on correct semantic naming. 

https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

monopoint (approved):
:+1:

PR #2557, opened 2022-04-26
chone: New component for Traveling Zones

jorelosorio:
Closes https://github.com/AtB-AS/mittatb-app/issues/2458

This issue contains a new file with a new Zone component `ZoneItem` that reuses the existing `ButtonInput` component.

Dark mode and accessibility were tested.


gorandalum (approved):
Looks good! About the PR title:

- `chore` is ment for stuff like build system etc. This is a change that will be visible for users, so should use `feat`.

PR #2592, opened 2022-04-29
ci: invalidate ipa build cache when assets changes to include new images

mikaelbr:
Invalidates IPA cache when assets change. This will cause new images to be included in the bundle. Will invalidate the cache when assets change (changes in generated assets and other assets).

This is not an issue in store as it is not cached.


rosvik (approved):
Ser bra ut, bare du endrer base til `release/1.18` 👍

PR #2610, opened 2022-05-03
feat: New summary component in ticket purchase overview

gorandalum:
Tickets after implementing:
<div>
Screenshot 2022-05-04 at 21 19 50
Screenshot 2022-05-04 at 21 19 22
</div>

### DIFF
--- a/src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
+++ b/src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
@@ -87,5 +87,5 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     textAlign: 'center',
   },
   message: {textAlign: 'center'},
-  button: {marginTop: theme.spacings.medium},
+  button: {marginTop: theme.spacings.xLarge},
 }));
--- a/src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
+++ b/src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
@@ -1,0 +1,91 @@
+import ThemeText from '@atb/components/text';
+import {ActivityIndicator, StyleProp, View, ViewStyle} from 'react-native';
+import Button from '@atb/components/button';
+import {PurchaseOverviewTexts, useTranslation} from '@atb/translations';
+import {ArrowRight} from '@atb/assets/svg/mono-icons/navigation';
+import React from 'react';
+import {StyleSheet} from '@atb/theme';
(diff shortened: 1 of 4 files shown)


src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+type Props = {
+  price: number;
+  isLoading: boolean;
+  isError: boolean;
+  fromTariffZone: TariffZoneWithMetadata;
+  toTariffZone: TariffZoneWithMetadata;
+  userProfilesWithCount: UserProfileWithCount[];
+  preassignedFareProduct: PreassignedFareProduct;
+  travelDate?: string;
```
gorandalum, line 23: This might rise the question "Why not just pass the `toPaymentFunction` from the overview-screen? Then we don't need this many parameters".

My thought is that the Overview-components (Summary, ZoneItem, and others we will make) encapsulates the logic about where they are going to navigate. They are not (as of now) reusable components, so we don't need to concern the parent component about specifying where the component should navigate. For the summary component it makes it necessary to pass along quite a few parameters, but this will not be the case for the other components.

src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
```diff
@@ -0,0 +1,91 @@
...
+  );
+}
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  price: {
+    marginBottom: theme.spacings.medium,
+    textAlign: 'center',
+  },
+  message: {textAlign: 'center'},
+  button: {marginTop: theme.spacings.medium},
```
monopoint, line 90: comparing screen shots with figma, it looks like this spacing is too small.

monopoint (changes requested):

src/screens/Ticketing/Purchase/Overview/components/Summary.tsx
gorandalum: Great! Fixed now and updated pics.

src/translations/screens/subscreens/PurchaseOverview.ts
```diff
@@ -1,5 +1,6 @@
 import {translation as _} from '../../commons';
 import orgSpecificTranslations from '@atb/translations/utils';
+import {formatDecimalNumber} from '@atb/utils/numbers';
```
rosvik, line 3: Denne trengs vel ikke å importeres her?

```suggestion
```

PR #2611, opened 2022-05-03
refactor: Some fixes to the zones component

gorandalum:
After implementing the Summary-component i saw some small changes that
could be done to make the implementations more aligned.

- Renamed to 'Zones'. Follow the pattern with components named with
  CamelCase. Also not necessary with suffix "Item" or "Component" since
  it is inside a components-folder.
- Moved specifying the top margin from inside the component to the
  overview screen. Components should generally not handle margins around
  the component, this should be handled where the component is used.
- Added "accessible: true" on the section. Without this the
  "accessibilityRole: 'button'" was not working as intended, atleast on
  iOS.


jorelosorio:
@gorandalum good work, Looks good!

jorelosorio (approved):

PR #2616, opened 2022-05-04
feat: add ticket duration selection component

### LINKED ISSUE #2457: Komponent for valg av periodelengde
### Bakgrunn
Nytt design for valg av periodelengde i forbindelse med billettkjøp 2.0

Brukes på:
Periodebilletter
Ungdomsbillett
Klippekort
24 timersbillett
Sommerpass

https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App-1.1?node-id=7657%3A65776

### Løsning
En "karusell" over mulige periodelengdevalg. Bare én kan velges om gangen. Scrolles horisontalt hvis skjermbredden ikke kan vise hele. Bør løses som en generisk flervalgs karusell, og ikke tett knyttet opp mot billettprodukter.

I dagens billettkjøpsflyt velger man billettprodukt, og og ikke periodebillettlengde. For å løse denne endringen på en god måte må man også få inn billettlengde på produktene i referansedata i Firestore. Merk da at valg av periodelengde rent teknisk blir valg av et spesifikt billettprodukt.

rosvik:
Ny komponent for valg av periodelenge. ([Figma](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App-1.1?node-id=6600%3A44952))

Pluss noen andre småting:
- Billetteringskjermen ligger nå i en scrollview, slik at man kan se alt også på mindre skjermer.
- Billettnavn vises ikke lengre for enkeltbilletter.
- Gjør margin over og under bittering 2.0 -komponenter konsistente (alle med medium margin over og under)

## Demo

https://user-images.githubusercontent.com/1774972/166684321-91176820-ddee-41c2-a8a8-e52989626176.mp4

closes #2457


src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+    marginRight: -theme.spacings.medium,
+  },
+  durationContentContainer: {
+    flexDirection: 'row',
+    marginLeft: theme.spacings.medium,
+    paddingRight: theme.spacings.medium,
+  },
+  durationChip: {
+    justifyContent: 'center',
+    borderRadius: theme.border.radius.circle * 2,
```
gorandalum, line 126: Should probably just increase `theme.border.radius.circle` in the design-system to be like `1000` (Not necessary to do now).

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    marginVertical: theme.spacings.medium,
+  },
+  durationScrollView: {
+    marginTop: theme.spacings.medium,
+    marginBottom: theme.spacings.xLarge,
+    marginLeft: -theme.spacings.medium,
+    marginRight: -theme.spacings.medium,
```
gorandalum, line 117: This kinda makes a tight coupling between this component and the parent component, as it is dependant on the parent component using the same margin size around this component. Even though this component is not ment to be reused multiple places, we should still try to not create too tight bindings between parent and child. I would rather make the component be "full-width" in the parent, and remove this negative margins.

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <ScrollView
+        style={styles.durationScrollView}
+        contentContainerStyle={styles.durationContentContainer}
+        horizontal={true}
+        showsHorizontalScrollIndicator={false}
+      >
+        {selectableProducts.map((fp, i) => (
+          <DurationChip
+            color={color}
+            text={getReferenceDataName(fp, language)}
```
gorandalum, line 49: In the Figma-sketches the product duration is used, not product name (e.g. "30 dager", not "30-dagersbillett"). The field `durationDays` is added to the `PreassignedFareProduct` in Firestore in staging to enable this. I don't know if there has been any office discussions changing this, but if not the implementation should be changed to be aligned with how it is in Figma and described in the issue.

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+      >
+        {text}
+      </ThemeText>
+    </TouchableOpacity>
+  );
+}
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    marginVertical: theme.spacings.medium,
```
gorandalum, line 111: Margins should preferably be styled from outside the component. See the other comment about this.

gorandalum (changes requested):
Some nitpicking you can take a look at and decide what is worth changing. The major point is whether the `DurationChip` should show duration ("30 dager") or product name ("30-dagersbillett").

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const styles = useStyles();
+  const {theme} = useTheme();
+  const {t} = useTranslation();
+
+  const currentColor =
+    theme.interactive[color][selected ? 'active' : 'default'];
+
+  // To make items with and without border the same size, we subtract the border
+  // width from the padding when there is a border.
+  const borderOffset = selected ? theme.border.width.medium : 0;
```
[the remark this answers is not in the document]
rosvik: Good point!

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
rosvik: This was also my first thought, but it introduced a lot of extra complexity at the parent level. We would need to split the parent container in three parts, which has an effect on flex functionality as well as padding between components. Also if the component is full width, there still has to be something that handles the horizontal padding of the header ("Select duration"). 

So while I agree that the current implementation isn't a great solution, I think we should stick to handling it within the component. One solution is also to pass the needed overflow margin as a parameter, to make it more universal. Or do you have any other ideas?

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
rosvik: 👍

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
rosvik: Agreed!

src/screens/Ticketing/Purchase/Overview/components/zone-item.tsx
```diff
@@ -71,6 +71,9 @@ export default function ZoneItem({
 }
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    marginVertical: theme.spacings.medium,
+  },
```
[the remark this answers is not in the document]
rosvik: Made a suggestion to how this should be handled for all components. See ae152883e3ccba2777fce4adb3f5b5cbda2b0f9c.

It's pretty clear from figma that every selection component should have 12px above and below, so I thought it would be clean to implement it the same way in the app.

Screenshot 2022-05-05 at 11 09 47

src/screens/Ticketing/Purchase/Overview/components/DurationSelection.tsx
gorandalum: That's ok. We'll keep it as it is for now.

src/screens/Ticketing/Purchase/Overview/components/zone-item.tsx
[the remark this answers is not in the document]
gorandalum: It's better this way when it is styled from outside the component :+1:

We'll keep it like this, even though It is kind of in conflict with the principle that styles and components should be a 1-to-1 relationship. The principle is there to ensure that styles are named based on in which component they are used (like "purchaseButton", "noTicketWarning", "headerTitle" etc.) and not have generic naming based on what they are doing (like "disabled", "bottomMargined", "strikeThrough" etc.). Even though this is not 1-to-1 relationship, it is not a totally generic naming.

gorandalum (approved):
Great! Good work :+1:

_Disclaimer: I haven't tested accessibility_

PR #2626, opened 2022-05-09
feat: add StartTimeSelection component

### LINKED ISSUE #2459: Komponent for å velge oppstartstidspunkt
### Bakgrunn

Brukes for å velge oppstartstidspunkt for en gitt billett. Står default på Nå. Ved valg av "Senere" får man opp eksisterende dato/tidsvelger. Deretter vises også evt. dato og tidspunkt som er satt når man har endret til senere.

Brukes på:
- Periodebilletter
- Enkeltbillett (i fremtiden?)
- Ungdomsbillett
- 24 timersbillett
- Sommerpass

### Løsningsforslag
Trolig gjenbruk av reisetidspunktvelger fra Avganger-skjerm

rosvik:
Added new StartTimeSelection component, as part of ticketing 2.0. This includes a new generic component `SegmentedControl`.  I also removed the "now" option from TravelDateSheet, since this is now handled from the ticketing view.

closes #2459 

## Dark/light / english/norwegian

collage

## Demo

https://user-images.githubusercontent.com/1774972/167390465-b256a601-8383-41c2-a5ee-2fd50c058ce9.mp4

# SegmentedControl

I have tried making the "SegmentedControl" component as generic as possible, to allow other uses in the future. (E.g. for the departures v1 time selection)

Different colors:

collage

Enabled / disabled state:

collage

### DIFF
--- a/src/components/segmented-control/index.tsx
+++ b/src/components/segmented-control/index.tsx
@@ -49,6 +49,9 @@ export default function SegmentedControl({
             accessible={true}
             accessibilityRole="radio"
             accessibilityState={{selected}}
+            accessibilityLabel={
+              option.subtext ? option.text + '. ' + option.subtext : option.text
+            }
             accessibilityHint={option.accessibilityHint}
             style={[
               styles.optionBox,
--- a/src/components/segmented-control/index.tsx
+++ b/src/components/segmented-control/index.tsx
@@ -1,0 +1,109 @@
+import React from 'react';
+import {InteractiveColor} from '@atb/theme/colors';
+import {View, ViewStyle, StyleProp, TouchableOpacity} from 'react-native';
+import {StyleSheet, useTheme} from '@atb/theme';
+import ThemeText from '../text';
+
+export type SegmentOptions = {
+  onPress: () => void;
+  text: string;
+  subtext?: string;
+  selected?: boolean;
+  accessibilityHint?: string;
+};
+
+type SegmentedControlProps = {
+  color: InteractiveColor;
+  activeIndex: number;
+  options: SegmentOptions[];
+  style?: StyleProp<ViewStyle>;
+};
+
+export default function SegmentedControl({
+  color,
+  activeIndex,
+  options,
+  style,
+}: SegmentedControlProps) {
+  const styles = useStyles();
+  const {theme} = useTheme();
+
+  const interactiveColor = theme.interactive[color];
+
+  return (
+    <View
+      style={[
+        styles.segmentedControl,
+        {backgroundColor: interactiveColor.default.background},
+        style,
+      ]}
+    >
+      {options.map((option, i) => {
+        const selected = activeIndex === i;
+        const currentColor = interactiveColor[selected ? 'active' : 'default'];
+        const borderWidth = selected ? theme.border.width.medium : 0;
+
+        return (
+          <TouchableOpacity
+            onPress={option.onPress}
+            accessible={true}
+            accessibilityRole="radio"
+            accessibilityState={{selected}}
+            accessibilityHint={option.accessibilityHint}
+            style={[
+              styles.optionBox,
+              {
+                backgroundColor: currentColor.background,
+                borderColor: interactiveColor.outline.background,
+                borderWidth,
+                // To make items with and without border the same size, we
+                // subtract the border width from the padding.
+                padding: theme.spacings.medium - borderWidth,
+              },
+            ]}
+          >
+            <ThemeText
+              type={selected ? 'body__secondary--bold' : 'body__secondary'}
+              style={[styles.optionText, {color: currentColor.text}]}
+            >
+              {option.text}
+            </ThemeText>
+            {option.subtext && (
+              <ThemeText
+                type="body__secondary"
+                style={[styles.optionText, {color: currentColor.text}]}
+              >
+                {option.subtext}
+              </ThemeText>
+            )}
+          </TouchableOpacity>
+        );
+      })}
+    </View>
+  );
+}
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  segmentedControl: {
+    borderRadius: 12,
+    flexDirection: 'row',
+    justifyContent: 'center',
+  },
+  multiSelectContainer: {
+    marginTop: theme.spacings.medium,
+    borderRadius: 12,
+    flexDirection: 'row',
+    justifyContent: 'center',
+  },
+  optionBox: {
+    flexGrow: 1,
+    width: 0,
+    justifyContent: 'center',
+    margin: theme.spacings.xSmall,
+    borderRadius: theme.border.radius.regular,
+  },
+  optionText: {
+    textAlign: 'center',
+    flexBasis: 0,
+  },
+}));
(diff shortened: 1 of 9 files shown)


src/components/segmented-control/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+type SegmentedControlProps = {
+  color: InteractiveColor;
+  activeIndex: number;
+  options: SegmentOptions[];
+  enabled?: boolean;
+  style?: StyleProp<ViewStyle>;
+};
+
+export default function SegmentedControl({
```
gorandalum, line 23: Great work on creating this reusable component! Should it maybe be named something with "radio" and each option have `accessibilityRole: 'radio'`? 🤔

src/components/segmented-control/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+        },
+        style,
+      ]}
+    >
+      {options.map((option, i) => {
+        const selected = activeIndex === i;
+        const currentColor = interactiveColor[selected ? 'active' : 'default'];
+        const textColor = enabled
+          ? currentColor.text
+          : theme.text.colors.disabled;
```
gorandalum, line 51: Will this work? My concern is that `theme.text.colors.disabled` will always be black in light mode (and vice versa), regardless of the background it is placed on?

strombraaten:
@rosvik In the video, 17 seconds in, and you've chosen "Start now", and you want to go to payment, does the "time picker" still pop up? Or did you just tap "Later" there?

gorandalum (approved):
Looks good! A couple comments you can look at. 

I have done a quick review, but have not run it to test it or check accessibility. If I get the time I will do that later, if no one beats me to it.

rosvik:
@strombraaten I tapped "later". It doesn't pop up when you go to payment.

strombraaten:
@rosvik I assume that the 24-hour ticket will be moved to its own "card" later on (?), like it's in the design sketches, but for now can we change the fact that it says "1 days / 1 dager" to 1day/dag?

image

strombraaten:
Let me know if this is out of scope, but there are some differences between the bottom sheets, as well as the field for selecting the date and time. That last one might be utilising some native design choices that I don't know if we can change as easily though?

### So if you want me to be nitpicky I'd do the following:

- [ ] Change "Cancel" to [the close-icon](https://www.figma.com/file/e8MWjmPKrisUDbT1zc0fvv/Icons-1.0?node-id=7%3A90)
- [ ] Date picker fields -> Add stroke/outline/border + Change background color to white
- [ ] Bottom sheet 👉 Add line above header to indicate "swipability"
- [ ] Bottom sheet 👉 Round off corners on the top corners of the panel according to sketches
- [ ] Add [confirm-icon](https://www.figma.com/file/e8MWjmPKrisUDbT1zc0fvv/Icons-1.0?node-id=102%3A63) to the "Confirm selection"-button

image

Skjermbilde 2022-05-09 kl  12 57 08

strombraaten (changes requested):
Mentioned which changes that should be made above 👆

src/components/segmented-control/index.tsx
rosvik: - accessibilityRole: can do!
- New naming: Suggestions? "Segmented controls" is whats used in figma. "RadioSegments" maybe?

gorandalum:
@strombraaten, two comments about your comments 😅 
- About text/icon in header, there was a process surrounding this earlier (like 1+ year ago) with the designers (mainly Petter) and I think some usability-testing, where it was decided to always use texts instead of icons. Is this changed?
- Bottom sheet is not swipeable. It is a little complex to implement, so it has not been prioritized.

rosvik:
@strombraaten I know there has been a discussion about this in the past, but I can't find it right now. If I remember correctly, the bottom sheet from the sketches (with border radius and swipe) is based on a native component that isn't (or wasn't) accessibile, which is why we implemented our own bottom sheet. 

> Date picker fields -> Add stroke/outline/border + Change background color to white

These are native iOS/android components, so I don't think we can change the styling to what's in the sketches. 

> Add confirm-icon to the "Confirm selection"-button

Can do 👍

src/components/segmented-control/index.tsx
gorandalum: Yeah. RadioSegments is good :+1:

rosvik:
And the last comment: 

> I assume that the 24-hour ticket will be moved to its own "card" later on (?), like it's in the design sketches, but for now can we change the fact that it says "1 days / 1 dager" to 1day/dag?

This was mentioned by Tor Kjetil as well ([here](https://github.com/AtB-AS/mittatb-app/issues/2457#issuecomment-1118498480)). My thought was that it's unnecessary to add complexity to handle something that won't be visible in production.

strombraaten:
> @strombraaten, two comments about your comments 😅
> 
> * About text/icon in header, there was a process surrounding this earlier (like 1+ year ago) with the designers (mainly Petter) and I think some usability-testing, where it was decided to always use texts instead of icons. Is this changed?
> * Bottom sheet is not swipeable. It is a little complex to implement, so it has not been prioritized.

Great!

1. I wasn't aware of that. I would thing that as long as the icon itself has a alternative text saying "close" instead of an actual word saying the same thing, that it shouldn't really matter, but I haven't tested that, so for now let's just stick to how it is.
2. Regarding the bottom sheet, if it's complex I understand that it hasn't been prioritized. Is there an issue for this though? I understand the need to get something out quickly, but at the same time I would definitely say that this is a part of ticketing v2, just like the other components. And it seems like it's been up for discussion previously, so it might be helpful to at least visualize the need for it. Seeing that this would most likely be a reusable component that we'll use in different areas.

strombraaten:
> @strombraaten And the last comment:
> 
> > I assume that the 24-hour ticket will be moved to its own "card" later on (?), like it's in the design sketches, but for now can we change the fact that it says "1 days / 1 dager" to 1day/dag?
> 
> This was mentioned by Tor Kjetil as well ([here](https://github.com/AtB-AS/mittatb-app/issues/2457#issuecomment-1118498480)). My thought was that it's unnecessary to add complexity to handle something that won't be visible in production.

Yeah, that makes sense 👍

strombraaten (approved):

rosvik:
@strombraaten 

> Is there an issue for this though?

Couldn't find any, but please make one! It would be nice if we could use a pre-made component for this, so the main issue might be to find something that fits our needs.

1. Swipe to dismiss
2. Similar design to the sketches: Border radius, dismiss button
3. React native, with support for both android and iOS
4. Accessible / Support for screen readers

gorandalum:
@rosvik Regarding pre-made-components a good handfull of them were tested before writing our own. All of them fell short on some stuff like:
- The design could not be changed enough to our liking
- Animation-speed
- Greyed out background which would dismiss it if pressed
- Accessibility
- +++

All of them had more than one issue, with the accessibility a glaring issue in all of them. 

We now have a battle-tested bottom-sheet which works for both iOS and Android, so I think the best way forward is adding swipeability to that one, and rather trying to find a premade react-native-swipe-lib or something. Doing research and testing bottom-sheet-libraries was **very** time-consuming, since it's also includes trying to design them, animate them, add accessibility to them, etc., just to find out if they are working or not.

strombraaten:
No worries!

PR #2631, opened 2022-05-09
refactor: hide StartTimeSelection for single tickets

rosvik:
I forgot to add logic to hide the StartTimeSelection component from the single ticket purchase view in #2626. This PR fixes that.


gorandalum (approved):
Something to note if not discussed with designers: 

Ticketing v1 had the start-time-selection-button disabled on single-tickets. This was so the text "Oppstart nå" would still be visible also on single tickets, even though it could not be changed.

PR #2639, opened 2022-05-10
refactor: disable login for NFK

toretefre:
Solves #2642 

NFK does not have the updated backend platform to support deletion of user profiles which is to come in 1.19.

Therefore NFK wants to avoid users signing in and creating user profiles. This is done by adding a `enable_login` value to Remote Config which seems to already be present in the configurations. If not present, `enable_login` defaults to `true` to avoid any consequences for AtB.

While we are already hiding things from the profile menu, the `Default traveller` is also hidden when `enable_ticketing` is `false`.


toretefre:
Awaiting merge of `release/1.19` into master to avoid merge conflicts.

PR #2648, opened 2022-05-11
refactor: fix translation in english

Nisha1306:
The Norwegian version states that minors over 15 will be fined. In English its mentioned minors under 15 are fined. The Norwegian version is correct, hence fixed the english version.


rosvik (approved):
Good catch! 👍

monopoint (approved):
:+1:

PR #2663, opened 2022-05-12
refactor: add interactive colors to feedback component

toretefre:


src/components/feedback/GoodOrBadButton.tsx
```diff
@@ -80,21 +78,21 @@ const useButtonStyle = StyleSheet.createThemeHook((theme) => ({
...
     borderRadius: theme.border.radius.regular,
-    borderColor: theme.static.background[themeColor].background,
+    borderColor: theme.interactive.interactive_2.default.background,
   },
   selectedButton: {
     backgroundColor: hexToRgba(
-      theme.static.background[selectedThemeColor].background,
+      theme.interactive.interactive_2.active.background,
       0.2,
     ),
```
rosvik, line 94: I don't think you need to add transparency here anymore

src/components/feedback/RenderQuestions.tsx
```diff
@@ -110,17 +110,17 @@ const useAlternativeStyle = StyleSheet.createThemeHook((theme) => ({
...
     borderRadius: theme.border.radius.regular,
-    borderColor: theme.static.background.background_0.background,
+    borderColor: theme.interactive.interactive_2.default.background,
   },
   checked: {
     backgroundColor: hexToRgba(
-      theme.static.background.background_accent_3.background,
+      theme.interactive.interactive_2.active.background,
       0.2,
     ),
```
rosvik, line 122: Same here

src/components/feedback/GoodOrBadButton.tsx
toretefre: Got it from here where 20% seems to be specified, but I can see that the colors are a bit lighter than what shows in Figma. Will try without transparency and see how it works :)

image

rosvik: Hmm, that might be an outdated color or something 🤔 I think interactive_2.active should be right without transparency.

rosvik (commented):
👍

rosvik (approved):

toretefre:
Looking better without transparency both for AtB and NFK. Merging 👌 

Simulator Screen Shot - iPhone 13 Pro - 2022-05-13 at 15 18 50

Simulator Screen Shot - iPhone 13 Pro - 2022-05-13 at 15 19 02

HanneLH:
Regression test on iPhone 11 Pro - Ok.

Issue #2695, opened 2022-05-18
Show cancelled trips in Departures


Nisha1306:
Related: #1127 

## Description

Show cancellations in the new Departures view in the APP 

## Example

<img width="1259" alt="image" src="https://user-images.githubusercontent.com/29625161/168980216-35427c0f-6c25-4f74-acc5-cb5ca28ddf13.png">

## Acceptance criteria

- Add a warning-icon

- Add a text box in the details view to say that the departure has been cancelled

- Strikethrough datetime also in details view

HanneLH:
Tested on **iPhone 11 Pro** and the cancelled trips shows in the departures section and voice over is working ass well. 
![IMG_0033](https://user-images.githubusercontent.com/104772494/169284630-0ea6c788-a1d1-41a8-9644-291e23436f40.PNG)
![IMG_0032](https://user-images.githubusercontent.com/104772494/169284645-6ba7cd95-524a-44d6-a6de-3ec9b713adc9.PNG)
![IMG_0031](https://user-images.githubusercontent.com/104772494/169284650-891882a6-b843-48c5-9378-5fc5ac7201de.PNG)
![IMG_0030](https://user-images.githubusercontent.com/104772494/169284652-4f29302a-27e9-48eb-a581-23972df41085.PNG)
To verify that the trips was cancelled entur.no was used.
![Skjermbilde 2022-05-19 kl  13 41 19](https://user-images.githubusercontent.com/104772494/169285167-367668df-fb89-4504-a267-b8062a0a9b55.png)

PR #2696, opened 2022-05-18
feat: show cancelled trips in departures section

Nisha1306:
resolves https://github.com/AtB-AS/mittatb-app/issues/2695


monopoint (changes requested):
components/cancelled-departure should be moved to TripDetails/Components
also we should move Pagination to TripDetails/components because it's not really reusable outside trips

Nisha1306:
> components/cancelled-departure should be moved to TripDetails/Components also we should move Pagination to TripDetails/components because it's not really reusable outside trips

Migrated to relevant directories 👍

monopoint:
Talk Back does not inform users that the departure is cancelled:
image

Nisha1306:
> Talk Back does not inform users that the departure is cancelled: image

Fixed this 👍

monopoint (approved):
great! looks good!

PR #2705, opened 2022-05-18
test: e2e tests fixes

tormoseng:
There are som failing e2e test scenarios on GH Actions:
- travelsearch: should list recent locations
    - the elements for tapping is hiding behind the keyboard. Solved by resetting the search history.
- ticketing: should get an offer for a single ticket
    - add the missing test-ID offerTotalPriceText
- ticketing: should be able to change travellers
    - add the missing test-ID offerTotalPriceText
- ticketing: should be able to change zone
    - tapping a zone removes the keyboard, so another tap is in order to actually tap the chosen zone


rosvik (approved):
👍

PR #2706, opened 2022-05-18
refactor: future proof date-fns and shortWaitTime

monopoint:
Date-fns show warnings that it will now longer support dates in string type.

updatet isShortWaitTime to parse any strings and updated tests with a "weird" case. 

image


Nisha1306 (approved):
👍

PR #2716, opened 2022-05-19
fix: remove redundant accessible element for better usability

Nisha1306:
As of now we have two accessible elements which give two different kind of information, while only one of them is important. Getting rid of redundant one and moving the informative message to outer accessible element.


rosvik (approved):

monopoint (approved):
👍

PR #2733, opened 2022-05-24
Conditional iOS-build

toretefre:


toretefre:
Is there any way to test this without merging, @gorandalum ?

gorandalum (approved):

gorandalum:
Nice. Might be able to test with a smaller script which contains the logic and something like:
```
on:
  push:
    branches:
      - tefre/conditional-build-workflow
```

But also ok to just merge.

toretefre:
Did not work, reverting. For now I think I can start the job, and just cancel the AtB build.

`The workflow is not valid. .github/workflows/build-store-ios.yml (Line: 14, Col: 9): Unrecognized named-value: 'matrix'. Located at position 2 within expression: (matrix.org == 'atb' && !contains(github.ref, 'nfk') || (matrix.org == 'nfk' && !contains(github.ref, 'atb'))`

gorandalum:
I see that other `if`s in the code doesn't have curly brackets, but is written like this:
```
if: matrix.org == 'atb'
```

PR #2735, opened 2022-05-24
Revert "build: conditional build to store based on org and release tag"

toretefre:
Reverts AtB-AS/mittatb-app#2733


mikaelbr (approved):
~See #2733 ~ (already mentioned 😄 )

PR #2737, opened 2022-05-25
fix: add accessibilty hint and resolve label correctly

Nisha1306:
Missed the accessibility hint from last fix and also resolve the cancellation info correctly


monopoint (approved):
:+1:

PR #2738, opened 2022-05-25
build: cancel org specific jobs

toretefre:
v2 of #2733

### DIFF
--- a/.github/workflows/build-store-ios.yml
+++ b/.github/workflows/build-store-ios.yml
@@ -14,6 +14,9 @@ jobs:
     timeout-minutes: 90
     runs-on: macos-11
     steps:
+      - name: Check if release tag is org specific and exit for other orgs
+        if: ${{ (matrix.org == 'atb' && contains(github.ref, 'nfk')) || (matrix.org == 'nfk' && contains(github.ref, 'atb')) }}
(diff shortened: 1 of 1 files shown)


toretefre:
Would a `exit(0)` be better to avoid a potentially large number of failed jobs, or is it more understandable and explicit as is?

toretefre:
Found an example of exit usage here, should we do the same maybe? 

https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-steps-context

mikaelbr:
Exit 0 would probably be better, yes. But maybe then it's more difficult to stop the following steps? 🤔

toretefre:
Aha, exit 1 will lead to remaining steps being cancelled, while exit 0 means successful and make remaining steps being executed? 

In that case exit 1 would be better, yes.

Are we sure that exit 1 in the first step stops remaining steps from being executed, or do we need to make something else actually fail?

toretefre:
[This solution](https://stackoverflow.com/a/68777117) also might be possible, however it seems we might need to add an `if`-check on every subsequent step. 

The post above it by the way seems to confirm that `exit 1` will fail all remaining job steps.

PR #2739, opened 2022-05-25
refactor: replace NFK traffic situation url

toretefre:
Avoids showing atb.no references in NFK app.

image (2)
image (1)


toretefre:
Unchanged for AtB: 
image

toretefre:
NFK text may change after input from PM, but nothing technical should be changed.

mikaelbr (approved):

PR #2740, opened 2022-05-27
feat/moved-24h-ticket-out-of-periodic-one

Nisha1306:
This PR is to extract 24-hour ticket out of Periodic ticket. As 24h ticket is still not considered a product and we have agreed upon not to change the way we receive these product details from Entur, I have termed it as Independent Periodic ticket which we can extend later on if we want to extract more tickets like weekly, monthly out of periodic section.

### DIFF
--- a/src/reference-data/defaults/preassigned-fare-products.json
+++ b/src/reference-data/defaults/preassigned-fare-products.json
@@ -31,6 +31,68 @@
     }
   },
   {
+    "id": "ATB:PreassignedFareProduct:1fb2d3e4",
+    "version": "",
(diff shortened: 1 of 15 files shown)


src/reference-data/defaults/preassigned-fare-products.json
```diff
@@ -30,6 +30,68 @@
       ]
     }
   },
+  {
+    "id": "ATB:PreassignedFareProduct:1fb2d3e4",
```
Nisha1306, line 34: I have updated this from staging. Does this have any consequences?

PR #2742, opened 2022-05-30
build: Avoid changelog error after org-specific releases

toretefre:
Skips nfk-only releases when finding latest releasetag, so that all changes are included in the changelog of next common release.

### DIFF
--- a/tools/release/changelog.config.js
+++ b/tools/release/changelog.config.js
@@ -1,7 +1,7 @@
 const {execSync} = require('child_process');
 
 const latestReleaseTag = execSync(
-  `git tag --sort=creatordate |  grep -E '(^alpha-.*)|(^v.*)|(.*nfk.*)' | tail -1`,
+  `git tag --sort=creatordate |  grep -E '/(?!(.*atb.*)|(.*nfk.*))((alpha-.*)|(v.*))' | tail -1`,
 )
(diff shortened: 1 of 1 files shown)


tools/release/changelog.config.js
```diff
@@ -1,7 +1,7 @@
 const {execSync} = require('child_process');
 
 const latestReleaseTag = execSync(
-  `git tag --sort=creatordate |  grep -E '(^alpha-.*)|(^v.*)' | tail -1`,
+  `git tag --sort=creatordate |  grep -E '/(?!(.*(atb|nfk).*))((alpha-.*)|(v.*))' | tail -1`,
```
mikaelbr, line 4: Looks good. Maybe not strictly necessary to check if there are characters before? Or is the thought to support things like `realease/` prefix?

mikaelbr:
FYI @gorandalum also.

mikaelbr (approved):

gorandalum (approved):
Good :+1:

tools/release/changelog.config.js
toretefre: The thought is to support anything really, just including an org-check, in case the naming of the releases would change in the future. Are you referring to the removal of the `^`-s when talking about checking characters before?

PR #2747, opened 2022-06-02
fix: add 24 hour ticket in recent purchase

Nisha1306:
Now when we have added 24 hour ticket as a new ticket type, it got missed from recent purchases. So have added it here.

### DIFF
--- a/src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
+++ b/src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
@@ -45,7 +45,8 @@ export const RecentTickets = () => {
         if (
           ticketType === 'single' ||
           ticketType === 'period' ||
-          ticketType === 'carnet'
+          ticketType === 'carnet' ||
+          ticketType === 'hour24'
         )
           return true;
         else return false;


src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
```diff
@@ -45,7 +45,8 @@ export const RecentTickets = () => {
         if (
           ticketType === 'single' ||
           ticketType === 'period' ||
-          ticketType === 'carnet'
+          ticketType === 'carnet' ||
+          ticketType === 'hour24'
```
gorandalum, line 49: Can't this be removed all together? 🤔  This should just be handled by the `productIsSellableInApp`.

gorandalum (approved):
Looks good, but I don't see why the ticket type check is necessary. Just remove it?

src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
gorandalum: Or let me think. Maybe it is necessary when adding new products 🤔  Just leave it 😅

Nisha1306: yeah...if we add new products other than these. so not touching it for now :)

PR #2750, opened 2022-06-09
Replace hardcoded tariff zone in getFeatureFromVenue

toretefre:
Solves #73

Still an issue with train stopPlaces in Trøndelag, but solves critical issue for orgs other than AtB.


src/api/geocoder.ts
```diff
@@ -46,7 +46,7 @@ export async function getFeatureFromVenue(
       lon: venue.coordinates.longitude,
       limit: 1,
       layers: ['venue'],
-      tariff_zone_authorities: 'ATB',
```
gorandalum, line 49: To be more precise, "AtB" is not a tariff zone, but the authority (organization) for the tariff zone collection. Maybe use `TARIFF_ZONE_AUTHORITY` for the env variable as well.

gorandalum (approved):

src/api/geocoder.ts
toretefre: Ah, yes, will rename to `TARIFF_ZONE_AUTHORITY` 👍

PR #2756, opened 2022-06-15
feat: do-not-update-travel-search-after-every-20m

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/1653

### DIFF
--- a/src/screens/Nearby/state.ts
+++ b/src/screens/Nearby/state.ts
@@ -43,7 +43,6 @@ export type DepartureDataState = {
   cursorInfo: DepartureGroupMetadata['metadata'] | undefined;
   lastRefreshTime: Date;
   searchTime: SearchTime;
-  fromLocation?: Location | undefined;
 };
 
 const initialState: Omit<
@@ -308,7 +307,6 @@ const reducer: ReducerWithSideEffects<
         ...state,
         isLoading: false,
         locationId: action.locationId,
-        fromLocation: action.fromLocation,
         data: action.reset
           ? action.result.data
           : (state.data ?? []).concat(action.result.data),
--- a/src/screens/Nearby/state.ts
+++ b/src/screens/Nearby/state.ts
@@ -381,7 +381,7 @@ export function useDepartureData(
     [location?.id, favoriteDepartures],
   );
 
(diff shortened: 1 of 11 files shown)


gorandalum:
I haven't tried running this, but my initial thought when seeing this change is that the geolocation is now updated continuously, which will make the app reload even more often. My feeling is that this also must be changed at the places in the code where we are listening on geolocation coordinates changes to perform updates/reloads.

Nisha1306:
> I haven't tried running this, but my initial thought when seeing this change is that the geolocation is now updated continuously, which will make the app reload even more often. My feeling is that this also must be changed at the places in the code where we are listening on geolocation coordinates changes to perform updates/reloads.

Not really. i tried this. If you set it to zero, it wont update the location at all. And if we get rid of this all together, it defaults to 100m. 
```distanceFilter (m) - The minimum distance from the previous location to exceed before returning a new location. Set to 0 to not filter locations. Defaults to 100m.```

**Comment by gorandalum** [comment]
Nice. What you can test is that when manually performing a new search, then the updated phone coordinates should be used. For example if a user makes a travel search, moves 30 meters, then refreshes the travel search.

Good tool for testing this in iOS simulator:
https://github.com/MobileNativeFoundation/set-simulator-location

**Comment by Nisha1306** [comment]
@gorandalum I think it should be fixed now for assistant as well as departures v1 and v2. Please have a look.

File: src/screens/Nearby/state.ts [hunk 1]
```diff
@@ -404,7 +408,9 @@ export function useDepartureData(
     [location?.id, favoriteDepartures],
   );
 
-  useEffect(refresh, [location?.id]);
+  useEffect(refresh, [
+    location?.resultType === 'geolocation' ? null : location?.id,
+  ]);
```
gorandalum, line 413: I not sure if this will work as intended. Remember that manually refreshing the departures list (in v1) or nearby stops list (in v2) should use the current position for the user.

gorandalum (changes requested):

Nisha1306:
@gorandalum  I think this should be fine now. 😅

gorandalum (approved):

PR #2758, opened 2022-06-20
feat: add repeat purchase header to differentiate correctly between actions

Nisha1306:
Ref: https://github.com/AtB-AS/kundevendt/issues/1666

Before: 

image

After:

image (1)

Figma Link: https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App-1.2?node-id=6648%3A45714

### DIFF
--- a/src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
+++ b/src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
@@ -64,9 +64,6 @@ export const RecentTickets = () => {
     <View
       style={{backgroundColor: theme.static.background.background_1.background}}
     >
-      <ThemeText type="body__secondary" style={styles.header}>
-        {t(RecentTicketsTexts.repeatPurchase.label)}
-      </ThemeText>
       {loading && (
         <View
           style={{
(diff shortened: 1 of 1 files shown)


gorandalum:
Could you add before/after pictures in this PR? It's nice for small changes like this, then we don't need to run it locally to look at the changes.

src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
```diff
@@ -64,6 +64,15 @@ export const RecentTickets = () => {
     <View
       style={{backgroundColor: theme.static.background.background_1.background}}
     >
+      <ThemeText
+        type="body__secondary"
+        style={{
+          marginTop: theme.spacings.xLarge,
+          marginLeft: theme.spacings.xLarge,
+        }}
```
gorandalum, line 72: Set these fields in `useStyles` hook instead. I see there are other places where this hasn't been followed in this file, but I think it is good to do it "correctly" even though.

Nisha1306:
@gorandalum  done..could you please review again.

src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
```diff
@@ -64,6 +64,9 @@ export const RecentTickets = () => {
     <View
       style={{backgroundColor: theme.static.background.background_1.background}}
     >
+      <ThemeText type="body__secondary" style={styles.header}>
+        {t(RecentTicketsTexts.repeatPurchase.label)}
+      </ThemeText>
```
gorandalum, line 69: Now this heading is shown also when there is no recent purchases, which will probably look a little odd. Maybe move it below into the `!loading && !!recentTickets.length` check?

gorandalum (changes requested):

src/screens/Ticketing/Tickets/RecentTickets/RecentTickets.tsx
Nisha1306: Oopsie..ofcourse :) Fixed.

gorandalum (approved):

PR #2759, opened 2022-06-20
refactor: Use new mobiletoken sdk

gorandalum:
Fixes AtB-AS/kundevendt#1678


gorandalum:
No need to review changes under `.yalc`.

If GitHub in the browser is slow because of the amount of changes, it may be better to look at it locally in the IDE.

PR #2763, opened 2022-06-21
refactor: Update tokens endpoints to v3

gorandalum:


Nisha1306 (approved):
M sure you have a valid reason for it but just wondering why did we switch to v3 directly? 
Update: I mean instead of v2

PR #2773, opened 2022-06-23
build: add nfk gradle properties

toretefre:
Adds a copy of `env/atb/gradle.properties` with the correct nfk values.


mikaelbr (approved):
Probably ok? 😄

PR #2787, opened 2022-06-29
feat: show specific warning messages if the relevant token is not the one in hand

Nisha1306:
Solves: https://github.com/AtB-AS/kundevendt/issues/220

Scenarios covered:

Case 1: Info on the ticket itself about errors (internet lack ex.)
Case 2: Info on the ticket itself if not token
Case 3: Info on the ticket itself when tokens are missing
Case 4: Info that the t: card is used as a travel document inside the ticket itself
Case 5: Info about mobile token is used as a travel document inside the ticket itself

Eg Case 5:

Screenshot_20220629-154134

### DIFF
--- a/src/screens/Ticketing/Ticket/TicketInfo.tsx
+++ b/src/screens/Ticketing/Ticket/TicketInfo.tsx
@@ -9,7 +9,7 @@ import {
   getReferenceDataName,
 } from '@atb/reference-data/utils';
 import {StyleSheet, useTheme} from '@atb/theme';
-import {PreactivatedTicket, useTicketState} from '@atb/tickets';
+import {PreactivatedTicket} from '@atb/tickets';
 import {TicketTexts, useTranslation} from '@atb/translations';
 import React, {ReactElement} from 'react';
 import {View} from 'react-native';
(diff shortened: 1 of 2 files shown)


src/screens/Ticketing/Ticket/TicketInfo.tsx
```diff
@@ -96,6 +103,49 @@ export const TicketInfoView = (props: TicketInfoViewProps) => {
...
+      <ThemeText type="body__secondary">
+        {t(TicketTexts.warning.unableToRetrieveToken)}
+      </ThemeText>
+    );
+  if (remoteTokens && remoteTokens.length === 0)
+    return (
+      <ThemeText type="body__secondary">
+        {t(TicketTexts.warning.noTokenFound)}
+      </ThemeText>
+    );
```
gorandalum, line 124: An empty `remoteTokens?` list will mae `isError` `true`, so this clause is redundant. Just showing the `unableToRetrieveToken` is good enough, as something has gone wrong.

gorandalum (approved):
Nice! Just a note about some code that can be removed.

src/screens/Ticketing/Ticket/TicketInfo.tsx
Nisha1306: Ohh Oka. I will get rid of this. It was in one of the cases mentioned in the issue, so I thought its a valid scenario :)

src/screens/Ticketing/Ticket/TicketInfo.tsx
```diff
@@ -169,12 +204,10 @@ const TicketInfoTexts = (props: TicketInfoViewProps) => {
           {tariffZoneSummary}
         </ThemeText>
       )}
-      {showTravelCardActiveWarning && (
-        <View style={styles.notInspectableWarning}>
-          <ThemeIcon svg={Warning} style={styles.notInspectableWarningIcon} />
-          <ThemeText isMarkdown={true}>
-            {t(TicketTexts.ticketInfo.travelcardIsActive)}
-          </ThemeText>
+      {warning && (
```
Nisha1306, line 207: @Gøran Dalum  I haven't put this under any flag/condition like "show only for new tokens" or something. Is there anything like that, which should to be considered?

gorandalum: Not that I can think of now

PR #2790, opened 2022-06-30
feat: improve warning message while buying ticket when current device is not inspectable

Nisha1306:
Fixes: https://github.com/AtB-AS/kundevendt/issues/220#issuecomment-1156542681 

Before: 
Before

After:
After


src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -194,7 +197,11 @@ const PurchaseOverview: React.FC<OverviewProps> = ({
           <MessageBox
             isMarkdown={true}
             containerStyle={styles.warning}
-            message={t(PurchaseOverviewTexts.notInspectableTokenDeviceWarning)}
+            message={t(
+              PurchaseOverviewTexts.notInspectableTokenDeviceWarning(
+                getDeviceNameForCurrentToken(token, remoteTokens),
```
mikaelbr, line 202: What happens here if we cannot find a device name? Haven't there been cases where the device name is unknown?

src/mobile-token/utils.ts
```diff
@@ -18,5 +19,13 @@ export const isMobileToken = (remoteToken?: RemoteToken) =>
 export const getDeviceName = (remoteToken?: RemoteToken) =>
   remoteToken?.keyValues?.find((kv) => kv.key === 'deviceName')?.value;
 
+export const getDeviceNameForCurrentToken = (
+  activatedToken?: ActivatedToken,
+  remoteTokens?: RemoteToken[],
+) =>
+  getDeviceName(
+    remoteTokens?.find((token) => token.id === activatedToken?.getTokenId()),
```
mikaelbr, line 27: activatedToken doesn't have `keyValues` itself?

Nisha1306: No, it doesnt.

Nisha1306:
> The message box for not inspectable token warning comes (mainly) when the inspectable token is another token than the current native token. Here `getDeviceNameForCurrentToken` gets the name for the current native token, but should get the name of the current inspectable token.

But if we read the message, isnt it correct that we tell the user that they can not use the current device they have in hand. Or have I interpreted it wrong? Isn't it @hildeor ?

src/mobile-token/utils.ts
```diff
@@ -22,5 +23,13 @@ export const hasNoTokenType = (remoteToken?: RemoteToken) =>
 export const getDeviceName = (remoteToken?: RemoteToken) =>
   remoteToken?.keyValues?.find((kv) => kv.key === 'deviceName')?.value;
 
+export const getDeviceNameForCurrentToken = (
+  activatedToken?: ActivatedToken,
+  remoteTokens?: RemoteToken[],
+) =>
```
gorandalum, line 29: Can contemplate changing this to take in `tokenId` as param instead of `ActivatedToken`, as that would make it more reusable.

gorandalum (approved):
Looks good! :+1:

PR #2791, opened 2022-06-30
chore: Yalc add after git crypt unlock for iOS

gorandalum:


mikaelbr (approved):
Makes sense. Looks like it's running also

PR #2792, opened 2022-07-01
feat: Add mobile token fallback to static qr

gorandalum:
If mobile token fails the app should fallback to static qr which is
present on the fare contract in Firestore. This is toggled by the flag
'enable_token_fallback' in RemoteConfig.

Some other changes:
- Generally don't show errors when mobiletoken fails when mobile token
  fallback is enabled.
- Update mobile token state in Intercom, to be used in determining how
  many users experience mobile token errors.


gorandalum:
I'm not completely "happy" with the changes. The logic for simple error messages is pretty complicated as it is based on multiple factors:
- is mobile token enabled for the user in remote config
- is mobile token working as intended
- if fallback enabled in remote config
- what is the ticket type if mobile token is not enabled

This logic is present several places where we show warnings or error messages to the user if not working as intended. The logic could probably have been rewritten so it is handled more centralized, and added tests to ensure that it is handled correctly.

I have not prioritized this now because of the many story points in the sprint, and after summer we should be able to remove some logic when mobile token is enabled for all users. And later when mobiletoken is stable enough the fallback can also be removed, simplifying this quite a bit.

src/screens/Ticketing/Ticket/Details/QrCode.tsx
```diff
@@ -123,39 +160,39 @@ const useQrCode = (
...
+//       <MessageBox
+//         type={'error'}
+//         title={t(TicketTexts.details.qrCodeErrors.generic.title)}
+//         message={t(TicketTexts.details.qrCodeErrors.generic.text)}
+//         onPress={retry && (() => retry(true))}
+//         onPressText={retry && t(TicketTexts.details.qrCodeErrors.generic.retry)}
+//       />
+//     </Sections.GenericItem>
+//   );
+// };
```
jorelosorio, line 178: Perhaps remove all these comments if are not going to be used.

jorelosorio (commented):
@gorandalum Is quite hard to follow the logic, do you have some examples of how to test this? and I have a couple of questions:

- Is the new TokenMobileSDK running always by default?
- By enabling fallback disable everything or just displays a different QR if the new token Azteca QR code fails?
- What if the entire mobile SDK token fails and needs to be turned off, is there a code to handle that?
- In which scenarios could the QR code fail?

src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -60,14 +60,21 @@ const PurchaseOverview: React.FC<OverviewProps> = ({
...
+    fallbackEnabled,
+  } = useMobileTokenContextState();
   const tokensEnabled = useHasEnabledMobileToken();
   const {customerProfile} = useTicketState();
   const hasProfileTravelCard = !!customerProfile?.travelcard;
 
   const showProfileTravelcardWarning = !tokensEnabled && hasProfileTravelCard;
 
-  const showNotInspectableTokenWarning = tokensEnabled && !deviceIsInspectable;
+  const showInspectableTokenWarning =
```
Nisha1306, line 74: could this be simplified to **tokensEnabled && (mobileTokenError ? !fallbackEnabled : !deviceIsInspectable);** or something?

src/screens/Ticketing/Ticket/Details/QrCode.tsx
```diff
@@ -123,39 +160,39 @@ const useQrCode = (
...
-      />
-    </Sections.GenericItem>
-  );
-};
-const QrCodeDeviceNotInspectable = ({
-  inspectableToken,
-}: {
-  inspectableToken: RemoteToken;
-}) => {
+// const QrCodeError = ({retry}: {retry?: (forceRestart: boolean) => void}) => {
```
Nisha1306, line 164: redundant?

src/screens/Ticketing/Ticket/Details/QrCode.tsx
```diff
@@ -5,39 +5,76 @@ import {TicketTexts, useTranslation} from '@atb/translations';
...
-  if (deviceIsInspectable) return <QrCodeSvg />;
+  if (!mobileTokenEnabled) return 'static';
 
-  const inspectableToken = findInspectable(remoteTokens);
-  if (inspectableToken) {
-    return <QrCodeDeviceNotInspectable inspectableToken={inspectableToken} />;
-  }
+  if (isLoading) return 'loading';
+  if (isError) return 'static';
+  if (deviceIsInspectable) return 'mobiletoken';
```
Nisha1306, line 65: After seeing so many conditions it has become a little difficult to differentiate between several terminologies like isInspectable and deviceIsInspectable? Not sure if its just me 😅

gorandalum:
> @gorandalum Is quite hard to follow the logic, do you have some examples of how to test this

@tormoseng is pretty experienced in testing this and provoking error situatation, so he can probably do that.

> Is the new TokenMobileSDK running always by default?

The new mobile token sdk is running as long as the user is enrolled into mobile token, from the period tickets invitation code.  There is no references to the old mobile token sdk now. The logic whether mobile token is enabled for the user is in the `useHasEnabledMobileToken` hook:
```ts
export function useHasEnabledMobileToken() {
  const {customerProfile} = useTicketState();
  const {enable_period_tickets} = useRemoteConfig();

  if (Platform.OS !== 'android' && DeviceInfo.isEmulatorSync()) {
    return false;
  }

  return customerProfile?.enableMobileToken || enable_period_tickets;
}
```

> By enabling fallback disable everything or just displays a different QR if the new token Azteca QR code fails?

Enabling fallback means that if the mobile token sdk fails then the user will get a static qr. It is mainly the code in the `useQrCodeStatus` hook that handles it. It will return `static` if mobile token is not enabled, something goes wrong with mobile token, and alsoas the default return value of the method:
```ts
const useQrCodeStatus = (
  validityStatus: ValidityStatus,
  isInspectable: boolean,
) => {
  ...
  if (!mobileTokenEnabled) return 'static';
  ...
  if (isError) return 'static';
  ...
  return 'static';
};
```
Also the `fallback` enabled value will affect whether some error messages are shown or not.

> What if the entire mobile SDK token fails and needs to be turned off, is there a code to handle that?

We can turn of mobile token with the flag `enable_period_tickets` in RemoteConfig. Also we can turn of fallback with the flag `enable_token_fallback` in RemoteConfig.

> In which scenarios could the QR code fail?

Network errors, implementation bugs, unsupported devices, ++. We saw during implementation of the sdk in the task force that we got several different errors, and will probably find new ones when releasing it to lots of users.

src/screens/Ticketing/Ticket/Details/QrCode.tsx
```diff
@@ -179,6 +216,33 @@ const QrCodeLoading = () => {
...
+const StaticQr = ({fc}: {fc: FareContract}) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+  const [qrCodeSvg, setQrCodeSvg] = useState<string>();
+
+  useEffect(() => {
+    if (fc.qrCode) {
+      qrcode.toString(fc.qrCode, {type: 'svg'}).then(setQrCodeSvg);
+    }
+  }, [fc.qrCode, setQrCodeSvg]);
```
Nisha1306, line 228: why does **setQrCodeSvg** have to be in the dependency array here? 🤔

src/screens/Profile/TravelToken/index.tsx
```diff
@@ -40,7 +40,7 @@ export default function TravelCard({navigation}: TravelCardScreenProps) {
         leftButton={{type: 'back'}}
       />
       <ScrollView style={styles.scrollView}>
-        <TravelTokenBox showIfThisDevice={true} />
+        <TravelTokenBox showIfThisDevice={true} alwaysShowErrors={true} />
```
Nisha1306, line 43: what is the purpose of alwaysShowErrors if it is always going to be true? 🤔

src/screens/Ticketing/Ticket/Details/QrCode.tsx
Nisha1306: sounds much clearer :)

src/screens/Ticketing/Ticket/Details/QrCode.tsx
gorandalum: This code was there from before and just moved by me. It doesn't have to be, but we haven't any coding guideline regarding this for our projects.

I tried to find if there was any best-practice regarding this, but nothing clear cut. Quote [Dan Abramov](https://github.com/reactjs/reactjs.org/issues/4506):
> … you can add it for consistency but it doesn't really matter one way or another.

src/screens/Profile/TravelToken/index.tsx
gorandalum: The `TravelTokenBox` is used in other parts in the code (not touched by this PR) where the `alwaysShowErrors` is not added, and thus `false`.

src/travel-token-box/index.tsx
```diff
@@ -145,13 +152,17 @@ const TravelDeviceTitle = ({
   }
 };
 
-const ErrorMessages = (remoteTokens?: RemoteToken[]) => {
+const ErrorMessages = (
+  fallbackEnabled: boolean,
```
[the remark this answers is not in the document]
gorandalum: Yeah. Moving this.

src/screens/Ticketing/Ticket/Details/QrCode.tsx
gorandalum: I actually will keep this for now. We have Figma design for the error, we have been through deciding norwegian/english texts for the error, and it is ment to be readded when fallback is removed. If this is still not reenabled in 2023 you can say "I told you so" 😅

src/screens/Ticketing/Ticket/Details/QrCode.tsx
gorandalum: I actually will keep this for now. We have Figma design for the error, we have been through deciding norwegian/english texts for the error, and it is ment to be readded when fallback is removed. If this is still not reenabled in 2023 you can say "I told you so" 😅

src/screens/Ticketing/Ticket/Details/QrCode.tsx
Nisha1306: hmmm.. I atleast dont really do that. 😅

src/screens/Profile/TravelToken/index.tsx
Nisha1306: makes sense 👍  Just so many flags 😅

Nisha1306 (approved):
👏

src/screens/Profile/TravelToken/index.tsx
jorelosorio: @gorandalum Thanks for the explanation, of course, there will more to improve but as is now I think is fine, @Nisha1306 also pointed out good things here :) I do not have more to add than trying to test this solution deeply.

jorelosorio:
@gorandalum Thanks for the explanation, of course, there will more to improve but as is now I think is fine, @Nisha1306 also pointed out good things here :) I do not have more to add than trying to test this solution deeply.

jorelosorio (approved):

PR #2795, opened 2022-07-01
Update project.pbxproj to include CoreNFC.framework as optional

jorelosorio:
It will fix an issue with devices that don't support this feature. Despite the code handling, this seems like including the framework and making it optional will not break the app.


tormoseng:
What is the consequence of not including it - like on the iPhone 6?

jorelosorio:
>

jorelosorio:
> What is the consequence of not including it - like on the iPhone 6? Do we use NFC anywhere now?

If we do not include it (I mean the framework) then it will fail as is happening now, however, in the code, there is a condition to check whether the NFC is supported or not but still, we need to include it.

We are not using NFC but is needed for the attestation, it gets the status of this, enabled, disabled or not supported. but not a real implementation of it.

tormoseng (approved):
LGTM! Will be interesting to try it out in staging.

gorandalum (approved):

PR #2796, opened 2022-07-04
feature: add timeout to mobile token api calls

Nisha1306:
Fixes: https://github.com/AtB-AS/kundevendt/issues/1904

### DIFF
--- a/src/mobile-token/tokenService.ts
+++ b/src/mobile-token/tokenService.ts
@@ -1,5 +1,5 @@
 import {RemoteTokenService} from '@entur/atb-mobile-client-sdk/token/token-state-react-native-lib/src';
-import {createClient} from '@atb/api/client';
+import client from '@atb/api/client';
 import {
   ActivateResponse,
   CompleteResponse,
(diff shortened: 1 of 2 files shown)


src/mobile-token/tokenService.ts
```diff
@@ -28,6 +29,8 @@ type ListTokensResponse = {
   tokens: RemoteToken[];
 };
 
+const client = createClient(API_BASE_URL, 15000);
```
jorelosorio, line 32: It is ok if we set this value from some config?

Nisha1306: you mean the timeout? Just wondering if we really need it?

gorandalum (approved):
Great if its this easy. Have you tested it with both request timeout and connection timeout?

About timeout from config: Maybe later, would probably need to be done through RemoteConfig.

This change creates two axios instances in the app with the same base url, which probably is not a problem, just may seem a little "strange". Another option would just be to add the timeout on the requests in `tokenService`, like this:
```ts
client
      .post<InitResponse>('/tokens/v3/init', data, {
        headers: {
          [CorrelationIdHeaderName]: traceId,
          [IsEmulatorHeaderName]: String(isEmulator),
        },
        authWithIdToken: true,
        timeout: 15000,
      })
```
Not saying we should change it or that it is better, just mentioning it as an alternative. You can decide Nisha 😄

Nisha1306:
> > connection timeout
> 
> @gorandalum Right, I actually started with adding it for each request, but for some reason I tried creating another client and forgot to move back 😅 . I would add it back i.e. separately for each request. 👍
> 
> I checked only for response timeout by reducing the timeout value to too low to get a response. I couldn't simulate connection timeout though.

Tested connection timeout as well. looks good to me 👍

PR #2799, opened 2022-07-04
feat: Add abt-token-client npm lib

gorandalum:
- Use abt-token-client-sdk from Entur private registry.
- Add an encrypted .npmrc file for authorizing with Entur private registry
- Also need to run git-crypt-unlock before tsc and lint
- Removed all references yalc, both in code and during build
- The token-core-ios-lib is not published to an online repository, and is fetched through git
- Some changes to abt-token-client-sdk applied in patches to not needing to wait for them to be applied and released
- Refresh remote config in debug menu
- Some bug fixes


mikaelbr:
I wonder if maybe the npmrc should be generated outside and not included as part of the repo?

gorandalum:
> I wonder if maybe the npmrc should be generated outside and not included as part of the repo?

Yeah, open for suggestions. My first thought was to have it as a part of `yarn setup` to generate it from GH secrets, then I remembered `setup` is run after `yarn install`. In the workflows it is simple to generate it, but wanted it to also work locally.

gorandalum:
But this uses the atb-build identity-token also locally. Maybe it would be just better to say "put your personal Entur jfrog credentials in your npmrc file" for local development.

package.json
```diff
@@ -26,10 +26,11 @@
     "@atb-as/theme": "^6.0.0-alpha.7",
     "@bugsnag/react-native": "^7.12.0",
     "@bugsnag/source-maps": "^2.3.1",
+    "@entur/abt-mobile-client-sdk": "^1.0.5-rc.0",
+    "@entur/abt-token-server-node-lib": "^1.0.0",
```
mikaelbr, line 30: Is this included for the types only? 🤔

mikaelbr (approved):
Seems sensible enough. I haven't checked the code changes, but only the npmrc, gradle.properties and pod.lock changes. I think all secrets are already set for both AtB and NFK also, so it should work as is?

package.json
gorandalum: Yeah. Is going to get it's own types-lib, but not done yet by Entur. Can probably move it to dev dependencies?

mikaelbr: Yep. Has probably no implications for this tho, but technically "more correct"

gorandalum:
I changed `ENTUR_REGISTRY_PASSWORD` to `ENTUR_REGISTRY_TOKEN`, since the access token may also be used as password, and it's working for both npm and gradle. Talked with Tore and added it for NFK.

PR #2800, opened 2022-07-05
fix: make onboarding flow responsive

Nisha1306:
Fox for comments on https://github.com/AtB-AS/kundevendt/issues/434


monopoint (dismissed):

monopoint (changes requested):
needs a more robust strategy

PR #2808, opened 2022-07-07
fix: Change order of the terms t:kort and mobilen

jorelosorio:
The translations have been changed with the right order for t:kort and mobilen / t:card and phone.


HanneLH (approved):
🤩

Nisha1306 (approved):

PR #2809, opened 2022-07-07
refactor: Check mobile token user change before retrieving token

gorandalum:
Retrieving token has possible side effects like renewing,
activation, etc. This should not be done on the old token if there has
been any user change.

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -96,7 +96,7 @@ const MobileTokenContextProvider: React.FC = ({children}) => {
    * Load/create native token and handle the situations that can arise.
    *
    * - First check if there has been a user change. If there has, then a new
-   *   token should always be created (skip to last step)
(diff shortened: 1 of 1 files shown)


tormoseng (approved):
LGTM!

src/mobile-token/MobileTokenContext.tsx
```diff
@@ -95,10 +95,10 @@ const MobileTokenContextProvider: React.FC = ({children}) => {
   /**
    * Load/create native token and handle the situations that can arise.
    *
-   * - First retrieve the token from the client-sdk. If necessary, the token
-   *   will be renewed while getting it.
-   * - If a new user has logged in then the old token will be wiped both
-   *   natively and remote.
+   * - First check if there has been a user change. If there has, then a new
+   *   token should always be created (skip to last step)
```
monopoint, line 99: nitpicking but "."

monopoint (approved):

PR #2814, opened 2022-07-12
fix: handle payment callback from Android urls

toretefre:
Should solve remaining Android bug from https://github.com/AtB-AS/kundevendt/issues/108. Explanation can be found in issue.


toretefre:
In addition to the comment made in the [issue](https://github.com/AtB-AS/kundevendt/issues/108#issuecomment-1181792934), I am experiencing some unstability.

During a purchase, the app is sometimes crashing either while reserving a ticket, or during redirect after completing payment. For the latter event, the app is recovering after some seconds, and the ticket purchase completes while the app is recovering in a magical fashion. Will examine a bit further on those bugs, and see if this is newly introduced for everyone or if it's just something weird with my device.

toretefre:
Cannot recreate any of the bugs experienced yesterday, and don't think anything in this PR changes that, so think this is ready for review and testing.

monopoint (approved):
LGTM

PR #2815, opened 2022-07-13
feat: remove payment options from profile

monopoint:
resolves https://github.com/AtB-AS/kundevendt/issues/106

image

### DIFF
--- a/src/screens/Profile/PaymentOptions/index.tsx
+++ b/src/screens/Profile/PaymentOptions/index.tsx
@@ -23,6 +23,8 @@ import MessageBox from '@atb/components/message-box';
 import {StackNavigationProp} from '@react-navigation/stack';
 import {RootStackParamList} from '@atb/navigation';
 import useFontScale from '@atb/utils/use-font-scale';
+import Bugsnag from '@bugsnag/react-native';
+import {useIsFocused} from '@react-navigation/native';
 
 type PaymentOptionsProps = {
   navigation: StackNavigationProp<RootStackParamList>;
@@ -33,38 +35,49 @@ export default function PaymentOptions({navigation}: PaymentOptionsProps) {
   const {theme} = useTheme();
   const {t} = useTranslation();
   const {user} = useAuthState();
+  const isFocused = useIsFocused();
 
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [storedCards, setStoredCards] = useState<RecurringPayment[]>([]);
 
-  async function getRecurringPaymentOptions(): Promise<
-    Array<RecurringPayment>
-  > {
+  async function getRecurringPaymentOptions(): Promise<RecurringPayment[]> {
     if (!user || user.isAnonymous) return [];
-    const recurringOptions: Array<RecurringPayment> =
-      await listRecurringPayments();
-    return [...recurringOptions.reverse()];
+    try {
+      const recurringOptions = await listRecurringPayments();
+      return recurringOptions.reverse();
+    } catch (error: any) {
+      Bugsnag.notify(error);
+      return [];
+    }
   }
 
-  /*
-   * Refresh stored cards when screen gets focus
-   */
   useEffect(() => {
-    const unsubscribe = navigation.addListener('focus', refreshCards);
-    return unsubscribe;
-  }, [navigation]);
+    if (isFocused) {
+      refreshCards();
+    }
+  }, [isFocused]);
 
   async function handleRemovePayment(paymentOption: RecurringPayment) {
     setIsLoading(true);
-    await deleteRecurringPayment(paymentOption.id);
-    refreshCards();
+    try {
+      await deleteRecurringPayment(paymentOption.id);
+    } catch (error: any) {
+      Bugsnag.notify(error);
+    } finally {
+      refreshCards();
+    }
   }
 
   async function refreshCards() {
     setIsLoading(true);
-    const remoteOptions = await getRecurringPaymentOptions();
-    setStoredCards(remoteOptions);
-    setIsLoading(false);
+    try {
+      const remoteOptions = await getRecurringPaymentOptions();
+      setStoredCards(remoteOptions);
+    } catch (error: any) {
+      Bugsnag.notify(error);
+    } finally {
+      setIsLoading(false);
+    }
   }
 
   return (
@@ -184,9 +197,7 @@ const useStyle = StyleSheet.createThemeHook((theme: Theme) => ({
     marginVertical: theme.spacings.xSmall,
     borderRadius: theme.border.radius.regular,
     backgroundColor: theme.static.background.background_0.background,
-    paddingHorizontal: theme.spacings.medium,
-    paddingTop: theme.spacings.medium,
-    paddingBottom: theme.spacings.medium,
+    padding: theme.spacings.medium,
   },
   cardTop: {
     flex: 1,
@@ -194,7 +205,6 @@ const useStyle = StyleSheet.createThemeHook((theme: Theme) => ({
     alignItems: 'center',
     justifyContent: 'space-between',
   },
-  cardName: {},
   cardIcons: {
     flex: 1,
     flexDirection: 'row',
--- a/src/screens/Profile/PaymentOptions/index.tsx
+++ b/src/screens/Profile/PaymentOptions/index.tsx
@@ -84,11 +84,7 @@ export default function PaymentOptions({navigation}: PaymentOptionsProps) {
 
   return (
     <View style={style.container}>
-      <View
-        onLayout={(e) => {
-          console.log(JSON.stringify(e.nativeEvent.layout));
-        }}
-      >
+      <View>
         <FullScreenHeader
           style={{elevation: 1000, zIndex: 1000}}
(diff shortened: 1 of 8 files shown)


src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    if (!user || user.isAnonymous) return [];
+    const recurringOptions: Array<RecurringPayment> =
+      await listRecurringPayments();
+    return [...recurringOptions.reverse()];
+  }
+
+  /*
+   * Refresh stored cards when screen gets focus
+   */
+  useEffect(() => {
```
mikaelbr, line 52: I think there's a `useInFocus` hook defined for somewhere?

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <View accessibilityLiveRegion="polite">
+          <MessageBox
+            message={t(PaymentOptionsTexts.noStoredCards)}
+            withMargin={true}
+            containerStyle={{marginTop: theme.spacings.xLarge}}
+          />
+        </View>
+      )}
+
+      <ScrollView style={style.contentMargin}>
```
mikaelbr, line 87: If we used the FlatList instead we'd get pull to refresh which I guess is a more common way to refresh data on apps

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [storedCards, setStoredCards] = useState<RecurringPayment[]>([]);
+
+  async function getRecurringPaymentOptions(): Promise<
+    Array<RecurringPayment>
+  > {
+    if (!user || user.isAnonymous) return [];
+    const recurringOptions: Array<RecurringPayment> =
+      await listRecurringPayments();
+    return [...recurringOptions.reverse()];
```
mikaelbr, line 46: The `[...obj]` copying here is just extra work for copying where it has no real effect. listRecurringPaymets returns newly created data and isn't shared anywhere. Also `.reverse()` mutates the array itself before copying it. If you really wanted to reverse a copy it would have to be `[...recurringOptions].reverse()`, but still I don't think this is needed.

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  async function handleRemovePayment(paymentOption: RecurringPayment) {
+    setIsLoading(true);
+    await deleteRecurringPayment(paymentOption.id);
+    refreshCards();
+  }
+
+  async function refreshCards() {
+    setIsLoading(true);
+    const remoteOptions = await getRecurringPaymentOptions();
```
mikaelbr, line 65: What happens here if this requests fails? I usually do a `try {setLoading(true) } catch (e) {handleError(e)} finally { setLoading(false) }` pattern to ensure that you always make sure to unset loading.

src/screens/Ticketing/Purchase/utils.ts
```diff
@@ -29,3 +31,27 @@ export const getPurchaseFlow = (
...
+
+export function getExpireDate(iso: string): string {
+  let date = parseISO(iso);
+  let year = date.getFullYear();
+  let month = date.getMonth();
+  if (month === 0) {
+    month = 12;
+    year--;
+  }
+  return `${month < 10 ? '0' + month : month}/${year.toString().slice(2, 4)}`;
```
mikaelbr, line 43: Wouldn't it be easier to just use date-fns here to avoid padding and date oddities?

src/screens/Ticketing/Purchase/utils.ts
```diff
@@ -29,3 +31,27 @@ export const getPurchaseFlow = (
...
+}
+
+export function getPaymentTypeName(paymentType: PaymentType) {
+  switch (paymentType) {
+    case PaymentType.Visa:
+      return 'Visa';
+    case PaymentType.Mastercard:
+      return 'MasterCard';
+    case PaymentType.Vipps:
```
mikaelbr, line 52: We technically have also BankAxept and American Express. Not something to handle here probably but more FYI

src/tickets/api.ts
```diff
@@ -36,6 +36,12 @@ export async function listRecurringPayments(): Promise<RecurringPayment[]> {
   return response.data;
 }
 
+export async function deleteRecurringPayment(paymentId: number) {
+  const url = `ticket/v2/recurring-payments/${paymentId}`;
+  const response = await client.delete<void>(url, {authWithIdToken: true});
+  return;
```
mikaelbr, line 42:
```suggestion
  await client.delete<void>(url, {authWithIdToken: true});
```

mikaelbr (changes requested):
Looks mostly good 🙌  Only thing I think needs fixing is error handling. Now there's a case when failing to get recurring payment it will load forever and not really give an indication of failure. I also think pull to refresh using FlatList or what ever would be a more common pattern than refreshing on focus, but that's kind of subjective I guess.

src/tickets/api.ts
```diff
@@ -36,6 +36,12 @@ export async function listRecurringPayments(): Promise<RecurringPayment[]> {
   return response.data;
 }
 
+export async function deleteRecurringPayment(paymentId: number) {
+  const url = `ticket/v2/recurring-payments/${paymentId}`;
+  const response = await client.delete<void>(url, {authWithIdToken: true});
```
Nisha1306, line 41: The request keeps on loading if delete request fails. Missing error handling?

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    paddingTop: theme.spacings.medium,
+    paddingBottom: theme.spacings.medium,
+  },
+  cardTop: {
+    flex: 1,
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+  },
+  cardName: {},
```
Nisha1306, line 197: redundant?

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  contentMargin: {
+    margin: theme.spacings.medium,
+  },
+  card: {
+    marginVertical: theme.spacings.xSmall,
+    borderRadius: theme.border.radius.regular,
+    backgroundColor: theme.static.background.background_0.background,
+    paddingHorizontal: theme.spacings.medium,
+    paddingTop: theme.spacings.medium,
+    paddingBottom: theme.spacings.medium,
```
Nisha1306, line 189: can be shorthand to **padding : theme.spacings.medium**. Doesn't do a lot but less lines of code :)

src/screens/Ticketing/Purchase/utils.ts
monopoint: didn't really consider the implementation. I merely moved the function from component to utils.
That said, I agree.

mikaelbr: This is the perfect time for [boy/girl scout driven development](https://dzone.com/articles/the-boy-scout-software-development-principle)!

src/screens/Profile/PaymentOptions/index.tsx
monopoint: I don't see the scenario where the user would want to refresh manually, or have any reason to do so.
The reason we do a refresh on focus is to sync if the user navigates *back* to "removePaymentOptions" after storing a new card, because the screen would steill be mounted.

src/screens/Profile/PaymentOptions/index.tsx
monopoint: yeah good point.

src/screens/Profile/PaymentOptions/index.tsx
monopoint: I'll check it out

src/screens/Profile/PaymentOptions/index.tsx
monopoint: I agree. This was a copy paste from elsewhere in the app. I assumed it did this for a reason. 😄

src/screens/Profile/PaymentOptions/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+  useEffect(() => {
+    if (isFocused) {
+      refreshCards();
+    }
+  }, [isFocused]);
+
+  async function handleRemovePayment(paymentOption: RecurringPayment) {
+    setIsLoading(true);
+    try {
+      await deleteRecurringPayment(paymentOption.id);
```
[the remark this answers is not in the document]
monopoint: Like this?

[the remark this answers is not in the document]
mikaelbr: Yeah, sure. Is that handled now? Couldn't see any in the code unless bugsnag notify handles it?

Nisha1306 (approved):
👍

PR #2819, opened 2022-07-14
fix: hide token warnings if mobile token not enabled

toretefre:
Should solve https://github.com/AtB-AS/kundevendt/issues/2029


toretefre:
Are you agreeing that this is how it should work, @gorandalum?

gorandalum (approved):
This works, however I think it would be clearer to just have `if (!hasEnabledMobileToken) return null;` as soon as possible in the file, as all warning messages are related to mobile token.

gorandalum (approved):

PR #2824, opened 2022-07-15
feat: Navigate to the correct screen after mobile token onboarding

gorandalum:


Nisha1306 (approved):

Nisha1306:
Note: Made some changes after the last fixes we did on this task. This should be working now. Should be tested a bit more before raising a PR.

PR #2825, opened 2022-07-18
feat: redirect user to consequences of buying tickets anonymously

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/667

The Consequences of buying anonymously opens from three places -

1. Generic onboarding
2. Logout 
3. Warning in Tickets page


src/screens/AnonymousTicketPurchase/ConsequencesScreen.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <Consequence
+            value={t(AnonymousTicketPurchases.consequences.messages[0])}
+            icon={
+              <Phone fill={theme.static.background.background_0.background} />
+            }
+          />
+          <Consequence
+            value={t(AnonymousTicketPurchases.consequences.messages[1])}
+            icon={
+              <Phone fill={theme.static.background.background_0.background} />
```
Nisha1306, line 67: icon will be updated after https://github.com/AtB-AS/design-system/pull/58 is merged.

mikaelbr (changes requested):
Looks mostly good! I see there are some conflicts, but not too complex ones. I had some comments on the design for the consequences with icon sizes and padding. Also, I think for OOS partners using the feature toggle for activated tickets is important, for this entire feature probably.

src/screens/AnonymousTicketPurchase/components/Actions.tsx
```diff
@@ -0,0 +1,210 @@
...
+        afterLogin: {
+          routeName: 'TabNavigator',
+        },
+      },
+    });
+  };
+
+  const logout = async () => {
+    setIsLoading(true);
+    await logoutAndWipeTokens();
```
mikaelbr, line 49: Can `signOut` fail here? Causing `setIsLoading(false)` never to be set?

src/screens/Profile/Home/index.tsx
```diff
@@ -192,14 +187,8 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
...
-                    destructiveArrowFunction: async () => {
-                      try {
-                        // On logout we delete the user's token
-                        await wipeToken();
-                      } catch (err: any) {
-                        Bugsnag.notify(err);
-                      }
-                      return signOut();
+                    destructiveArrowFunction: () => {
+                      navigation.navigate('ConsequencesFromLogout');
```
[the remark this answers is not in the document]
toretefre: The modal is quite new and there wasn't exactly a long strategic design phase before it was implemented, so if the consequences page is more informative, it is of course possible to remove it:

https://github.com/AtB-AS/mittatb-app/pull/2754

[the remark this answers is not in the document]
Nisha1306: Good catch! Guess we can get rid of the confirmation. I can talk with Christian.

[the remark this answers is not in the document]
Nisha1306: Discussed and confirm. removing it.

src/screens/Onboarding/IntercomInfo.tsx
```diff
@@ -38,9 +42,9 @@ export default function IntercomInfo() {
       <View style={styles.bottomView}>
         <Button
           interactiveColor="interactive_1"
-          onPress={finishOnboarding}
+          onPress={() => navigation.navigate('ConsequencesFromOnboarding')}
```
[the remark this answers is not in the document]
Nisha1306: Ooohh..dint know about that. Will add it conditionally then.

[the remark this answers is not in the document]
Nisha1306: So I would just add it in onboarding and logout. Rest trigger points are already covered under the flag.

src/screens/AnonymousTicketPurchase/components/Actions.tsx
Nisha1306: mmm might be..we are wiping tokens and then signing as anonymous.  I can put it inside a try-catch and set an error if it fails.

src/screens/Profile/Home/index.tsx
```diff
@@ -169,7 +185,11 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
               <Sections.LinkItem
                 text={t(ProfileTexts.sections.account.linkItems.logout.label)}
                 icon={<ThemeIcon svg={LogOut} />}
-                onPress={() => navigation.navigate('ConsequencesFromLogout')}
+                onPress={() =>
+                  enable_ticketing
+                    ? navigation.navigate('ConsequencesFromLogout')
+                    : logoutAndWipeTokens()
```
Nisha1306, line 191: I have not spent a lot of time adding a loader here. because then we need to decide where we need to add the loader on the profile page. It's a flow that would only happen when we disable ticketing. So I thought it was not worth spending time on. Also, it would be introducing a logic that we practically dont use anywhere.

src/screens/Profile/Home/index.tsx
```diff
@@ -169,29 +169,7 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
...
-                    confirmAlertString: t(
-                      ProfileTexts.sections.account.linkItems.logout.alert
-                        .confirm,
-                    ),
-                    destructiveArrowFunction: () => {
-                      navigation.navigate('ConsequencesFromLogout');
-                    },
-                  })
-                }
+                onPress={() => navigation.navigate('ConsequencesFromLogout')}
```
mikaelbr: How will this be when ticketing is featured toggled out?

Nisha1306: I have fixed this already https://github.com/AtB-AS/mittatb-app/pull/2825/commits/09435fcc40d15454c1b86ec0f5c281925c8a54f8#diff-ad7eb34b5f49ab9a495ab3088652f0b4dfcfabcbc9981909c1c431b3afd48b65R189
Is this some old changeset? 🤔

mikaelbr: Ah, yes, I was going commit by commit. But it still removes confirmation but looking more into it the existing confirmation is more about ticketing also, so just not having it when disabling ticketing is probably fine 👍

src/login/LoginOnboarding.tsx
```diff
@@ -78,7 +78,9 @@ export default function LoginOnboarding({
         </View>
         <TouchableOpacity
           style={styles.laterButton}
-          onPress={navigation.goBack}
+          onPress={() => {
+            navigation.navigate('ConsequencesFromLoginOnboarding');
```
mikaelbr, line 82: Should this do something special when enable_ticketing is false? 🤔

Nisha1306: It's the log-in onboarding for periodic tickets. So it's actually not needed.  
`Log in to purchase 7, 30, 60, 90 or 180-day tickets`

gorandalum (approved):
Wow, a lot of code and complexity. Not a critique of the implementation though, as it seems pretty clean and readable. However the complexity keeps on piling up in the app, and when evaluating the cost vs benefit of a task the complexity should also be taken into consideration. For example was it necessary to include this warning when login out, when they get it when buying tickets anonymously? 

Also after some time we should merge the general onboarding with the mobiletoken onboarding, to reduce complexity in the app.

tldr:
Good work!

Nisha1306:
> Wow, a lot of code and complexity. Not a critique of the implementation though, as it seems pretty clean and readable. However the complexity keeps on piling up in the app, and when evaluating the cost vs benefit of a task the complexity should also be taken into consideration. For example was it necessary to include this warning when login out, when they get it when buying tickets anonymously?
> 
> Also after some time we should merge the general onboarding with the mobiletoken onboarding, to reduce complexity in the app.
> 
> tldr: Good work!

Honestly, I also think that the warning of the anonymous consequences pops up a bit too much now. Especially when you test it, you feel that it's there on every other click. Might be we can take this back to @Sebstorvik, while it is only in staging, and let him have a look at how it looks in the app and see if we can get rid of it at a couple of places?

gorandalum:
Discussed with Nisha:
She will show it to Sebastian locally, so we have a faster feedback loop than doing it in staging.

Nisha1306:
> Discussed with Nisha: She will show it to Sebastian locally, so we have a faster feedback loop than doing it in staging.

So we decided on removing it from two out of four places i.e. Logout & log in onboarding.

gorandalum (approved):

PR #2826, opened 2022-07-19
feat: use different accessibility labels for different stop icons

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/111

Please find some examples of how the accessibility labels would sound like :

- Tram stop Skansen, Trondheim. Activate to use this location.
- Bus stop Skansen, Trondheim. Activate to use this location.
- Airport Trondheim lufthavan, Størdal. Activate to use this location.
- Location Skansen, Trondheim. Activate to use this location. (This would be used when the resulted location is not a specific stop but just a location)


rosvik (approved):
LGTM 👍

PR #2827, opened 2022-07-22
feat: modernise active/expired ticket view

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/330

**Note:** 
Looks like a big PR, but a lot of changes are lifting and shifting to smaller components instead. 😅 

**Screenshots:** 

- Active Tickets and details: 

<div>
image
image
image
</div>

- Under Processing Tickets: 

<div>
image
</div>

- Future Tickets: 

<div>
image
</div>

- Expired Tickets: 

<div>
image
</div>


Nisha1306:
I was not aware that we show carnet tickets in the App. So I need to make some modifications for that as well. SO marking this as a draft for time being.

src/translations/screens/Ticket.ts
```diff
@@ -139,6 +137,10 @@ const TicketTexts = {
         `Bruk ${deviceName} når du reiser`,
         `Bring ${deviceName} when you travel`,
       ),
+    carnetWarning: _(
+      'Vennligst bytt til t:card for å kunne bruke dette hullkortet.',
```
gorandalum, line 141: `t:card` -> `t:kort`
`hullkortet` -> `klippekortet`

gorandalum (approved):
Looks good! And good structure on the components. If you feel up for it, I think you also now can just remove everything regarding `summerPass`. Summer pass is not close to implementation, and I think having it "half-way" in the code gives more clutter than benefit.

src/translations/screens/Tickets.ts
```diff
@@ -119,6 +119,14 @@ const TicketsTexts = {
       ),
       transportModes: _('Flere reisemåter', 'Several travel modes'),
     },
+    carnet: {
```
gorandalum, line 122: Is this used? We are not selling carnets in the app.

mikaelbr: But we are showing them if sold on the webshop?

Nisha1306: We are not selling but we are showing and hence we are using transport modes. The rest of them are not being used actively right now, But since I have modified the ticket component and we are resolving the title of tickets dynamically using `ticket type` and carnet is a type of ticket, typescript gives an error. So had to add values for carnet as well. 
`t(TicketsTexts.availableTickets[ticketType].title);`
Thought it should be no harm as carnet should be added in app eventually. But the other option is to exclude carnet from ticket component and remove here as well.

PR #2828, opened 2022-07-27
feat: add and remove favourites in departures v2

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/149

**Note:** 
Used most of the code from Favourites implemented in Departures v1 except the Toggle favourite Button. Old one is based on old types taken from SDK/used in old departures. 
Dint modify the old one because it is anyway going to be discarded soon. Also dint want to disturb the old flow and hence kept it as-is for testing behaviour of new one as compared to old one. 

**Screenshots:** 
<div>
image
image
image
image
</div>


rosvik (approved):
Great work! LGTM 👍

PR #2829, opened 2022-08-01
feat: add separate tab for favourite tabs

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/151

Every stop that has one or more lines marked as Favourite, should appear in the favorites tab. 

<div>

</div>

### DIFF
--- a/src/screens/Departures/AllStopsOverview.tsx
+++ b/src/screens/Departures/AllStopsOverview.tsx
@@ -1,102 +1,62 @@
-import {Location as LocationIcon} from '@atb/assets/svg/mono-icons/places';
-import SimpleDisappearingHeader from '@atb/components/disappearing-header/simple';
-import ScreenReaderAnnouncement from '@atb/components/screen-reader-announcement';
-import {LocationInput, Section} from '@atb/components/sections';
-import ThemeIcon from '@atb/components/theme-icon';
-import FavoriteChips from '@atb/favorite-chips';
-import {GeoLocation, Location} from '@atb/favorites/types';
-import {
-  RequestPermissionFn,
-  useGeolocationState,
-} from '@atb/GeolocationContext';
-import {useOnlySingleLocation} from '@atb/location-search';
-import {RootStackParamList} from '@atb/navigation';
-import {StyleSheet} from '@atb/theme';
+import {useGeolocationState} from '@atb/GeolocationContext';
+import React, {useEffect, useMemo, useState} from 'react';
 import {NearbyTexts, useTranslation} from '@atb/translations';
+import {useOnlySingleLocation} from '@atb/location-search';
 import {
   CompositeNavigationProp,
   RouteProp,
   useIsFocused,
 } from '@react-navigation/native';
-import {StackNavigationProp} from '@react-navigation/stack';
-import React, {useEffect, useMemo, useState} from 'react';
-import {View} from 'react-native';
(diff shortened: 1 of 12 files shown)


src/screens/Departures/AllStopsOverview.tsx
```diff
@@ -1,102 +1,62 @@
...
-};
-
-const PlacesOverview: React.FC<PlacesOverviewProps> = ({
-  requestGeoPermission,
-  currentLocation,
-  hasLocationPermission,
-  navigation,
-}) => {
+  const requestGeoPermission = requestPermission;
+  const currentLocation = location || undefined;
```
mikaelbr, line 53: Are these needed or could we just either rename in destructuring or just use the names from `useGeolocationState`?

src/screens/Departures/AllStopsOverview.tsx
```diff
@@ -216,7 +175,6 @@ const PlacesOverview: React.FC<PlacesOverviewProps> = ({
     <SimpleDisappearingHeader
       onRefresh={refresh}
       isRefreshing={isLoading}
-      leftButton={leftButton}
```
mikaelbr, line 219: Should it no longer show icon on service disruption? Is this intended or just that the icon is not updated in the design?

mikaelbr: Ah. I see it is used in the tab navigator instead. Should this then actually be `SimpleDisappearingHeader`? I get it is easy to use to refresh, but when removing header title and left button here the idea of DisappearingHeader goes away and it is now more used like a FlatList with pull to refresh? I think there might be better components for this instead of in a way deluding the concept of `SimpleDisappearingHeader`

src/screens/Departures/FavouriteStopsOverview.tsx
```diff
@@ -0,0 +1,88 @@
...
+      latitude &&
+      longitude &&
+      coordinatesDistanceInMetres(currentLocation.coordinates, {
+        latitude: latitude,
+        longitude: longitude,
+      })
+    );
+  };
+
+  const favouriteStopsDetails = state.data?.stopPlaces
```
mikaelbr, line 48: This might be something that could be wrapped either in a custom hook or just a useMemo directly. If there are a lot of calculations and a lot of renders.

mikaelbr (changes requested):
Looks good 🚀  Only thing I was thinking is that I feel we here are using `SimpleDisappearingHeader` of convenience not strictly by correctness. It has almost the functionality we need, but not strictly the correct semantics. So it's increasing the complexity of `SimpleDisappearingHeader`, it is changing its intention.

src/screens/Departures/AllStopsOverview.tsx
mikaelbr: Testing it more now and seeing real use. Scratch this comment. It makes sense with favorites and input box

mikaelbr (approved):
> Testing it more now and seeing real use. Scratch comment on SimpleDisappearingHeader. It makes sense with favorites and an input box

src/screens/Departures/AllStopsOverview.tsx
Nisha1306: Yes, I started with not using SimpleDisappearingHeader first but realized that it was doing a lot behind the scenes. So kept it as-is.

PR #2832, opened 2022-08-02
feat: navigate to correct departure versioned screen on quay select in trip details

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/2064

### DIFF
--- a/src/screens/TripDetails/DepartureDetails/index.tsx
+++ b/src/screens/TripDetails/DepartureDetails/index.tsx
@@ -42,6 +42,7 @@ import {getServiceJourneyMapLegs} from '@atb/api/serviceJourney';
 import {ServiceJourneyMapInfoData_v3} from '@atb/api/types/serviceJourney';
 import {useFirestoreConfiguration} from '@atb/configuration/FirestoreConfigurationContext';
 import CancelledDepartureMessage from '@atb/screens/TripDetails/components/CancelledDepartureMessage';
+import {usePreferenceItems} from '@atb/preferences';
 
 export type DepartureDetailsRouteParams = {
   items: ServiceJourneyDeparture[];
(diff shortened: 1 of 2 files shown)


rosvik:
Looks good, but could you verify that this doesn't introduce a bug I had a while back?

1. Enable departures v1
2. Navigate to the details screen (tap a departure then a bus stop)
3. Switch to the profile tab and enable departures v2
4. Go back to departures, and verify that you end up at the old departure screen, and that the v2 details view is not visible.

And the other way around

1. Enable departures **v2**
2. Navigate to the details screen (tap a stop place, a departure then a bus stop)
3. Switch to the profile tab and enable departures **v1**
4. Go back to departures, and verify that you end up at the old departure screen, and that the **v1** details view is not visible.

Nisha1306:
> Looks good, but could you verify that this doesn't introduce a bug I had a while back?
> 
> 1. Enable departures v1
> 2. Navigate to the details screen (tap a departure then a bus stop)
> 3. Switch to the profile tab and enable departures v2
> 4. Go back to departures, and verify that you end up at the old departure screen, and that the v2 details view is not visible.
> 
> And the other way around
> 
> 1. Enable departures **v2**
> 2. Navigate to the details screen (tap a stop place, a departure then a bus stop)
> 3. Switch to the profile tab and enable departures **v1**
> 4. Go back to departures, and verify that you end up at the old departure screen, and that the **v1** details view is not visible.

mmmm checked..the behaviour is a little different in both the scenarios though. 
Scenario 1:  On going back to departures it renders the trip details screen(which is still relevant) instead of old v1 departure details screen. However on click of any stop on the trips it now navigates to the correct departures screen.
Scenario 2: On going back to departures it renders the departures v1 screen from scratch. 

The behaviour is same in master as well, hence can say this is not a bug introduced by this change but still looks inconsistent. Should be fixed but maybe as a separate issue?

rosvik:
Alright, sounds fine to me 👍

rosvik (approved):

PR #2836, opened 2022-08-04
feat: set markdown as true to allow links in alert boxes

Nisha1306:
Solves a part of https://github.com/AtB-AS/kundevendt/issues/1997 i.e. Allow links on alert boxes. 

In order to allow link in alert boxes we have to allow markdown in alerts, so that they get rendered properly. Also we need to set alerts in Firestore accordingly i.e. in Markdown syntax.

` For details see [atb.no/driftsavvik](https://www.atb.no/driftsavvik)` and not ` For details see atb.no/driftsavvik.`

https://user-images.githubusercontent.com/29625161/182830754-0211a2cd-0601-455c-8e61-1cf8a0727679.mov

### DIFF
--- a/src/alerts/AlertBox.tsx
+++ b/src/alerts/AlertBox.tsx
@@ -28,6 +28,7 @@ const AlertBox = ({alertContext, style}: Props) => {
       title={getReferenceDataText(alert.title ?? [], language)}
       message={getReferenceDataText(alert.body, language)}
       type={alert.type}
+      isMarkdown={true}
     />
   );
 };
--- a/src/alerts/AlertBox.tsx
+++ b/src/alerts/AlertBox.tsx
@@ -6,6 +6,12 @@ import {getReferenceDataText} from '@atb/reference-data/utils';
 import {StyleProp, ViewStyle} from 'react-native';
 import {Theme, useTheme} from '@atb/theme';
 
+type AlertWithMarkdown = {
+  lang: string;
+  value: string;
+  valueWithMarkdown?: string;
+};
+
 type Props = {
   alertContext?: AlertContext;
   style?: StyleProp<ViewStyle>;
@@ -26,12 +32,20 @@ const AlertBox = ({alertContext, style}: Props) => {
     <MessageBox
       containerStyle={style}
       title={getReferenceDataText(alert.title ?? [], language)}
-      message={getReferenceDataText(alert.body, language)}
+      message={getReferenceDataText(
+        markdownPreferredText(alert.body),
+        language,
+      )}
       type={alert.type}
       isMarkdown={true}
     />
   );
 };
+
+const markdownPreferredText = (alertsWithMarkdown: AlertWithMarkdown[]) =>
+  alertsWithMarkdown.map((item) => {
+    return {lang: item.lang, value: item.valueWithMarkdown || item.value};
+  });
 
 const isValidAlert = (theme: Theme, alert?: Alert) => {
   if (!alert) return false;
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -3,7 +3,6 @@ import {Polygon} from 'geojson';
 export type LanguageAndText = {
   lang: string;
   value: string;
-  valueWithMarkdown?: string;
 };
 
 export type PreassignedFareProductType =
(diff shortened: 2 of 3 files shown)


rosvik:
Hmm, this won't work when the same alerts are shown for people with old versions of the app. (And soon also the webshop? @toretefre) Maybe we should add a markdown-specific field in firestore to provide two versions of the text?

The data in firestore could possibly look something like

```
active: false
context: "travel"
type: "warning" 
body:
	lang: "nob"
		value: "For detaljer se atb.no/driftsavvik"
	lang: "eng"
		value: "For details see atb.no/driftsavvik"
bodyMarkdown:
	lang: "nob"
		value: "For detaljer se [atb.no/driftsavvik](https://atb.no/driftsavvik)"
	lang: "eng"
		value: "For details see [atb.no/driftsavvik](https://atb.no/driftsavvik)"
title:
	lang: "nob"
		value: "Trondheim S er stengt for linje 24"
	lang: "eng"
		value "Trondheim S is closed for line 24"
```

Nisha1306:
> 

Oh yes. It's of course not backward compatible. So we can add one more field like `valueWithMarkdown` in the firestore data like you suggested. I think I would give with the second option you mentioned above a try and see how it looks.

toretefre:
I can easily refactor the webshop alerts into supporting a very similar (identical?) format. 

Originally I went for supporting only one link on a separate line (and field), but I can refactor into using markdown so that we can customise a bit more, as long as the users writing the alerts are comfortable with the formatting.

image

image

Unrelated question: is it correct that `travel` and `ticketing` are the only contexts currently in use? Should there be a `global` or `everywhere` option, or is that the default if context is not provided?

Nisha1306:
> I can easily refactor the webshop alerts into supporting a very similar (identical?) format.
> 
> Originally I went for supporting only one link on a separate line (and field), but I can refactor into using markdown so that we can customise a bit more, as long as the users writing the alerts are comfortable with the formatting.
> 
> image
> 
> image
> 
> Unrelated question: is it correct that `travel` and `ticketing` are the only contexts currently in use? Should there be a `global` or `everywhere` option, or is that the default if context is not provided?

There is also general. `export type AlertContext = 'general' | 'ticketing' | 'travel';  It applies to everywhere used if the context is General I think. 
`

rosvik:
@toretefre Ref. https://github.com/AtB-AS/mittatb-app/pull/1006

>  If the context is travel the alert is shown on the Assistant and Nearby screen, if ticketing the alert is shown on Ticket screen (including Overview and Confirmation screen), and if general the alert will be shown on all mentioned screens.

So it would make sense to show "ticketing" and "general" in the webshop? Not sure if "travel" makes sense before there is a travel planner for web.

toretefre:
> @toretefre Ref. #1006
> 
> > If the context is travel the alert is shown on the Assistant and Nearby screen, if ticketing the alert is shown on Ticket screen (including Overview and Confirmation screen), and if general the alert will be shown on all mentioned screens.
> 
> So it would make sense to show "ticketing" and "general" in the webshop? Not sure if "travel" makes sense before there is a travel planner for web.

I think it depends on whether or not we would like to have separate alert collections for the webshop. Currently I have implemented it using separate collections, as there might be webshop issues that does not apply in the app and vice versa, then there will not be a need to handle `travel` alerts yet.

rosvik:
@toretefre It would probably be a good idea to be able to choose platform (webshop / app / both), but from the messages I can imagine, they usually apply to both. E.g. "Free tickets this weekend", "Changes to how youth tickets work on monday", "New time tables for summer june 21.", "Vipps is having issues with payment". 

Maybe we should handle it with a field as well, so that you don't have to duplicate alerts when they are added to firestore? Something like 

```
active: false
context: "travel"
type: "warning" 
platform: "app" //"app" | "webshop" | "general"
body:
	lang: "nob"
		value: "For detaljer se atb.no/driftsavvik"
		markdown: "For detaljer se [atb.no/driftsavvik](https://atb.no/driftsavvik)"
	lang: "eng"
		value: "For details see atb.no/driftsavvik"
		markdown: "For details see [atb.no/driftsavvik](https://atb.no/driftsavvik)"
title:
	lang: "nob"
		value: "Trondheim S er stengt for linje 24"
	lang: "eng"
		value "Trondheim S is closed for line 24"
```

for example.

toretefre:
> Maybe we should handle it with a field as well, so that you don't have to duplicate alerts when they are added to firestore?

Yup, if we're already doing not-so-backwards-compatible-changes we can absolutely do that, good idea!

rosvik (approved):
Looks good 👍

PR #2837, opened 2022-08-05
fix: add info box instead of text only

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2079

Before: 
<div>
  
  
</div>

After: 
<div>
  
  
</div>

### DIFF
--- a/src/screens/Ticketing/Tickets/TicketsScrollView.tsx
+++ b/src/screens/Ticketing/Tickets/TicketsScrollView.tsx
@@ -86,7 +86,11 @@ const TicketsScrollView: React.FC<Props> = ({
           />
         )}
         {!fareContracts?.length && !reservations?.length && (
-          <ThemeText style={styles.noTicketsText}>{noTicketsLabel}</ThemeText>
+          <MessageBox
+            containerStyle={styles.messageBox}
+            type="info"
+            message={noTicketsLabel}
+          />
         )}
         {reservations?.map((res) => (
           <TicketReservation key={res.orderId} reservation={res} />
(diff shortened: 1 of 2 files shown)


src/screens/Ticketing/Tickets/TicketsScrollView.tsx
```diff
@@ -86,7 +86,11 @@ const TicketsScrollView: React.FC<Props> = ({
           />
         )}
         {!fareContracts?.length && !reservations?.length && (
-          <ThemeText style={styles.noTicketsText}>{noTicketsLabel}</ThemeText>
+          <MessageBox
+            containerStyle={styles.messageBox}
```
mikaelbr, line 90: I'm guessing this can be a normal use case here. We have horizontal padding but want to have padding under. The component isn't really working as intended when we have to add style to it, but also may be a problem with React Natives layout model not having the ability to do proper flex gaps.

Not necessary for this task, but I think we should reevaluate if we need to change the API of MessageBox to support margin bottom as a common case. Maybe with something like `margin: 'horizontal' | 'bottom' | 'bottom-horizontal';` etc.

mikaelbr (approved):
Looks good 🍰

src/screens/Ticketing/Tickets/TicketsScrollView.tsx
Nisha1306: Yes, I think so too. But maybe not for this task at we would have to verify all the places it is being used. Don't want to increase the scope of this, so not touching this one.

PR #2843, opened 2022-08-12
feature toggle + blank slates

monopoint:
toggle in Profile to activate new front page
Navigation and blank pages for Dashboard and Travel Search

### DIFF
--- a/src/screens/Profile/Home/index.tsx
+++ b/src/screens/Profile/Home/index.tsx
@@ -80,7 +80,7 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
 
   const {
     setPreference,
-    preferences: {newDepartures},
+    preferences: {newDepartures, newFrontPage},
   } = usePreferences();
 
   function copyInstallId() {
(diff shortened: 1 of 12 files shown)


jorelosorio (commented):
In English, the TabBar name is `Assistant` but the title is `Travel Search`, I saw that in NO-locale it is fine in both places `Reisesøk`, it is OK? or Should be different?

Simulator Screen Shot - iPhone 13 Pro - 2022-08-12 at 13 40 56

monopoint:
> In English, the TabBar name is `Assistant` but the title is `Travel Search`, I saw that in NO-locale it is fine in both places `Reisesøk`, it is OK? or Should be different?
Yeah, I'm not really sure what it should be called. In the figma sketches it's heading title is "travel search", but in reallity it's more of a dashboard. I guess this needs some discussion, but for right now it's really not that important

jorelosorio (approved):

src/screens/Profile/Home/index.tsx
```diff
@@ -269,6 +269,22 @@ export default function ProfileHome({navigation}: ProfileScreenProps) {
...
+              analytics().logEvent('toggle_beta_frontPage', {
+                toggle: newFrontPage ? 'enable' : 'disable',
+              });
+              updateMetadata({
+                'AtB-Beta-Frontpage': newFrontPage ? 'enabled' : 'disabled',
+              });
+              setPreference({newFrontPage});
+            }}
+          />
+
```
rosvik, line 287: We should put this under "Development menu" instead of "New features" so it's not visible to end users when we do a release. We can move it to here once its ready for beta testing.

PR #2846, opened 2022-08-12
fix: update behaviour on no favorite departures

rosvik:
https://github.com/AtB-AS/kundevendt/issues/150

- Adds a message to departures that suggest toggling favorites off when there are no results.
- Minimizes quays instead of hiding them when favorites are toggled.

https://user-images.githubusercontent.com/1774972/184355562-0fd0508c-7ea4-4510-a321-75d148c85da4.mp4


mikaelbr (approved):
Looks like an improvement I think! 👍

PR #2845, opened 2022-08-12
fix: handling of app scheme for nfk and atb

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/1920

Came across this issue when testing for https://github.com/AtB-AS/kundevendt/issues/1776 and https://github.com/AtB-AS/kundevendt/issues/2114.

Sets up:

### AtB
```
# Dev
atb-dev://

# Staging
atb-staging://

# Production
atb://
```

### NFK

```
# Dev
nfk-dev://

# Staging
nfk-staging://

# Production
nfk://
```

Based on `APP_SCHEME` in env-folders. This allow us to override based on env and org and easily add support for other orgs.

[See proposed docs](https://github.com/AtB-AS/mittatb-app/blob/b5c73ddddbce5e2c879f33d641ef84444d29b443/docs/OrgConfig.md)

---

Tested on:

- [x] iOS
- [x] Android

This can be a big issue if anyone has both Reis and AtB app running.

---

## How to test locally:

Make sure you have the latest changes and run setup: E.g. `yarn setup dev atb`

```
npx uri-scheme open 'atb-dev://ticketing?transaction_id=321321&payment_id=321321' --android
```

```
npx uri-scheme open 'atb-dev://ticketing?transaction_id=321321&payment_id=321321' --ios
```

## How to test in staging

Completing a ticket purchase should redirect back to active tickets. Using a proxy it should say something like `atb-staging://...`

PR #2848, opened 2022-08-16
feat: add login via vipps

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/1991 

Note: The _remembered flow_ remembers the user even when a new user has logged in to vipps. This is yet to be taken care of.


src/auth/AuthContext.tsx
```diff
@@ -101,6 +101,10 @@ const getAuthenticationType = (
 
 export type ConfirmationErrorCode = 'invalid_code' | 'unknown_error';
 export type PhoneSignInErrorCode = 'invalid_phone' | 'unknown_error';
+export type VippsSignInErrorCode =
+  | 'access_denied'
+  | 'outdated_app_version'
+  | 'unknown_error';
```
Nisha1306, line 107: Don't think we have many of the error states that make sense to the user. So it's unknown_error most of the time.

Nisha1306:
Also, the vipps login should be under a feature I think. Any suggestions?

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,215 @@
...
+    setIsLoading(true);
+    try {
+      await authorizeUser(setIsLoading);
+    } catch (err) {
+      setError('unknown_error');
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(
```
monopoint, line 63: this should probably provide a cleanup function

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,215 @@
...
+        setError('unknown_error');
+        setIsLoading(false);
+      }
+    };
+    if (authorizationCode) {
+      signInVippsUser();
+    }
+  }, [authorizationCode]);
+
+  useEffect(() => {
```
monopoint, line 100: this should probably provide a clean up function

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    backgroundColor: theme.static.background[themeColor].background,
+    flex: 1,
+  },
+  scrollView: {
+    marginHorizontal: theme.spacings.large,
+    flexDirection: 'column',
+    paddingVertical: 24,
```
monopoint, line 195: should probably use theme spacing

src/translations/screens/subscreens/Login.ts
```diff
@@ -1,6 +1,22 @@
...
+    selectLoginMethod: _('Velg innloggingsmetode', 'Select login method'),
+    or: _('Eller logg inn med', 'or Log in with'),
+    options: {
+      phoneAndCode: _(
+        'Mobilnummer og engangskode',
+        'Mobile number and one-time code',
+      ),
+      vipps: {
+        labelPart1: _('Logg inn med ', 'Log in with '),
+        labelPart2: _('ipps ', 'ipps'),
```
monopoint, line 15: probably the "i" shouldn't be here

monopoint (dismissed):
some comments

src/login/LoginOptionsScreen.tsx
Nisha1306: true that..should remove the subscription.

src/login/LoginOptionsScreen.tsx
Nisha1306: I think it already returns the function to unsubscribe from the listener.

src/login/ConfirmCode.tsx
```diff
@@ -178,6 +179,8 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
   },
   scrollView: {
     padding: theme.spacings.large,
+    flexGrow: 0.5,
```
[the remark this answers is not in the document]
Nisha1306: It's a leftover of me experimenting with something. 😅 I have fixed it.

package.json
```diff
@@ -78,11 +78,13 @@
     "react": "17.0.2",
     "react-native": "0.66.3",
     "react-native-bootsplash": "^3.2.4",
+    "react-native-custom-tabs": "^0.1.7",
```
Nisha1306, line 81: Need to get rid of this as we are using `react-native-inappbrowser-reborn` instead.

package.json
```diff
@@ -78,11 +78,13 @@
...
     "react-native-dash": "^0.0.11",
     "react-native-device-info": "^8.3.1",
     "react-native-geolocation-service": "^5.2.0",
     "react-native-gesture-handler": "^1.10.3",
     "react-native-get-random-values": "^1.7.0",
+    "react-native-inappbrowser-reborn": "^3.7.0",
```
mikaelbr, line 87: Just out of curiosity, why is this needed? don't we have existing webview implementations that does the same or is this different?

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,227 @@
...
+  useEffect(
+    () =>
+      navigation.addListener('focus', () => {
+        setIsLoading(false);
+      }),
+    [navigation],
+  );
+
+  const signInUsingCustomToken = async (token: string) => {
+    const response = await signInWithCustomToken(token);
```
mikaelbr, line 72: Hm. Due to the try/catch inside `signInWithCustomToken` and not returning the underlying promise, won't the error in line 76 always return`unknown_error` here? Meaning we'll never show specific error cases.

package.json
Nisha1306: We do have Webview implementations in the App, however, Vipps documentation asks the apps to use Custom tabs and not Webviews. https://github.com/vippsas/vipps-login-api/blob/master/vipps-login-api.md#using-vipps-login-in-native-applications 
So this basically opens Chrome Custom Tabs for Android & Safari for iOS.

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,227 @@
...
+    if (!response.error) {
+      navigation.navigate('Profile');
```
gorandalum, line 74: Should use the `afterLogin` param here to decide where to go.

src/translations/screens/subscreens/Login.ts
```diff
@@ -30,6 +45,18 @@ const LoginTexts = {
       '...to store payment cards for later use. Soon, login will store personal preferences to your profile so that you can carry them with you when swapping phones.',
     ),
   },
+  vipps: {
+    errors: {
+      outdated_app_version: _(
+        'Er du sikker på at telefonnummeret er korrekt?',
+        'Are you certain the phone number is correct?',
+      ),
```
gorandalum, line 53: The `outdated_app_version` error gives "Are you certain the phone number is correct"?

gorandalum (changes requested):
Looks mostly good. As discussed the navigation should be fixed when merging the consequences which includes updated onboarding flow.

src/login/LoginOptionsScreen.tsx
Nisha1306: Well yes, so this is the case where the user has given consent and we are in the after redirect to app flow. I am not sure if we will have any errors here that would make sense to the user. So we show a generic error to the user and ask them to try again.

src/translations/screens/subscreens/Login.ts
Nisha1306: Oos..copy paste mistake 😅  Would fix this.

src/login/LoginOptionsScreen.tsx
Nisha1306: About the actual error, I think the response error handler in the axios client would notify bugsnag about it. Should we do more?

mikaelbr: I think there are cases where Vipps returns errors that the users should know about or that we should handle in some case: https://github.com/vippsas/vipps-login-api/blob/master/vipps-login-api.md#custom-error-codes

mikaelbr: Ah yeah. I read the flow wrong and thought of this as Vipps errors, but these are Firebase errors. There might be some errors there also that the user should know about, but also I was thinking more here https://github.com/AtB-AS/mittatb-app/pull/2848/files#diff-55e5a9f1f6d5f5e6abc246e3841ca0c8eb69c79e42d7b6e9746238d8f45d297fR91

Where it also seems maybe like it's just general "some error"?

Nisha1306: you are right, I did go through that list and thought of handling these errors but I don't remember why I ended up translating all of them to unknown_error.  🥺
```
export type VippsSignInErrorCode =
  | 'access_denied'
  | 'outdated_app_version'
  | 'unknown_error';
```
However, I only thought of handling these three errors. Also probably should handle `login_required`. But shouldn't rest of [them](https://github.com/vippsas/vipps-login-api/blob/master/vipps-login-api.md#custom-error-codes) translate to unknown_error to the end user? What do you think? Anyway, I think I should give this some more thought first and then get back 😅

mikaelbr:
@Nisha1306 If this is under development again i propose moving it so to clearer show what is in actual review and not

Nisha1306:
> @Nisha1306 If this is under development again i propose moving it so to clearer show what is in actual review and not

mmm not exactly under development but I need to look into suggestions yes. But can be moved to Development for better picture 👍

src/login/LoginOptionsScreen.tsx
Nisha1306: Yeah..I would atleast handle those then 👍

Nisha1306: I went through the code again and realised that these errors are already handled. They should be part of the query param in the callback URL if any of them occur. See https://github.com/AtB-AS/mittatb-app/pull/2848/files#diff-55e5a9f1f6d5f5e6abc246e3841ca0c8eb69c79e42d7b6e9746238d8f45d297fR114

As far as [this](https://github.com/AtB-AS/mittatb-app/pull/2848/files#diff-55e5a9f1f6d5f5e6abc246e3841ca0c8eb69c79e42d7b6e9746238d8f45d297fR91) is concerned, it should happen only when Bff is not able to redirect the user to the app due to some reason. And I think it should be nothing but an unknown error for the user. 

I think we should probably go with this error handling and see what scenarios we get during testing. cc @tormoseng  @HanneLH

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,227 @@
...
+      }
+    };
+    if (authorizationCode) {
+      signInVippsUser();
+    }
+  }, [authorizationCode]);
+
+  useEffect(() => {
+    const vippsCallbackHandler = async (event: any) => {
+      if (event.url.includes('atb://auth/vipps')) {
```
mikaelbr, line 102: This should be updated according to this: https://github.com/AtB-AS/mittatb-app/pull/2845
```suggestion
      if (event.url.includes(`${APP_SCHEMA}://auth/vipps`)) {
```

mikaelbr: Also `VIPPS_CALLBACK_URL` in the BFF

src/login/LoginOptionsScreen.tsx
mikaelbr: I see! Good catch.

src/login/LoginOptionsScreen.tsx
Nisha1306: @torfjor  Could you please set the env variable accordingly? Might be changes needed at vipps end as well?

Nisha1306: think` atb-staging://` and `atb:// `should be enough

src/components/vipps-login-button/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const VippsLoginButton = ({
+  onPress,
+  disabled,
+}: {
+  onPress: () => {};
+  disabled: boolean;
+}) => {
+  const {t} = useTranslation();
+  const style = useStyle();
+  const getVippsLabel = () => (
```
mikaelbr, line 18: Why is this defined as a function instead of just used directly?

src/components/vipps-login-button/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    >
+      {getVippsLabel()}
+    </TouchableOpacity>
+  );
+};
+
+const useStyle = StyleSheet.createThemeHook((theme: Theme) => ({
+  container: {
+    flexDirection: 'row',
+    backgroundColor: '#FF5B24',
```
mikaelbr, line 40: Should this be included fromt he theme or as a named color or something? Ah, it's probably the "vipps" color? Can be nice to have it as a constant variable at the top or something

src/login/LoginOptionsScreen.tsx
```diff
@@ -0,0 +1,230 @@
...
+        setIsLoading(true);
+        const code = parseUrl(event.url).query.code;
+        if (code) {
+          const state = parseUrl(event.url).query.state;
+          const initialState = await storage.get('vipps_state');
+          if (initialState?.toString() !== state?.toString()) {
+            setError('unknown_error');
+            setIsLoading(false);
+          }
+          setAuthorizationCode(code.toString());
```
mikaelbr, line 115: This should be called and changing the code even if `initialState?.toString() !== state?.toString()`? No consequences for when calling `getOrCreateVippsUserCustomToken` above?

mikaelbr (approved):
Haven't tested it out locally (dev is in a state of non-functioning due to upgrades etc). But seems to make sense by code and the things that has been discussed has been fixed. Gj 💪

src/components/vipps-login-button/index.tsx
Nisha1306: yes it could be..it was needed earlier but not anymore..can fix it 👍

src/login/LoginOptionsScreen.tsx
Nisha1306: You are so right! This should not be called. I started with throwing the error when the state dint match so it should have stopped the flow. But now when we are just setting the state, this should be done in else block

gorandalum:
After a talk with Nisha we landed upon providing the redirect url from the app to the bff when creating the Vipps login session. This makes us able to have correct handling of callback also in the dev app when using the staging bff.

gorandalum (approved):

PR #2849, opened 2022-08-16
Compact ticket display on front page

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/1880

Screenshot_1660639548
Screenshot_1660639579
Screenshot_1660639631


src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -0,0 +1,141 @@
...
+type Props = {
+  onPressDetails?: (orderId: string) => void;
+};
+
+const CompactTickets: React.FC<Props> = ({onPressDetails}) => {
+  const itemStyle = useStyles();
+
+  // TODO: Verify if this is the correct way to update time-to-time the expiration time of the tickets.
+  const [now, setNow] = useState<number>(Date.now());
+  useInterval(() => setNow(Date.now()), 1000);
```
jorelosorio, line 34: I noticed that the ticket is not updating to present an updated text for validity, I made this solution based on current solutions in the code. 

NOTE: The details view is not being updated when the time changes. (Maybe another task)

src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const isA = isValidRightNowFareContract(a);
+    const isB = isValidRightNowFareContract(b);
+
+    if (isA === isB) return 0;
+    if (isA) return -1;
+    return 1;
+  });
+
+  // TODO: Double check if this is how it should be implemented, using the customer profile!
+  const hasActiveTravelCard = !!customerProfile?.travelcard;
```
jorelosorio, line 49: This is a required parameter in order to reuse the main component here, however, I saw this in other places, but I want to make sure it is correct here in this context.

rosvik: Looking at the logic inside `isInspectableTicket()`, this is only used when mobile token is not enabled, so this looks correct.

monopoint (approved):

src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <ThemeText
+        type="body__secondary"
+        color="secondary"
+        style={itemStyle.sectionText}
+        accessibilityLabel={t(TicketsTexts.header.title)}
+      >
+        {t(TicketsTexts.header.title)}
+      </ThemeText>
+      {activeFareContracts?.map((fareContract, index) => {
+        const travelRights =
```
mikaelbr, line 71: I think it's totally fine to extract this as its own component in the same file. It's easier to separate responsibilities and change over time. Here we essentially have 2 setup/initializer phases. One in the "List" component and one in the "Item" component. Further down we use a lot of variables, it's harder to control "which is closures" and which is dependencies to the sub component.

src/screens/Dashboard/CompactTickets.tsx
mikaelbr: We talked about moving all interactions on travelcard over to checking tokens instead of customer profiles. However, the webshop at least hasn't done that transition. But that is a improvement over time

src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -0,0 +1,141 @@
...
+          travelRights.map((tr) => tr.userProfileRef),
+          userProfiles,
+        );
+
+        return (
+          <Sections.Section withPadding>
+            <Sections.GenericClickableItem
+              onPress={() =>
+                onPressDetails && onPressDetails(fareContract.orderId)
+              }
```
mikaelbr, line 111: Very nit-picky here, but I think this can be written simpler:

```suggestion
              onPress={() => onPressDetails?.(fareContract.orderId)}
```

src/screens/Ticketing/Ticket/TicketInfo.tsx
```diff
@@ -249,38 +361,42 @@ const TicketInspectionSymbol = ({
...
-        },
-        isValid &&
-          !isInspectable && {
-            ...styles.textContainer,
-            borderColor: theme.static.status.warning.background,
+    <View style={styles.symbolContainer}>
+      <View
+        style={[
+          showAsInspectable && styles.symbolIcon,
+          isValid && {
```
mikaelbr, line 368: I know this isn't a change from before but national park rules and all, this looks like something that really could use some naming and abstractions. Hard to read the logic here.

src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -0,0 +1,141 @@
...
+        const fromTariffZone = findReferenceDataById(tariffZones, firstZone);
+        const toTariffZone = findReferenceDataById(tariffZones, lastZone);
+        const preassignedFareProduct = findReferenceDataById(
+          preassignedFareproducts,
+          productRef,
+        );
+        const userProfilesWithCount = mapToUserProfilesWithCount(
+          travelRights.map((tr) => tr.userProfileRef),
+          userProfiles,
+        );
```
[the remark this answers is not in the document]
mikaelbr: I think mostly my point is that this is a lot to follow when debugging and reading. You have to trace a lot of variables and mentally keep track of them. To me, that's a sign of a need for abstraction.

mikaelbr (approved):
Haven't tested it locally, but from the screen shots it looks good 👍 Some nitpicky/reflection comments that you could consider, but functionally it looks sound.

Feels like at some point we're not hitting the proper level of abstraction for ticket information. Things like validity headings, time calculations, tariffzones always involve a lot of logic to use. Which is also error prone in this case I think. Should at some point™ evaluate if we can introduce some more "low level atom components" and "molocyle components".

jorelosorio:
@mikaelbr Thanks for the feedback! I am going to take your comments and create an issue for refactoring the layer of abstraction of ticketing info, seems like we are repeating the code in multiple places, there are other implementations using memorized, would be nice to have it easier and unified when re-using it on different components.

PR #2850, opened 2022-08-16
feat: opt in Dashboard as front page.

monopoint:
There's a bit of code duplication now, since travel search is in 3 different screens, and property transport is done via navigation


src/screens/Dashboard/Dashboard.tsx
```diff
@@ -1,36 +1,165 @@
...
+import ThemeIcon from '@atb/components/theme-icon';
+import FavoriteChips from '@atb/favorite-chips';
+import {useFavorites} from '@atb/favorites';
+import {useLocationSearchValue} from '@atb/location-search';
+import {useGeolocationState} from '@atb/GeolocationContext';
+import {DashboardParams} from '.';
+import {SelectableLocationData} from '@atb/location-search/types';
+import Bugsnag from '@bugsnag/react-native';
+import {StaticColorByType} from '@atb/theme/colors';
+import {useDoOnceWhen} from '../utils';
```
mikaelbr, line 34: should this be `@atb/..` something?

src/screens/Dashboard/TripSearch.tsx
```diff
@@ -1,53 +1,535 @@
...
+  getSearchTimeLabel,
+  SearchTime,
+  useSearchTimeValue,
+} from '../Assistant/journey-date-picker';
+import {TFunc} from '@leile/lobo-t';
+import {isInThePast} from '@atb/utils/date';
+import useTripsQuery from '../Assistant/use-trips-query';
+import {StaticColorByType} from '@atb/theme/colors';
+import ScreenReaderAnnouncement from '@atb/components/screen-reader-announcement';
+import Results from '../Assistant/Results';
```
mikaelbr, line 44: I would duplicate this from Assistant and have Dashboard _not_ be dependent on anything from Assistant. Keeping them separated and "deletable"

src/screens/Dashboard/index.tsx
```diff
@@ -1,30 +1,62 @@
...
 import {createStackNavigator, TransitionPresets} from '@react-navigation/stack';
-import Dashboard from './Dashboard';
 import TripSearch from '@atb/screens/Dashboard/TripSearch';
+import {TripDetailsRoot} from '../TripDetails';
+import {SearchTime} from '../Nearby/types';
+import {Location} from '@atb/favorites/types';
+import {DetailsRouteParams} from '../TripDetails/Details';
+import JourneyDatePicker, {
+  DateTimePickerParams,
+} from '../Assistant/journey-date-picker';
```
mikaelbr, line 10: Same as the other stuff I think this should be duplicated in the transition/feature toggle phase.

src/screens/Dashboard/index.tsx
```diff
@@ -1,30 +1,62 @@
...
+import DashboardRoot from './Dashboard';
 
 export type DashboardParams = {
-  Dashboard: undefined;
-  TripSearch: undefined;
+  DashboardRoot: {
+    fromLocation: Location | undefined;
+    toLocation: Location | undefined;
+    searchTime: SearchTime | undefined;
+  };
```
mikaelbr, line 18: This type could be a named concept. But we can handle it when refactoring next week

mikaelbr (changes requested):
Some general comments. Didn't go too detailed as we've talked about refactoring when mob programming next week. But I think things imported from `@atb/screens/Assistant` should be duplicated and this holds entirely separate from integrating with the existing Assistant screen. 

Also I think there's been some discussion on SimpleDisappearingHeader here, and I don't know the conclusion there. But personally, I feel like scrolling away the search bar is almost required for smaller screens and allows people to focus better on looking through the search result.

Also name of PR should be changed to be a `feat: ` with some semantic explanation on what we're doing with the new front page.

src/screens/Dashboard/Dashboard.tsx
```diff
@@ -1,36 +1,165 @@
...
+  const currentLocation = location || undefined;
+
+  useDoOnceWhen(
+    () => setUpdatingLocation(true),
+    !Boolean(currentLocation) && hasLocationPermission,
+  );
+  useDoOnceWhen(
+    () => setUpdatingLocation(false),
+    Boolean(currentLocation) && hasLocationPermission,
+  );
```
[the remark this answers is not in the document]
monopoint: I think most of how we handle location stuff could be simplified or streamlined. now there are hook chains and side effect chains that are a bit hard to follow.
We can take a look at it in the refactor session.

src/screens/Dashboard/index.tsx
monopoint: :+1:

src/screens/Dashboard/TripSearch.tsx
```diff
@@ -1,53 +1,535 @@
...
+            text={getSearchTimeLabel(searchTime, timeOfLastSearch, t, language)}
+            accessibilityHint={t(TripSearchTexts.dateInput.a11yHint)}
+            interactiveColor="interactive_0"
+            onPress={onSearchTimePress}
+            testID="assistantDateTimePicker"
+          />
+        </View>
+      </View>
+
       <ScrollView
```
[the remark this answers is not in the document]
monopoint: I guess so.
It's even worse when the header has notifications / disruptions
Make the entire header part of scroll view maybe?

src/screens/Dashboard/TripSearch.tsx
monopoint: ok

jorelosorio (approved):

src/screens/Dashboard/TripSearch.tsx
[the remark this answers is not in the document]
mikaelbr: Example from departuers here now also: 
Screenshot 2022-08-17 at 14 08 51

Looks pretty rediculous to be honest 😂 I'd argue that messages should also should disappear but I also see why they are placed there.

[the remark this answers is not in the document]
monopoint: I agree.
but I think maybe this is out of scope for this issue?

PR #2852, opened 2022-08-17
chore: updates @react-native-firebase/* dependencies

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/1776 

Upgrades all `@react-native-firebase/*` dependencies. Should update dependencies causing issue with reconnect in android.

Tested on iOS and Android locally. There might be some changes and seen some reports that it might break Android on the device. So should be tested thoroughly here.


Nisha1306 (approved):
🤞

PR #2853, opened 2022-08-17
fix: Fix sca exemption flow

gorandalum:
Sca exemption flow was not working as the bankid authentication was "hidden" from
the user. The code was simplified some to fix this, and will now only have a
screen saying "loading", and not inspect redirect urls to find out the appropriate
loading screen text.

Fixes AtB-AS/kundevendt#2114

### DIFF
--- a/src/screens/Ticketing/Purchase/Payment/CreditCard/use-terminal-state.ts
+++ b/src/screens/Ticketing/Purchase/Payment/CreditCard/use-terminal-state.ts
@@ -1,10 +1,6 @@
 import {AxiosError} from 'axios';
-import {useCallback, useEffect, useReducer, useRef, useState} from 'react';
-import {
-  WebViewError,
-  WebViewErrorEvent,
-  WebViewNavigation,
-} from 'react-native-webview/lib/WebViewTypes';
+import {useCallback, useEffect, useState} from 'react';
+import {WebViewNavigation} from 'react-native-webview/lib/WebViewTypes';
 import {parse as parseURL} from 'search-params';
 import {ErrorType, getAxiosErrorType} from '@atb/api/utils';
 import {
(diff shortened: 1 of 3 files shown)


toretefre:
Had a look at this in the morning, since I worked on the same code earlier. It looks like a good simplification, but I don't think I'm comfortable approving it alone.

gorandalum:
Just a small note on what was actually fixed. Earlier the bankid authentication was hidden from the user as the `loadingState` was interpreted wrong or had the wrong value. Instead of deepdiving into getting the correct `loadingState`, I simplified it to just be a parameter `isLoading` which is `true` until first `onLoadEnd`. After that there is no more in-app loading screens, and we no longer have complex code to try to maintain a local state of where in the purchase process the customer is.

src/screens/Ticketing/Purchase/Payment/CreditCard/use-terminal-state.ts
```diff
@@ -148,85 +66,50 @@ export default function useTerminalState(
...
-    // for each type of resource, html, assets, etc
-    // so we have a "loading guard" here
-
-    if (initialLoadRef.current) {
-      initialLoadRef.current = false;
-      dispatch({type: 'TERMINAL_LOADED'});
-    } else if (!redirectToPaymentCaptureRef.current) {
-      if (
-        url.includes('/ticket/v2/payments/') ||
-        url.includes('atb://ticketing')
```
Nisha1306, line 212: we are not using this anymore? does it have any consequences? was it redundant?

gorandalum: It was pretty complex code. My strategy was to remove as much as possible, and see if it still works 😅 It was probably not redundant with the old implementation where several loading states was tracked, but with this new implementation it should no longer be necessary.

Nisha1306 (approved):
LGTM! 🚀

PR #2854, opened 2022-08-17
feat: disable favorites toggle on quays without favorites

rosvik:
follow up to https://github.com/AtB-AS/mittatb-app/pull/2846

After a discussion with @ckbilstad, we decided to disable favorites filtering on quays that has no favorites. Previously, the list would display as empty if the toggle was enabled.

https://user-images.githubusercontent.com/1774972/185149741-529992f6-b485-420e-a5e4-4bcb50c2c510.mp4

Acceptance criteria:
- [ ] Only quays with favorites gets the toggle
- [ ] Toggling other places does not filter departures in the quay without favorites
- [ ] Adding the first favorite for a quay makes the toggle appear. It inherits the toggle state from the other views. 
- [ ] Removing every favorite from a quay makes the toggle disappear

### DIFF
--- a/src/screens/Departures/QuayView.tsx
+++ b/src/screens/Departures/QuayView.tsx
@@ -1,14 +1,13 @@
 import {Place, Quay} from '@atb/api/types/departures';
 import {useFavorites} from '@atb/favorites';
-import {SearchTime} from '@atb/screens/Departures/utils';
+import {SearchTime, hasFavorites} from '@atb/screens/Departures/utils';
 import {StyleSheet} from '@atb/theme';
-import React, {useEffect} from 'react';
+import React, {useEffect, useState} from 'react';
 import {RefreshControl, SectionList, SectionListData, View} from 'react-native';
 import DateNavigation from './components/DateNavigator';
 import FavoriteToggle from './components/FavoriteToggle';
 import QuaySection from './components/QuaySection';
 import {useQuayData} from './state/quay-state';
-import {hasFavorites} from './StopPlaceView';
 
 export type QuayViewParams = {
   quay: Quay;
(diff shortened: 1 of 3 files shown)


mikaelbr:
Is this still a draft, or should it be reviewed?

rosvik:
@mikaelbr Thanks for the reminder, it should be ready for review now. Wanted to test the logic properly in simulator before the weekend, and after the weekend it went straight to the glemmebok.

gorandalum:
The code looks good, just one question, mostly for @ckbilstad:
Is this ok flow for people with screen readers? I've heard before (and I think Hilde followed this principle) that you shouldn't dynamically add/remove elements from a page/screen (at least after loading it), since it is confusing for those with screen readers. It should rather be disabled, with a screen reader message why it is disabled.

I searched the big Interwebs now and didn't find an official clarification for if this is a standard best practice or not, so I think we can make our own decision whether it should be hidden or disabled.

ckbilstad:
I dont have any previous experience with the specific screen reader issue. Based on the information provided, i suggest we keep it as is for now (if @rosvik dont have more details). This might be subject to change as part of the review of favorites as a whole down the road.

gorandalum:
My suggestion is to keep the code mostly as it is now, just disable the toggle instead, and add a a11y message why it is disabled. "No relevant stops" or something like it.

ckbilstad:
Sounds like a plan @gorandalum

rosvik:
This gets complicated in practice, so I needed som time to think about this. I have two questions:

1. Should the toggle _always_ be visible then? Also when the entire stop has no favorites? If so, the state of the switch also becomes visible even when it's not possible to change.
2. When a quay doesn't have favorites, the toggle state could still be true (because the state is shared between screens), but we want to load all departures without filtering. Should the state be shown as disabled in that case, and then enabled when you navigate to a quay with favorites?

If the answer to both of these questions are yes, then:

This means that we need three states 
- "global toggle state": if the toggle is enabled for the current stop
- "local toggle state": if the app is filtering departures in the current view
- "interaction enabled toggle state": if the toggle component should be enabled for interaction from the user

When a stop loses all of its favorites, the global toggle state _could_ be set to false.

gorandalum:
Yes this get's complicated fast! I think for now it is better to have it as it is, since it also how it is in v1. The global toggle state is remembered, even though you don't have any favourites.

PR #2857, opened 2022-08-17
fix: stop details should be updated to empty array when input is empty

Nisha1306:
Even when requested location ids were empty we were sending a request to BFF, which in return gives a 400, but doesn't do anything with the existing state. This results in an incorrect state in favorites tab and probably elsewhere as well.

So, here we set the output to empty when the input is empty without even calling BFF.

### DIFF
--- a/src/screens/Departures/state/stop-place-details-state.ts
+++ b/src/screens/Departures/state/stop-place-details-state.ts
@@ -59,6 +59,14 @@ const reducer: ReducerWithSideEffects<
         },
         async (state, dispatch) => {
           if (!action.locations) return;
+          if (action.locations.length === 0) {
+            dispatch({
+              type: 'UPDATE_STOP_DETAILS',
+              ids: action.locations,
+              result: {stopPlaces: []},
+            });
+            return;
+          }
           try {
             const result = await getStopsDetails({ids: action.locations});
             dispatch({
--- a/src/screens/Departures/FavouriteStopsOverview.tsx
+++ b/src/screens/Departures/FavouriteStopsOverview.tsx
@@ -72,7 +72,7 @@ const FavouriteStopsOverview = ({navigation}: RootProps) => {
     });
   };
 
-  return favouriteStopsDetails ? (
+  return favouriteStopsDetails && favouriteStopsDetails.length > 0 ? (
     <StopPlaces
       header={t(DeparturesTexts.resultType.favourites)}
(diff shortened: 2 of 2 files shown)


src/screens/Departures/state/stop-place-details-state.ts
```diff
@@ -59,6 +59,14 @@ const reducer: ReducerWithSideEffects<
         },
         async (state, dispatch) => {
           if (!action.locations) return;
+          if (action.locations.length === 0) {
+            dispatch({
+              type: 'UPDATE_STOP_DETAILS',
+              ids: action.locations,
+              result: {stopPlaces: []},
```
mikaelbr, line 66: This could be solved instead using the same code as below, or?

```ts
onst result = await getStopsDetails({ids: action.locations});
            dispatch({
              type: 'UPDATE_STOP_DETAILS',
              ids: action.locations,
              result: !action.locations.length ? {stopPlaces: []} : result
            });
```

mikaelbr: I'd also say that maybe passing the entire result here is kind of tightly coupling the dispatcher/reducer with service layer, but it's not strictly necessary to fix.

Nisha1306: BFF gives a 400 and goes to the catch block. SO I thought why to even call it.

mikaelbr: Hmm. Ok. I would say that there's kind of an odd flow here then. We're calling something we know will fail and still update the state with something that should strictly speaking be the default? I think a cleaner flow would be:

1. Sane defaults with empty lists
2. If locations, dispatch LOAD.

But I'm perfectly happy with just patching this like you have now.

mikaelbr (approved):

src/screens/Departures/state/stop-place-details-state.ts
Nisha1306: I think the empty list is actually a default already. 
But the problem occurs when 
1. We call BFF with a valid value i.e. non-empty list, 
2. Get valid data back and update stat the result
3. Call BFF again, but with an empty array which is an invalid input
4. Get error from BFF and land in catch
5. The state which now should be an empty list remains untouched leading to an incorrect state.

So an alternative would be to have this in the catch but then we have to figure out the type of error and then only do it.  😅

PR #2863, opened 2022-08-18
feat: add global message data validation

rosvik:
resolves https://github.com/AtB-AS/kundevendt/issues/2219
fixes https://github.com/AtB-AS/kundevendt/issues/2200

Adds type validation of global messages similarly to how it's implemented in the webshop.

Also fixes that duplicate messages are shown in departures v2.


toretefre (approved):
Neat!

PR #2865, opened 2022-08-18
feat: Buy ticket button

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/1880

If no tickets are available shows a button to buy.

Screenshot_1660812504

### DIFF
--- a/src/translations/screens/Tickets.ts
+++ b/src/translations/screens/Tickets.ts
@@ -7,6 +7,9 @@ const TicketsTexts = {
   header: {
     title: _('Billetter', 'Tickets'),
   },
+  dashboard: {
+    label: _('Kjøp billetter', 'Buy tickets'),
+  },
   buyTicketsTab: {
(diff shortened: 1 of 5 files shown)


src/translations/screens/Tickets.ts
```diff
@@ -7,6 +7,9 @@ const TicketsTexts = {
   header: {
     title: _('Billetter', 'Tickets'),
   },
+  dashboard: {
+    label: _('Kjøp billetter', 'Buy tickets'),
+  },
```
gorandalum, line 12: I think this is the wrong translation file to put this text, as it is used in Dashboard components and not in Tickets components.

gorandalum (approved):
Just commented a small change you can do.

src/translations/screens/Tickets.ts
jorelosorio: Fixed :)

PR #2868, opened 2022-08-18
chore: adds logging on "old" reservations

mikaelbr:
Related to debugging https://github.com/AtB-AS/kundevendt/issues/1776

Add log for every reservation that is older than 1 minute ago at the time of discovery.

PR #2875, opened 2022-08-23
chore: Revert "chore: adds logging on "old" reservations (#2868)"

mikaelbr:
This reverts commit 7979014df9b797f59c5d63c4cb89eb4a22323318.

This seems to produce a lot of false positives and make more noise than actual valuable logging after running a while.


gorandalum (approved):
Alright. Seems like a good decision.

PR #2876, opened 2022-08-23
chore: add nfk test device

toretefre:
Should add test device of Amp team to Reis staging, just added a line with UUID and name.


HanneLH (approved):
🌸

PR #2881, opened 2022-08-24
chore: correct nfk test device name

toretefre:
When running `fastlane ios update_devices`, I'm getting the error `"The Device Name **** is too large. Device Name cannot exceed 50 characters."` 

Therefore shortening it a bit


HanneLH (approved):
🌸

PR #2882, opened 2022-08-24
Workaround for higher time-out

jorelosorio:
It handles "large" time-out times when building, something built can be slow due to a large number of dependencies or limited resources in the host machine, so is good to have these.


mikaelbr (approved):
I don't know if there is any other limitations for GH Actions or anything, but worth a shot in those extreme cases.

PR #2883, opened 2022-08-24
feat: Select favourite departures in dashboard

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/1877

Missing:
- A nice and shiny button on the front page. Thinking the button should be a separate PR following Espens PR on displaying favourite departures on the dashboard to avoid merge conflicts. For testing purposes, the button in the Profile section can be used for opening the bottom sheet.

<details>

image

simulator_screenshot_3753F2B4-F431-4377-823E-3245DE855A30

simulator_screenshot_D4B01FB6-910B-43BC-BBCC-CA477E6297E4

This should not be possible to get from the front page, as the button to open the bottom sheet will be hidden, but needs to be included if opening from the profile page:

image

</details>


src/screens/Assistant/SelectFavouritesBottomSheet.tsx
```diff
@@ -0,0 +1,239 @@
...
+                </View>
+              );
+            })}
+          {!favouriteItems && (
+            <MessageBox type="info">
+              <ThemeText>
+                {t(SelectFavouriteDeparturesText.noFavourites.text)}
+              </ThemeText>
+            </MessageBox>
+          )}
```
toretefre, line 168: This one needs a check on length as the default is `[]`, will fix that (hopefully before any of you get to review 😅 

(the same applies for the other check)

toretefre: Should be good now 👌

toretefre:
This is ready for review now 👍

src/translations/screens/Profile.ts
```diff
@@ -88,6 +88,13 @@ const ProfileTexts = {
             'Activate to edit favourite departures',
           ),
         },
+        frontpageFavourites: {
+          label: _('Avganger på forside', 'Dashboard departures'),
```
Nisha1306, line 92: Should this be `Departures on the front page` instead?

Nisha1306:
@toretefre  the toggle doesn't look very much toggle-ey in dark mode (at least in android ) unlike your screenshots 🤔

Screenshot_20220830-125053

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
```diff
@@ -0,0 +1,249 @@
...
+            <View>
+              {favouriteItems &&
+                favouriteItems.map((departureDetails, index) => {
+                  return (
+                    <View key={departureDetails.id}>
+                      <SelectableFavouriteDeparture
+                        handleSwitchFlip={handleSwitchFlip}
+                        favouriteId={departureDetails.id}
+                        active={
+                          frontPageFavouriteDepartures.find(
```
Nisha1306, line 144: Can be simplified to  
```
  !!frontPageFavouriteDepartures.find(
                              (departure) => departure.id === departureDetails.id,
                          )
```

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
```diff
@@ -0,0 +1,249 @@
...
+      </ScrollView>
+
+      <FullScreenFooter>
+        <Button
+          interactiveColor="interactive_0"
+          text={t(SelectFavouriteDeparturesText.confirm_button.text)}
+          accessibilityHint={t(
+            SelectFavouriteDeparturesText.confirm_button.a11yhint,
+          )}
+          onPress={() => {
```
Nisha1306, line 188: can be simplified to  `onPress={close}` :)

Nisha1306:
> > @toretefre the toggle doesn't look very much toggle-ey in dark mode (at least in android ) unlike your screenshots 🤔
> 
> Thats not a very toggly toggle, will check out what has gone wrong when copying the styling from another place in the code 😅

👍  rest looks good to me

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+    if (!favouriteInFrontpageFavourites) {
+      const favourite = favoriteDepartures.find(
+        (departure) => departure.id === favouriteId,
+      );
+      if (!favourite) return;
+      addFrontPageFavouriteDeparture(favourite);
+    } else if (favouriteInFavourites) {
+      removeFrontPageFavouriteDeparture(favouriteInFrontpageFavourites);
+    }
```
rosvik, line 113: Having a hard time understanding what's going on here. Tried to formulate some questions, but it may be good to try to do a refactor to make it more readable.

- Should `favouriteInFavourites` be a boolean? (You can use `.some()` instead of `.find()` if so. Also, is it strictly necessary to check if the favorite exists before you (try to) remove it? Could this work just as well?
```ts
else {
  removeFrontPageFavouriteDeparture(favouriteInFrontpageFavourites);
}
```

- Also, what's the difference between `favouriteItems` and `favoriteDepartures`? And why do you need to search for the input departure id in both?

src/translations/screens/subscreens/SelectFavouriteDeparturesTexts.ts
```diff
@@ -0,0 +1,38 @@
...
+  switch: {
+    a11yhint: _(
+      'Aktiver for å vise eller skjule denne favorittavgangen på forsiden',
+      'Activate to show or hide this favourite departure on the front page.',
+    ),
+  },
+  noFavourites: {
+    text: _(
+      'Vi fant ingen favorittavganger. For å legge til en favorittavgang må du aktivere ny avganger-visning, og klikke på stjernen ved siden av en avgang.',
+      'We found no favourite departures. To add a favourite departure, you may activate the new Departures-view, and then press the star next to a departure.',
```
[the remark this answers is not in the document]
toretefre: After a closer look, it absolutely does! 😅 

I think my train of thoughts here was that departures v1 did not have favourites, but that was of course version 2 until just now, I just forgot while writing this message 😄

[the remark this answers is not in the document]
toretefre: Will update the message

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
toretefre: Ah, that's true! Fixing, thanks :)

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
toretefre: Absolutely, fixing that now. Tried to remember that one, but no success at that time 😅

src/translations/screens/Profile.ts
toretefre: We're moving it to developer options, so don't think I will spend time changing it, but it might be more user-friendly for end users :)

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
toretefre: I flipped the whole function, so I think it should be more readable now. Any opinions, @rosvik?

```
  const handleSwitchFlip = (favouriteId: string) => {
  // Checks if favourite is currently selected as a front page favourite
    const favouriteInFrontpageFavourites = frontpageFavouriteItems.find(
      (item) => item.id === favouriteId,
    );

    // if its already in the frontpage list, it should be removed
    if (favouriteInFrontpageFavourites) {
      removeFrontPageFavouriteDeparture(favouriteInFrontpageFavourites);
    } else {
    // if its not part of the frontpage list, it should probably be added
      const favourite = favoriteDepartures.find(
        (departure) => departure.id === favouriteId,
      );
     // there is also the crazy chance of the departure not being part of the ordinary favourite list anymore, 
     // so we will need to check if the favourite has been found, we would also get a type error if we don't
      if (favourite) addFrontPageFavouriteDeparture(favourite);
    }
  };
```

monopoint: looks like you are using both "Favorite" and "Favourite" as spelling. maybe go for just one, and maybe use the  more international "Favourite"

monopoint:
what happens if a user removes a favourite, will the front page favourites be updated as well?

toretefre:
> looks like you are using both "Favorite" and "Favourite" as spelling. maybe go for just one, and maybe use the more international "Favourite"

Yeah, that's an issue. Thing is, the already existing context for favourites is written using the American spelling favorite, while the app language is British English where "favourite" is the correct spelling. 

So either that means rewriting unrelated code to using the British English "favourite" spelling, or using both spellings @monopoint  😬

toretefre:
> what happens if a user removes a favourite, will the front page favourites be updated as well? looks like we need to remove the front page favourite when the favourite is removed.

Good catch! Will need to remove frontpage favourite at the time of deleting the favourite. Will implement shortly

src/screens/Assistant/SelectFavouritesBottomSheet.tsx
rosvik: I like what you did in the branch now 💯

rosvik (approved):

toretefre:
> @toretefre the toggle doesn't look very much toggle-ey in dark mode (at least in android ) unlike your screenshots 🤔

This seems to be an issue throughout the app, and will be handled separately: https://github.com/AtB-AS/kundevendt/issues/2359

PR #2887, opened 2022-08-24
refactor/fix carnet warning

Nisha1306:
Solves suggestions on https://github.com/AtB-AS/kundevendt/issues/330 

- Show "Bring your travel card" warning for carnet tickets as well
- Modify the zones text as per the tickets
- Add round borders for carnet view

### DIFF
--- a/src/translations/screens/subscreens/TariffZones.ts
+++ b/src/translations/screens/subscreens/TariffZones.ts
@@ -14,8 +14,8 @@ const TariffZonesTexts = {
         _(`Reise i 1 sone (${zoneName})`, `Travel in 1 zone (${zoneName})`),
       multipleZone: (zoneNameFrom: string, zoneNameTo: string) =>
         _(
-          `Reise fra sone ${zoneNameFrom} til sone ${zoneNameTo}`,
-          `Travel from zone ${zoneNameFrom} to zone ${zoneNameTo}`,
+          `Reise i 2 soner (${zoneNameFrom} - ${zoneNameTo})`,
+          `Travel in 2 zones (${zoneNameFrom} - ${zoneNameTo})`,
         ),
     },
   },
--- a/src/screens/Ticketing/Ticket/Carnet/CarnetDetails.tsx
+++ b/src/screens/Ticketing/Ticket/Carnet/CarnetDetails.tsx
@@ -45,7 +45,7 @@ export function CarnetDetails(props: {
   );
 
   return (
-    <>
+    <Sections.Section>
       <Sections.GenericItem>
         {fareContractValidityStatus !== 'valid' ? (
           <ValidityHeader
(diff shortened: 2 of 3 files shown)


src/translations/screens/subscreens/TariffZones.ts
```diff
@@ -14,8 +14,8 @@ const TariffZonesTexts = {
         _(`Reise i 1 sone (${zoneName})`, `Travel in 1 zone (${zoneName})`),
       multipleZone: (zoneNameFrom: string, zoneNameTo: string) =>
         _(
-          `Reise fra sone ${zoneNameFrom} til sone ${zoneNameTo}`,
-          `Travel from zone ${zoneNameFrom} to zone ${zoneNameTo}`,
+          `Reise i 2 soner (${zoneNameFrom} - ${zoneNameTo})`,
+          `Travel in 2 zones (${zoneNameFrom} - ${zoneNameTo})`,
```
mikaelbr, line 18: Can't this be multiple zones? 🤔 From B to C2 is travelling through several zones in a way

Nisha1306: mmm yes, but if the customer would buy a ticket from one zone to two another zone, so aren't those the zones they care about?

mikaelbr: But the text is "Travel *in* 2 zones" but in reality, you can travel *in* multiple zones in that case? I think this might be somewhat misleading/unintuitive. I also *think* you could get off the bus within a zone and travel there before going further. In single tickets that might sound weird but with periodic tickets that makes sense.

Nisha1306: oh..so when we say "Travel from zone A to E2", what does it really mean?

Nisha1306: btw we discussed this here and I think we are not 100% clear on how multi-zone periodic tickets work. So we decided to switch it back to what we have now to avoid confusion for the customers. 👍

mikaelbr: Should verify this with Sebastian et.al. Albeit more implicit that's at least "more correct"™

Nisha1306: yep...would do that once @Sebstorvik  is here.

gorandalum: Yeah, technically if we buy a ticket from the south of Trøndelag to the north of Trøndelag, like C3 to E2, the ticket will actually get all the zones C3, B2, A. C6, D, E2 and be validated to green in all of them. Just for the app we simplify this by just using start and end zone. So travel in 2 zones is not correct phrasing.

gorandalum: But since we can't say what is the start zone and what is end zone, maybe something like "travel between C3 and E2" is better than "travel from C3 to E2"

Nisha1306: I discussed this with Sebastian. Technically yes, the ticket is valid for intermediate stops as well. However, regarding the wording, he would say he would discuss it with Gunn and get back. 
I am closing this PR for now, as the rest of the small commits have been taken care of in another PR.

PR #2890, opened 2022-08-25
feat: Redirect the user to purchase overview and prefill fields with the previously purchased ticket

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/1774

### DIFF
--- a/src/screens/Ticketing/Purchase/Overview/index.tsx
+++ b/src/screens/Ticketing/Purchase/Overview/index.tsx
@@ -104,25 +104,25 @@ const PurchaseOverview: React.FC<OverviewProps> = ({
     preferences: {defaultUserTypeString},
   } = usePreferences();
 
-  const defaultUserType: UserProfileTypeWithCount = {
+  const defaultPreSelectedUser: UserProfileTypeWithCount = {
     userTypeString: defaultUserTypeString ?? userProfiles[0].userTypeString,
     count: 1,
   };
 
-  const defaultUsers = params.userProfilesWithCount?.map(
-    (up: UserProfileTypeWithCount): UserProfileTypeWithCount => {
+  const preSelectedUsers = params.userProfilesWithCount?.map(
+    (up: UserProfileWithCount): UserProfileTypeWithCount => {
       return {userTypeString: up.userTypeString, count: up.count};
     },
   );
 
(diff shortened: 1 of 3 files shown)


src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -82,19 +82,33 @@ const PurchaseOverview: React.FC<OverviewProps> = ({
...
+    defaultUserProfilesWithCount.forEach((u) => (u.count = 0));
+    params.userProfilesWithCount?.forEach((u: UserProfileWithCount) => {
+      defaultUserProfilesWithCount
+        .filter(
+          (upc: UserProfileWithCount) => upc.userTypeString == u.userTypeString,
+        )
+        .forEach((v: UserProfileWithCount) => (v.count = u.count));
+    });
+  }
+
```
gorandalum, line 111: I think we can try to simplify this a little bit, also the logic in `useDefaultUserProfilesWithCount`. The logic should be something like this:
- Get the user profiles from Firestore
- Filter by the limitations on the preassigned fare product
- If userProfilesWithCount in screen parameter: Apply them
- If no userProfilesWithCount in screen parameter: Apply default
- Use this as the initial state for travellerSelection

gorandalum (changes requested):
Good work, and everything looks correct! Just suggested some refactoring to make this more maintainable for future developers, as both the already existing `useDefaultUserProfilesWithCount` and the new `filter`-/`forEach`-logic is quite complex.

src/screens/Ticketing/Purchase/Overview/index.tsx
jorelosorio: I improved this

gorandalum: Slightly better, although still kinda complex, but haven't seen into if it actually possible to do in a less complex manner. Approving now!

gorandalum (approved):

src/screens/Ticketing/Purchase/Overview/index.tsx
jorelosorio: I agree it is complex, I did the easiest possible with the current code solution. I have also changed the naming of some vars for a better understanding of what is going on.

PR #2892, opened 2022-08-25
feat: warn other device is inspectable on confirmation screen

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/2216

Screenshot 2022-08-25 at 12 53 28

https://user-images.githubusercontent.com/1774972/186646349-5d33175f-21ca-4f63-a6f7-511825bd6fc2.MP4

### DIFF
--- a/src/screens/Ticketing/Purchase/Confirmation/index.tsx
+++ b/src/screens/Ticketing/Purchase/Confirmation/index.tsx
@@ -34,6 +34,16 @@ import {useAuthState} from '@atb/auth';
 import {usePreviousPaymentMethod} from '../saved-payment-utils';
 import MessageBoxTexts from '@atb/translations/components/MessageBox';
 import {useFirestoreConfiguration} from '@atb/configuration/FirestoreConfigurationContext';
+import {
+  useHasEnabledMobileToken,
+  useMobileTokenContextState,
+} from '@atb/mobile-token/MobileTokenContext';
+import {
+  findInspectable,
+  getDeviceName,
+  isTravelCardToken,
+} from '@atb/mobile-token/utils';
+import {getOtherDeviceIsInspectableWarning} from '../../Ticket/utils';
 
 export type RouteParams = {
   preassignedFareProduct: PreassignedFareProduct;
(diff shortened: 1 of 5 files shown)


src/screens/Ticketing/Purchase/Confirmation/index.tsx
```diff
@@ -34,6 +34,16 @@ import {useAuthState} from '@atb/auth';
...
+import {
+  useHasEnabledMobileToken,
+  useMobileTokenContextState,
+} from '@atb/mobile-token/MobileTokenContext';
+import {
+  findInspectable,
+  getDeviceName,
+  isTravelCardToken,
+} from '@atb/mobile-token/utils';
+import {getOtherDeviceIsInspectableWarning} from '../../Ticket/utils';
```
toretefre, line 46: Not sure what import guidelines we are following, and the pros and cons of the possible options, but my personal feeling is that multiple navigation dots can be less readable than explicit references.

```suggestion
import {getOtherDeviceIsInspectableWarning} from '@atb/screens/Ticketing/Ticket/utils';
```

src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -44,7 +43,7 @@ import StartTimeSelection from './components/StartTimeSelection';
 import TravellerSelection from './components/TravellerSelection';
 import FullScreenFooter from '@atb/components/screen-footer/full-footer';
 import {getPurchaseFlow} from '../utils';
-import {getDeviceNameForCurrentToken} from '@atb/mobile-token/utils';
+import {getOtherDeviceIsInspectableWarning} from '../../Ticket/utils';
```
toretefre, line 46: Same as https://github.com/AtB-AS/mittatb-app/pull/2892/files#r954826095

```suggestion
import {getOtherDeviceIsInspectableWarning} from '@atb/screens/Ticketing/Ticket/utils';
```

toretefre (approved):
Not able to buy tickets in staging right now to verify, but from the code and the video I think it looks great. 

Added a suggestion on import styling, but I guess that's nitpicking and personal preferences, so no need to accept it! 🚀

src/screens/Ticketing/Purchase/Confirmation/index.tsx
rosvik: I was actually considering the same thing. And I'm not sure what's best, but since both these files are part of screens/Ticketing, maybe this is fine as is? I dont think we have a policy on this, so I just stuck with whatever was auto generated ¯\_(ツ)_/¯

rosvik: (Turns out, shrugging is hard in markdown: `¯\\\_(ツ)\_/¯` -> ¯\\\_(ツ)\_/¯)

toretefre: I had a feeling they were auto generated 😄 

As long as you think it is the most readable way, it works just fine! 👌

rosvik:
I made a few changes since that demo was recorded, so I'm waiting with merge until the problems with offers in staging are resolved and I get to verify this.

src/screens/Ticketing/Purchase/Confirmation/index.tsx
rosvik: Maybe not more readable in this case, so I'm not against changing it this once, but it will be a lot more effort trying to enforce something other than what's auto generated in the future, so I'm leaning towards keeping it.

PR #2894, opened 2022-08-25
fix: add spinner during logout

Nisha1306:
Solves https://github.com/AtB-AS/kundevendt/issues/1906

https://user-images.githubusercontent.com/29625161/187071549-4150508a-03af-414c-9707-2c6e87c8f65c.mov

https://user-images.githubusercontent.com/29625161/187071551-dcfd54b9-a15f-432a-b7d0-a18baa64b587.mov

### DIFF
--- a/src/components/activity-indicator-overlay/index.tsx
+++ b/src/components/activity-indicator-overlay/index.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import {ActivityIndicator, View} from 'react-native';
+import {StyleSheet} from '@atb/theme';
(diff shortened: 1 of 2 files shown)


mikaelbr:
I think it's more common to have a modal and loading indicator within a box 🤔 But I also think this is a very edge feature that we don't need to spend too much time on, so this looks fine 😅  The ActivityIndicator component has some accessibility built in announcing loading for screen readers, no?

mikaelbr (approved):

PR #2896, opened 2022-08-26
fix: update validation of GlobalMessage data

rosvik:
Noticed a weakness in the data validation of global messages, where if `type` or `active` is set to some value not in the Status-type, the invalid value would slip trough, eventually leading to the app crashing.

This would be an issue if a typo is made in firestore, or if we add more status types in the future. Since the bug is part of v1.23, it will stay an issue for users on that version of the app.

Also related to https://github.com/AtB-AS/kundevendt/issues/2190


toretefre (approved):
Good catch! 🙈

PR #2897, opened 2022-08-29
refactor: navigate to correct stack after login

Nisha1306:
Solves bug in https://github.com/AtB-AS/kundevendt/issues/2032

### DIFF
--- a/src/screens/AnonymousTicketPurchase/components/Actions.tsx
+++ b/src/screens/AnonymousTicketPurchase/components/Actions.tsx
@@ -1,119 +1,39 @@
 import Button from '@atb/components/button';
 import AnonymousTicketPurchases from '@atb/translations/screens/subscreens/AnonymousTicketPurchases';
 import React from 'react';
-import {useFinishOnboarding} from '@atb/screens/Onboarding/use-finish-onboarding';
 import {useTranslation} from '@atb/translations';
 import {StyleSheet} from '@atb/theme';
-import {useRemoteConfig} from '@atb/RemoteConfigContext';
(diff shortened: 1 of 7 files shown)


src/screens/AnonymousTicketPurchase/components/Actions.tsx
```diff
@@ -27,6 +27,10 @@ const Actions = ({
       params: {
         afterLogin: {
           routeName: 'TabNavigator',
+          routeParams:
+            callerRoute === 'ConsequencesFromTicketPurchase'
+              ? {screen: 'Ticketing'}
+              : {},
```
gorandalum, line 33: Generally the coupling between Actions-component and the components which uses it feels little iffy. They are not decoupled, but a pretty tight coupling as the Actions-component is aware of its users and has business logic dependant on it. If possible, and not to much work, I would suggest a refactor of this so the Actions-component is more decoupled from its usages.

src/screens/Ticketing/Purchase/index.tsx
```diff
@@ -55,6 +56,7 @@ type TicketPurchaseRootProps = {
 };
 
 export default function PurchaseStack({route}: TicketPurchaseRootProps) {
+  useGoToMobileTokenOnboardingWhenNecessary();
```
gorandalum, line 59: Should this then be removed from the `TabNavigator`? 🤔

gorandalum (approved):
Good! Some comments you can take a look at.

gorandalum (approved):

PR #2898, opened 2022-08-29
refactor: check if ticket has travelright in compact ticket

rosvik:


src/screens/Dashboard/CompactTickets.tsx
```diff
@@ -79,6 +79,7 @@ const CompactTickets: React.FC<Props> = ({
         // TODO: Move all this initialization into a better layer of abstraction!
         const travelRights =
           fareContract.travelRights.filter(isPreactivatedTicket);
+        if (travelRights.length < 1) return undefined;
```
mikaelbr, line 82: Maybe this is better as `null`? More common way to return non-existing component

```suggestion
        if (travelRights.length < 1) return null;
```

mikaelbr (changes requested):
See comment. Think it should be `return null`

mikaelbr (approved):
I'm guessing not showing anything when there is no travel right makes sense!

PR #2899, opened 2022-08-29
fix: improve multiple zone text in valid and buy tickets

Nisha1306:
Solves zone text as part of https://github.com/AtB-AS/kundevendt/issues/330

Before: 

<div>
  
  
</div>

After: 

<div>
  
  
</div>

### DIFF
--- a/src/translations/screens/subscreens/TariffZones.ts
+++ b/src/translations/screens/subscreens/TariffZones.ts
@@ -14,23 +14,12 @@ const TariffZonesTexts = {
         _(`Reise i 1 sone (${zoneName})`, `Travel in 1 zone (${zoneName})`),
       multipleZone: (zoneNameFrom: string, zoneNameTo: string) =>
         _(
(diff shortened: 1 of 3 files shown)


src/translations/screens/subscreens/TariffZones.ts
```diff
@@ -14,23 +14,12 @@ const TariffZonesTexts = {
...
-        return _(
-          `Reise i ${numberOfZones} soner`,
-          `Travel in ${numberOfZones} zones`,
-        );
-      }
-
-      return _(`Reise i 1 sone`, `Travel in 1 zone`);
-    },
-  },
+  zoneTitle: _(`Reisesone(r)`, `Travel zone(s)`),
```
Nisha1306, line 22: Showing between or from on title was repetitive. So gave a static text to it irrespective of selected number of zones.

gorandalum (approved):

PR #2902, opened 2022-08-30
feat: Favourites widget

monopoint:
resolves issue: https://github.com/AtB-AS/kundevendt/issues/1878

Image


src/screens/Dashboard/DeparturesWidget.tsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    return () => {
+      if (interval) {
+        clearInterval(interval);
+      }
+    };
+  }, [polling]);
+
+  // do polling only when screen has focus and user has faourites.
```
rosvik, line 59: If we're unsure about whether to use "favorite" or "favourite", we can always settle with "faourite" 😆

```suggestion
  // do polling only when screen has focus and user has favourites.
```

rosvik (approved):
Looks good 💪

src/screens/Dashboard/DeparturesWidget.tsx
monopoint: :D fixed

PR #2903, opened 2022-08-30
fix: add app version as header to all api calls

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2346


Nisha1306:
> Very good and efficient! Some notes:
> 
> The title should be "refactor:" and not "fix:" as this is not a bug fix that is relevant for the release notes.
> 
> Rather than using "Solves" in the description you can use "Resolves", which will make the PR and issue better linked: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Yeah.. Forgot to have refactor instead of fix. Remember you had mentioned before 😅 Next time 👍

PR #2905, opened 2022-08-31
refactor: Compact ticket fixes

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/1880

Screenshot_1661938031


gorandalum (approved):
Seems ok! Remember: The PR prefix "fix" is for bugfixes that are relevant for our users, and if using "fix" the title needs to describe what is fixed for the user.

In this case the fix is not relevant for the users, since the new front page hasn't been released yet to them. So in stead the prefix "refactor" should be used. You can edit it when squashing and merging.

PR #2907, opened 2022-08-31
refactor: blank favourites list when user deletes all favourites

monopoint:
when the user updates her favourites we do a data request to BFF to update the list.
but we make no request if the user has no favourites.
this causes the list to display the last favourite, even if it is removed.

this PR blanks the list when the user has no favourites.

also refactored better variabel names


rosvik (approved):
💯

jorelosorio (approved):

PR #2908, opened 2022-08-31
refactor: Merged ZoneSymbol and NotInspectableSymbol

gorandalum:
They were two components with two different implementations, although
they should look mostly the same. Also they were in two separate places
in the GUI, but after talking with Christian, it was decided they can be
presented at the same place.

Fixes https://github.com/AtB-AS/kundevendt/issues/330


gorandalum:
Changed in Figma so the not-inspectable symbol is at the same place as the inspectable symbol:
Screenshot 2022-08-31 at 15 43 16

gorandalum:
Etter fix:
Screenshot 2022-08-31 at 16 21 27

jorelosorio (approved):
LGTM

rosvik (approved):
Looks good! 💯

PR #2910, opened 2022-09-01
fix: add correct information regarding train/night buses for 24 hour pass

mortennordseth:
Fixes https://github.com/AtB-AS/kundevendt/issues/2247. 

Added missing information regarding train/night buses to the purchase view for 24 hour pass. 
I also extracted the function getTrainTicketNoticeText in a seperate utils file as the same check is done for multiple views (ticket details and purchase view).


rosvik (approved):
💯

PR #2912, opened 2022-09-01
refactor: Compact ticket non inspectable and accessibility fixes

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/1880

### DIFF
--- a/src/screens/Ticketing/Ticket/CompactTicketInfo.tsx
+++ b/src/screens/Ticketing/Ticket/CompactTicketInfo.tsx
@@ -21,10 +21,17 @@ import LoadingSymbol from './Component/LoadingSymbol';
 import * as Sections from '@atb/components/sections';
 import {screenReaderPause} from '@atb/components/accessible-text';
 import InspectionSymbol from '@atb/screens/Ticketing/Ticket/Component/InspectionSymbol';
-import {isInspectable} from '@atb/mobile-token/utils';
+import {
+  findInspectable,
+  getDeviceName,
+  isInspectable,
+  isMobileToken,
+} from '@atb/mobile-token/utils';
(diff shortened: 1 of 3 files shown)


src/screens/Ticketing/Ticket/CompactTicketInfo.tsx
```diff
@@ -148,9 +165,21 @@ export const getTicketInfoTexts = (
     conjunction,
     serialComma: false,
   });
-  const timeUntilExpire = t(TicketTexts.validityHeader.valid(durationText));
-
-  var accessibilityLabel = timeUntilExpire + screenReaderPause;
+  var timeUntilExpireOrInvalid: string;
+
+  if (isMobileToken(inspectableToken) && !isInspectable) {
```
gorandalum, line 170: This isn't ment to be showed if t:card is the inspectable token? 🤔

jorelosorio: The accessibility changes when the token is valid or inspectable on another device. These conditions evaluate that, but do you see something is missing? Please let me know the logic about tickets is a quite complex to follow 😅

gorandalum: It was not regarding accessibility label, but regarding the `timeUntilExpireOrInvalid` value. By just reading it, it seems that if the inspectable token is a t:card, then the normal validity header "Expires in ..." is shown. I don't know if this is intended or not.

jorelosorio: Hey! After our talk I made improvements in the overall code, now it looks slightly better and fixed with the invalid ticket!

gorandalum (approved):

PR #2913, opened 2022-09-01
refactor: Fix wrong text color on non-filled inspection symbol

gorandalum:
Resolves https://github.com/AtB-AS/kundevendt/issues/330

### DIFF
--- a/src/components/theme-icon/theme-icon.tsx
+++ b/src/components/theme-icon/theme-icon.tsx
@@ -23,17 +23,18 @@ const ThemeIcon = ({
   svg,
   colorType,
   size = 'normal',
+  fill,
   ...props
 }: ThemeIconProps): JSX.Element => {
   const {theme, themeName} = useTheme();
 
-  const fill = getFill(theme, themeName, colorType);
+  const fillToUse = fill || getFill(theme, themeName, colorType);
 
   const fontScale = useFontScale();
   const iconSize = theme.icon.size[size] * fontScale;
 
   const settings = {
-    fill,
+    fill: fillToUse,
     height: iconSize,
     width: iconSize,
     ...props,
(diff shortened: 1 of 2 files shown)


src/components/theme-icon/theme-icon.tsx
```diff
@@ -23,17 +23,18 @@ const ThemeIcon = ({
   svg,
   colorType,
   size = 'normal',
+  fill,
   ...props
 }: ThemeIconProps): JSX.Element => {
   const {theme, themeName} = useTheme();
 
-  const fill = getFill(theme, themeName, colorType);
+  const fillToUse = fill || getFill(theme, themeName, colorType);
```
gorandalum, line 31: When using `fill: undefined` on `ThemeIcon` it didn't fallback to the fill based on the theme, but it just used no fill and the icon was not visible.

src/screens/Ticketing/Ticket/Component/InspectionSymbol.tsx
```diff
@@ -76,8 +76,7 @@ const InspectableContent = ({
       <ThemeText
         type="body__primary--bold"
         allowFontScaling={false}
-        style={styles.symbolZones}
-        color={themeColor}
+        color={shouldFill ? themeColor : undefined}
```
jorelosorio, line 79: Instead of undefined could it be the default theme colour?

gorandalum: The `ThemeText` component gives the appropriate text color based on theme, as opposed to just using `Text`, and if necessary the color can be overridden with the `color` parameter. So I don't think it is necessary to specifically send in the default theme color here, as `ThemeText` falls back to using it when the color parameter is undefined.

PR #2916, opened 2022-09-06
refactor: Duplicate requests for departure v2

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/2366

A short explanation of the changes:

Due to many `hooks` being called for filtering favourites, time pickers, and more.. was producing multiple calls to the server end-point at the same time with the `useEffect`, to avoid this the `state` value from  `ReducerWithSideEffects` is compared while `isLoading` (Requesting data) otherwise then proceed to load the data.

### DIFF
--- a/src/screens/Departures/state/stop-place-state.ts
+++ b/src/screens/Departures/state/stop-place-state.ts
@@ -108,6 +108,10 @@ const reducer: ReducerWithSideEffects<
 > = (state, action) => {
   switch (action.type) {
     case 'LOAD_INITIAL_DEPARTURES': {
+      if (state.isLoading == true) {
+        return NoUpdate();
+      }
+
       // Update input data with new date as this
       // is a fresh fetch. We should fetch the latest information.
       const queryInput: QueryInput = {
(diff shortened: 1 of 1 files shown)


src/screens/Departures/state/stop-place-state.ts
```diff
@@ -278,15 +286,15 @@ export function useStopPlaceData(
       }),
     [stopPlace.id, favoriteDepartures, showOnlyFavorites],
   );
-  useEffect(refresh, [stopPlace.id, startTime]);
+  useEffect(dispatchLoadInitialDepartures, [stopPlace.id, startTime]);
```
Nisha1306, line 289: Why are we not using refresh here? Are we not losing the load departures being wrapped in a callback hook because of this?

jorelosorio: I am not the best at using react! 😅. But, I have noticed that `refresh` is a `useCallback` implementation with its own dependencies and is called inside a `useEffect` that also contains its own dependencies and looks a bit odd, according to the documentation https://reactjs.org/docs/hooks-reference.html#usecallback the use of `useCallback` are mainly used to be returned by the child view to trigger updates, so no make sense double call it inside the same child. So because `refresh` is still returned to the parent the hook implementation is not being lost, I have tried and have not seen anything failed or strange but I might be wrong. (Long explanation)

Nisha1306: I think it would be fine here as refresh was anyway not being called if anything other than stopPlace.id or startTime is changing, but probably at the other place you still want to keep refresh. SO its wrapped in a callback and return memoised value unless the refresh dependencies are changed.

src/screens/Departures/state/stop-place-state.ts
```diff
@@ -278,15 +286,15 @@ export function useStopPlaceData(
...
   useEffect(() => {
     if (!state.tick) {
       return;
     }
     const diff = differenceInMinutes(state.tick, state.lastRefreshTime);
 
     if (diff >= HARD_REFRESH_LIMIT_IN_MINUTES) {
-      refresh();
+      dispatchLoadInitialDepartures();
```
Nisha1306, line 297: here i mean :)

src/screens/Departures/state/stop-place-state.ts
Nisha1306: there should be a failed result as a consequence, but only calling expensive functions when they dont need to be I think :)

src/screens/Departures/state/stop-place-state.ts
jorelosorio: Based on our discussion then I see that using the refresh callback in the useEffect is Ok as well, so I have reverted my changes.

Nisha1306 (approved):

PR #2944, opened 2022-09-15
refactor: Add remote config toggle for new frontpage

gorandalum:
Developers may still override this value in the developer menu.

Fixes: https://github.com/AtB-AS/kundevendt/issues/2489


Nisha1306 (approved):

rosvik (approved):
We've used the "dashboard" naming in the code now, and "front page" towards users. Would maybe be more correct with `enable_dashboard` and so on? But I guess this is a somewhat temporary config value, so not a big issue.

Otherwise, code looks ok to me 👍

PR #2945, opened 2022-09-15
feat: add new tab for map in bottom navigator

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2373

**Note:**
1. Controlled by Remote config
2. Falls back to default location which is Trondheim center if user hasnt given location permission to App
3. Have extracted the same MAP component as used in other places
4. Toggle option in developer menu
5. Map is yet not supported for screen reading and large fonts

<div>
  
  
</div>


src/screens/Map/MapScreen.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Map from '@atb/components/map/Map';
+import {useGeolocationState} from '@atb/GeolocationContext';
+import {FOCUS_ORIGIN} from '@atb/api/geocoder';
+
+export function MapScreen() {
+  const {location} = useGeolocationState();
+  const currentLocationCoordinates = location?.coordinates || FOCUS_ORIGIN;
```
Nisha1306, line 8: yet to test with if user hast allowed App to use  current location

src/translations/dictionary.ts
```diff
@@ -14,6 +14,7 @@ const dictionary = {
       `Aktiver for reisesøk. Aktiver to ganger for å nullstille søket.`,
       'Activate for search assistant. Activate twice to reset search.',
     ),
+    map: _(`Kart`, 'Map'),
```
Nisha1306, line 17: would move these translations to a separate relevant named file in a separate PR

Nisha1306:
Need to make slight changes here..hence converting to draft for now.

rosvik (approved):
Looks good 👍

PR #2946, opened 2022-09-15
test: add junit output from e2e tests

tormoseng:
For easy retrieval of e2e results there is a need to create a report after the e2e tests. The reporter `jest-junit` provides a report that can be fetched through GH APIs and post-processed afterwards.


rosvik (approved):
LGTM 👍

PR #2960, opened 2022-09-21
feat: Show aztec code instead of qr code

gorandalum:
Aztec codes is faster and less error-prone to scan for good scanners, and we have been told by Entur that this is the recommended barcode format.

In general we now instead of 'qr code' use the term 'barcode' in the app, as it is probably more understandable than 'aztec code'.

Also tried to simplify the countdown logic, and removed the countdown text from screen readers.

Resolves https://github.com/AtB-AS/kundevendt/issues/1681


rosvik (approved):
LGTM!

PR #2961, opened 2022-09-21
ci: change android build to ubuntu

mikaelbr:
Changes Android builds to use ubuntu-latest and removes some unnecessary running code.

### DIFF
--- a/.github/workflows/build-staging-android.yml
+++ b/.github/workflows/build-staging-android.yml
@@ -48,8 +48,10 @@ jobs:
       - name: Install node_modules
         if: steps.modules-cache.outputs.cache-hit != 'true'
         run: yarn install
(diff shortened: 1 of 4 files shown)


.github/workflows/build-staging-android.yml
```diff
@@ -48,60 +48,62 @@ jobs:
       - name: Install node_modules
         if: steps.modules-cache.outputs.cache-hit != 'true'
         run: yarn install
-      - name: Bundle install, fastlane dependencies
-        run: bundle install
+      - uses: ruby/setup-ruby@v1
```
gorandalum, line 51: Maybe add a name for GH Actions GUI? Something like:
`- name: Download ruby and run bundle install`

gorandalum (approved):

PR #2972, opened 2022-09-23
refactor: refactor ticket overview

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/2495

Refactors ticket purchase overview, centered around `src/screens/Ticketing/Purchase/Overview/index.tsx`. Mostly moving of code around, so this might look like a bigger change than it is.

- `Overview/index.tsx` now mainly handles setting of default values which are then passed to a new component `TicketDetailsSelection`.
- `TicketDetailsSelection` handles state of `useOfferState` parameters, and shows the relevant set of selection components.
- Message boxes for the overview page are moved to their own component: `PurchaseMessages.tsx`.
- Adds `durationSelectionEnabled` value to PurchaseFlow, as well as more explicit handling of product type in `getPurchaseFlow`
- Moves `useTariffZoneFromLocation` to `Ticketing/utils.ts`, since it's used several places in ticketing.


src/screens/Ticketing/Purchase/Overview/index.tsx
```diff
@@ -87,11 +35,13 @@ const PurchaseOverview: React.FC<OverviewProps> = ({
     .filter(productIsSellableInApp)
     .filter((product) => product.type === productType);
 
-  const [preassignedFareProduct, setPreassignedFareProduct] = useState(
-    params.preassignedFareProduct ?? selectableProducts[0],
-  );
+  const preassignedFareProduct =
+    params.preassignedFareProduct ?? selectableProducts[0];
```
rosvik, line 39: Typescript gives this the type `PreassignedFareProduct`. Is it not more accurate with `PreassignedFareProduct | undefined`? It works ok now, but the app was crashing for me while the react navigation updates was ongoing, because preassignedFareProduct was undefined.

src/screens/Ticketing/Purchase/utils.ts
```diff
@@ -16,19 +32,35 @@ export type PurchaseFlow = {
...
+    case 'single':
+      return {
+        travellerSelectionMode: 'multiple',
+        travelDateSelectionEnabled: false,
+      };
+    case 'carnet':
+      return {
+        travellerSelectionMode: 'single',
+        travelDateSelectionEnabled: false,
+      };
```
rosvik, line 63: Carnet is not available for purchase in the app, but added for completeness sake, instead of having a fallback. Also means the same function could be used in the webshop.

src/screens/Ticketing/Purchase/utils.ts
```diff
@@ -14,21 +23,42 @@ export type PurchaseFlow = {
...
+        travellerSelectionMode: 'multiple',
+        travelDateSelectionEnabled: false,
+        durationSelectionEnabled: false,
+      };
+    case 'carnet':
+      return {
+        travellerSelectionMode: 'single',
+        travelDateSelectionEnabled: false,
+        durationSelectionEnabled: false,
+      };
```
rosvik, line 61: Carnet is not available for purchase in the app, but added for completeness sake, instead of having a fallback. Also means the same function could be used in the webshop.

src/screens/Ticketing/Purchase/TariffZones/search/TariffZoneResults.tsx
```diff
@@ -10,7 +10,7 @@ import {TariffZoneSearchTexts, useTranslation} from '@atb/translations';
 import insets from '@atb/utils/insets';
 import React from 'react';
 import {ScrollView, TouchableOpacity, View} from 'react-native';
-import {useTariffZoneFromLocation} from '@atb/screens/Ticketing/Purchase/Overview';
+import {useTariffZoneFromLocation} from '../../utils';
```
mikaelbr, line 13: Think this should be absolute with `@atb...`

src/screens/Ticketing/Purchase/Overview/TicketDetailsSelection.tsx
```diff
@@ -0,0 +1,155 @@
...
+import useOfferState from './use-offer-state';
+
+export type TicketDetailsSelectionProps = {
+  preassignedFareProduct: PreassignedFareProduct;
+  refreshOffer?: boolean;
+  selectableTravellers: UserProfileWithCount[];
+  selectableProductType?: PreassignedFareProductType;
+  fromTariffZone: TariffZoneWithMetadata;
+  toTariffZone: TariffZoneWithMetadata;
+  travelDate?: string;
```
Nisha1306, line 30: Is this and some other props redundant here?

src/translations/screens/subscreens/PurchaseConfirmation.ts
```diff
@@ -46,6 +46,10 @@ const PurchaseConfirmationTexts = {
         `This ticket will be valid from ${time}.`,
       ),
   },
+  travelDate: {
```
Nisha1306, line 49: I see that this is moved from another place. Did we miss the a11y label in the transition?

src/screens/Ticketing/Purchase/Overview/components/PurchaseMessages.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {TariffZoneWithMetadata} from '../../TariffZones';
+
+export type PurchaseWarningsProps = {
+  preassignedFareProduct: PreassignedFareProduct;
+  fromTariffZone: TariffZoneWithMetadata;
+  toTariffZone: TariffZoneWithMetadata;
+};
+
+const PurchaseMessages: React.FC<PurchaseWarningsProps> = ({
+  preassignedFareProduct,
```
Nisha1306, line 25: Do we need to pass complete fare product here? maybe just pass the preassignedFareProduct type instead?

src/screens/Ticketing/Purchase/Overview/components/PurchaseMessages.tsx
```diff
@@ -0,0 +1,103 @@
...
+          message={getTrainTicketNoticeText(t, preassignedFareProduct.type)}
+          type="info"
+        />
+      )}
+    </>
+  );
+};
+
+const shouldShowValidTrainTicketNotice = (
+  PreassignedFareProductType: PreassignedFareProductType,
```
Nisha1306, line 86: should start from a small case :) Also, do we need the method at all or just create a variable in the component itself and calculate it there?

src/screens/Ticketing/Purchase/utils.ts
```diff
@@ -14,21 +21,42 @@ export type PurchaseFlow = {
...
+
+  /**
+   * Whether the customer should be able to select between a set of durations
+   * for the ticket, based on `fareproduct.durationDays`.
+   */
+  durationSelectionEnabled: boolean;
 };
 
 export const getPurchaseFlow = (
   product: PreassignedFareProduct,
```
Nisha1306, line 33: maybe just pass the type here as a prop?

src/screens/Ticketing/Purchase/Overview/TicketDetailsSelection.tsx
```diff
@@ -0,0 +1,155 @@
...
+};
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  selectionComponent: {
+    marginVertical: theme.spacings.medium,
+  },
+  selectionLinks: {margin: theme.spacings.medium},
+  totalSection: {flex: 1, textAlign: 'center'},
+  toPaymentButton: {marginHorizontal: theme.spacings.medium},
+  warning: {
```
Nisha1306, line 148: vi har noen ekstra css her (Testing my Norsk skills simultaneously 😉 )

src/screens/Ticketing/Purchase/Overview/TicketDetailsSelection.tsx
Nisha1306: or maybe we want use it as default is user doesn't update it 🤔

src/screens/Ticketing/Purchase/TariffZones/search/TariffZoneResults.tsx
rosvik: Not sure what the logic is but, VSCode suggests `../../utils` here by default. Do we want to manually update the defaults in cases like these? It's slightly more work if we can't change default settings, but I agree it's more readable with `@atb...`.

gorandalum (changes requested):
Looks pretty good! 👍 

I think there is a "leap" in naming from `Overview` to `TicketDetailsSelection` which is not completely intuitive. The `TicketDetailsSelection` is all the components for the `Overview`, and as such it is pretty much two sides of the same coin and should be named more accordingly.

If possible I think it would be beneficial to do it the other way around. That we have the components (which is now in `TicketDetailsSelection` directly in the `Overview`-prop, but the complex "setup-logic" moved out to a (pretty large) custom hook. I think most often when we are modifying the `Overview`-component it will be the React-components, error messages, etc. that will be changed, and not the setup-logic. Because of this it would be better that the setup-logic is "hidden" from the developer. As it is now when the developer is going to the `Overview`-component he is met by the complex setup-logic, and after reading it they find out that they need to navigate inside the `TicketDetailsSelection`.

rosvik:
@gorandalum 

> If possible I think it would be beneficial to do it the other way around.

Looking at it now, that makes so much sense! On it.

src/translations/screens/subscreens/PurchaseConfirmation.ts
rosvik: I think the a11yHint was just a remnant after we redid the ticketing flow a while back, which is why it was placed in PurchaseOverview instead of PurchaseConfirmation. After the change it's not clickable anymore, which is why it doesn't need a a11yHint.

src/screens/Ticketing/Purchase/utils.ts
rosvik: Good call! Updated now.

src/screens/Ticketing/Purchase/Overview/TicketDetailsSelection.tsx
rosvik: This _was_ true, but I redid all of this now 😅

rosvik:
Your suggestions should be implemented now 😊 @gorandalum @Nisha1306

gorandalum (approved):

PR #2975, opened 2022-09-27
fix: Posisjonsoppdatering

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/2244


src/screens/Departures/FavouriteStopsOverview.tsx
```diff
@@ -53,7 +53,7 @@ const FavouriteStopsOverview = ({navigation}: RootProps) => {
         if (edgeB.node?.distance === undefined) return -1;
         return edgeA.node?.distance > edgeB.node?.distance ? 1 : -1;
       });
-  }, [state.data]);
+  }, [state.data, location]);
```
Nisha1306, line 56: isn't state.data enough here? The func doesn't really depend on the location 🤔

jorelosorio: Location was required to be part of this when the user selects that wants to see the favourites instead, but the distance is based on location, so that is location needed here to be updated as soon it moves.

Nisha1306: Ohh..the getDistanceFromCurrentLocation is using location 👍

Nisha1306 (approved):

PR #2977, opened 2022-09-27
feat: handle rail replacement bus services

toretefre:
Solves https://github.com/AtB-AS/kundevendt/issues/2566

This has not been designed by Christian (but he has agreed on trying it in production), so have a short look at the visual stuff as well :) 

**Features**

Show yellow warning icon indicating a situation in result list:

image

In trip overview, say that trip includes rail replacement bus. And also inform which leg is rail replacement bus:

simulator_screenshot_304DD0A4-3DF7-451A-804A-B9BBE2F5AF03

### DIFF
--- a/src/screens/Assistant/ResultItem.tsx
+++ b/src/screens/Assistant/ResultItem.tsx
@@ -74,6 +74,22 @@ function getFirstQuayName(legs: Leg[]) {
   const publicCodeOutput = fromQuay.publicCode ? ' ' + fromQuay.publicCode : '';
   return fromQuay.name + publicCodeOutput;
 }
+const WarnWhenRailReplacementBus: React.FC<{
+  tripPattern: TripPattern;
+}> = ({tripPattern}) => {
+  const {t} = useTranslation();
+  const styles = useThemeStyles();
+  return tripPattern.legs.some(
+    (leg) => leg.transportSubmode === TransportSubmode.RailReplacementBus,
+  ) ? (
+    <Warning
+      accessibilityLabel={t(
+        AssistantTexts.results.resultItem.tripIncludesRailReplacementBus,
+      )}
+      style={styles.warningIcon}
+    />
+  ) : null;
+};
 const ResultItemHeader: React.FC<{
   tripPattern: TripPattern;
   strikethrough: boolean;
@@ -83,9 +99,6 @@ const ResultItemHeader: React.FC<{
   const durationText = secondsToDurationShort(tripPattern.duration, language);
   const startTime = tripPattern.legs[0].expectedStartTime;
   const endTime = tripPattern.legs[tripPattern.legs.length - 1].expectedEndTime;
-  const tripIncludesRailReplacementBus = tripPattern.legs.some(
-    (leg) => leg.transportSubmode === TransportSubmode.RailReplacementBus,
-  );
 
   return (
     <View style={styles.resultHeader}>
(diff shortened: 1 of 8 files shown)


src/screens/Dashboard/ResultItem.tsx
```diff
@@ -81,6 +83,9 @@ const ResultItemHeader: React.FC<{
   const durationText = secondsToDurationShort(tripPattern.duration, language);
   const startTime = tripPattern.legs[0].expectedStartTime;
   const endTime = tripPattern.legs[tripPattern.legs.length - 1].expectedEndTime;
+  const resultIncludesRailReplacementBus = tripPattern.legs.some(
```
gorandalum, line 86:
```suggestion
  const tripIncludesRailReplacementBus = tripPattern.legs.some(
```

src/screens/Dashboard/ResultItem.tsx
```diff
@@ -113,6 +118,14 @@ const ResultItemHeader: React.FC<{
         </AccessibleText>
       </View>
 
+      {resultIncludesRailReplacementBus && (
```
gorandalum, line 121:
```suggestion
      {tripIncludesRailReplacementBus && (
```

src/screens/TripDetails/components/TripSection.tsx
```diff
@@ -141,6 +142,15 @@ const TripSection: React.FC<TripSectionProps> = ({
             <SituationMessages mode="no-icon" situations={leg.situations} />
           </TripRow>
         )}
+        {leg.transportSubmode === TransportSubmode.RailReplacementBus && (
+          <TripRow rowLabel={<ThemeIcon svg={Warning} />}>
+            <MessageBox type="warning" icon={null}>
+              <ThemeText>
+                {t(TripDetailsTexts.messages.departureIsRailReplacementBus)}
+              </ThemeText>
+            </MessageBox>
```
gorandalum, line 151: I think this should use `message` param instead of `children` to get the styling correct. Is the image in the PR from the actual running app? 🤔 Since I think this will give white text in dark mode.

gorandalum (changes requested):
Also can you include it for the old frontpage, since it will probably be pretty quick to do there as well? There is some talk now about that we might delay the new frontpage because of a departures v1 bug.

src/screens/TripDetails/components/TripSection.tsx
toretefre: Good catch, had not commited line 149 for some reason, you're absolutely correct. Anyways, I will update to using the message param. 

image

toretefre:
> Also can you include it for the old frontpage, since it will probably be pretty quick to do there as well? There is some talk now about that we might delay the new frontpage because of a departures v1 bug.

Good idea, didn't realize these were separate actually. Seems to be working there as well now.

src/screens/Assistant/ResultItem.tsx
```diff
@@ -113,6 +118,15 @@ const ResultItemHeader: React.FC<{
...
       </View>
 
+      {tripIncludesRailReplacementBus && (
+        <Warning
+          accessibilityLabel={t(
+            AssistantTexts.results.resultItem.tripIncludesRailReplacementBus,
+          )}
+          style={styles.warningIcon}
+        />
+      )}
```
mikaelbr, line 128: Things like this could also be it's own component where the logic is encapsulated. Naming it something like `WarnWhenRailReplacementBus` or something.

toretefre: I don't think it necessarily makes things less complex, but at least it makes the return function cleaner. Did not extract it into a separate file or something, as the `Assistant` version is to be removed some day, and the texts are fetched from two different text files. 

Was this in line with what you had in mind, @mikaelbr?

mikaelbr: Does it need to have separate texts? My point is, _is_ it the same or does it just look the same? Translated texts are an abstraction and not semantically difference. If it _is_ the same, you can have a component shared across and have that component it's own text file.

toretefre: I've extracted the encapsulated logic where the texts actually are the same, and put it into a separate component structure. Starting to feel a bit like over-engineering something quite simple, but the code looks cleaner in the original file at least.

rosvik (approved):
LGTM 👍

PR #2978, opened 2022-09-28
refactor: improve screen reader and rejected ticket details UI

Nisha1306:
Closes comments on https://github.com/AtB-AS/kundevendt/issues/2090

1. Made the view accessible so that screen reader takes all the three details together
2. Updated view to be in sync with the one in ticket details screen

Now:
  

Before:


tormoseng (approved):
LGTM!

PR #2979, opened 2022-10-03
Update build-store-ios.yml

jorelosorio:
Github actions are failing at https://github.com/AtB-AS/mittatb-app/actions/runs/3158145874/jobs/5142031779#step:10:23
which means there is something wrong with the library https://github.com/bruceadams/get-release.

I saw this library also being used on android without issues, however, for iOS it was there but non-used.

NOTE: Recently there was an upgrade in the OS version, it might break the library in the iOS build, but based on what the library does and the usage it has (none) is safe to remove.

This PR does:

- Remove unused action that among other things is causing a fail.


mikaelbr (approved):
Would be nice to pass in release data again at some point, but OK enough

PR #2983, opened 2022-10-05
feat: disable map when using screen reader

rosvik:
resolves https://github.com/AtB-AS/kundevendt/issues/2474

Disables map page when screen reader is enabled, and shows this screen instead:

collage

### DIFF
--- a/src/screens/Map/components/MapDisabledForScreenReader.tsx
+++ b/src/screens/Map/components/MapDisabledForScreenReader.tsx
@@ -1,0 +1,34 @@
+import ThemeText from '@atb/components/text';
+import {StyleSheet} from '@atb/theme';
+import {SafeAreaView} from 'react-native-safe-area-context';
+import {Map as MapImage} from '@atb/assets/svg/color/images';
+import {MapTexts, useTranslation} from '@atb/translations';
+
+export function MapDisabledForScreenReader() {
+  const styles = useStyles();
+  const {t} = useTranslation();
+  return (
+    <SafeAreaView style={styles.container}>
+      <MapImage style={styles.illustration} />
+      <ThemeText type="body__primary--bold" style={styles.header}>
+        {t(MapTexts.disabledForScreenReader.title)}
+      </ThemeText>
+      <ThemeText>{t(MapTexts.disabledForScreenReader.description)}</ThemeText>
+    </SafeAreaView>
+  );
+}
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {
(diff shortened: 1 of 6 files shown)


rosvik:
I noticed that in light mode, the status bar becomes invisible towards the background. This is also a problem for the map screen in general, where a dark status bar would give more contrast towards the map. 



Do we have any option to set the status bar color?  I tried adding

```tsx
<StatusBar barStyle="dark-content"></StatusBar>
```

to the map page, but this seemed to set the color globally.

src/screens/Map/components/MapDisabledForScreenReader.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const {t} = useTranslation();
+  return (
+    <SafeAreaView style={styles.container}>
+      <ThemedMapImage />
+      <ThemeText type="body__primary--bold" style={styles.header}>
+        {t(MapTexts.disabledForScreenReader.title)}
+      </ThemeText>
+      <ThemeText style={styles.description}>
+        {t(MapTexts.disabledForScreenReader.description)}
+      </ThemeText>
```
gorandalum, line 18: I'll suggest that we add an accessible `View` around all this, with an a11y label reading both the title and description.

gorandalum (dismissed):
I haven't tested this, so it might be working ok, but even though I think it is better to handle this more explicitly.

Normally when people with screen reader select a tab the focus is changed to the header title. On this page there is no header, so we should probably change the focus to the message about the map not being enabled for screen reader users. This can be done with the `useFocusOnLoad` hook.

src/screens/Map/components/MapDisabledForScreenReader.tsx
rosvik: Updated here 1c95aae72a955775f427cf5e021c2fd92c5dd842.

gorandalum: Nice. Tested on Android?

rosvik: I don't have a device to test accessibility on android. Is there any known differences between iOS and android here that I should be aware of?

gorandalum: Yes 😅 Should try to borrow one from Tor.

rosvik: Works well on Android ✅

tormoseng (approved):
LGTM!

PR #2987, opened 2022-10-05
fix: conditionally navigate to departures v2 from TripDetails

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/2615


toretefre (approved):

toretefre:
Tested fine, code looks gode as well 👌

PR #2988, opened 2022-10-05
fix: reset dashboard travel search when navigating

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/2576

This resets the travel search in dashboard at the same time as navigating to the trip search screen. This means we don't have to bother with reseting the search field when navigating back to the dashboard  (in cases like swiping back / android back button / double tapping the navbar).


mikaelbr (approved):
It's something iffy with updating the state we are reacting on, but after discussing this with @rosvik and trying to find any edge-cases it seems this would work fine. Might be good to make a separate `reset` function

PR #2993, opened 2022-10-10
fix: use stopPlace param when navigating from TripDetails

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/481

This solves the "departures v1-part" of https://github.com/AtB-AS/mittatb-app/pull/2990. This is done by supporting `StopPlace` as param for location in departures v1.

This also removes the need for `getFeatureFromVenue()`, and the `bff/v1/geocoder/features` endpoint, which is marked as deprecated in https://github.com/AtB-AS/atb-bff/pull/168.


src/departure-list/DeparturesList.tsx
```diff
@@ -31,7 +32,7 @@ export default function DeparturesList({
   isLoading = false,
   error,
   isInitialScreen,
-  currentLocation,
```
gorandalum, line 34: Try to avoid `currentLocation` here, as the term is mainly used throughout the app for the current location of the app.

src/screens/Nearby/state.ts
```diff
@@ -439,3 +434,28 @@ export function useDepartureData(
...
+    return locationOrStopPlace.resultType === 'search'
+      ? locationOrStopPlace
+      : {
+          layer: 'address',
+          coordinates: locationOrStopPlace.coordinates,
+        };
+  }
+
+  // Should not be possible, but for typing there needs to be a fallback
+  return {id: '', layer: 'venue'};
```
gorandalum, line 460: Can't this be avoided by just using 
```
if ('latitude' in locationOrStopPlace) {
    ...
} else {
    ...  
}

gorandalum (approved):
Looks good! Some minor comments you can take a look at.

src/screens/Nearby/state.ts
rosvik: I though so as well, but TypeScript doesn't seem to realize that it `locationOrStopPlace` has the `Location` type in that case.

Screenshot 2022-10-11 at 09 41 46

src/departure-list/DeparturesList.tsx
rosvik: Ah, think I got them all now 1a9825ef75144d93f94bad32fb243104c575d920.

src/screens/Nearby/state.ts
gorandalum: It's because `latitude` is optional in `StopPlace`. Check on `resultType` instead, which is only in `Location`.

rosvik: Aha 💡 Updated here: https://github.com/AtB-AS/mittatb-app/pull/2997

PR #2998, opened 2022-10-11
fix: update searchTime on refresh of trip search

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/2181

When search time "now" is selected, refreshing the Assistant (or TripSearch when front page is enabled) will also update the searchTime date.


gorandalum (approved):
Looks good! Longer term I think this is the same as I mentioned regarding the geolocation, that the object shouldn't actually include the time when option is `now`. The date should be added when the search is actually performed.

However, this will affect more places, so not something we should do now.

PR #3002, opened 2022-10-12
feat: show toggle limit when user tries to toggle token

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2511

1. Calls abt-token-server API (PR: https://github.com/AtB-AS/abt-token-server/pull/32) to get toggle limit for the user.
2. Max limit is configured in firestore ;  configuration => other => tokenToggleMaxLimit
3. If tokenToggleMaxLimit is not set in firestore no limitation would be applied on toggle
4. Added new point in the FAQ regarding token toggle.

<div>
  
  

</div>

### DIFF
--- a/src/screens/Profile/TravelToken/index.tsx
+++ b/src/screens/Profile/TravelToken/index.tsx
@@ -2,7 +2,7 @@ import FullScreenHeader from '@atb/components/screen-header/full-header';
 import {StyleSheet, Theme} from '@atb/theme';
 import {TravelTokenTexts, useTranslation} from '@atb/translations';
 import TravelTokenBox from '@atb/travel-token-box';
-import React, {useEffect, useState} from 'react';
+import React from 'react';
 import {View} from 'react-native';
 import {ScrollView} from 'react-native-gesture-handler';
 import {ProfileScreenProps} from '../types';
@@ -20,7 +20,7 @@ export default function TravelCard({navigation}: TravelCardScreenProps) {
   const {isError, isLoading} = useMobileTokenContextState();
   const screenHasFocus = useIsFocused();
   const shouldFetchTokenDetails = screenHasFocus && !isError && !isLoading;
-  const {showLoader, toggleLimit, maxToggleLimit} = useTokenToggleDetails(
+  const {shouldShowLoader, toggleLimit, maxToggleLimit} = useTokenToggleDetails(
     shouldFetchTokenDetails,
   );
   return (
@@ -34,7 +34,7 @@ export default function TravelCard({navigation}: TravelCardScreenProps) {
         <ChangeTokenAction
           onChange={() => navigation.navigate('SelectTravelToken')}
           toggleLimit={toggleLimit}
-          showLoader={showLoader}
+          shouldShowLoader={shouldShowLoader}
         />
         <FaqSection toggleMaxLimit={maxToggleLimit} />
       </ScrollView>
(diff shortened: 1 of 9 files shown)


src/screens/Profile/TravelToken/index.tsx
```diff
@@ -1,22 +1,46 @@
...
 export default function TravelCard({navigation}: TravelCardScreenProps) {
   const styles = useStyles();
   const {t} = useTranslation();
+  const [showLoader, setShowLoader] = useState<boolean>(false);
+  const [toggleLimit, setToggleLimit] = useState<number | undefined>();
+  const [maxToggleLimit, setMaxToggleLimit] = useState<number | undefined>();
+
+  const {isError, isLoading, getTokenToggleDetails} =
+    useMobileTokenContextState();
+  const screenHasFocus = useIsFocused();
```
gorandalum, line 25: Maybe more correct to use [`useFocusEffect`](https://reactnavigation.org/docs/use-focus-effect) instead of [`useIsFocused`](https://reactnavigation.org/docs/use-is-focused) here, since the `screenHasFocus` triggers a `useEffect`.

src/screens/Profile/TravelToken/index.tsx
```diff
@@ -1,22 +1,46 @@
...
+          setToggleLimit(toggleMaxLimit - toggledCount);
+        }
+        setMaxToggleLimit(toggleMaxLimit);
+      }
+      setShowLoader(false);
+    };
+    if (!isError && !isLoading) {
+      fetchToggleLimit();
+    }
+  }, [getTokenToggleDetails, screenHasFocus]);
```
gorandalum, line 43: Maybe extract this is a custom hook, making the `TravelTokenScreen` a little less complex, and the developer can themselves figure out if they need/want to deep dive into the custom hook to see the implementation for fetching the toggle details.

gorandalum (approved):

src/screens/Profile/TravelToken/index.tsx
Nisha1306:
```
  const shouldFetchTokenDetails = !isError && !isLoading;
  const [showLoader, setShowLoader] = useState<boolean>(false);
  const [toggleLimit, setToggleLimit] = useState<number | undefined>();
  const [maxToggleLimit, setMaxToggleLimit] = useState<number | undefined>();

  useFocusEffect(useCallback(
      () => {
        const {showLoader, toggleLimit, maxToggleLimit} = useTokenToggleDetails(
            shouldFetchTokenDetails,
        );
        setShowLoader(showLoader)
        setMaxToggleLimit(maxToggleLimit)
        setToggleLimit(toggleLimit)
      },
      [shouldFetchTokenDetails],
  ))
```

I tried to change it. This will now not re-render the component on every focus, but will still do when shouldFetchTokenDetails changes which I believe should not change a lot. Although when we change to useFocusEffect, we have to maintain the state here as well along with the hook so the code becomes slightly verbose rather than what it is now.  How do you think it looks here ☝️  vs what it is now?

gorandalum: No worries. Keep it at it is!

gorandalum (approved):
Looks good! 

One thing you can contemplate to change is the naming `showLoader`. I searched the codebase and found no other uses of `showLoader`, but multiple uses of `isLoading`. The naming `isLoading` is also more inline with that the variable is a representation of the state, and not a function as one might think when the variable starts with the verb "show".

Nisha1306:
> showLoader

The reason I used showLoader is that earlier this logic was in the ChangeTokenAction initially and there was already isLoading from useMobileTokenContextState in the scope. But now when I have moved it somewhere else, I can use isLoading there, but would leave the prop name in ChangeTokenAction as **shouldShowLoader** :)

PR #3004, opened 2022-10-12
feat: Clicking a bus stop should open a bottom sheet with departures v2

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/2375


src/components/bottom-sheet/BottomSheetContainer.tsx
```diff
@@ -1,15 +1,27 @@
 import React, {ReactNode} from 'react';
+import {ViewStyle} from 'react-native';
 import {useWindowDimensions, View} from 'react-native';
 
+export enum BottomSheetSize {
+  auto = 0,
+  compact = 0.5,
+  cover = 0.8,
+}
```
gorandalum, line 9: Can't we just have a parameter `maxHeight` as a number parameter instead? I think this is more intuitive to understand when developing.

src/components/bottom-sheet/AnimatedBottomSheet.tsx
```diff
@@ -1,14 +1,21 @@
...
 
 export default function AnimatedBottomSheet({
   animatedOffset,
   children,
   onLayout,
+  style,
 }: {
   animatedOffset: Animated.Value;
   children: ReactNode;
+  style?: ViewStyle;
```
gorandalum, line 18: If we can avoid it, I would like to not be able to pass `style` from outside. This makes it easier to maintain the same appearance for all bottom-sheets in the app.

src/components/map/DeparturesDialogSheet.tsx
```diff
@@ -0,0 +1,124 @@
...
+    padding: theme.spacings.medium,
+  },
+  title: {
+    marginLeft: theme.spacings.medium,
+    marginBottom: theme.spacings.medium,
+  },
+  roundEdgesOnTop: {
+    borderTopLeftRadius: theme.border.radius.circle,
+    borderTopRightRadius: theme.border.radius.circle,
+  },
```
gorandalum, line 121: All bottom-sheets in the app should look the same. If we want round edges, then do it somewhere it effects all bottom sheets.

src/components/map/Map.tsx
```diff
@@ -81,8 +116,53 @@ const Map = ({initialLocation, selectionMode, onLocationSelect}: MapProps) => {
...
+                        navigateToQuay(stopPlace, quay);
+                      }}
+                      ref={focusRef}
+                    />
+                  ),
+                  closeRef,
+                  false,
+                  bottomSheetStyles.roundEdgesOnTop,
+                );
+              }
```
gorandalum, line 165: This is too long to have inline in a `onClick` handler. Also we have a hook called `use-selected-feature-change-effect`, which is triggered by the `onMapClick`. Opening the bottom sheet is also an effect which is triggered by a map click, so I think it fits into that hook.

src/components/map/use-selected-feature-change-effect.tsx
```diff
@@ -73,7 +80,16 @@ const useSelectedFeatureChangeEffect = (
...
+                const bbox = getBoundingBox(mapLines);
+                const northEast: Coordinates = {
+                  longitude: bbox.xMin,
+                  latitude: bbox.yMin,
+                };
+                const southWest: Coordinates = {
+                  longitude: bbox.xMax,
+                  latitude: bbox.yMax,
+                };
+                fitBounds(northEast, southWest, mapCameraRef, padding);
```
gorandalum, line 92: Nice, this seems to be working correctly. I think we should do something like this also when pressing the `My location` button, since now it centres the my location to the middle of the screen, which is pretty much on the border between map and bottom-sheet.

src/components/screen-header/HeaderButton.tsx
```diff
@@ -39,21 +39,23 @@ const HeaderButton: React.FC<HeaderButtonProps> = (buttonProps) => {
 };
 
 export type HeaderButtonWithoutNavigationProps = {
-  text: string;
+  text?: string;
+  icon?: React.ReactNode;
```
gorandalum, line 43: Sometime ago it was decided that we always use text in headers in stead of icons. Talked with Christian, and it was decided we still follow this rule also for the departures bottom sheet. He will update the sketches to use "Close"/"Lukk" instead.

gorandalum (changes requested):
Some comments. Also when using I found that it will be necsessary to open the bottom sheet again when navigating back from `DepartureDetails` or `PlaceScreen`. Will probably be a little hassle to get it to work, but I think it is necessary.

src/components/bottom-sheet/BottomSheetContainer.tsx
jorelosorio: It is fixed to just use `max-height value` as a per cent value of the screen.

jorelosorio: Fixed

src/components/bottom-sheet/AnimatedBottomSheet.tsx
jorelosorio: Fixed

src/components/map/DeparturesDialogSheet.tsx
jorelosorio: Fixed

src/components/map/Map.tsx
jorelosorio: @gorandalum do you mean, to move all this inside onMapClick?

src/components/map/use-selected-feature-change-effect.tsx
jorelosorio: Sure, I will do it for other places but perhaps in another task.

src/components/map/Map.tsx
jorelosorio: I have moved to another place using hooks, let me know what do you think ;)

gorandalum (approved):

PR #3008, opened 2022-10-18
feat: set status bar content color on map screen

rosvik:
Follow up to https://github.com/AtB-AS/kundevendt/issues/2474

In order to make the status bar more visible on the map page, we need to update the status bar content color when one of the screens are focused. My suggestion is to create a `StatusBarOnFocus` component which wraps `useIsFocused`, but not sure it's the best solution, so open to feedback.

https://user-images.githubusercontent.com/1774972/195813383-ef1e154c-d485-4801-ab44-d319b6a2b486.mp4


gorandalum (approved):

HanneLH:
Tested on Android and iOS ✅

PR #3010, opened 2022-10-19
build: Update register scripts

gorandalum:
The field 'application_version_id' is changed to 'version_name'. In
addition we start using the field 'version_code' for Android, which is
necessary for Play Integrity.

Resolves https://github.com/AtB-AS/kundevendt/issues/2687


Nisha1306 (approved):
✅

rosvik (approved):

PR #3011, opened 2022-10-19
feat: sort dashboard departures by distance

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/2595

Sorts dashboard favorite departures by distance to the user geolocation. If location permission is not given, the app keeps the default order.

### DIFF
--- a/src/screens/Dashboard/DeparturesWidget.tsx
+++ b/src/screens/Dashboard/DeparturesWidget.tsx
@@ -40,7 +40,7 @@ const DeparturesWidget: React.FC = () => {
 
   const sortedStopPlaceGroups = location
     ? state.data?.sort((a, b) =>
-        sortStopsByDistance(a.stopPlace, b.stopPlace, location.coordinates),
+        compareStopsByDistance(a.stopPlace, b.stopPlace, location.coordinates),
       )
     : state.data;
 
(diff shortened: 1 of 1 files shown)


src/screens/Dashboard/DeparturesWidget.tsx
```diff
@@ -120,6 +124,29 @@ const DeparturesWidget: React.FC = () => {
   );
 };
 
+function sortStopsByDistance(
```
gorandalum, line 127: This function does not sort, so I think something like `compareStopsByDistance` would be better.

src/screens/Dashboard/DeparturesWidget.tsx
```diff
@@ -120,6 +124,29 @@ const DeparturesWidget: React.FC = () => {
...
+    geolocation,
+  );
+  const distanceToB = haversineDistance(
+    {lat: b.latitude, lon: b.longitude},
+    geolocation,
+  );
+
+  if (distanceToA > distanceToB) return 1;
+  if (distanceToA < distanceToB) return -1;
+  return 0;
```
gorandalum, line 147: Not necessary to change, just FYI. The compare function logic is dependant on if the returned value is positive, negative or zero. Using `1`, `0` and `-1` is most normal when comparing non-number values. Here where we have number values it can just be `return distanceToA - distanceToB`.

gorandalum (approved):

src/screens/Dashboard/DeparturesWidget.tsx
rosvik: That makes a lot more sense 😅 Updated.

PR #3012, opened 2022-10-20
build: Adds a fix to the keychain for Github Actions

jorelosorio:
Short explanation: The keychain that stores the certificates for signing is expired before the compilation ends, causing an endless loop in the linking process. Ensuring more time `timeout` in the creation guarantee to be available.


mikaelbr:
Does this mean that when it takes time it just goes on forever? is it possible to instead get it to explode? Other than that I can't think of any reason to not increase the timeout here.

mikaelbr (approved):

PR #3022, opened 2022-10-25
Revert "fix: upgrade theme and assets to remove nfk build error"

toretefre:
Android staging build fails: https://github.com/AtB-AS/mittatb-app/actions/runs/3315306866/jobs/5482349523#step:16:2472

This probably won't change a thing, but it may be worth testing to make sure these aren't the reason for the issues.

Reverts AtB-AS/mittatb-app#3020

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "link-assets": "react-native link"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


gorandalum:
This should just be closed @toretefre ?

toretefre:
Yes, it was unrelated to the issues.

PR #3026, opened 2022-10-26
refactor: Map departures fixes

jorelosorio:
This PR fixes all the issues mentioned on https://github.com/AtB-AS/kundevendt/issues/2375


src/components/bottom-sheet/AnimatedBottomSheet.tsx
```diff
@@ -45,7 +45,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     width: '100%',
     position: 'absolute',
     bottom: 0,
-    borderTopLeftRadius: theme.border.radius.regular,
-    borderTopRightRadius: theme.border.radius.regular,
+    borderTopLeftRadius: theme.border.radius.circle,
+    borderTopRightRadius: theme.border.radius.circle,
```
rosvik, line 49: Looking through the issue comments, didn't you settle on keeping the previous border radius here?

src/components/map/hooks/use-trigger-camera-move-effect.tsx
```diff
@@ -40,7 +41,13 @@ export const useTriggerCameraMoveEffect = (
...
   const {height: bottomSheetHeight} = useBottomSheet();
   const {minHeight: tabBarMinHeight} = useBottomNavigationStyles();
-  const paddingBottom = bottomSheetHeight - tabBarMinHeight;
+  var paddingBottom = bottomSheetHeight - tabBarMinHeight;
+
+  paddingBottom =
+    // Workaround for Android when font and screen size changes!
+    Platform.OS == 'android' && PixelRatio.get() > 3.0
+      ? paddingBottom / PixelRatio.get()
+      : paddingBottom;
```
rosvik, line 50: This looks kind of hacky 🤔 Is there something special about pixel ration being > 3, or could we solve it in a more generic way? Would it work to always divide by pixel ratio, regardless of amount?

But if this is the best solution, this might be a more readable way to write this 😊

```suggestion
  let paddingBottom = bottomSheetHeight - tabBarMinHeight;

  // Workaround for Android when font and screen size changes
  if (Platform.OS == 'android' && PixelRatio.get() > 3.0) {
    paddingBottom = paddingBottom / PixelRatio.get();
  }
```

src/components/map/utils.ts
```diff
@@ -87,14 +96,22 @@ export const getFeaturesAtCoordinates = async (
     coords.longitude,
     coords.latitude,
   ]);
+  var boundingOffset = 15;
```
rosvik, line 99:
```suggestion
  let boundingOffset = 15;
```

src/components/screen-header/index.tsx
```diff
@@ -170,9 +170,12 @@ export default ScreenHeader;
 
 const useHeaderStyle = StyleSheet.createThemeHook((theme) => ({
   container: {
-    padding: theme.spacings.medium,
-    borderTopLeftRadius: theme.border.radius.regular,
-    borderTopRightRadius: theme.border.radius.regular,
+    paddingLeft: theme.spacings.medium,
+    paddingRight: theme.spacings.medium,
+    paddingBottom: theme.spacings.large,
+    paddingTop: theme.spacings.large,
```
rosvik, line 176:
```suggestion
    paddingHorizontal: theme.spacings.medium,
    paddingVertical: theme.spacings.large,
```

src/components/bottom-sheet/AnimatedBottomSheet.tsx
jorelosorio: Yes, however, I showed it to @ckbilstad and he said we can use that borders, even for other bottom sheets.

src/components/map/hooks/use-trigger-camera-move-effect.tsx
jorelosorio: Yes, it looks hacky! Indeed! but after many tests, I realize that Android breaks when PixelRatio is bigger than 3.0, I could show you, and maybe together we could find something better :)

src/components/map/components/DeparturesDialogSheet.tsx
```diff
@@ -58,28 +70,34 @@ const DeparturesDialogSheet = ({
...
-            isFocused={false}
-            searchTime={searchTime}
-            setSearchTime={setSearchTime}
-            showOnlyFavorites={showOnlyFavorites}
-            setShowOnlyFavorites={setShowOnlyFavorites}
-            testID="departuresContentView"
-            allowFavouriteSelection={false}
-          />
+        {!isSearching && !isLoading ? (
+          stopPlace && (stopPlace.quays?.length ?? 0 > 0) ? (
```
rosvik, line 78: I think we can make the logic more readable here. 

1. Update `isLoading` to `const isLoading = state.isLoading || isSearching;`, so you can check just the one variable here.
2. `stopPlace && (stopPlace.quays?.length ?? 0 > 0)` can be simplified to `!!stopPlace?.quays?.length`

rosvik (approved):
Looks good 👍 Just one more comment I missed the first time around.

src/components/map/hooks/use-trigger-camera-move-effect.tsx
gorandalum: We looked at it today and got it working without the "magic number".

gorandalum (approved):
Remember to not use "fix" for fixes on unreleased features. This hasn't been released to the users yet, so it is not relevant as a bugfix in the release notes. Instead use "refactor", you may change it when you squash and merge.

src/components/map/components/DeparturesDialogSheet.tsx
jorelosorio: Fixed!

src/components/map/hooks/use-trigger-camera-move-effect.tsx
```diff
@@ -158,10 +148,32 @@ export const moveCameraToCoordinate = (
...
+    const headerHeight = StatusBar.currentHeight ?? 0;
+
+    // This is the base level for the padding, where the values are divided by the device font scale to match the map unit.
+    const scaledBottomSheetPadding = bottomSheetHeight - tabBarMinHeight;
+    return [
+      padding + headerHeight, // Subtract the header height on Android since it is not see through
+      padding,
+      padding + scaledBottomSheetPadding,
+      padding,
+    ].map((p) => p / PixelRatio.getFontScale()) as MapboxGL.Padding;
```
adriawh, line 170: @jorelosorio do you remember why you are dividing on the font scale here?

jorelosorio: As far I remember, the scales inside the map are not based on dpi so by using / by that factor becomes somehow pixel units, not point units.

PR #3028, opened 2022-10-27
feat: improve favourites in the app

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2592
Dependent on https://github.com/AtB-AS/atb-bff/pull/171

<div>
  
  
  
  
  
  
</div>


src/screens/Departures/StopPlaceView.tsx
```diff
@@ -3,20 +3,24 @@ import Feedback from '@atb/components/feedback';
...
+import FavoriteToggle from '@atb/screens/Departures/components/FavoriteToggle';
+import DateSelection from '@atb/screens/Departures/components/DateSelection';
+import {PlaceScreenMode} from '@atb/screens/Departures/PlaceScreen';
 
 type StopPlaceViewProps = {
   stopPlace: Place;
   showTimeNavigation?: boolean;
   navigateToQuay: (quay: Quay) => void;
-  navigateToDetails: (
+  navigateToDetails?: (
```
gorandalum, line 23: `navigateToDetails` is mandatory if `mode` is `'Departure', but should not be provided if `mode` is `Favourite`. It would be beneficial if this was reflected in the typing as well.

src/screens/Departures/components/EstimatedCallItemWrapperView.tsx
```diff
@@ -0,0 +1,33 @@
+import {PlaceScreenMode} from '@atb/screens/Departures/PlaceScreen';
+import {TouchableOpacity, View} from 'react-native';
+import {StyleSheet} from '@atb/theme';
+import {ReactNode} from 'react';
+
+const EstimatedCallItemWrapperView = ({
```
gorandalum, line 6: Could consider making this a more generic component, as I feel the naming `EstimatedCallItemWrapperView` is kinda hard to understand, and a developer have to read the code to get an idea what it actually does. 

Maybe something like:
```
const TouchableOrView = ({onClick, style, children}) => 
  onClick
    ? <TouchableOpacity onPress={onClick} style={style}>{children}</TouchableOpacity>
    : <View style={style}>{children}</View>
```

You don't necessarily have to do this now, we don't want PRs to drag out forever. You can decide based on how you feel about the suggestion 😄

src/screens/Departures/state/quay-state.ts
```diff
@@ -355,7 +362,7 @@ async function fetchEstimatedCalls(
       id: quay.id,
       startTime: queryInput.startTime,
       numberOfDepartures: queryInput.numberOfDepartures,
-      timeRange: timeRange,
+      timeRange: queryInput.timeRange,
```
gorandalum, line 365: Remove the existing `timeRange` variable?

gorandalum (approved):

src/screens/Departures/state/quay-state.ts
Nisha1306: I dint know that you had commented,  I have force pushed this fix to previous commit itself 😅

src/screens/Departures/components/EstimatedCallItemWrapperView.tsx
Nisha1306: I did think about this, but then I couldn't think of any scenario where this would be re-used. So I decided to make it component-specific, but from the perspective of how intuitive it is, I can rename it :)

src/screens/Dashboard/DeparturesWidget.tsx
```diff
@@ -10,15 +10,26 @@ import {useGeolocationState} from '@atb/GeolocationContext';
...
 import React, {useEffect, useRef} from 'react';
 import {ActivityIndicator, Linking, TouchableOpacity, View} from 'react-native';
 import {useFavoriteDepartureData} from './state';
+import * as Sections from '@atb/components/sections';
+import ThemeIcon from '@atb/components/theme-icon';
+import {NavigationProp} from '@react-navigation/native';
 
-const DeparturesWidget: React.FC = () => {
+type Props = {
+  navigation: NavigationProp<any>;
```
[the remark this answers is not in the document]
Nisha1306: passed onEditFavouriteDeparture and onAddFavouriteDeparture instead of navigation object

src/screens/Departures/components/EstimatedCallItem.tsx
```diff
@@ -23,20 +23,24 @@ import {Warning} from '../../../assets/svg/color/situations';
 import ToggleFavouriteDeparture from '@atb/screens/Departures/components/ToggleFavouriteDeparture';
 import DeparturesTexts from '@atb/translations/screens/Departures';
 import {isToday, parseISO} from 'date-fns';
+import {useOnMarkFavouriteDepartures} from '@atb/screens/Departures/components/use-on-mark-favourite-departures';
+import {StopPlacesMode} from '@atb/screens/Departures/types';
+import {TouchableOpacityOrView} from '@atb/components/touchable-opacity-or-view/TouchableOpacityOrView';
```
gorandalum, line 28: Shouldn't import directly from file. Export it in an index file, and then use 
```import {TouchableOpacityOrView} from '@atb/components/touchable-opacity-or-view```

gorandalum (approved):

PR #3031, opened 2022-10-28
refactor: Make location-search module up-to-date with new coding guidelines

gorandalum:
- Naming with correct suffixes
- Only accessed through index.ts
- Navigation/Routing types and helpers in its own file ´navigation-types.ts`

Inline with the [coding guidelines](https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md).


rosvik (approved):
LGTM 👍

HanneLH:
Tested on Android and iOS and the Location search work as before ✅

PR #3032, opened 2022-10-31
feature: add more filters to global messages

Nisha1306:
Closes https://github.com/AtB-AS/kundevendt/issues/2326

Includes:

 - Filtering global messages by app version
 - Filtering global messages by app Platform i.e. ios, android
 - Some refactoring
 
image


src/global-messages/converters.ts
```diff
@@ -59,6 +73,11 @@ function mapToContexts(data: any): GlobalMessageContext[] | undefined {
     .filter(Boolean) as GlobalMessageContext[];
 }
 
+function isAppPlatformValid(platforms: AppPlatformType[]) {
+  if (!platforms) return true;
+  return !!platforms.find((platform) => platform === Platform.OS);
```
Nisha1306, line 78: Wondering if we should make this comparison case insensitive. wdut?

gorandalum: Yeah, seems like a good idea :+1:

src/global-messages/types.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type GlobalMessageType = {
+  id: string;
+  active: boolean;
+  title?: LanguageAndText[];
+  body: LanguageAndText[];
+  type: Statuses;
+  context: GlobalMessageContext[];
+  isDismissable?: boolean;
+};
```
gorandalum, line 34: I think it would be slightly better if this type is derived from `GlobalMessageRaw`. You can look into `Omit`.

src/global-messages/types.ts
```diff
@@ -0,0 +1,34 @@
+import {LanguageAndText} from '@atb/reference-data/types';
+import {Statuses} from '@atb/theme';
+
+export type AppPlatformType = 'ios' | 'android';
+
+export type GlobalMessageContext =
```
gorandalum, line 6: I think it would be better if this is named `GlobalMessageContextType`. This would also make it less likely to be confused with the `GlobalMessagesContext`.

gorandalum (changes requested):
Nice! Just some small comments.

Also good that some improvements are done to the module, like the `GlobalMessage`/`GlobalMessageType` naming. When you are at it, I think you can do the last changes necessary for making this an "up-to-date" module with the latest guidelines. So make an `index.ts` file and make sure all imports are through `@atb/global-messages`.

Also remember to update the documentation here:
https://github.com/AtB-AS/docs-private/blob/main/How-to-update-info-messages-in-app.md

rosvik:
If I understand this correctly, all of the messages with device and version numbers set will still be shown on older versions of the app (Less than v1.28, if this becomes part of the next release). Should we try to make it possible to include/exclude these versions?

src/global-messages/converters.ts
```diff
@@ -1,32 +1,46 @@
...
   if (!result) return;
 
   const body = mapToLanguageAndTexts(result.body);
   const title = mapToLanguageAndTexts(result.title);
   const context = mapToContexts(result.context);
   const type = mapToMessageType(result.type);
   const isDismissable = result.isDismissable;
+  const appVersionMin = result.appVersionMin;
+  const appVersionMax = result.appVersionMax;
+  const platforms = result.appPlatform;
```
gorandalum, line 37: `appPlatforms`?

Nisha1306:
> If I understand this correctly, all of the messages with device and version numbers set will still be shown on older versions of the app (Less than v1.28, if this becomes part of the next release). Should we try to make it possible to include/exclude these versions?

Nisha1306:
@rosvik  @gorandalum Updated to version 2. Also updated the documentation. https://github.com/AtB-AS/docs-private/blob/main/How-to-update-info-messages-in-app.md

rosvik (approved):
Looks good 💯

gorandalum (approved):

PR #3033, opened 2022-10-31
refactor: Map departures request fixes

jorelosorio:


src/components/map/components/DeparturesDialogSheet.tsx
```diff
@@ -43,17 +43,82 @@ const DeparturesDialogSheet = ({
...
+    data: stopDetailsData,
+    isLoading: isStopDetailsLoading,
+    error: stopDetailsError,
+  } = stopDetailsState;
+
+  const stopPlace = stopDetailsData?.stopPlaces?.[0];
+  const isLoading = isStopDetailsLoading || isGeocoderSearching;
+  const didLoadingDataFail = !!geocoderError || !!stopDetailsError;
+
+  const memoizedComponents = useMemo(() => {
```
gorandalum, line 66: Why use `useMemo` here? Seems a little strange as there are no expensive calculations. From the React doc for ùseMemo`:
> This optimization helps to avoid expensive calculations on every render.

gorandalum (approved):

src/components/map/components/DeparturesDialogSheet.tsx
jorelosorio: Is not expensive but avoids re-calculation 😅, I took the concept from other places like `src/components/bottom-sheet/BottomSheetContext.tsx`, do you want me to change it to a function instead?

gorandalum: In the bottom-sheet there is some technicality with calculating the animation and stuff like that which makes it necessary. I think here we can just skip `useMemo`. Using `useMemo` is not without risk, as dependencies can be forgotten causing bugs in the implementation.

jorelosorio: Fixed :)

gorandalum (approved):

PR #3038, opened 2022-11-02
feature: Map bottom sheet should have travel to/from buttons

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/2376


src/utils/navigation.tsx
```diff
@@ -33,12 +33,12 @@ export function settingToRouteName(
...
   switch (setting) {
     case 'assistant':
-      return 'Assistant';
+      return 'Dashboard';
     case 'departures':
       return 'Nearest';
     case 'ticketing':
       return 'Ticketing';
     default:
-      return 'Assistant';
```
jorelosorio, line 42: @gorandalum something is going on here after we removed the Assistant, I have replaced it with Dashboard the returns but not quite sure, could you please take a look?

jorelosorio: @gorandalum together with that, there is a compilation error in src/screens/AnonymousPurchase/ConsequencesScreen.tsx

jorelosorio:
@gorandalum because now assistant was removed and changed to dashboard there are many places where it is referenced as:

- src/screens/MobileTokenOnboarding/types.ts
- src/global-messages/GlobalMessagesContext.tsx 

and many others. I am not sure if all this will affects the app.

src/components/map/components/DeparturesDialogSheet.tsx
```diff
@@ -133,6 +152,30 @@ const DeparturesDialogSheet = ({
         }}
       />
       <View style={styles.departuresContainer}>
+        <View style={styles.buttonsContainer}>
+          <View style={styles.travelButton}>
+            <Button
+              text={t(AssistantTexts.location.travelFrom.title)}
```
gorandalum, line 158: Shouldn't use texts from `AssistantTexts` in the departures bottom sheet.

src/translations/screens/Assistant.ts
```diff
@@ -8,6 +8,12 @@ const AssistantTexts = {
     },
   },
   location: {
+    travelFrom: {
+      title: _(`Reise fra`, `Travel from`),
```
gorandalum, line 12: Reise -> Reis

src/components/map/components/DeparturesDialogSheet.tsx
```diff
@@ -150,6 +193,24 @@ const useBottomSheetStyles = StyleSheet.createThemeHook((theme) => ({
   departuresContainer: {
     flex: 1,
   },
+  buttonsContainer: {
+    padding: theme.spacings.medium,
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'center',
```
gorandalum, line 200: Are `alignItems` and `justifyContent` necessary here?

src/screens/Map/MapScreen.tsx
```diff
@@ -34,6 +34,18 @@ export const MapScreen = ({navigation}: MapScreenProps<'MapScreen'>) => {
       activeItemIndex: 0,
     });
   };
+  const navigateToTripSearch: TravelFromAndToLocationsCallback = (
+    location,
+    destination,
+  ) => {
+    navigation.navigate({
+      name: 'TripSearch' as any,
```
gorandalum, line 42: Shouldn't use `any` here. Needs to go to `TripSearch` through the `Dashboard`. 
```
navigation.navigate('Dashboard', {
  screen: 'TripSearch'
  params: ... 
})
```

src/utils/navigation.tsx
gorandalum: In the `ConsequencesScreen` you need to navigate to the `DashboardRoot` through both the `TabNavigator` and `Dashboard` first.

gorandalum (changes requested):
This looks good! Some comments, but nothing that needs major refactorings 😅  Also try to be precise with the margins/paddings, I see some deviations from the design sketches which I'm sure will be commented during testing.

src/screens/Map/MapScreen.tsx
jorelosorio: @gorandalum how do I navigate to `TripSearch` through the `Dashboard` and continue using the merge feature in params as is now?

src/components/map/components/DeparturesDialogSheet.tsx
jorelosorio: In the design, the text appears to be centred.
image

src/utils/navigation.tsx
jorelosorio: Do you have any ideas on how to do it? I am a bit confused here.

jorelosorio:
> 

There is a container with a margin of 12 and text inside with padding 12 plus 24 which is the value required for the text:

image

Or were you talking about the spacing of the buttons? Could you please be more specific about where did you see the deviations, then I could fix them there?

src/components/map/components/DeparturesDialogSheet.tsx
gorandalum: But this is the `buttonsContainer`?

src/screens/Map/MapScreen.tsx
gorandalum: Oh, yeah, it is the other `navigate`-method when using `merge`. You can try something like this then:
```
navigation.navigate({
  name: 'Dashboard',
  params: {
    screen: 'TripSearch',
    params: ...
  },
  merge: true,
})
```

src/utils/navigation.tsx
gorandalum: Something like this? It is written for hand, so might be some details that is wrong, but maybe enough to get the gist of it?
```
navigation.navigate("TabNavigator", {
  screen: 'Dashboard',
  params: {
    screen: 'DashboardRoot',
  }
})
```

src/screens/Map/MapScreen.tsx
```diff
@@ -38,6 +38,18 @@ export const MapScreen = ({navigation}: MapScreenProps<'MapScreen'>) => {
       activeItemIndex: 0,
     });
   };
+  const navigateToTripSearch: TravelFromAndToLocationsCallback = (
+    location,
+    destination,
+  ) => {
+    navigation.navigate({
```
jorelosorio, line 45: @gorandalum I have tried the other way around as you suggested but, the parameters inside `TripSearch` is not merged.

I tried: 

```
navigation.navigate({
      name: 'Dashboard',
      params: {
        screen: 'TripSearch',
        params: {
          [destination]: location,
        },
      },
      merge: true,
    });
```
But if you leave it as is now I have noticed that the `TripSearch` is already present in the stack because now Dashboard becomes the default when the app loads. 
I didn't find any other way to get the previous parameter on that screen, so without the merge option, it always replaces the last from selection to the current position, which breaks the flow we want.

gorandalum: If it can't be done then we should navigate directly. But `any` should be used even though, so the NavigationProp have to be updated so the `TripSearch` is a visible screen. I'll take a look now.

gorandalum: I couldn't get it to work 🤔 Let's hear what they have to say:
https://github.com/react-navigation/react-navigation/issues/10995

gorandalum: But for now just make it work with this. Try to remove `any` by using `CompositeScreenProps` so the `TripSearch` is available for the `navigation` object.

gorandalum (approved):

PR #3041, opened 2022-11-03
feat: Add departures v2 onboarding

gorandalum:
Resolves https://github.com/AtB-AS/kundevendt/issues/2721

Screenshots in Norwegian, English and max fontsize:
<div>

</div>


src/screens/Departures/DeparturesOnboardingScreen.tsx
```diff
@@ -0,0 +1,93 @@
...
+          </View>
+          <ThemeText color={themeColor} style={styles.body} isMarkdown={true}>
+            {t(DeparturesTexts.onboarding.body.part2)}
+          </ThemeText>
+        </View>
+        <Button
+          text={t(DeparturesTexts.onboarding.button)}
+          onPress={async () => {
+            await storage.set(
+              '@ATB_has_read_departures_v2_onboarding',
```
jorelosorio, line 59: It would be nice if we have this value inside something else and not here directly in the view, I saw other places uses a concept like StoreKey. I think is better to change this value in one place than searching in all the source code in the future.

gorandalum: Good suggestion! 👍 I created a new enum key for this in the storage-file now. I will for now not "move" the other storage keys into the new enum, as it would require more regression testing.

jorelosorio:
@gorandalum the branch has conflicts with master

gorandalum:
> @gorandalum the branch has conflicts with master

Yes, I know. I'll fix before merge.

jorelosorio (approved):

PR #3042, opened 2022-11-04
Widget: Fetch and show departure data for closest favorite

adriawh:


adriawh:
Closes https://github.com/AtB-AS/kundevendt/issues/2588

PR #3046, opened 2022-11-08
refactor: Change departures headers to full width

gorandalum:
This is to fix the issue with the missing rounded corner on the first most departure.

### DIFF
--- a/src/screens/Departures/components/QuaySection.tsx
+++ b/src/screens/Departures/components/QuaySection.tsx
@@ -81,7 +81,6 @@ export default function QuaySection({
     <View testID={testID}>
       <Sections.Section withPadding withBottomPadding>
         <Sections.GenericClickableItem
-          type="inline"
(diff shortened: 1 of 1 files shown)


gorandalum:
After the change the header is now full width:
<div>


</div>

rosvik (approved):

gorandalum:
Acceptance criteria:
- [ ] Should be full width in all situations
- [ ] Should work when description message spans multiple lines
- [ ] Should work with large text
- [ ] Should work on both iOS and Android

gorandalum:
Is (or will be) updated in Figma here:
https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?node-id=6242%3A28888

HanneLH:
Test on iOS (v.1.28 (1667919856)) og Android (v.1.28 (16679197771)) ✅

PR #3048, opened 2022-11-08
Widget now looks like figma design

adriawh:
Added: 
- Looks like the Figma design for light and dark mode
- Adjusts to different system font sizes
- Added the two supported families(widget sizes) medium and small
- Shows the correct icon for transport mode
- Shows day of departure if the departure is in a future day


adriawh:
Closes https://github.com/AtB-AS/kundevendt/issues/2589

ios/departureWidget/Views/SmallView.swift
```diff
@@ -0,0 +1,32 @@
...
+                    .multilineTextAlignment(.center)
+                    .foregroundColor(Color("LineInformationTextColor"))
+            }
+            Spacer()
+
+            HStack {
+                ForEach(viewModel.departures(numberOfdepartures: 2), id: \.self) { time in
+                    DepartureTileView(date: time)
+                }
+            }.scaledToFit().minimumScaleFactor(0.1)
```
jorelosorio, line 28: I think we could reuse the content inside the body in another view that could be reused in SmallView.swift and MediumView.swift seems pretty similar.

jorelosorio (changes requested):

ios/departureWidget/Views/SmallView.swift
adriawh: yes, agreed. Will fix

jorelosorio (approved):

jorelosorio:
Good work @adriawh the widget looks nice :)

PR #3051, opened 2022-11-09
refactor: Add new section message tem

gorandalum:
Support title and message, and the screen reader will read the status message type before reading the text content.

This is to be used when adding situation messages on departures v2.

Screenshot 2022-11-09 at 10 24 16


rosvik (approved):
Looks good 👍

PR #3052, opened 2022-11-10
feat: upgrade theme package to get new NFK colours

toretefre:
Updating theme package, to include new NFK destructive interactive colours in the 1.28 version of app.

Should include the following changes in the theme package (only NFK colours as far as I can see):

- https://github.com/AtB-AS/design-system/commit/395f3eaa76b380ae6111892afd68310047acc3de
- https://github.com/AtB-AS/design-system/commit/734ffae25faa33fd6b66fa3d3c0638d3bce58996
- https://github.com/AtB-AS/design-system/commit/9ddd692faa9d76987823f766ea779eb52cf6b673

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "link-assets": "react-native link"
   },
   "dependencies": {
-    "@atb-as/theme": "^6.1.0-alpha.1",
+    "@atb-as/theme": "^6.1.3-alpha.1",
     "@bugsnag/react-native": "^7.17.3",
     "@bugsnag/source-maps": "^2.3.1",
     "@entur-private/abt-mobile-client-sdk": "^1.1.6",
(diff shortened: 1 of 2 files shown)


mikaelbr:
If this has been tested locally with the alpha release I propose we go from alpha to current (so just good old `v6.1.3`).

mikaelbr:
Alpha is meant to test out in real usage from package that the changes work on different systems (webshop, app). And when we verify that it is in fact working, I think un-alphaing it is good (this has somewhat been "misused" until now, but no time like the present to fix it :D )

rosvik:
I knew we've had this discussion before!

Screenshot 2022-11-10 at 13 46 20

Do you want to try updating, Tore? Looks good otherwise.

rosvik (approved):

toretefre:
May the force be with me!

toretefre:
The force .. was with me

toretefre:
Looking good locally both for NFK and AtB, merging.

PR #3054, opened 2022-11-11
build: deleteDebugFilesForVariant true when store build

gorandalum:


toretefre (approved):
Potentially solving https://github.com/AtB-AS/kundevendt/issues/2823

PR #3060, opened 2022-11-16
refactor: Restructure situations module

gorandalum:
- The situations module is restructured to be inline with the coding guidelines.
- The LanguageAndText utils and types were moved from reference-data module to the translations module, as that is a more correct placement for them.
- Renamed 'LanguageAndText' to 'LanguageAndTextType' to be inline with the naming convention for types.
- 'LanguageAndTextType' now also contains the typing for the structure used in situations message, where the language field is named 'language' instead of 'lang', and the fields are optional.


src/reference-data/utils.ts
```diff
@@ -1,54 +1,29 @@
...
+  getTextForLanguage,
+  LanguageAndTextType,
+  Language,
+} from '@atb/translations';
 
 /**
  * Wrapper for getting the name of a NeTeX entity in the given language.
+ *
+ * The name should always be present, however we fall back to "Unknown" so we
+ * don't get any unexpected errors if the code. If we actually end up getting
```
rosvik, line 12:
```suggestion
 * don't get any unexpected errors in the code. If we actually end up getting
```

src/situations/SituationWarningIcon.tsx
```diff
@@ -0,0 +1,23 @@
+import {Warning} from '@atb/assets/svg/color/situations';
+import React, {ComponentProps} from 'react';
+import {hasSituations} from './utils';
+import {Situation as Situation_v1} from '@atb/sdk';
+import {Situation as Situation_from_Trips} from '@atb/api/types/trips';
+
+type Props = {
+  situations: Situation_v1[] | Situation_from_Trips[];
```
rosvik, line 8: Use the new `SituationsType` here?

rosvik (approved):
Looks good! 💯

PR #3062, opened 2022-11-17
refactor: Fix QuaySection border bottom error

gorandalum:
There was an error where not only the last item of a QuaySection had bottom border-radius.

Before/After
<div>

</div>


HanneLH:
Tested/Verified on iOS and Android ✅

HanneLH:
Verifisert i prod (test Flight) for iOS (V.1.28 (1668690378)) ✅

PR #3066, opened 2022-11-18
refactor: Date change didn't refresh departure search

gorandalum:
Restructured the logic in quay-state and stop-place-state so (hopefully) only one effect for reloading state is fired when anything changes.

Resolves https://github.com/AtB-AS/kundevendt/issues/2875


rosvik (approved):
👍👍

PR #3069, opened 2022-11-21
test: handle departure onboarding and ticket change

tormoseng:
The new departure onboarding has to be clicked in order to run the departure v2 tests. In addition, the ticket tab is renamed "Active" from previous "Valid".

### DIFF
--- a/e2e/utils/departures.ts
+++ b/e2e/utils/departures.ts
@@ -1,7 +1,7 @@
 import {by, element, expect} from 'detox';
 import {tapById} from './interactionHelpers';
 import {expectIdToHaveText, expectToBeVisibleByText} from './expectHelpers';
-import { chooseSearchResult, idExists, setInputById } from "./commonHelpers";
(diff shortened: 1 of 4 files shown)


tormoseng:
Note! The tests will still fail due to en error in the ticket offer when mixing adult with any other category. This is handled elsewhere.

e2e/utils/departures.ts
```diff
@@ -1,7 +1,7 @@
 import {by, element, expect} from 'detox';
 import {tapById} from './interactionHelpers';
 import {expectIdToHaveText, expectToBeVisibleByText} from './expectHelpers';
-import {chooseSearchResult, setInputById} from './commonHelpers';
+import { chooseSearchResult, idExists, setInputById } from "./commonHelpers";
```
rosvik, line 4: Is your linter configured correctly? This line seems inconsistent with the rest of the file.

tormoseng: Looks like I forgot to run it on the changed file... My bad!

rosvik (approved):

PR #3072, opened 2022-11-21
refactor: Update gql types from bff

gorandalum:
Main change is reusing the same shared SituationFragment for all queries
asking for situation messages.

### DIFF
--- a/src/api/types/generated/fragments/situations.ts
+++ b/src/api/types/generated/fragments/situations.ts
@@ -1,0 +1,9 @@
+import * as Types from '@atb/api/types/generated/journey_planner_v3_types';
+
+export type SituationFragment = {
+  id: string;
+  situationNumber?: string;
+  reportType?: Types.ReportType;
+  summary: Array<{language?: string; value: string}>;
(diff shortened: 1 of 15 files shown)


src/api/types/generated/fragments/situations.ts
```diff
@@ -0,0 +1,9 @@
+import * as Types from '@atb/api/types/generated/journey_planner_v3_types';
+
+export type SituationFragment = {
+  id: string;
+  situationNumber?: string;
+  reportType?: Types.ReportType;
+  summary: Array<{language?: string; value: string}>;
+  description: Array<{language?: string; value: string}>;
+};
```
rosvik, line 9: Nitpick: Should the file be called `situation.ts` instead of plural `situations.ts`?

rosvik (approved):
LGTM 👍

jorelosorio (approved):

src/api/types/generated/fragments/situations.ts
gorandalum: Might be better with singular, but it is also used plural in bff so I don't change this now.

PR #3078, opened 2022-11-22
feat: Not repeat duplicate situations on departure details

gorandalum:
In stead of using the situations from the estimated call, only the situations from the quay is show in the departure details screen.

Also removed the type 'SituationsType' as it was redundant. Just as easy to do 'SituationType[]'.

Screenshot before/after:
<div>

</div>


rosvik (approved):
Apologies for late review, but looks good 💯

PR #3083, opened 2022-11-25
refactor: Situation messages improvements

gorandalum:
- More correct color coding for situation messages
- Cancelled departures now gives red/error level
- Internal info messages from AtB is now blue/info level
- Not show multiple situation messages in same box
- Situation summary preferred over description

Resolves https://github.com/AtB-AS/kundevendt/issues/2916


src/screens/TripDetails/DepartureDetails/index.tsx
```diff
@@ -148,14 +149,17 @@ export default function DepartureDetails({navigation, route}: Props) {
...
+            />
+          )}
+
+          {someLegsAreByTrain && (
+            <MessageBox
+              containerStyle={styles.messageBox}
+              type="info"
+              message={t(DepartureDetailsTexts.messages.collabTicketInfo)}
             />
           )}
```
rosvik, line 164: Just a heads up @ingvisim: Looks like this will create conflicts with what you have been working on. Probably more places as well, like in `TripMessages`.

rosvik (approved):
Looks good! Is this something that should be updated in the design system as well? From what I've seen, there's only examples of warnings, but now there can also be error and info icons and messages.

PR #3085, opened 2022-11-28
refactor: Refactor departure details implementation

gorandalum:
Removed the CallGroup related logic. I think it made the code more complex than it needed to be. In stead now the group (and other metadata) is added directly on the estimated call.

This also made it possible to store whether the estimated call is start/end of service journey and start/end of group directly in the metadata, which encapsulates the complexity. The users of the data don't need to find out this themselves, but can just get it from the metadata.

Resolves https://github.com/AtB-AS/kundevendt/issues/2916


rosvik (approved):
Clean! ✨

PR #3087, opened 2022-11-29
feat: add fare product type configurations from firestore

rosvik:
Just a draft PR for now. Having problems with my development environment so haven't tested the implementation of this, but maybe enough to show how it'll be used.

closes https://github.com/AtB-AS/kundevendt/issues/2895

### DIFF
--- a/src/configuration/converters.ts
+++ b/src/configuration/converters.ts
@@ -51,6 +51,10 @@ function mapToFareProductTypeConfig(
   const configuration = mapToFareProductConfigSettings(config.configuration);
   if (!configuration) return;
 
+  if (!isArray(config.transportModes)) return;
+  if (!config.transportModes.every((value: any) => typeof value === 'string'))
+    return;
+
   return {
     type: fcType,
     name,
--- a/src/configuration/converters.ts
+++ b/src/configuration/converters.ts
@@ -42,10 +42,9 @@ function mapToFareProductTypeConfig(
     return;
   }
 
-  const fcTypes = ['single', 'period', 'hour24', 'night', 'carnet'];
+  const fcTypes = ['single', 'period', 'hour24', 'night'];
   const fcType = mapToStringAlternatives<FareProductType>(config.type, fcTypes);
   if (!fcType) {
-    notifyWrongConfigurationType(config.type, 'type', fcTypes);
     return;
   }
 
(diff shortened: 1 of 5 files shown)


jorelosorio (approved):

src/configuration/converters.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (!fcType) return;
+
+  const configuration = mapToFareProductConfigSettings(config.configuration);
+  if (!configuration) return;
+
+  return {
+    type: fcType,
+    name,
+    description,
+    transportModes: config.transportModes,
```
rosvik, line 58: Should probably validate `transportModes` too. Maybe check that it is an array of strings.

rosvik (commented):
Looks good I think, though there are some things we need to consider before this becomes part of a release. Maybe not critical, so we can consider merging this, so it no longer blocks other issues, then update this in other PRs.

- Are the defaults in `fare-product-type-config.json` how we want them?
- Me and @gorandalum discussed that when there is an incorrect fareProductTypeConfigs in firestore, we should log it to bugsnag, and make sure that we notice when it happens. While it should not be a problem to add new fields or ticket products in the future.
- I can have a look at documenting the configurations in docs-private, since they may be long-lived and used by different people and orgs.

src/configuration/converters.ts
jorelosorio: @rosvik `fare-product-type-config.json` is based on the actual values on the remote-config branch.
Do you want to record each wrong value or one bugsnag report to the entire configuration?

rosvik: > `fare-product-type-config.json` is based on the actual values on the remote-config branch.

Yes, it's probably fine. I was thinking if we want to show every product if the device doesn't have access to Firestore, or if we should hide periodic tickets for example. I don't know, but shouldn't be a blocker for this PR 😊

> Do you want to record each wrong value or one bugsnag report to the entire configuration?

I'm not sure how it should work, but I guess it would be nice if we had an indication on what part is wrong.

rosvik:
Can't approve my own PR, but I would if I could! 😄

src/configuration/converters.ts
jorelosorio: @rosvik I made fixes to include detailed bugsnag descriptions if something is wrong. Please take a look!

PR #3088, opened 2022-11-29
fix: Show realtime even though before aimed time

gorandalum:
When the expected time was before aimed time, the aimed time was
shown. I am not sure if this was intentional or not. But when searching
for trips all legs of the trip is based on the realtime data, so we need
to show the expected time for all legs.

Resolves https://github.com/AtB-AS/kundevendt/issues/2928


gorandalum:
Before/After
<div>


</div>

rosvik (approved):

PR #3094, opened 2022-11-30
refactor: MessageBox improvements

gorandalum:
- Adhere to the coding guidelines, with names export through index.ts.
- Remove possiblity for sending in content with the children parameter instead of message and title. It was not used many places in the app, and we have previously had some bugs by this as the text needs to be manually themed according to the box color.
- Remove the isDismissable parameter, as it is enough to show the dismiss button based on if the onDismiss function is provided or not.
- Renamed containerStyle parameter to style, to follow the preferred naming pattern for the view style parameter.
- Removed withMargin parameter, as it instead should be applied through the style parameter.
- Restructured the code in the MessageBox component.

https://github.com/AtB-AS/kundevendt/issues/2908


rosvik (approved):
Looks good 💯

PR #3098, opened 2022-12-01
fix: show correct information message when starting on foot

ingvisim:
Show correct information message when trip has train in zone a and starts on foot

https://github.com/AtB-AS/kundevendt/issues/2728


rosvik (approved):
💯

PR #3099, opened 2022-12-02
test: disable test due to entur error

tormoseng:
There is currently an error when choosing a combination of adult and other categories when buying tickets. This is due to the offer response from Entur. Until this is fixed, some assertions are disabled to make the e2e test run again.


HanneLH (approved):
🌸

PR #3100, opened 2022-12-02
feat: show realtime icon on front page

ingvisim:
Show realtime icons on departures on front page and remove "Ca.".

<details>
<summary>Screenshot</summary>

Simulator Screen Shot - iPhone 14 - 2022-12-02 at 14 19 11
</details>

Close https://github.com/AtB-AS/kundevendt/issues/2947

### DIFF
--- a/src/translations/screens/Nearby.ts
+++ b/src/translations/screens/Nearby.ts
@@ -171,14 +171,11 @@ const NearbyTexts = {
         ),
       upcomingRealtimeAccessibilityLabel: (time: string) =>
         _(
-          `Kommende avgang har sanntid på ${time}.`,
+          `Kommende avgang sanntid ${time}.`,
           `Next departure realtime ${time}.`,
         ),
       upcomingAccessibilityLabel: (time: string) =>
-        _(
-          `Kommende avgang har rutetid på ${time}.`,
-          `Next departure route time ${time}.`,
-        ),
+        _(`Kommende avgang ${time}.`, `Next departure ${time}.`),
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Looks good.

You can do a checkup on how it works with the screen reader. If i remember correctly we previously added prefix "rutetid" to the non-realtime-departures. Now the logic has kinda changed, it is the realtime-departures that have added extra information in the form of a realtime icon. Maybe the screen reader now should say "sanntid" on the realtime-departures, and no prefix on the non-realtime-departures?

src/translations/screens/Nearby.ts
```diff
@@ -167,22 +167,22 @@ const NearbyTexts = {
...
         _(
           `Avgangen ${time} har trolig passert plattformen.`,
-          `Departure ${time} has likely left the plattform.`,
+          `Departure ${time} has likely left the platform.`,
         ),
       upcomingRealtimeAccessibilityLabel: (time: string) =>
         _(
-          `Kommende avgang estimert til ${time}.`,
-          `Next departure estimated in ${time}.`,
+          `Kommende avgang har sanntid på ${time}.`,
```
gorandalum, line 174: Maybe just have `Kommende avgang sanntid ${time}` here? Then we reinforce the pattern that we say "sanntid" right before announcing the time if it's realtime. I don't think "har sanntid på" gives enough extra information to deviate from the pattern. As always, feel free to argue ;)

gorandalum (approved):
Good work! 👍

PR #3103, opened 2022-12-05
build: Copy .netrc in ios setup

gorandalum:
I have no idea why this is now necessary, but wasn't before 🤔 Or to be more precise, I don't fully understand why it was working before without copying .netrc before installing pods.

I see the macos-12 runner changed version from 20221121.1 to 20221127.5 when this started happening, but don't see anything relevant in the changelogs that could have caused this.

It seems like the build started failing exactly on the day two years after the mapbox token was created was just a coincidence.


rosvik (approved):
Strange stuff 🤔 Hopefully this works!

PR #3104, opened 2022-12-06
chore: add hermes to "Link Binary With Libraries"

rosvik:
Fixes a crash that seems to appear when building iOS with Xcode 14.1 on Intel macs. Adds hermes to this section to the Xcode project: 

image


gorandalum (approved):
As it seems to be necessary on your computer we will add it. Maybe we can reevaluate later when all developers have mac chip.

PR #3107, opened 2022-12-06
feat: Show notices on Departure Details

gorandalum:
To enable this we use the new serviceJourney/{id}/calls endpoint in the bff. This also gets the relevant notices on service journey, journey pattern and line level. In addition show the situations on each estimated call, but filter out those who are shown at the top. This is for removing line- and serviceJourney-level situations which are duplicated for each estimated call.


rosvik (approved):
LGTM

PR #3109, opened 2022-12-07
feat: Show notices on trips

gorandalum:


toretefre:
Looking good, notices in place for NFK at least 🎉 

Should we display an icon in the Assistant results, indicating that there are notices in the details?

image

image

toretefre:
That should work really well, good work! 👏

toretefre (approved):

PR #3111, opened 2022-12-07
feat: Show info icon if trip or call has notices

gorandalum:
On trip and estimated call we show a blue or yellow info icon if it has any situations. Now if the trip or estimated call has any notices the blue icon will show.

This is fixed for Assistant on the new frontpage (but not the old one), on departures v2, and on departures v1 on the frontpage.


toretefre (approved):
Looking good in NFK app as well, thank you! 👌

rosvik (approved):
💯

PR #3112, opened 2022-12-07
feat: Make boarding info less intrusive

gorandalum:
Instead of using message boxes, just have the boarding info as secondary text under the quay name. Also skip redundant boarding info on first and last stop of service journey.

Also fixed som padding on the TripRow for situation message, so it is a little closer to the quay name above.

Resolves https://github.com/AtB-AS/kundevendt/issues/2992

Before/After:
<div>

</div>


rosvik (approved):
Nice, good update! 💯

PR #3113, opened 2022-12-08
Ingvisim/add icon size to button

ingvisim:
Add option to set size of icons on buttons to support small realtime icons.

image

https://github.com/AtB-AS/design-system/pull/92

https://github.com/AtB-AS/kundevendt/issues/2950
https://github.com/AtB-AS/kundevendt/issues/2947

### DIFF
--- a/src/chat/ContactSheet.tsx
+++ b/src/chat/ContactSheet.tsx
@@ -137,7 +137,7 @@ const ContactItem: React.FC<ContactProps> = ({
         text={buttonText}
         accessibilityHint={accessibilityHint}
         onPress={onPress}
-        rightIcon={icon}
+        rightIcon={icon && {svg: icon}}
       />
     </View>
   );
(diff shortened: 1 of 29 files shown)


gorandalum (changes requested):
Very nice that you have reused the typing from Theme.

I think I would rather have the icon fields grouped together, more like this:

```ts
type ButtonIconProps = {
  svg: ({fill}: {fill: string}) => JSX.Element;
  size?: keyof Theme['icon']['size'];
}

export type ButtonProps = {
  ...;
  leftIcon?: ButtonIconProps;
  rightIcon?: ButtonIconProps;
}
```

In this way the typing is more precise, and it is not possible with illogical combinations like setting `leftIconSize` even though `leftIcon` is not set. I know it will probably generate a lot of changes in the code, but I think it's worth it.

gorandalum (approved):
Great! :+1:

PR #3114, opened 2022-12-08
refactor: Message box updated according to design

gorandalum:
- The whole box is pressable
- Larger insets for the dismiss button
- Improved for screen reader

Based on updates in Figma:
https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components?node-id=109%3A0&t=0EafJVyPUQPApAoc-1

### DIFF
--- a/src/components/message-box/MessageBox.tsx
+++ b/src/components/message-box/MessageBox.tsx
@@ -1,13 +1,35 @@
 import React from 'react';
-import {StyleProp, TouchableOpacity, View, ViewStyle} from 'react-native';
+import {
+  Linking,
+  StyleProp,
+  TouchableOpacity,
+  View,
+  ViewStyle,
+} from 'react-native';
 import {Statuses, StyleSheet, useTheme} from '@atb/theme';
 import ThemeText from '@atb/components/text';
 import ThemeIcon from '@atb/components/theme-icon';
 import MessageBoxTexts from '@atb/translations/components/MessageBox';
 import {useTranslation} from '@atb/translations';
 import {Close} from '@atb/assets/svg/mono-icons/actions';
-import {screenReaderPause} from '@atb/components/accessible-text';
 import {messageTypeToIcon} from '@atb/utils/message-type-to-icon';
+import {TouchableOpacityOrView} from '@atb/components/touchable-opacity-or-view';
+import insets from '@atb/utils/insets';
+import {screenReaderPause} from '@atb/components/accessible-text';
+
+/**
+ * Configuration for how the onPress on the message box should work. The
+ * configuration is split into action (function) or url (string) so that we can
+ * decide correct accessibility role on the message box.
+ *
+ * Note that for the screen reader the value in the `text` field will be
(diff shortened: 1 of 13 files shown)


src/components/message-box/MessageBox.tsx
```diff
@@ -1,24 +1,45 @@
...
   type: Statuses;
   title?: string;
   message: string;
   noStatusIcon?: boolean;
   onDismiss?: () => void;
-  onPress?: () => void;
-  onPressText?: string;
   isMarkdown?: boolean;
   style?: StyleProp<ViewStyle>;
+  onPressConfig?: OnPressConfig;
```
gorandalum, line 42: Feel free to suggest better naming

src/components/message-box/MessageBox.tsx
```diff
@@ -39,53 +59,78 @@ export const MessageBox = ({
...
   };
+
+  const onPress =
+    onPressConfig &&
+    ('action' in onPressConfig
+      ? onPressConfig.action
+      : () => Linking.openURL(onPressConfig.url));
+
+  const a11yLabel = [title, message, onPressConfig?.text]
+    .filter((s): s is string => !!s)
```
jorelosorio, line 70: There are other cases for this kind of filter where `.filter(Bool)` is used instead.

src/components/message-box/MessageBox.tsx
```diff
@@ -39,53 +59,78 @@ export const MessageBox = ({
...
+          <TouchableOpacity
+            onPress={onDismiss}
+            accessible={true}
+            accessibilityLabel={t(MessageBoxTexts.dismiss.allyLabel)}
+            accessibilityRole="button"
+            hitSlop={insets.all(theme.spacings.medium)}
+          >
+            <ThemeIcon fill={textColor} svg={Close} />
+          </TouchableOpacity>
+        </View>
```
rosvik, line 131: What's the reason for the added `View` tag here?

rosvik (approved):
LGTM 👍

src/components/message-box/MessageBox.tsx
gorandalum: Good suggestion! I feel this would be just as good, but still not the perfect solution that I was dreaming of 😅 Will not use time on changing anything now.

src/components/message-box/MessageBox.tsx
gorandalum: When using `.filter(Boolean)` it doesn't work as a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) since we are missing the `s is string` part. This would make `a11yLabel` a `(string | undefined)[]` and not `string[]`.

src/components/message-box/MessageBox.tsx
gorandalum: When the `TouchableOpacity` is in a `View` which has `flex: 1` it grows to full height.

With the added `View` / Without the added `View`:
<div>


</div>

PR #3126, opened 2022-12-14
refactor: Only proved quay id to PlaceScreen

gorandalum:
The parameter selectedQuay to the PlaceScreen was an object. This was not beneficial as it is better that the PlaceScreen itself is in charge of what data is shown. Now when just passing the id, then the quay data from the stop-departures query are shown, and not the quay data available at the screen where it linked to the PlaceScreen.


rosvik (approved):
Nice! A lot cleaner 👍

PR #3131, opened 2022-12-14
fix: handle undefined firestore feedback values

rosvik:
Fixes an error when submitting feedback, caused by `isReprompt` / `answeredAtDisplayCount` was undefined, which isn't supported in firestore, unless `ignoreUndefinedProperties` is set to true.

~Fixed by initializing the value to `1`, and providing a fallback. I could have set `answeredAtDisplayCount` to be required [here](https://github.com/AtB-AS/mittatb-app/blob/rosvik/1996-error-in-feedback-components-for-ios/src/components/feedback/index.tsx#L81) in `VersionStats` as well, but since there is data saved to user's devices where it is undefined, I think we should still support it, and rather use a fallback.~

fixes https://github.com/AtB-AS/kundevendt/issues/1996

### DIFF
--- a/src/components/feedback/index.tsx
+++ b/src/components/feedback/index.tsx
@@ -106,7 +106,7 @@ export const Feedback = ({
   const incrementCounterAndSetDisplayStats = async () => {
     if (feedbackConfig) {
       const defaultDisplayStatObject: VersionStats = {
-        answeredAtDisplayCount: undefined,
+        answeredAtDisplayCount: 1,
         doNotShowAgain: false,
         displayCount: 1,
         surveyVersion: feedbackConfig.surveyVersion,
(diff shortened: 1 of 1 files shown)


rosvik:
I could have set `answeredAtDisplayCount` to be required [here](https://github.com/AtB-AS/mittatb-app/blob/rosvik/1996-error-in-feedback-components-for-ios/src/components/feedback/index.tsx#L81) in `VersionStats` as well, but since there is data saved to user's devices where it is undefined, I think we should still support it, and rather use a fallback.

ingvisim (approved):

gorandalum (commented):
Did you look into using [`ignoreUndefinedProperties`](https://firebase.google.com/docs/reference/js/v8/firebase.firestore.Settings#ignoreundefinedproperties) instead?

rosvik:
@gorandalum Aha, I didn't know that was an option! Updated now.

gorandalum (approved):

PR #3143, opened 2022-12-16
feat: Verify that night tickets can only be purchased at night

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/2837

This solution uses the following rules:

- A time when night tickets will be sold in normal conditions (day to day), from weekends or weekdays in a range of hours.
- There are exceptions where one or a range of days are closed.
- Enable at now take precedence over regular hours.


jorelosorio:
Was replaced by: https://github.com/AtB-AS/mittatb-app/pull/3208

PR #3154, opened 2022-12-20
Change student max age in app for NFK and FRAM

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/3043

Changes text for student max age at NFK and FRAM (they're the same)

Verified for NFK: 
Simulator Screen Shot - iPhone 13 Pro - 2022-12-20 at 12 16 45

### DIFF
--- a/src/translations/screens/subscreens/TicketTraveller.ts
+++ b/src/translations/screens/subscreens/TicketTraveller.ts
@@ -1,5 +1,6 @@
 import {translation as _} from '../../commons';
 import {PreassignedFareProductType} from '@atb/reference-data/types';
+import {APP_ORG} from '@env';
 
 enum TravellerType {
   adult = 'ADULT',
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Approved, but I kinda don't like it (not something to do with your change). Both that:
- These texts really should be from Firestore. I don't actually know why we have them in the app.
- We already have the `orgSpecificTranslations` for overriding specific texts for organizations. However I see that it will be more code changes to utilise this here, as the setup differentiates quite a bit from a normal `Texts`-file.

toretefre:
> * These texts really should be from Firestore. I don't actually know why we have them in the app.

Agree, I initially thought these were indeed fetched from Firestore, so I was surprised to see this. That's why I asked you for a review. 😄  

Maybe I should make an issue to get these from Firestore?

gorandalum:
Yeah. You can do that :+1:

toretefre:
Created PR now @gorandalum 

https://github.com/AtB-AS/kundevendt/issues/3100

PR #3158, opened 2022-12-21
refactor: Support for AbortController in use-pollable-resource

gorandalum:
The reason for this is so the hook doesn't return outdated data. If something like the currently selected quay is changed on the screen using use-pollable-resource, then a request in progress for the previously selected quay should be aborted and not update state.

This functionality may be utilized by passing a data-retrieval function that takes in an AbortSignal as argument.

### DIFF
--- a/src/utils/use-pollable-resource.ts
+++ b/src/utils/use-pollable-resource.ts
@@ -6,11 +6,11 @@ type PollableResourceOptions<T, E> = {
   initialValue: T;
   pollingTimeInSeconds?: number;
   disabled?: boolean;
-  filterError?(error: E): boolean;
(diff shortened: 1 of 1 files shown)


rosvik (approved):
Nice! Not sure how and when to use this, so I'll try to remember asking you about this next year 😅

gorandalum:
> Nice! Not sure how and when to use this, so I'll try to remember asking you about this next year 😅

At least this is better than the CancelToken that was used before, as [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) is an official standard, and pretty straightforward to use.

PR #3171, opened 2022-12-28
refactor: Text component refactored

gorandalum:
Refactored to be inline with the coding guidelines.


gorandalum:

PR #3173, opened 2022-12-28
refactor: Components refactored according to guidelines

gorandalum:
Lots of changes, but should not be any major risk as it is mainly moving and renaming, and errors would be found by TypeScript.


tormoseng:
👍

PR #3174, opened 2022-12-29
refactor: Map, Header and Sections according to guidelines

gorandalum:


gorandalum:
Refactor of the last remaining components according to the coding guidelines.

PR #3177, opened 2023-01-03
refactor: FixedSwitch improvements 

gorandalum:
Everywhere the FixedSwitch was used there was need for scaling it
with font scale etc. This is now moved into the FixedSwitch so the
scaling is done in just one place.

Also add the possibility for setting an interactive color for the
switch.

Affects mainly three places:
- The switch in select favourites
- The switch on InfoToggle (like on show travellers info text)
- The switches under "My profile"

### DIFF
--- a/src/components/switch/FixedSwitch.tsx
+++ b/src/components/switch/FixedSwitch.tsx
@@ -1,12 +1,26 @@
 import useDelayGate from '@atb/utils/use-delay-gate';
 import React, {useEffect, useState} from 'react';
-import {Switch, SwitchProps} from 'react-native';
+import {AccessibilityProps, Switch, SwitchProps, Platform} from 'react-native';
+import {StyleSheet, useTheme} from '@atb/theme';
+import useFontScale from '@atb/utils/use-font-scale';
+import {InteractiveColor} from '@atb/theme/colors';
+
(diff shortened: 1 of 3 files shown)


rosvik (commented):
A downside to scaling buttons like this with text size, is that the clickable area gets small when a user sets their text size to smaller default. Could we make the clickable area bigger than the button area, or set a minimum size?

rosvik:
Some screenshots for reference:

<div>



</div>

Small text:

gorandalum:
I talked quite a bit with @ckbilstad about this. Generally the switch is not "correctly" used in our app. The native UI switch is not ment to be scaled, and since it needs to be done with transform instead of height/width, it messes up the layout. This is especially when increasing font scale to max, where the switch just grows and goes outside its parent component.

We concluded we want to look into a react-native-switch library to use instead, to have better control of the styling, sizing and layout. However we didn't want to use time on it in this task. 

> Could we make the clickable area bigger than the button area, or set a minimum size?

It seems that when having a clickable area which programatically updates the switch state, the animation is not triggered. I think I'll set a minimum font-scale of 1 as a temporary fix for preventing the switch to become to small.

rosvik (approved):

PR #3187, opened 2023-01-04
refactor: Remove expand option in ActionSectionItem

gorandalum:
Two main reasons:
- The ActionSectionItem was quite large and complex, with options for expandable, checkable and toggleable. It is better if these are seperate components.
- We already have the ExpandableSectionItem which should instead be used for this.

### Acceptance criteria
This affects two places, the FAQ in SelectTokenScreen, and the expandable stop place name in v1 departures.

- [ ] Should look and feel as before
- [ ] Should work good with screen reader


rosvik (approved):

HanneLH:
Test on Android v.1.31(1672922528) and iOS v.1.31(1672922545) 
Affect FAQ in SelectTokenScreen, and the expandable stop place name in v1 departures: 
- [ ]  Should look and feel as before
- [ ] Should work good with screen reader

PR #3188, opened 2023-01-06
Added deep linking to app, and fixed view for no favorites

adriawh:
Widget now redirects user to departures view of the way shown in the widget. If the user has no favorite departures, there is now a new view, and the user is sent directly to add a favorite.


jorelosorio (approved):
Good work ;)

PR #3190, opened 2023-01-06
refactor: Fix some filters design flaws

gorandalum:
- Add section seperator under heading for select favourites bottom sheet so it looks the same as the filters bottom sheet.
- Some changes to the internal section label so text doesn't overflow.
- Fixed a bug where removing a filter option from Firestore made the "all" button work incorrectly.

### DIFF
--- a/src/screens/Dashboard/SelectFavouritesBottomSheet.tsx
+++ b/src/screens/Dashboard/SelectFavouritesBottomSheet.tsx
@@ -142,9 +142,10 @@ const SelectFavouritesBottomSheet = ({
 
             <View>
               {updatedFavorites &&
-                updatedFavorites.map((favorite, index) => {
+                updatedFavorites.map((favorite) => {
                   return (
                     <View key={favorite.id}>
+                      <SectionSeparator />
                       <SelectableFavouriteDeparture
                         handleSwitchFlip={handleSwitchFlip}
                         favouriteId={favorite.id}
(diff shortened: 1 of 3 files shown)


src/screens/Dashboard/SelectFavouritesBottomSheet.tsx
```diff
@@ -142,9 +142,10 @@ const SelectFavouritesBottomSheet = ({
 
             <View>
               {updatedFavorites &&
-                updatedFavorites.map((favorite, index) => {
+                updatedFavorites.map((favorite) => {
                   return (
                     <View key={favorite.id}>
+                      <SectionSeparator />
```
jorelosorio, line 148: Does this mean that the first will contains a separator on the top always? Before it was like the last one doesn't include a separator at the end.

rosvik (approved):
LGTM

src/screens/Dashboard/SelectFavouritesBottomSheet.tsx
gorandalum: Yes. The last one still shouldn't have a separator, but the first one should have at top. Context:
https://github.com/AtB-AS/kundevendt/issues/3105#issuecomment-1373604879

jorelosorio: @gorandalum Last question but is the header who should have it statically (the separator) or should the list include it a the top? Because if you do scroll now the separator will be moved along the list.

gorandalum: Talked with Jorge. It is fine as it is now as long as the header-title of the section also scrolls.

PR #3191, opened 2023-01-06
refactor: Add ToggleSectionItem

gorandalum:
The old 'toggle' mode on ActionSectionItem had some screen reader challenges. This is fixed in this commit, alongside some restructuring.

- Created a new ToggleSectionItem instead of using ActionSectionItem.
- Renamed FixedSwitch to Toggle. This is because toggle is used as the word for the component in design-sketches/Figma, and it is slightly more intuitive.
- Parameter names are now 'value' and 'onValueChange' so it is the same parameters as using the Toggle (previously FixedSwitch) directly.
- When using screen reader the whole section item is now selected, and some extra measures has been added to ensure the accessibility state is read correctly.

### DIFF
--- a/src/components/sections/items/ActionSectionItem.tsx
+++ b/src/components/sections/items/ActionSectionItem.tsx
@@ -12,12 +12,10 @@ import {ThemeIcon} from '@atb/components/theme-icon';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
 import {useSectionStyle} from '../use-section-style';
-import {InternalLabeledSectionItem} from './InternalLabeledSectionItem';
-import {FixedSwitch} from '@atb/components/switch';
 import {InteractiveColor} from '@atb/theme/colors';
 import {SvgProps} from 'react-native-svg';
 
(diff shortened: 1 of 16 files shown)


src/components/sections/items/ActionSectionItem.tsx
```diff
@@ -12,12 +12,10 @@ import {ThemeIcon} from '@atb/components/theme-icon';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
 import {useSectionStyle} from '../use-section-style';
-import {InternalLabeledSectionItem} from './InternalLabeledSectionItem';
-import {FixedSwitch} from '@atb/components/switch';
 import {InteractiveColor} from '@atb/theme/colors';
 import {SvgProps} from 'react-native-svg';
 
-type ActionModes = 'check' | 'toggle';
+type ActionModes = 'check';
```
rosvik, line 18: Do we need the `ActionModes` prop when there is only one option? Is it just in case we add more in the future?

rosvik (approved):
Looks good 👍

PR #3193, opened 2023-01-09
build: register nfk app versions

toretefre:
Registering NFK iOS and Android app versions with Entur, while also uploading source maps for NFK Android.

### DIFF
--- a/.github/workflows/build-store-android.yml
+++ b/.github/workflows/build-store-android.yml
@@ -87,12 +87,10 @@ jobs:
           GOOGLE_PLAY_SERVICE_ACCOUNT_PATH: google-play-service-account.json
           ANDROID_PACKAGE_NAME: ${{ env.ANDROID_APPLICATION_ID }}
       - name: Upload bundle and source maps
-        if: matrix.org == 'atb'
         run: bash ./scripts/android/upload-sourcemaps.sh
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-store-android.yml
```diff
@@ -87,12 +87,10 @@ jobs:
           GOOGLE_PLAY_SERVICE_ACCOUNT_PATH: google-play-service-account.json
           ANDROID_PACKAGE_NAME: ${{ env.ANDROID_APPLICATION_ID }}
       - name: Upload bundle and source maps
-        if: matrix.org == 'atb'
         run: bash ./scripts/android/upload-sourcemaps.sh
         env:
           BUGSNAG_API_KEY: ${{ secrets.BUGSNAG_API_KEY }}
       - name: Register app version
-        if: matrix.org == 'atb'
```
mikaelbr, line 95: This will probably also build for FRAM at some point then, but maybe that doesn't matter too much?

mikaelbr (approved):

.github/workflows/build-store-android.yml
toretefre: To my understanding this was done because the NFK app would be released without ticketing, and I'm changing now since next version for NFK should include ticketing. 

For FRAM we will release with ticketing from day one, so I don't think it should be an issue?

PR #3197, opened 2023-01-09
fix: Night tickets should be presented on Recent Purchases view

jorelosorio:
Mentioned on https://github.com/AtB-AS/kundevendt/issues/3068

This PR shows night tickets on recent purchases.

Screenshot_1673276259

### DIFF
--- a/src/screens/Ticketing/FareProducts/use-recent-fare-contracts.ts
+++ b/src/screens/Ticketing/FareProducts/use-recent-fare-contracts.ts
@@ -117,6 +117,10 @@ const mapBackendRecentFareContracts = (
     userProfiles,
   );
 
+  if (!userProfilesWithCount?.length) {
+    return null;
+  }
+
   if (fareProductTypeConfig.configuration.zoneSelectionMode === 'none') {
     return {
       preassignedFareProduct,
@@ -135,12 +139,7 @@ const mapBackendRecentFareContracts = (
     recentFareContract.zones.slice(-1)[0],
   );
 
-  if (
-    !preassignedFareProduct ||
-    !fromTariffZone ||
-    !toTariffZone ||
-    !userProfilesWithCount?.length
-  ) {
+  if (!fromTariffZone || !toTariffZone) {
     return null;
   }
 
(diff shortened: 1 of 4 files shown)


src/screens/Ticketing/FareProducts/use-recent-fare-contracts.ts
```diff
@@ -98,6 +100,32 @@ const mapBackendRecentFareContracts = (
...
+  );
+
+  if (fareProductTypeConfig.configuration.zoneSelectionMode === 'none') {
+    return {
+      preassignedFareProduct,
+      fromTariffZone: undefined,
+      toTariffZone: undefined,
+      userProfilesWithCount,
+    };
+  }
```
gorandalum, line 127: Should also check `userProfilesWithCount.length` before returning?

src/screens/Ticketing/FareProducts/use-recent-fare-contracts.ts
```diff
@@ -107,11 +135,6 @@ const mapBackendRecentFareContracts = (
...
   );
 
-  const userProfilesWithCount = mapUsers(
-    recentFareContract.users,
-    userProfiles,
-  );
-
   if (
     !preassignedFareProduct ||
     !fromTariffZone ||
```
gorandalum, line 140: Maybe the check on `preassignedFareProduct` and `userProfilesWithCount?.length` is redundant here as it is already checked further up?

gorandalum (changes requested):
Looks good. Just a small check that I think should be in place so we don't get any surprises later on.

rosvik (approved):
LGTM 👍

src/screens/Ticketing/FareProducts/use-recent-fare-contracts.ts
jorelosorio: Fixed now, please take a look ;)

gorandalum (approved):

HanneLH:
Tested [through](https://github.com/AtB-AS/kundevendt/issues/3068)

PR #3198, opened 2023-01-09
refactor: New button type pill

gorandalum:
New button type pill according to how it is specced in Figma. Also some refactoring of the in-app design system to have a better representation of the different buttons.

Examples:
Screenshot 2023-01-09 at 16 40 19


rosvik (approved):
Looks good! Would be nice to have the design system changes be more dynamic (and not 800 lines of code 😅), but a lot better than nothing 👍

PR #3201, opened 2023-01-10
feat: Add visual representation of applied filters

gorandalum:
When a filter is applied to a travel search we show a button (of type pill) representing the applied filter. The button is clickable, and if clicked will remove the filter.

In addition there is a small notification icon added to the filter button icon if any filters are applied. The same is done to the travel time icon if the time is something else than 'now'.

Also some refactoring in how the selected filters are modelled. Instead of using an array containing just the selected filters, we now instead add field "selected" which may be true or false on each of the available filters.

Screenshot 2023-01-10 at 15 49 04


src/translations/screens/TripSearch.ts
```diff
@@ -117,6 +117,10 @@ const TripSearchTexts = {
         'Prøv å justere på sted eller tidspunkt.',
         'Try adjusting your time or location input.',
       ),
+      genericHintWithFilters: _(
+        'Prøv å justere på sted, filtere eller tidspunkt.',
```
rosvik, line 121: I think this is more correct grammar ([filter in dictionary](https://ordbokene.no/bm/search?q=filter&scope=ei))

```suggestion
        'Prøv å justere sted, filter eller tidspunkt.',
```

rosvik (approved):
LGTM 👍

jorelosorio:
IMG_E8A79388B586-1

In both iOS and Android, the notification icon is too small to be noticed, I would like to suggest increase 2 points the sizes in the `getIndicatorSize` function.

Tested on a real device. (iOS)

src/screens/Dashboard/SelectedFiltersButtons.tsx
```diff
@@ -0,0 +1,61 @@
...
+        }
+        text={text}
+        accessibilityHint={t(TripSearchTexts.filters.selection.a11yHint)}
+        onPress={resetTransportModes}
+        interactiveColor="interactive_0"
+        active={true}
+        leftIcon={{svg: Bus}}
+        rightIcon={{svg: Close}}
+      />
+    </View>
```
jorelosorio, line 51: It feels a bit strange close (reset) filters by tapping the button and not in the close icon, it might be something to consider for changing but is not a blocker for this PR.

jorelosorio:
When I apply filters there is a button called `Lagre valg` / `Save selection` but then I was expecting to have the selection back when reopening the app, should it be changed to just apply filters?

jorelosorio (approved):
I approved it however there are two small comments after I tested the functionality it might be taken into consideration.

gorandalum:
> In both iOS and Android, the notification icon is too small to be noticed, I would like to suggest increase 2 points the sizes in the getIndicatorSize function.
> 
> Tested on a real device. (iOS)

It is 6px in the design sketches. What do you think @ckbilstad when you look at the screenshot? It stands out a little more in the sketches with the spacing around the notification indicator, but this is hard to do when adding the indicator programatically.

> It feels a bit strange close (reset) filters by tapping the button and not in the close icon, it might be something to consider for changing but is not a blocker for this PR.

That the whole button is clickable was decided during the design process as far as I remember. You can correct me if I'm wrong @ckbilstad. Didn't find it written down anywhere when doing a quick search.

> When I apply filters there is a button called Lagre valg / Save selection but then I was expecting to have the selection saved and pre-filled back when reopening the app but seems like not.
Should it be changed to just apply filters then?

There is a task for this: https://github.com/AtB-AS/kundevendt/issues/3108

gorandalum:
> It is 6px in the design sketches. What do you think @ckbilstad when you look at the screenshot? It stands out a little more in the sketches with the spacing around the notification indicator, but this is hard to do when adding the indicator programatically.

After discussion with @ckbilstad we improved this a little by having a 2px spacing around the icons to make it "pop" out a little more
Screenshot 2023-01-12 at 15 50 23

PR #3202, opened 2023-01-11
refactor: Icon for unknown transportation mode

gorandalum:


src/components/transportation-icon/utils.ts
```diff
@@ -30,7 +30,8 @@ export function getTransportModeSvg(mode?: AnyMode, subMode?: AnySubMode) {
       return TransportIcons.Walk;
     case 'metro':
       return EnturTransportIcons.Subway;
+    case 'unknown':
     default:
-      return undefined;
+      return TransportIcons.Unknown;
```
rosvik, line 35: I'm slightly worried that the UFO will show up in places we don't want it when we use it as fallback, but not sure it's a real concern 🤔 Are we ok with it appearing in departures or ticketing for edge cases?

rosvik (approved):

PR #3203, opened 2023-01-12
Added dynamic icons

adriawh:


jorelosorio (changes requested):
I got the icon but was not able to see the background color as before, is there an extra set up that I need to do?

Simulator Screen Shot - iPhone 14 - 2023-01-17 at 11 24 11

jorelosorio (approved):
@adriawh Looks good, nice to see it a bit more dynamic.

PR #3209, opened 2023-01-13
feat: Add check for min and max version in fare product

ingvisim:
Add check for min and max version in fare product

Close https://github.com/AtB-AS/kundevendt/issues/2914

### DIFF
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -20,6 +20,8 @@ export type PreassignedFareProduct = {
   alternativeNames: LanguageAndTextType[];
   limitations: {
     userProfileRefs: string[];
+    appVersionMin: string;
+    appVersionMax: string;
   };
 };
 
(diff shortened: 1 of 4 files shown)


src/utils/compare-version.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (A[i] !== B[i]) {
+      return parseInt(A[i]) > parseInt(B[i]) ? 1 : -1;
+    }
+  }
+  if (A.length === B.length) {
+    return 0;
+  } else if (A.length > B.length) {
+    return 1;
+  }
+  return -1;
```
gorandalum, line 14: Probably don't need to return `-1`, `0` and `1` as explicitly.

I think `return parseInt(A[i]) - parseInt(B[i])` and `return A.length - B.length` should work just as well.

src/reference-data/types.ts
```diff
@@ -20,6 +20,8 @@ export type PreassignedFareProduct = {
   alternativeNames: LanguageAndTextType[];
   limitations: {
     userProfileRefs: string[];
+    appVersionMin: string;
+    appVersionMax: string;
```
gorandalum, line 24: These can be `undefined`?

gorandalum (changes requested):
Looks good! Just some small comments :+1:

gorandalum (approved):

PR #3210, opened 2023-01-16
fix: change AppTeamIdentifier into dynamic value

toretefre:
Changed value together with Jorge. Most likely just a field that was dynamic at some point, but become hardcoded for testing at some point.

Hopefully solves https://github.com/AtB-AS/kundevendt/issues/3257


jorelosorio (approved):

jorelosorio:
AtB must check everything is in place after change.

PR #3214, opened 2023-01-17
chore: Upgrade babel

ingvisim:
Upgrade babel dependencies


rosvik (approved):
Better to label these PRs `chore` I think, since there's no functional changes to src code.

PR #3215, opened 2023-01-17
chore: Upgrade firebase

ingvisim:
Upgrade firebase dependencies


jorelosorio:
@ingvisim I would like to suggest making a quick test to the updated branch, like login, if analytics are ok, change some configurations on the **firestore** and check if is in order.

gorandalum (approved):
As Jorge says, test a little locally before merging, but I'm guessing you do that already :+1:

jorelosorio (approved):

PR #3216, opened 2023-01-17
fix: Fix night ticket subtext

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/2837#issuecomment-1385326578


src/screens/Ticketing/Purchase/Overview/components/ProductSelectionByProducts.tsx
```diff
@@ -54,9 +54,19 @@ export default function ProductSelectionByProducts({
...
+            if (fp.warningMessage) {
+              const warningMessage = getTextForLanguage(
+                fp.warningMessage,
+                language,
+              );
+              return `${descriptionMessage}\n${warningMessage}`;
+            }
+
+            return descriptionMessage;
+          }}
```
gorandalum, line 69: No `description` on the `PreassignedFareProduct` shouldn't fallback to subtext "Unknown", but instead just not show a subtext. Also by the domain model it is possible for a `PreassignedFareProduct` to have a `warningMessage`, but no `description`, so the code should support that as well.

gorandalum (changes requested):

gorandalum (approved):

PR #3217, opened 2023-01-17
feat: Departure widget for iOS

adriawh:
## Widget for IOS 📱
Medium and small size widgets for showing scheduled departures for the closest favorite departure

### Features 📈:
- Shows a favorite departure with their times
- Updates of departure times based on location and intervals
- Updates when a user updates favorite departures in the app

> NOTE: The widget doesn't show real-time data, due to limitations on the Widget from Apple. Could maybe be worked around, but will need further research

#### If users have favorite departures: 
- Shows the scheduled times for the closest favorite departure
- Should update each time a departure has passed
- Redirects to the departure view, v1 or v2 (depending on the user's settings), with the departure shown in the widget
- The travel icon is only shown in the medium-size widget and should correspond to the travel type
- Should update to the closest favorite if the user moves more than 100 meters

#### If the user does not have favorite departures:
- Shows text and icon and informs that the user has no favorites
- Redirects directly to add favorite departures
- Updates when a user updates favorites in the app

### Code changes:
- Added code in the navigation too, deep linking from the widget
- Added `DISABLE_WIDGET` in the `.env`, which should be checked on the build, and remove the widget if needed. 
- A modification was made to `AsyncStorage` for handling iOS groups, that new project can be found at https://github.com/AtB-AS/async-storage. This dependency is now pointing to this new repo in the `package.json`
> It was needed for the best integration with the current solution and to keep support for Android. Many other libraries were evaluated, even building our own but due to backward compatibility we decided to keep the same with the variations that are needed for iOS
- The build process is changed a bit to handle better the app and their extensions weren't automatically managed by `Fastlane`
- The widget was developed based on Apple's restrictions as instructed on https://developer.apple.com/documentation/widgetkit/creating-a-widget-extension

## What to test ✅
- [ ] Widget is available to add from the widget selector, both medium and small
- [ ] The preview in the widget selector should show mock data of departures
- [ ] Adding the widget to the home screen
- [ ] Navigation in the app works as before(continuous)
- [ ] Updates when favorite departures are changed in the app
- [ ] Different font sizes
- [ ] Dark mode
- [ ] Test situations where there is no internet connection (For such cases `No departures` will be shown)
- [ ] Test situations where location has been disabled (For such cases widget shows the first departure without any order criteria)
- [ ] Looks like the [Figma sketches👀](https://www.figma.com/file/kpHgNuHVvIy0orv0YpWkyJ/Widget?node-id=0%3A1&t=9lVNrtT6nq9I0OqE-1)

#### If the user had previous storage
- [ ] All stored values are migrated when using old versions and have stored favorites and moved to the widget version.
- [ ] Widget has access to favorite departures stored before update with the widget

#### If the user does not have favorite departures:
- [ ] Shows view that informs the user of no favorites
- [ ] Redirects to add favorites
- [ ] Updates when a favorite departure is added to the app
- [ ] Text is in both English and Norwegian (However it works differently from the app, language changes are needed for the entire phone)

#### If the user has favorite departures:
- [ ] Shows the scheduled times for the closest departure( test by adding two departures at different locations)
- [ ] Shows the correct travel type icon for the travel type of the favorite
- [ ] Redirects to the departure shown in the widget in the preferred view of the app (v1 or v2)
- [ ] Updates when a departure has passed
- [ ] Changes to the closest favorite if the user moves more than 100 meters **For this point please refer to `What to expect to happen at point `4`.**

> As a general note, the fallback for favorite departure is the first on the list

### What to expect to happen❗️:
1. Because location will be fetched many times in the day, apple shows a couple of messages after time has passed by (This is completely normal when the location is shared with the widget) :

Widget specific             |  Use position in the background
Simulator Screen Shot - iPhone 14 - 2023-01-17 at 11 29 52  | Simulator Screen Shot - iPhone 14 - 2023-01-17 at 11 29 32

3. Due to the nature of the widget and the listening for location changes in the app, location permission is requested at the first install, not as it was before when it was needed.
4. When testing the change of favorite in the widget based on locations, is expected that the update is not immediate, it might happen if for example is tested inside a bus (Moving faster) the 100m range is not taken until an extra couple of minutes pass.
5. Widget sizes are defined by Apple, so sometimes texts don't fit or need to be cut off. Unfortunately, there is no way to customize widget sizes.


gorandalum (approved):
I've looked at the linking code, and it looks good enough :+1: I don't think it is necessary to invest significant time into how to do it without the custom `getStateFromPath` implementation. Is the code type safe, so if a screen changes name it will give a typescript error? (I have so many changes locally that I want to finish before changing context and running up this PR to test for myself 😅)

rosvik (approved):
Had just a quick skim through the code, but this looks good! Great job @adriawh 💯👏

HanneLH:
@adriawh - vil du merge ? 👼

PR #3219, opened 2023-01-18
chore: Add missing babel dependency

ingvisim:
Add missing babel dependency

https://github.com/AtB-AS/mittatb-app/pull/3214


jorelosorio (approved):
@ingvisim, it is only needed for development?

gorandalum:
> @ingvisim, it is only needed for development?

In general all dependencies used during building, like babel, is just dev dependencies.

PR #3236, opened 2023-01-23
fix: Small bugs in widget

adriawh:
Minor fixes on Widget📱:

1. Addresses all issues in this comment from @HanneLH: https://github.com/AtB-AS/kundevendt/issues/3280#issuecomment-1400206907
2. Fixes deeplinking, which got broken after refactoring of navigation.

Hannes comment with google translate: 

> A couple of things:
> 
> 1. If you have chosen the tram from e.g. Hospitalkirka, then I can't bring up favourites. This is pleasing both when I have a bus as a favorite as well and only the tram.
> 2. We are missing the icon for train, if it is selected as a favourite.
> 3. If you have selected the subway in Oslo as a favourite, the bus icon is displayed
> 4. If you have selected airport bus as a favourite, a green bus is shown, but here I expected the blue one that is in the app.
> 5. Small Widget, if there is no internet, you will lose the text regarding "No departures"


jorelosorio (approved):
Code looks good 👍

PR #3241, opened 2023-01-25
Show scooters in map

KristianSelnas:


src/vehicles/use-vehicles-enabled.ts
```diff
@@ -0,0 +1,18 @@
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {useDebugOverride} from '@atb/debug';
+import {StorageModelKeysEnum} from '@atb/storage';
+
+export const useIsVehiclesEnabled = () => {
+  const [debugOverride] = useVehiclesInMapDebugOverride();
+  const {enable_vehicles_in_map: enabledInRemoteConfig} = useRemoteConfig();
+
+  return debugOverride ?? enabledInRemoteConfig;
```
gorandalum, line 9: I think you need
```
  if (debugOverride !== undefined) {
    return debugOverride;
  }
  return enabledInRemoteConfig;
```

When `debugOverride` is set to false we want the vehicles functionality to be turned off, and not fallback to the value in RemoteConfig.

src/vehicles/use-vehicles.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }, [lon, lat, range, isVehiclesEnabled]);
+
+  const fetchVehicles = async (
+    {latitude, longitude}: Coordinates,
+    radius = 500,
+  ) => {
+    setLat(latitude);
+    setLon(longitude);
+    setRange(radius);
```
gorandalum, line 33: This shouldn't be done in the three set operations. Since the values are updated in three rounds it is a risk that the useEffect which has them as dependencies also trigger three times. Should instead have latitude, longitude and range in a single state object.

src/vehicles/use-vehicles.ts
```diff
@@ -0,0 +1,37 @@
...
+  const isVehiclesEnabled = useIsVehiclesEnabled();
+
+  useEffect(() => {
+    if (isVehiclesEnabled) {
+      getVehicles({lat, lon, range})
+        .then(toGeoJSONFeature)
+        .then(toFeatureCollection)
+        .then(setVehicles);
+    }
+  }, [lon, lat, range, isVehiclesEnabled]);
```
gorandalum, line 25: We probably need to do something with `AbortController` and the `useEffect` cleanup function, if not we risk race conditions. Like if you change regions quite fast, and the api response from the first region change returns after the api response from the second region, then the vehicles from the first region change will overwrite the vehicles from the second region change.

gorandalum (changes requested):
Great! A really good quality first PR :+1: Some comments on more advanced stuff, but I think it is good to just fix them now, as they might be forgotten if we delay them.

gorandalum:
Also added the rest of the developers as reviewers. Even though not everyone need to review it vigorously, it is good to just take a look so everyone has a change to get to know the new code.

src/vehicles/use-vehicles.ts
KristianSelnas: Yes, I suppose you are right. I'll see what I can come up with

KristianSelnas: Har lagt inn et forslag nå, men det er litt utfordrende å teste om det fungerer som forventet.

src/api/vehicles.ts
```diff
@@ -16,6 +22,9 @@ export const getVehicles = ({lat, lon, range}: GetVehiclesQueryVariables) => {
     formFactors: FormFactor.Scooter, //TODO: Read from variables
   });
   return client
-    .get<GetVehiclesQuery>(stringifyUrl(url, query))
+    .get<GetVehiclesQuery>(stringifyUrl(url, query), {
+      ...opts,
+      baseURL: 'http://localhost:8080',
```
gorandalum, line 27: ☝️

KristianSelnas: Doh! Takk

gorandalum (approved):

PR #3246, opened 2023-01-26
feat: improve departures when using large font

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/3197

**Updated navigation buttons in departures**
Excuse the horrific colors in the start

https://user-images.githubusercontent.com/1774972/214808711-2367ee12-cf79-4892-bd3d-d1fb0efb10d7.mp4

**Wrap transport mode text in ticket overview**

**Wrap total price on confirmation screen**

### DIFF
--- a/src/components/toggle/Toggle.tsx
+++ b/src/components/toggle/Toggle.tsx
@@ -51,7 +51,9 @@ export function Toggle({
 
 const useStyles = StyleSheet.createThemeHook(() => {
   const fontScale = useFontScale();
-  const scale = Math.max(fontScale, 1);
+
+  // Limit scale between 1, and 1.5
+  const scale = Math.min(1.5, Math.max(fontScale, 1));
   return {
     android: {
       transform: [{scale}],
--- a/src/place-screen/components/DateSelection.tsx
+++ b/src/place-screen/components/DateSelection.tsx
@@ -1,4 +1,4 @@
-import {StyleSheet, useTheme} from '@atb/theme';
+import {StyleSheet} from '@atb/theme';
 import React from 'react';
 import {View} from 'react-native';
 import {Button} from '@atb/components/button';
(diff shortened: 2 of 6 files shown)


oyvindmikkelsenatb:
Love the solution @rosvik! And I'm guessing the user who reported this will be really happy!

First a tiny details that's more a nice to have than a must: Moving the back button to show above the title ("Billetter" and "Prinses gate) on small screens. I would argue that it makes more sense have these navigational buttons above the title since navigation isn't connected to what the title refers to. In most other mobile and web apps the back and close icons are placed above the content.

Secondly, could we set the bottom navigation font size zoom to a maximum that does not breaks outside the background? I would argue that these buttons convey the meaning better with smaller texts that does not break than a big text that breaks.

gorandalum (approved):
Looks good! The only thing I'm a little unsure about is the hiding of "Next day"/"Previous day" based on font size. The WCAG doc gives these steps for testing text sizing:
> - Increase the text size of the content by 200%.
> - Check that no text is clipped, truncated, or obscured.

Here the text isn't clipped, truncated or obscured, it is intentionally removed.

However, I'm no WCAG expert so if the designers says the icon is self-explanatory enough, then I'm good with that :+1:

rosvik:
I think we can argue it's clear enough as is, looking at https://www.w3.org/WAI/tutorials/forms/labels/

image

But I guess that's up to interpretation. It's also an option to wrap the controls vertically. (Previous, date, and next -buttons on three lines) What do you think @oyvindmikkelsenatb / @ckbilstad ?

rosvik:
>  Moving the back button to show above the title ("Billetter" and "Prinses gate) on small screens.

I think this is a bigger task than it looks, as there seems to be quite a bit of effort put into the solution as it is. See https://github.com/AtB-AS/mittatb-app/blob/master/src/components/screen-header/ScreenHeader.tsx#L197:L197

Not sure it's worth doing, or if there's a good reason for it being the way it is.

oyvindmikkelsenatb:
**Arrows with labels**
Based on w3s text about lables opinion is that our arrows without lables is enough since the "calendar and time" button in the middle gives context.

**Back button**
I agree that moving the back button on top does not need to be worked on in this issue.

**Bottom navigation**
Will you be able to make the text not break @rosvik?

rosvik:
@oyvindmikkelsenatb Nice! I could have a look at limiting the nav bar text size, but will do in its own PR.

PR #3250, opened 2023-01-27
feat: deeplink from each departure on medium widget

adriawh:
Deeplinks in medium widget:
1. Can now click individual departures on the medium widget. 
2. This redirects you to the details screen of the clicked departure.

Closes https://github.com/AtB-AS/kundevendt/issues/3289

### DIFF
--- a/ios/departureWidget/Views/ChipView.swift
+++ b/ios/departureWidget/Views/ChipView.swift
@@ -1,15 +1,17 @@
 import SwiftUI
 
 struct ChipView: View {
-    let label: String
+    let departure: DepartureLinkLabel
 
     var body: some View {
-        Text(label)
-            .lineLimit(1)
-            .fixedSize()
-            .padding(8)
-            .background(Color("TimeTileBackgroundColor"))
-            .cornerRadius(8)
-            .font(DefaultFonts.bold)
+        Link(destination: URL(string: departure.link)!) {
+            Text(departure.label)
+                .lineLimit(1)
+                .fixedSize()
+                .padding(8)
+                .background(Color("TimeTileBackgroundColor"))
+                .cornerRadius(8)
+                .font(DefaultFonts.bold)
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/stacks-hierarchy/RootStack.tsx
```diff
@@ -63,7 +63,43 @@ export const RootStack = () => {
     const params = parse(path);
     let destination: PartialRoute<any>[] | undefined;
 
-    if (departuresV2Enabled) {
+    if (path.includes('details')) {
```
jorelosorio, line 66: The code looks good in general is would just suggest that:
```
{
          name: 'Departures_DepartureDetailsScreen',
          params: {
            activeItemIndex: 0,
            items: [
              {
                serviceJourneyId: params.serviceJourneyId,
                serviceDate: params.serviceDate,
                date: new Date(),
                fromQuayId: params.quayId,
              },
            ],
          },
```

could be added if details are present, otherwise, the code before that looks similar to the next condition.

jorelosorio: Is this details stuff applies to both v1 and v2? or is it just a v2 thing? in such cases, the `departuresV2Enabled` should be evaluated.

adriawh: Should be solved in the latest commit

ios/departureWidget/Views/ChipView.swift
```diff
@@ -1,15 +1,17 @@
...
 
     var body: some View {
-        Text(label)
-            .lineLimit(1)
-            .fixedSize()
-            .padding(8)
-            .background(Color("TimeTileBackgroundColor"))
-            .cornerRadius(8)
-            .font(DefaultFonts.bold)
+        Link(destination: URL(string: departure.link)!) {
```
jorelosorio, line 7: Would be nice to unwrap this `URL(string: departure.link)!` instead of force unwrap.

adriawh: This works?

jorelosorio (approved):

PR #3251, opened 2023-01-27
refactor: Add the option to force a build on staging

jorelosorio:
This adds a new input var to force a build in the staging action.


.github/workflows/build-staging-ios.yml
```diff
@@ -60,7 +69,7 @@ jobs:
           GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
       - name: Replace ipa bundle
-        if: steps.ipa-cache.outputs.cache-hit == 'true'
+        if: inputs.force-build == 'true' || steps.ipa-cache.outputs.cache-hit == 'true'
```
tormoseng, line 72: The other changes are logical to me, but what does `inputs.force-build == 'true' || ...` implies here?

jorelosorio: It should run it in any case but then I looked at another build, for example, the cache was changed and this is skipped, so I changed the logic a bit to fit.

tormoseng (approved):
LGTM!

tormoseng:
Merged on behalf of @jorelosorio

PR #3254, opened 2023-01-30
fix: Added "..." to the fade and found a better solution to how the fade is created. 

adriawh:
After some input from @cbrevik, @oyvindmikkelsenatb and user-tests we decided to add "..." to the fade. 

Also removes the small widget. All the code for it is till there, but it will not be available for the users.


jorelosorio (approved):
good work, we had a talk about what this implementation is about :)

PR #3255, opened 2023-01-30
refactor: Add info chip component

gorandalum:
Add a new shared component "InfoChip" which replaces the similar component used on recent fare product and on fare contract.

The InfoChip component is styled with the active color of an interactive color.

Screenshot 2023-01-31 at 11 08 52


rosvik:
Would it be a good idea to add a couple of examples of the component to the design system? (And also add some screenshots to the PR? 😇)

rosvik (approved):
Meant to leave a review as well. Code looks good 👍

PR #3257, opened 2023-01-31
fix: add nfk google project auth info

toretefre:
Values come from GoogleService-Info.plist, so would require extra steps to pass it in as a dynamic ENV value. Should not have any consequences to have more than one value here, added through Xcode.

Solves https://github.com/AtB-AS/kundevendt/issues/3358


gorandalum (changes requested):
In the script `override-config-files.sh` we have this section:
```
echo "Adding reversed client id as custom URL scheme to enable reCAPTCHA"
REVERSED_CLIENT_ID=`/usr/libexec/PlistBuddy -c "Print :REVERSED_CLIENT_ID" ./ios/atb/GoogleService-Info.plist`
/usr/libexec/PlistBuddy -c "Set :CFBundleURLTypes:1:CFBundleURLSchemes:0 $REVERSED_CLIENT_ID" ./ios/atb/Info.plist
```

This snippet replaces the Google reversed client id in the Info.plist when building for staging/store. Before it replaced the only existing dev value with the correct value for staging/store. With this change now the first dev value (the one for AtB) is replaced, while the dev value for nfk still remains in the staging/store build. 

Even though this will probably work, I think it would be cleaner and more correct to end up with just the staging/store value when building for staging/store. Could you take a quick look at updating the script to do this?

toretefre:
Think I've solved it now. Could we possibly also remove this part of the `override-config-files.sh` script now, or is there a reason to keep?

Was considering just removing the second array element with the `override-config-files.sh` script, but to me that's less clean than just relying on .env-values like we do for most other values.

gorandalum (changes requested):
The *GoogleService-Info.plist* is downloaded directly from Firebase, and added into the project. This file contains the `REVERSED_CLIENT_ID` which must be added to the *Info.plist*. 

The meaning of doing it like it was in the *override-config-files.sh* is that the *GoogleService-Info.plist* is the single source of truth for this value. With this change now we need to keep the value in *GoogleService-Info.plist* files and in the *.env* files in sync.

Even though it is less clean to remove the second array element with a script, I think it still is preferred as there is less risk later on that something wrong happens because we at some point forget to keep these values in sync.

I'm sorry for the back and forth 😅

toretefre:
> I'm sorry for the back and forth 😅

No worries, good to ensure quality in what we're doing!

I will then:
- Add dev reversed_ids directly from both AtB, NFK and later FRAM directly into the array in Xcode.
- Ensure the `override-config-files.sh` script removes all local entries when adding the staging / store value.

toretefre:
Should work as specified now. Deleting and recreating the array to support even more ids, such as FRAM, that will be inserted at a later stage.

gorandalum (approved):
:+1:

PR #3259, opened 2023-02-01
test: e2e update detox

tormoseng:
After dependency updates of the app, Detox is updated to v.20. Here, Detox is updated based on the migration guide. In addition, the dev version of the app is used since I didn't get the staging version to run in Detox.


scripts/override-environment.sh
```diff
@@ -33,7 +33,21 @@ else
...
+    head -2 .env | tail -1
+    echo "- dev env"
+    head -2 env/atb/dev/.env | tail -1
+    echo "- locale: "
+    locale
+
     cp $ENV_FOLDER/.env ios/Configs/Config.xcconfig
+    # TODO REMOVE PRINTING after testing
+    echo "- Config.xcconfig"
+    head -2 ios/Configs/Config.xcconfig | tail -1
```
rosvik, line 50: Should this be removed later, or was it included by mistake?

rosvik (approved):

scripts/override-environment.sh
tormoseng: Only for debug purposes :-)

tormoseng:
Then! Finally, after so much tries and tweaks, all tests run OK. Let's hope its not a once only. https://github.com/AtB-AS/mittatb-app/actions/runs/4194774122

It seems that it was the change from `macos-12` -> `macos-11` that did it. Xcode 13.2 instead of 14.0.

PR #3258, opened 2023-02-01
feat: Add support for wide transportation icons

ingvisim:
Add support for wide transportation icons for new travel search view

https://github.com/AtB-AS/kundevendt/issues/3288

PR #3263, opened 2023-02-01
chore: Move margin out of transportation icon component

ingvisim:
Make transportation icon take a style prop and move margin out

https://github.com/AtB-AS/mittatb-app/pull/3258#discussion_r1093077064

https://github.com/AtB-AS/kundevendt/issues/3288


gorandalum (approved):
The code looks good! I haven't tested the different places in the app, but from reading the code it seems that the "inhouse" margin from the `TransportationIcon` some places are not replaced in the new style param. You probably have control, just a heads up in case it was forgot somewhere.

PR #3265, opened 2023-02-01
fix: use theme color for login spinner

adriansberg:
There was no color defined for the spinner, so it defaulted to a gray color. I set the color to the primary text color since that has been used for the spinner in other places as well.

I also fixed the casing of the phone input heading, in english.

Android:

image
image (😅)

iOS:

image
image

Fixes https://github.com/AtB-AS/kundevendt/issues/3059

### DIFF
--- a/src/translations/screens/subscreens/Login.ts
+++ b/src/translations/screens/subscreens/Login.ts
@@ -61,7 +61,7 @@ const LoginTexts = {
       'Log in with a one-time code sent to your phone',
     ),
     input: {
-      heading: _('Mobilnummer', 'phone number'),
+      heading: _('Mobilnummer', 'Phone number'),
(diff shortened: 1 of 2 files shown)


src/translations/screens/subscreens/Login.ts
```diff
@@ -61,7 +61,7 @@ const LoginTexts = {
       'Log in with a one-time code sent to your phone',
     ),
     input: {
-      heading: _('Mobilnummer', 'phone number'),
+      heading: _('Mobilnummer', 'Phone number'),
```
rosvik, line 64: Nice! I've seen this so many times, but never when I was able to fix it, and then forgot 😅

rosvik (approved):

PR #3270, opened 2023-02-02
fix: Not log GRPC errors in axios client

gorandalum:
Some GRPC errors are not errors per se, but a part of a normal expected flow, for example the error signifying that the token needs renewal. Other errors that really are errors should be handled and logged in the mobile token specific code.

### DIFF
--- a/src/mobile-token/tokenService.ts
+++ b/src/mobile-token/tokenService.ts
@@ -1,6 +1,7 @@
 import {
   ActivatedToken,
   GrpcError,
+  GrpcErrorType,
   handleReattest,
   handleRemoteError,
   RemoteTokenServiceWithInitiate,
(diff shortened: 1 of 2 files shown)


gorandalum:
Tested locally. Before adding the `skipErrorLogging` two errors were logged on failed activation:
- `AxiosError: Request failed with status code 500`
- `Error: 16 UNAUTHENTICATED: Attestation failed: Unknown application for attestation: codespace ATB, applicationId no.mittatb.debug, applicationVersion 1.32`

After the change only one was logged:
- `Error: 16 UNAUTHENTICATED: Attestation failed: Unknown application for attestation: codespace ATB, applicationId no.mittatb.debug, applicationVersion 1.32`

src/mobile-token/tokenService.ts
```diff
@@ -201,6 +214,7 @@ const service: TokenService = {
             },
             authWithIdToken: true,
             timeout: 15000,
+            // skipErrorLogging: (err) => isGrpcErrorType(err.response?.data),
```
rosvik, line 217:
```suggestion
```

rosvik (approved):

src/mobile-token/tokenService.ts
gorandalum, line 217:
```suggestion
            skipErrorLogging: (err) => isGrpcErrorType(err.response?.data),
```

gorandalum: Opps. Should actually be like this:
```suggestion
            skipErrorLogging: (err) => isGrpcErrorType(err.response?.data),
```

PR #3275, opened 2023-02-03
fix: Disable widget listener on App if widget is disabled

jorelosorio:


gorandalum (approved):

tormoseng:
Tested the newest build, and location is no longer active when the app is in the background.

PR #3279, opened 2023-02-06
chore: add feature switch for realtime map

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/3403

image

### DIFF
--- a/src/components/map/hooks/use-realtime-map-enabled.ts
+++ b/src/components/map/hooks/use-realtime-map-enabled.ts
@@ -3,11 +3,11 @@ import {StorageModelKeysEnum} from '@atb/storage';
 import {useDebugOverride} from '@atb/debug';
 
 export const useRealtimeMapEnabled = () => {
-  const [debugOverride, setDebugOverride] = useRealtimeMapDebugOverride();
+  const [debugOverride] = useRealtimeMapDebugOverride();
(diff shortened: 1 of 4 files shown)


src/components/map/hooks/use-realtime-map-enabled.ts
```diff
@@ -0,0 +1,17 @@
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {StorageModelKeysEnum} from '@atb/storage';
+import {useDebugOverride} from '@atb/debug';
+
+export const useRealtimeMapEnabled = () => {
+  const [debugOverride, setDebugOverride] = useRealtimeMapDebugOverride();
+  const {enable_realtime_map: enabledInRemoteConfig} = useRemoteConfig();
+
+  if (debugOverride !== undefined) {
+    return [debugOverride, setDebugOverride];
```
gorandalum, line 10: Wondering about this `return [debugOverride, setDebugOverride]` here. This means the return type for `useRealtimeMapEnabled` will be `[boolean, (boolean) => void] | boolean`?

rosvik: Oups, had some uncommited changes

gorandalum (approved):

rosvik:
For testing: Not much to test here, since the buttons have no effect. Probably makes more sense to test later.

PR #3281, opened 2023-02-07
feat: Update screenreader

ingvisim:
Close https://github.com/AtB-AS/kundevendt/issues/3288


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -111,6 +111,7 @@ const Results: React.FC<Props> = ({
               }}
               searchTime={searchTime}
               testID={'tripSearchSearchResult' + i}
+              listPosition={i + 1}
```
rosvik, line 114: Think it makes more sense to name this `resultNumber` or something like that

rosvik (approved):
Nice!

PR #3283, opened 2023-02-08
feat: tick when focused

rosvik:
Suggested solution to https://github.com/AtB-AS/kundevendt/issues/3250

### DIFF
--- a/src/place-screen/hooks/use-quay-data.ts
+++ b/src/place-screen/hooks/use-quay-data.ts
@@ -293,6 +293,9 @@ export function useQuayData(
     [],
     !isFocused || mode !== 'Departure',
   );
+  useEffect(() => {
+    if (isFocused) dispatch({type: 'TICK_TICK'});
+  }, [isFocused]);
 
   return {
     state,
--- a/src/place-screen/hooks/use-stop-place-state.ts
+++ b/src/place-screen/hooks/use-stop-place-state.ts
@@ -287,6 +287,9 @@ export function useStopPlaceData(
     [],
     !isFocused || mode !== 'Departure',
   );
+  useEffect(() => {
+    if (isFocused) dispatch({type: 'TICK_TICK'});
+  }, [isFocused]);
 
   return {
     state,
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/use-favorite-departure-data.ts
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/use-favorite-departure-data.ts
@@ -274,6 +274,9 @@ export function useFavoriteDepartureData(
       loadInitialDepartures();
     }
   }, [state.tick, state.lastRefreshTime]);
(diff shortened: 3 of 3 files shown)


src/place-screen/hooks/use-quay-data.ts
```diff
@@ -293,6 +293,16 @@ export function useQuayData(
...
+  useEffect(() => {
+    if (!isFocused || !state.tick) return;
+
+    const timeSinceLastTick = differenceInSeconds(Date.now(), state.tick);
+    if (timeSinceLastTick / 60 >= HARD_REFRESH_LIMIT_IN_MINUTES) {
+      loadDepartures();
+    } else if (timeSinceLastTick >= updateFrequencyInSeconds) {
+      dispatch({type: 'LOAD_REALTIME_DATA', quay});
+    }
+  }, [isFocused]);
```
mikaelbr, line 305: These looks exactly the same with the same goal? I think if so this is the perfect opportunity to create a new custom hook? I think in general we can be better at having smaller custom hooks

rosvik: I though about it, but also thought it would be messy with all the params. But tried it now, and it turned out a lot cleaner ✨

src/utils/use-refresh-on-focus.ts
```diff
@@ -0,0 +1,23 @@
+import {useIsFocused} from '@react-navigation/native';
+import {differenceInSeconds} from 'date-fns';
+import {useEffect} from 'react';
+
+export function useRefreshOnFocus(
+  tick: Date | undefined,
```
gorandalum, line 6: Can this be called `lastUpdate` or something? If not I think it can be documented. It is kinda hard by reading just the parameters what `tick` actually means.

src/utils/use-refresh-on-focus.ts
```diff
@@ -0,0 +1,23 @@
...
+  useEffect(() => {
+    if (!isFocused || !tick) return;
+    const timeSinceLastTick = differenceInSeconds(Date.now(), tick);
+
+    if (timeSinceLastTick >= hardRefreshLimitInSeconds) {
+      hardRefresh();
+    } else if (timeSinceLastTick >= softRefreshLimitInSeconds) {
+      softRefresh();
+    }
+  }, [isFocused]);
```
gorandalum, line 22: Seems kinda strange to have so many parameters used in the `useEffect` which is not in the dependency list. Might be some real reason for it, but just commenting it here, so you can take an extra look if it is necessary to exclude them.

gorandalum (approved):
Looks good :+1: Some comments you can take a look at.

src/utils/use-refresh-on-focus.ts
gorandalum: Generally the rule is that they should be provided:
https://beta.reactjs.org/reference/react/useEffect#specifying-reactive-dependencies

React provides a lint rule "exhaustive-deps" to check for this, but we haven't enabled it since we break it so many places.

rosvik: 👍 Updated now

PR #3296, opened 2023-02-13
feat: support external links for ticketing info in my profile

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/3046

Makes it possible to provide urls in Firestore for the information that AtB are providing by embedding text into the code of the app. Should be unchanged for AtB if urls are not provided.

In the unlikely event that the app should function while being offline, a fallback generic page (found in Screenshots) have been implemented for other orgs than AtB.

<details>

NFK (notice external link icon):
9357A8B1-9C44-48F9-B56C-D269863D7905
AtB:
0936704C-7C8A-4440-B699-BC59EBAD8055

Norwegian fallback page:
9336CC3A-CC33-4A9F-A8AD-80F9C2C9B149
English fallback page:
BD29850A-1D1F-40AF-90A0-F2232FC45B73

</details>

### DIFF
--- a/src/configuration/FirestoreConfigurationContext.tsx
+++ b/src/configuration/FirestoreConfigurationContext.tsx
@@ -39,7 +39,7 @@ export type AppTexts = {
   discountInfo: LanguageAndTextType[];
 };
 
-type ProfileInfoUrls = {
+type ConfigurableLinks = {
   ticketingInfo: LanguageAndTextType[];
   termsInfo: LanguageAndTextType[];
   inspectionInfo: LanguageAndTextType[];
@@ -55,7 +55,7 @@ type ConfigurationContextState = {
   fareProductTypeConfigs: FareProductTypeConfig[];
   travelSearchFilters: TravelSearchFiltersType | undefined;
   appTexts: AppTexts | undefined;
-  profileInfoUrls: ProfileInfoUrls | undefined;
+  configurableLinks: ConfigurableLinks | undefined;
 };
 
 const defaultConfigurationContextState: ConfigurationContextState = {
@@ -68,7 +68,7 @@ const defaultConfigurationContextState: ConfigurationContextState = {
   fareProductTypeConfigs: defaultFareProductTypeConfig,
   travelSearchFilters: undefined,
   appTexts: undefined,
-  profileInfoUrls: undefined,
+  configurableLinks: undefined,
 };
 
 const FirestoreConfigurationContext = createContext<ConfigurationContextState>(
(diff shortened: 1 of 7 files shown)


gorandalum:
> Need to think about what will happen for offline users at NFK if we're using Firestore here. Will they be presented with the AtB content?

I wouldn't necessarily put too much thought into it. In general if something is not working with the Firestore connection then all sorts of error happens, which will make these wrong info texts the least of the problem. We have thought/planned just having a generic loading error at start if the firestore connection doesn't get up and running.

toretefre:
> I wouldn't necessarily put too much thought into it. In general if something is not working with the Firestore connection then all sorts of error happens, which will make these wrong info texts the least of the problem. We have thought/planned just having a generic loading error at start if the firestore connection doesn't get up and running.

Oh, okay, I would say that's a change of prerequisites for a bigger part of the app than just this little bit, but very good to know, thank you. Might just leave this as it is then.

Is there a separate issue for the proposed Firestore loading error functionality where we can have a look at the consequences, @gorandalum?

gorandalum:
> Oh, okay, I would say that's a change of prerequisites for a bigger part of the app than just this little bit, but very good to know, thank you. Might just leave this as it is then.
> 
> Is there a separate issue for the proposed Firestore loading error functionality where we can have a look at the consequences, @gorandalum?

Not a specific issue as of now.

toretefre:
Should be ready for both NFK and AtB now. For FRAM it should be to just add the orgSpecificTranslation for the generic fallback page. If AtB does not add urls to Firestore everything should behave as before.

In the unlikely event that anyone should be able to open the NFK app without a connection to Firestore (either now or in the future if we fix offline support), a generic page stating that current information can be found on the homepage of the authority should be available.

toretefre:
Skal fikse litt på layouten på den generiske skjermen så det ikke blir slik:

src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -49,6 +55,7 @@ type ConfigurationContextState = {
   fareProductTypeConfigs: FareProductTypeConfig[];
   travelSearchFilters: TravelSearchFiltersType | undefined;
   appTexts: AppTexts | undefined;
+  profileInfoUrls: ProfileInfoUrls | undefined;
```
gorandalum, line 58: I thin I would prefer just calling this "urls" or "links" in the firestore-configuration, to be able to reuse it for all links in the app.

gorandalum (approved):
Looks good. One comment about naming, but it is optional to agree 😅

src/configuration/FirestoreConfigurationContext.tsx
toretefre: Agree, renamed to `configurableLinks` and keeping `urls` in Firestore.

PR #3297, opened 2023-02-14
feat: Allow storing filters

ingvisim:
Make it possible to store filters for future travel searches

<details>
<summary>Screenshot</summary>

Simulator Screen Shot - iPhone 14 Pro - 2023-02-13 at 13 42 53
</details>

Close https://github.com/AtB-AS/kundevendt/issues/3108


src/travel-search-filters/storage.ts
```diff
@@ -0,0 +1,25 @@
+import storage, {StorageModelTypes} from '../storage';
+import {TransportModeFilterOptionWithSelectionType} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types';
+
+class FilterStore<T = TransportModeFilterOptionWithSelectionType> {
```
gorandalum, line 4: Do we need the generics here, since it is hardcoded to a specific type.

src/travel-search-filters/FiltersContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+const FiltersContext = createContext<FiltersContextState | undefined>(
+  undefined,
+);
+
+const FiltersContextProvider: React.FC = ({children}) => {
+  const [filters, setFiltersState] = useState<
+    TransportModeFilterOptionWithSelectionType[]
+  >([]);
+  async function populateFilters() {
+    const [filters] = await Promise.all([storedFilters.getFilters()]);
```
gorandalum, line 20: By first glanse I don't understand the need for `Promise.all`.

src/travel-search-filters/FiltersContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+    setFiltersState(filters ?? []);
+  }
+
+  useEffect(() => {
+    populateFilters();
```
gorandalum, line 25: Seems like `populateFilters` is only used here? In that case I think it is better to inline it. In general `useEffect` functions which runs only on first render shouldn't use reactive values.

src/travel-search-filters/FiltersContext.tsx
```diff
@@ -0,0 +1,51 @@
+import React, {createContext, useContext, useEffect, useState} from 'react';
+import {storedFilters} from './storage';
+import {TransportModeFilterOptionWithSelectionType} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types';
+
+type FiltersContextState = {
+  filters: TransportModeFilterOptionWithSelectionType[];
+  setFilters(
+    filters: TransportModeFilterOptionWithSelectionType[] | undefined,
+  ): Promise<void>;
```
gorandalum, line 9: I see that where this is used we don't await the promise. Since the caller doesn't care or want to know that this returns a `Promise`, maybe we can just make it a void-function?

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import {View} from 'react-native';
+import {Confirm} from '@atb/assets/svg/mono-icons/actions';
+import {StyleSheet} from '@atb/theme';
+
+type CheckedProps = {
+  checked: boolean;
+};
+
+export const SavedCheckbox: React.FC<CheckedProps> = ({checked}) => {
```
gorandalum, line 10: Just call this `Checkbox`?

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const styles = useStyles();
+  return (
+    <View
+      style={[
```
gorandalum, line 14: Should add `accessibilityRole` and such on this component. The containers wrapping this component as of now have the accessibility setup, but I think now that this Checkbox is a resusable component under `src/components` it should be made good enough to be a "standalone" checkbox. Maybe add it to the DesignSystem screen as well?

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+        styles.saveCheckbox,
+        checked ? styles.saveCheckboxChecked : styles.saveCheckboxDefault,
+      ]}
+    >
+      {checked ? <Confirm fill="white"></Confirm> : null}
```
gorandalum, line 19: `<Confirm fill="white" />`

gorandalum (dismissed):
Looks mainly good! Some comments to take a look at.

rosvik (approved):
LGTM 👍 Works well for me locally.

PR #3299, opened 2023-02-15
Click on Scooter to display details and switch to operator app

KristianSelnas:


src/components/map/utils.ts
```diff
@@ -60,6 +60,10 @@ export const findStopPlaceAtClick = async (
 export const isFeaturePoint = (f: Feature): f is Feature<Point> =>
   f.geometry.type === 'Point';
 
+export const isClusterFeature = (
+  feature: Feature,
+): feature is Feature<Point, Cluster> => feature.properties?.cluster;
```
gorandalum, line 65: Based on what is checked by this utility method we can't say with certainty that the `Feature` is of type `Point`, just that it has properties of type `Cluster`. In reality we might know that every `Feature` with `Cluster` properties also are of type `Point`, but I think it is that the utility methods are free of presumptions like this.

Maybe make the check `=> isFeaturePoint(feature) && feature.properties?.cluster`?

gorandalum (approved):
Looks good :+1:

KristianSelnas:
@gorandalum It's currently just a draft pr, just to make it simpler for my self to keep track of change. I'll notify you when it's ready for final review.

gorandalum:
> @gorandalum It's currently just a draft pr, just to make it simpler for my self to keep track of change. I'll notify you when it's ready for final review.

Aight. Can also "officially" make it a draft by using this:
Screenshot 2023-02-15 at 12 49 42

PR #3301, opened 2023-02-15
feat: Firestore traveler category description

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/3480 

Using the changes from https://github.com/AtB-AS/webshop2/pull/270, modified to work in the app. 

Ensures that passenger category descriptions for ticketing are used from Firestore referenceData if it exists.

### DIFF
--- a/src/reference-data/defaults/user-profiles.json
+++ b/src/reference-data/defaults/user-profiles.json
@@ -10,6 +10,16 @@
       "lang": "nob",
       "value": "Fra 20 til og med 66 år."
     },
+    "alternativeDescriptions": [
+      {
+        "lang": "eng",
+        "value": "20 to 66 years."
+      },
+      {
+        "lang": "nob",
+        "value": "20 til og med 66 år."
+      }
+    ],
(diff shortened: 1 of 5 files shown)


toretefre:
> `alternativeDescriptions` in bokmål and english (duplicate bokmål translations?).

This one was created with the linked webshop PR, and I think we should keep it consistent with the webshop. So either we need to rename it both in the app and the webshop, or we can keep it. What do you think?

rosvik (approved):
Although a bit hacky, considering this is consistent with the webshop, LGTM 👍

toretefre:
> Although a bit hacky, considering this is consistent with the webshop, LGTM 👍

A bit hacky it is, but as a lot of permanent code it is only intended to be there temporary 🎉 

As Gøran [mentioned in the webshop PR](https://github.com/AtB-AS/webshop2/pull/270#pullrequestreview-1281863422) there will come a rule engine handling this at a later stage.

Merging 🚀

PR #3303, opened 2023-02-16
fix: filter realtime departures by line

rosvik:
Together with https://github.com/AtB-AS/atb-bff/pull/224, this PR adds filtering on lines when fetching realtime departures on the dashboard, and when favorites are toggled in departures. (Also a bit of related refactoring)

resolves https://github.com/AtB-AS/kundevendt/issues/3252

### DIFF
--- a/src/api/departures/index.ts
+++ b/src/api/departures/index.ts
@@ -10,12 +10,12 @@ import {
   DeparturesRealtimeData,
   PaginationInput,
 } from '@atb/sdk';
-import {flatMap} from '@atb/utils/array';
+import {flatMap, uniqueItems} from '@atb/utils/array';
 import client from '../client';
 import {
   DepartureFavoritesQuery,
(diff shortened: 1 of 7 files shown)


src/api/departures/index.ts
```diff
@@ -37,30 +37,45 @@ export async function getDepartures(
...
-  if (!quayIds || quayIds.length === 0) return {};
-  const startTime = query.startTime;
+  if (query.quayIds.length === 0) return {};
 
   const params = build({
-    quayIds,
-    startTime,
-    limit: query.limitPerLine,
+    ...query,
+    quayIds: uniqueItems(query.quayIds),
```
gorandalum, line 74: We have an existing helper function `onlyUniques` that can be used like this:
`query.quayIds.filter(onlyUniques)`

gorandalum (approved):

PR #3307, opened 2023-02-16
Show vehicles only for larger zooms

KristianSelnas:


src/components/map/Map.tsx
```diff
@@ -48,7 +48,13 @@ export const Map = (props: MapProps) => {
   ) => {
     if (!props.vehicles) return;
     const [longitude, latitude] = feature.geometry.coordinates;
-    props.vehicles.fetchVehicles({longitude, latitude});
+    const zoom = feature.properties.zoomLevel;
+    const range = getVisibleRange(feature.properties.visibleBounds);
+    props.vehicles.fetchVehicles({
+      coordinates: {longitude, latitude},
+      zoom,
+      radius: range / 2,
```
gorandalum, line 56: Is there any limitation of this radius somewhere/somehow? If a user zooms out so all of Norway is visible we probably don't want all scooters in Norway returned from the query.

gorandalum (approved):

src/components/map/Map.tsx
KristianSelnas: This is handled by `MIN_ZOOM_LEVEL` in `use-vehicles.ts`

PR #3309, opened 2023-02-17
fix: widget options and fix in location battery drain

adriawh:
- Added widget options
- Small UI fixes from user tests
- Change in location usage, should fix battery drain


jorelosorio (approved):

tormoseng:
@adriawh @jorelosorio We're back to the state where the app is constantly using the location - even though I havn't added the widget. Not sure if the battery is affected yet, but at least iOS says that the app is using the location also with the app in background.

PR #3316, opened 2023-02-20
fix: enable realtime calls for Map bottom sheet

rosvik:
ref. https://github.com/AtB-AS/kundevendt/issues/3250#issuecomment-1434431109


ingvisim (approved):

tormoseng (approved):
LGTM!

PR #3319, opened 2023-02-20
fix: Make save-button part of scrollview

ingvisim:
https://github.com/AtB-AS/kundevendt/issues/3108#issuecomment-1437010479
https://github.com/AtB-AS/kundevendt/issues/3108#issuecomment-1437026694


rosvik (approved):
👍👍

PR #3321, opened 2023-02-22
fix: removed blue icon when app is running in background with widget enabled

adriawh:
Hopefully the last fix of the widget location and battery drain.


rosvik (approved):
LGTM 👍

PR #3326, opened 2023-02-23
feat: Add screenreader message about cancelled departures

ingvisim:
https://github.com/AtB-AS/kundevendt/issues/3271

### DIFF
--- a/src/departure-list/section-items/line.tsx
+++ b/src/departure-list/section-items/line.tsx
@@ -190,37 +190,40 @@ function getAccessibilityTextFirstDeparture(
   );
 
   const inPast = isInThePast(firstResult.time);
-  const upcoming = inPast
-    ? t(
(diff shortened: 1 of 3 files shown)


src/departure-list/section-items/line.tsx
```diff
@@ -217,7 +218,8 @@ function getAccessibilityTextFirstDeparture(
                     NearbyTexts.results.departure.nextAccessibilityRealtime(
                       labelForTime(i.time, searchDate, t, language, true),
                     ),
-                  )
+                  ) +
+                  (i.cancellation ? t(CancelledDepartureTexts.cancelled) : '')
```
rosvik, line 222: Should the cancellation label be dependent on realtime=true here?

rosvik (approved):

PR #3331, opened 2023-02-24
chore: Set up application for FRAM

adriansberg:
Adds dev and staging for FRAM.

Changes the configuration for iOS so that the bundle identifiers are set correctly and dynamically (https://github.com/AtB-AS/kundevendt/issues/3295). Also changes it so that the development team is set based on the .env. 

Note: all of the variables in .env need to be populated for running the app in Android.

Fixes https://github.com/AtB-AS/kundevendt/issues/3404, fixes https://github.com/AtB-AS/kundevendt/issues/3402


adriansberg:
In relation to https://github.com/AtB-AS/kundevendt/issues/3295, I did not touch the `.entitlements` files and application groups.

ios/atb.xcodeproj/project.pbxproj
```diff
@@ -776,7 +776,7 @@
 				CODE_SIGN_STYLE = Manual;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = "";
-				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = CSV447Q97L;
+				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = "$(IOS_DEVELOPMENT_TEAM_ID)";
```
mikaelbr, line 779: Wasn't it this that is replaced by fastline @toretefre ?

toretefre: Yes, but Fastlane does not help for local development, so this approach is better / complementary if it works

mikaelbr: @adriansberg is it verified with all orgs (atb/nfk)?

adriansberg: I guess what's set for development team doesn't matter locally then?

adriansberg: My previous comment was made with only seeing the first one somehow. Still somewhat relevant though, what does the development group do for local development? Have not verified it other than seeing that the values are set according to what's defined in the `.env` file.

jorelosorio (approved):
LGTM, I had a branch where I was planning to change a couple of things also. Nice work!

PR #3338, opened 2023-03-02
feat: Use goBack function to go back to correct page on cancel

ingvisim:
Use goBack function to go back to correct page on cancel, for instance back to trip details when purchasing ticket from travel search

Close https://github.com/AtB-AS/kundevendt/issues/3540


KristianSelnas:
Looks reasonable to me 👍

KristianSelnas (approved):

PR #3342, opened 2023-03-03
fix: Removed left padding on error message box in trip search

hanne-at-atb:
Closes: https://github.com/AtB-AS/kundevendt/issues/3579


ingvisim (changes requested):
This seems to remove the padding on other containers on the trip search page. Might have to create a new style for the message box with only bottom padding

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -73,7 +73,7 @@ const Results: React.FC<Props> = ({
 
   if (errorType) {
     return (
-      <View style={styles.container}>
```
jorelosorio, line 76: Looks good however `styles.container` was only used here, so it might not be in use by other components.

jorelosorio (approved):

ingvisim (approved):

PR #3343, opened 2023-03-06
feat: Scooter sheet redesign

KristianSelnas:
https://github.com/AtB-AS/kundevendt/issues/3545

### DIFF
--- a/src/vehicles/components/ScooterSheet.tsx
+++ b/src/vehicles/components/ScooterSheet.tsx
@@ -160,7 +160,7 @@ const getRange = (rangeInMeters: number, language: Language) => {
     rangeInMeters > 5000
       ? (rangeInMeters / 1000).toFixed(0)
       : formatDecimalNumber(rangeInMeters / 1000, language, 1);
-  return `ca ${rangeInKm} km`;
+  return `ca. ${rangeInKm} km`;
 };
 
 const useSheetStyle = StyleSheet.createThemeHook((theme) => ({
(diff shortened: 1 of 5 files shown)


src/vehicles/components/ScooterSheet.tsx
```diff
@@ -144,27 +160,34 @@ const getRange = (rangeInMeters: number, language: Language) => {
...
 
 const useSheetStyle = StyleSheet.createThemeHook((theme) => ({
   button: {
     marginTop: theme.spacings.medium,
   },
-  container: {
+  vehicleInfo: {
     flexGrow: 1,
     flexShrink: 0,
+    gapCol: theme.spacings.medium,
```
rosvik, line 173: Is this the property that doesn't work yet? If so, should it be removed for now, so it doesn't suddenly start to be behave in unexpected ways?

src/vehicles/components/ScooterSheet.tsx
```diff
@@ -144,27 +160,34 @@ const getRange = (rangeInMeters: number, language: Language) => {
     rangeInMeters > 5000
       ? (rangeInMeters / 1000).toFixed(0)
       : formatDecimalNumber(rangeInMeters / 1000, language, 1);
-  return `ca ${rangeInKm}km`;
+  return `ca ${rangeInKm} km`;
```
rosvik, line 163:
```suggestion
  return `ca. ${rangeInKm} km`;
```

rosvik (approved):
Looks good! Just a couple of nitpicks.

src/vehicles/components/ScooterSheet.tsx
KristianSelnas: You are right. This is just a left over from a previous iteration. I'll remove it.

PR #3344, opened 2023-03-06
fix: Use ticket start time of first non foot leg

ingvisim:
https://github.com/AtB-AS/kundevendt/issues/3539#issuecomment-1455874940


rosvik (approved):

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -186,7 +186,7 @@ function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
   if (!(enable_ticketing && !someLegsAreNotSingleTicket)) return;
 
   const tripStartWithBuffer = addMinutes(
-    parseISO(tripPattern.expectedStartTime),
+    parseISO(nonFootLegs[0]?.aimedStartTime),
```
jorelosorio, line 189: The `parseISO` expect a string as input, but the result we are sending might be nil / undefined, we do not know what it responds when that happens.

PR #3345, opened 2023-03-06
fix: Only put margin on icon in button if there is another element

ingvisim:
Fix issue where there is a margin on right side of buttons on map

Simulator Screen Shot - iPhone 14 Pro - 2023-03-06 at 14 14 13


rosvik (approved):
🙌

tormoseng:
LGTM!

PR #3348, opened 2023-03-07
fix: Background color in scooter sheet

KristianSelnas:


src/vehicles/components/ScooterSheet.tsx
```diff
@@ -175,8 +175,7 @@ const useSheetStyle = StyleSheet.createThemeHook((theme) => ({
   },
   vehicleInfoItem: {
     flex: 1,
-    gap: theme.spacings.medium,
-    backgroundColor: '#fff',
+    backgroundColor: theme.static.background['background_0'].background,
```
hanne-at-atb, line 178:
```suggestion
    backgroundColor: theme.static.background.background_0.background,
```
Any reason for not writing it like this? As used in for instance: https://github.com/AtB-AS/mittatb-app/blob/master/src/components/map/components/LocationBar.tsx#L172

hanne-at-atb (approved):

PR #3352, opened 2023-03-07
feat: Sell tickets from travel search for train in zone a

ingvisim:
https://github.com/AtB-AS/kundevendt/issues/3539#issuecomment-1458129539


src/travel-details-screens/components/use-can-sell-collab-ticket.ts
```diff
@@ -0,0 +1,24 @@
+import {TripPattern} from '@atb/api/types/trips';
+import {APP_ORG} from '@env';
+import {
+  Mode,
+  TariffZone,
+} from '@atb/api/types/generated/journey_planner_v3_types';
+
+export function useCanSellCollabTicket(tripPattern: TripPattern) {
```
rosvik, line 8: Am I missing something, or could this be turned into a regular function, not a hook? If so, it could probably be a part of utils.ts as well.

rosvik (approved):

PR #3355, opened 2023-03-08
feat: Scooter onboarding

KristianSelnas:


src/stacks-hierarchy/Root_MobileTokenOnboarding/MobileTokenOnboarding_FlexibilityInfoScreen.tsx
```diff
@@ -1,4 +1,4 @@
-import TravelPlanning from '@atb/assets/svg/color/images/TravelPlanning';
+import TravelPlanning from '@atb/assets/svg/color/images/light/TravelPlanning';
```
KristianSelnas, line 1: This image was moved, but light and dark images are currently similar. We need to devise a system for automatically selecting the correct image, as is currently done in the webshop. This is currently out of scope of this task.

rosvik: I agree, we should add better support for it. 

Is this the only asset that needs to be updated after the design system update? @mortennordseth

KristianSelnas: Yes, Morten claimed this was the only asset affected.

KristianSelnas: I've created an issue for this, btw: https://github.com/AtB-AS/kundevendt/issues/3621

rosvik (approved):

PR #3363, opened 2023-03-09
feat: Remote config for default map filter

KristianSelnas:


src/remote-config.ts
```diff
@@ -74,6 +75,11 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_from_travel_search_to_ticket: false,
   enable_vehicles_in_map: false,
   enable_vehicle_operator_logo: false,
+  default_map_filter: JSON.stringify({
+    vehicles: {
+      showVehicles: false,
+    },
+  }),
```
gorandalum, line 82: Could you describe a little how this model will grow in the future? When adding city bikes, car sharing, etc.

KristianSelnas: I haven't thought it through in detail, but I believe it can look something like this:
```
{
    vehicles: {
        showVehicles: true
    },
   bikes: {
       someFilterProperty: value
   },
   cars: {...),
   buses: {...)
}
```

gorandalum: It is probably pretty hard to specify a future proof model now, so maybe for now it would be better just to have `default_show_vehicles_in_map` or something in RemoteConfig which may just be `true`/`false`.

gorandalum (approved):
A comment about an alternative way to model default filter, but no strong opinions on whether it is worth doing or not.

src/remote-config.ts
KristianSelnas: I suggest we leave it like this now. I expect it will be refactored soon anyways.

PR #3365, opened 2023-03-09
fix: Start clustering vehicles at a higher zoom level

KristianSelnas:


src/components/map/components/Vehicles.tsx
```diff
@@ -27,7 +27,10 @@ export const Vehicles = ({mapCameraRef, vehicles, onPress}: Props) => {
       ref={shapeSource}
       shape={vehicles}
       tolerance={0}
+      buffer={0}
```
gorandalum, line 30: What happens if `buffer` and `tolerance` are removed?

gorandalum (approved):

src/components/map/components/Vehicles.tsx
KristianSelnas: `tolerance` definitely had an effect when I tested it. Less objects disappeared. I'm not sure about `buffer,` though. I can investigate a bit.

KristianSelnas: Removed the `buffer` option. Performance seems to be sufficient with default buffer value.

PR #3368, opened 2023-03-09
fix: Change letter spacing to avoid line break

ingvisim:
<details>
<summary>Screenshot after fix</summary>

Screenshot_1678369631
</details>

<details>
<summary>Screenshot before fix</summary>

Screenshot_1678371450
</details>


rosvik:
For context: Seems like some letter spacing values sometimes creates line break issues on some Android devices. Exactly which values creates issues seems arbitrary, so this PR updates it to a value that appear to create fewer issues. The issue was noticed on a Pixel 6 Pro, which works as expected now.

(https://github.com/facebook/react-native/issues/35039)

gorandalum (changes requested):
I think I would rather prefer a fix in `ThemeText` for this. Change the letter spacing if the device is Pixel 6 Pro (or a Pixel, or an Android, we decide how wide we want to target this fix).

gorandalum (approved):

HanneLH:
Tested ok på Android og iOS ✅

PR #3375, opened 2023-03-10
refactor: Added index-file to favorites and changed imports

hanne-at-atb:


src/favorites/FavouriteDepartureToggle.tsx
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 import {StyleSheet} from '@atb/theme';
 import SvgFavorite from '@atb/assets/svg/mono-icons/places/Favorite';
 import {StoredType} from '@atb/favorites/storage';
```
gorandalum, line 7: May change this import to just use `./storage`. I know this is inside same module, but if we later introduce eslint rule to assert correct imports this will probably fail.

gorandalum (approved):
Good job :+1:

PR #3378, opened 2023-03-13
fix: Fix Android dark mode toggle

ingvisim:
Make Android toggles visible when dark mode is set in app but not on phone.
<details>
<summary>Screenshot</summary>

Screenshot_1678702266
</details>

Close https://github.com/AtB-AS/kundevendt/issues/2359


gorandalum:
Seems ok for a quick fix now, but should at some point be changed to utilise colors from interactive colors theme palette. For AtB the color background_3 works well with the outline color, but there is no guarantee that this will work for other transport companies.

PR #3380, opened 2023-03-13
fix: Build issues for Intent and Widget

jorelosorio:
Short explanation: Widget and Intent were pointing to still to the pods even dough is not a dependency for them, so that was removed together with extra configurations, and now it builds correctly, issue was raised by the new Mapbox libraries.


adriawh (approved):
Tested the branch and was able to build the app for simulator and physical device. Nice work😎

rosvik (approved):
Looks good! Builds locally for me, and the widget works on my phone 🎉

PR #3381, opened 2023-03-13
chore: Add logging of auth error to bugsnag

ingvisim:
Log auth errors to bugsnag

Close https://github.com/AtB-AS/kundevendt/issues/3258

### DIFF
--- a/src/auth/AuthContext.tsx
+++ b/src/auth/AuthContext.tsx
@@ -137,7 +137,6 @@ export default function AuthContextProvider({children}: PropsWithChildren<{}>) {
         ) {
           return 'invalid_code';
         }
-        console.warn(error);
         if (isAuthError(error)) Bugsnag.notify(error);
         return 'unknown_error';
       }
--- a/src/auth/AuthContext.tsx
+++ b/src/auth/AuthContext.tsx
@@ -9,6 +9,7 @@ import React, {
 import auth, {FirebaseAuthTypes} from '@react-native-firebase/auth';
 import {useTranslation} from '../translations';
 import {updateMetadata} from '@atb/chat/metadata';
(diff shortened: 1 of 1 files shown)


src/auth/AuthContext.tsx
```diff
@@ -137,6 +138,7 @@ export default function AuthContextProvider({children}: PropsWithChildren<{}>) {
           return 'invalid_code';
         }
         console.warn(error);
```
rosvik, line 140: Could probably remove the console.warn, since bugsnag notify basically does the same.

src/auth/AuthContext.tsx
```diff
@@ -136,7 +137,7 @@ export default function AuthContextProvider({children}: PropsWithChildren<{}>) {
         ) {
           return 'invalid_code';
         }
-        console.warn(error);
+        if (isAuthError(error)) Bugsnag.notify(error);
```
jorelosorio, line 140: Wait!, I approved it, but now I saw something, hehe. 

```if (isAuthError(error)) Bugsnag.notify(error);``` Should go first before `if (isAuthError(error) && error.code === ERROR_INVALID_PHONE_NUMBER)` otherwise that case will be ignored and not logged.

**Comment by jorelosorio** [comment]
@ingvisim should this: 
```
const confirmCode = useCallback(
```

be logged?

src/auth/AuthContext.tsx
ingvisim: We might not be interested in knowing whenever someone types the wrong code. What we need for this issue is to know what kind of error type we get when it is not an invalid code, but the user still cannot log in

rosvik (approved):

PR #3384, opened 2023-03-13
feat: Retry Remote Config fetch when the request failed

gorandalum:
- Will retry fetching remote config values every 15 seconds until a
  successful fetch.
- Will only notify error to Bugsnag if 'enable_network_logging' is
  true. If it is not true, a breadcrumb will be sent instead.
- The default value for 'enable_network_logging' was changed to
  false as this is the value that we have in production today.

### DIFF
--- a/src/RemoteConfigContext.tsx
+++ b/src/RemoteConfigContext.tsx
@@ -1,10 +1,14 @@
 import React, {createContext, useContext, useEffect, useState} from 'react';
 import remoteConfig from '@react-native-firebase/remote-config';
-import {RemoteConfig, defaultRemoteConfig, getConfig} from './remote-config';
+import {defaultRemoteConfig, getConfig, RemoteConfig} from './remote-config';
 import Bugsnag from '@bugsnag/react-native';
 
 import {useAppState} from './AppContext';
 import {FeedbackConfiguration} from '@atb/components/feedback';
+import useInterval from '@atb/utils/use-interval';
+
+/** The retry interval when the Remote Config fetch failed */
+const RETRY_INTERVAL_MS = 15000;
 
 export type RemoteConfigContextState = RemoteConfig & {
   refresh: () => void;
@@ -37,6 +41,7 @@ function isUserInfo(a: any): a is UserInfoErrorFromFirebase {
 
 const RemoteConfigContextProvider: React.FC = ({children}) => {
   const [config, setConfig] = useState<RemoteConfig>(defaultRemoteConfig);
+  const [fetchError, setFetchError] = useState(false);
   const {isLoading: isLoadingAppState, newBuildSincePreviousLaunch} =
     useAppState();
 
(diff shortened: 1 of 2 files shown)


src/RemoteConfigContext.tsx
```diff
@@ -1,10 +1,14 @@
...
-import {RemoteConfig, defaultRemoteConfig, getConfig} from './remote-config';
+import {defaultRemoteConfig, getConfig, RemoteConfig} from './remote-config';
 import Bugsnag from '@bugsnag/react-native';
 
 import {useAppState} from './AppContext';
 import {FeedbackConfiguration} from '@atb/components/feedback';
+import useInterval from '@atb/utils/use-interval';
+
+/** The retry interval when the Remote Config fetch failed */
+const RETRY_INTERVAL_MS = 15000;
```
rosvik, line 11: Is 15s a bit much here? I guess we have to account for slow connections at times, but think I would expect the app to retry more often if I lost reception. (And better than nothing for sure, so just asking 😄)

rosvik (approved):

src/RemoteConfigContext.tsx
```diff
@@ -80,6 +91,8 @@ const RemoteConfigContextProvider: React.FC = ({children}) => {
     }
   }, [isLoadingAppState, newBuildSincePreviousLaunch]);
 
+  useInterval(fetchConfig, RETRY_INTERVAL_MS, undefined, !fetchError);
```
jorelosorio, line 94: I know there are some restriction on how often and how many times is allowed to request data from the firebase servers, because it might generate some throttles as described here: https://firebase.google.com/docs/remote-config/get-started?&platform=ios#throttling and some other restrictions for `fetchAndActivate` as mention in the documentation:

```If the time elapsed since the last fetch from the Firebase Remote Config backend is more than the default minimum fetch interval, configs are fetched from the backend.```

The documentation also mentions:

```
Don't rely on network connectivity to obtain Remote Config values. Do set in-app default parameter values so that your app always behaves as expected. You can periodically keep app and Remote Config backend default values in sync using [downloaded template defaults](https://firebase.google.com/docs/remote-config/templates#download_template_defaults).
```
They mention this Template defaults.

So, I am not sure if all this applies for this case because we do not know what `fetchAndActivate` register and / or knows there was a network issue, but I wanted to bring it.

This solution seems correct for me, just a bit aware of side cases.

src/RemoteConfigContext.tsx
gorandalum: Triggering it more often might be better for the user, but it is risky if we actually have some errors in our code, and end up triggering all the time on multiple devices. But I think I will invest some time into doing a quicker first retry, with exponential backoff so that it doesn't continuously trigger as often.

src/RemoteConfigContext.tsx
gorandalum: At least the retry mechanism is a "soft refresh", so it shouldn't end up being throttled since it will find the values in the cache first if the time since last fetch from backend hasn't surpassed the minimum fetch interval.

src/RemoteConfigContext.tsx
gorandalum: Added now. Did this locally in RemoteConfigContext for now, and we can extract it later if we want to use it more places.

rosvik (approved):

PR #3385, opened 2023-03-14
chore: Show purchase button based on total wait time

ingvisim:
Show purchase button if total wait time is less than an hour rather than for each individual leg

https://github.com/AtB-AS/kundevendt/issues/3539#issuecomment-1466324741

### DIFF
--- a/src/travel-details-screens/TripDetailsScreenComponent.tsx
+++ b/src/travel-details-screens/TripDetailsScreenComponent.tsx
@@ -230,7 +230,7 @@ function totalWaitTimeIsMoreThanAnHour(legs: Leg[]) {
 
 function getWaitTime(leg: Leg, nextLeg: Leg) {
   return nextLeg
-    ? secondsBetween(leg.expectedEndTime, nextLeg?.expectedStartTime)
+    ? secondsBetween(leg.expectedEndTime, nextLeg.expectedStartTime)
     : 0;
 }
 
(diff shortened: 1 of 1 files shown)


src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -221,11 +218,20 @@ function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
...
+      (sum, leg, currentIndex) =>
+        sum + getWaitTime(leg, legs[currentIndex + 1]),
+      0,
+    ) >= hoursToSeconds(1)
+  );
+}
+
+function getWaitTime(leg: Leg, nextLeg?: Leg) {
+  return nextLeg
     ? secondsBetween(leg.expectedEndTime, nextLeg?.expectedStartTime)
```
jorelosorio, line 233: `?` might not be needed on `nextLeg` since is validated in the condition.

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -221,11 +218,20 @@ function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
   };
 }
 
-function waitTimeIsMoreThanAnHour(leg: Leg, nextLeg?: Leg) {
-  const waitTimeInSeconds = nextLeg
+function totalWaitTimeIsMoreThanAnHour(legs: Leg[]) {
+  return (
+    legs.reduce(
+      (sum, leg, currentIndex) =>
+        sum + getWaitTime(leg, legs[currentIndex + 1]),
```
jorelosorio, line 225: It is curious, this doesn't break the app because this is a clear index overflow here when the `currentIndex` is in the last element. But it was like that before.

gorandalum (approved):

PR #3388, opened 2023-03-14
refactor: combine departures quay and stop place state

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/3654

Together with https://github.com/AtB-AS/atb-bff/pull/237, this replaces `use-quay-data.ts` and `use-stop-place-state.ts` with shared logic in `use-departures-data.ts`.


gorandalum (approved):
Really nice! 👍

PR #3389, opened 2023-03-15
fix: Fix several warnings

gorandalum:
### Solution
Some errors which shows up in the Metro when running up the simulator locally are fixed. These are mainly:
- React unique keys errors
- Removed usage of "setAppGroupName" in AsyncStorage since it gave "undefined is not a function". This method was no longer supported by the AsyncStorage lib, and also the error was spamming Bugsnag for every user.
- Removed some cyclical dependencies in the translations module.
- Turned off logging breadcrumbs to console when developing locally. This can easily be overrided in the `bugsnagConfig.ts`.

### Acceptance criteria
The app should work as before. Mainly two things in the app can be affected by this change, which is:
- [ ] Travel search results should look and behave as before, especially the horizontal list of buses/trains with the double dash (`--`) between.
- [ ] Local storage should work as before, and still be persisted when this change is applied.

### DIFF
--- a/src/diagnostics/bugsnagConfig.ts
+++ b/src/diagnostics/bugsnagConfig.ts
@@ -1,5 +1,11 @@
 import Bugsnag, {Event} from '@bugsnag/react-native';
 
+/**
+ * Set to true if you want Bugsnag breadcrumbs to be written to console.log
+ * locally
+ * */
+const LOG_BREADCRUMBS_LOCALLY = false;
+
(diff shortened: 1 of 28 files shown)


rosvik (approved):
Nice 🙌

gorandalum:
Great find! I'll look into it.

tormoseng:
@gorandalum Re-tested OK by installing the previous version, adding things to storage, and then updating to this (the latest fix). All values are preserved.

PR #3392, opened 2023-03-16
fix: Quick fix for compact map for iOS

gorandalum:
The fit bounds functionality on load was not working as expected
on iOS. A quick fix is triggering fitBounds after load.


jorelosorio (approved):
Nice, seems good for now, but would be nice more research for later.

PR #3397, opened 2023-03-17
build: Allow overridePins in build for pen-test

gorandalum:
I think this should work based on the documentation here:
https://developer.android.com/training/articles/security-config#certificates


gorandalum:
> Is this just a temporary workaround for the pentest, or should it be this way going forward?

☝️ I was too fast in creating PR. Did it from command line just from habit. Changed to draft now. Will build the APK from this branch.

gorandalum:
The build seemed to work for the penetration test done by Atea.

PR #3401, opened 2023-03-21
refactor: Refactor collapsedLeg to become reusable, renamed to CounterIconbox

hanne-at-atb:
Changed name of CollapsedLeg to CounterBox
Made an generic component and placed it in components folder


src/components/counter-box/CounterBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+    padding: theme.spacings.xSmall,
+    lineHeight: iconSizes.small,
+    type: 'label__uppercase',
+  };
+
+  const normalStyling: StylingCounter = {
+    padding: theme.spacings.small,
+    lineHeight: iconSizes.normal,
+    type: 'body__primary--bold',
+  };
```
gorandalum, line 38: I think we should support ´large´ as size here as well, since it is in the typing. If we don't want do to that now, the typing of the field size should probably be:
`size?: Exclude<keyof Theme['icon']['size'], 'large'>`

src/components/counter-box/CounterBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+  type: TextNames;
+}
+export const CounterBox = ({
+  count,
+  size = 'normal',
+  accessibilityLabel,
+}: {
+  count: number;
+  size?: keyof Theme['icon']['size'];
+  accessibilityLabel?: string;
```
gorandalum, line 20: I think it is better to include `AccessibilityProps` here instead of just `accessibilityLabel`. So the typing would be:
```ts
{
  count: number;
  size?: keyof Theme['icon']['size'];
} & AccessibilityProps
```

Look at `TouchableOpacityOrView` for an example.

src/components/counter-box/CounterBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const styling = isSmall ? smallStyling : normalStyling;
+
+  return (
+    <View
+      style={[
+        styles.transportationIcon,
+        {
+          padding: styling.padding,
+        },
```
gorandalum, line 48: I also think it is good practice to add a `style` param to the input parameters of `CounterBox` which is added to the style of this `View`.

src/components/counter-box/CounterBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+        }}
+      >
+        +{count}
+      </ThemeText>
+    </View>
+  );
+};
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  transportationIcon: {
```
gorandalum, line 67: Maybe rename this? We often just call it `container` if it is the base `View` of the component.

gorandalum (approved):
Looks good! I've added some comments which you may implement if you agree with them 😄

src/components/counter-container/CounterContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+export const CounterContainer = ({
+  count,
+  size = 'normal',
+  style,
+  ...a11yProps
+}: {
+  count: number;
+  size?: keyof Theme['icon']['size'];
+  style?: ViewStyle;
```
gorandalum, line 21: It is preferred to use `StyleProp<ViewStyle>` instead of just `ViewStyle`. I know it is not consistent in the app from before, but I'm just commenting to raise awareness :) 

The difference with using `StyleProp` is that the parameter works more as the `style` parameter on the "native" React components. For example `style={[style.first, style.second]}` will only work if using `StyleProp`.

src/components/counter-container/CounterContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+        }}
+      >
+        +{count}
+      </ThemeText>
+    </View>
+  );
+};
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  transportationIcon: {
```
gorandalum, line 83: Note that the previous comment
> "Maybe rename this? We often just call it container if it is the base View of the component."

was with regards to the style named `transportationIcon` and not the naming of the whole component. Not sure if that comment made you change the name to `CounterContainer` or if it was an objective decision based on `CounterContainer` being a better name that `CounterBox`. You are free to choose the name for the component which you think is best :)

gorandalum (approved):

PR #3402, opened 2023-03-21
fix: Stretched compact map fix

gorandalum:
Setting `useNativeDriver` to `true` seems to work when testing on iPhone 12 mini emulator. I am not exactly sure why, and probably not worth investing more time into understanding it since the map is going to be removed from this animated header really soon.

This change should be tested to see if it works, and if the scroll animation is good enough without it. More info about the setting can be found here:
https://reactnative.dev/blog/2017/02/14/using-native-driver-for-animated

A specific note about what happens with the value set to false:
> without the native driver it will always run a frame behind of the gesture because of the async nature of React Native.

Since this is a delayed/parallax type of animation, it hopefully isn't noticeable for the user. Should probably compare side by side with old animation which uses native driver.

Also set timeout value to 100 instead of 50 to work more often on slower phones.

### DIFF
--- a/src/components/disappearing-header/ContentWithDisappearingHeader.tsx
+++ b/src/components/disappearing-header/ContentWithDisappearingHeader.tsx
@@ -55,7 +55,7 @@ export function ContentWithDisappearingHeader({
         }
         onScroll={Animated.event(
           [{nativeEvent: {contentOffset: {y: scrollYRef}}}],
-          {useNativeDriver: true},
+          {useNativeDriver: false},
         )}
         contentContainerStyle={{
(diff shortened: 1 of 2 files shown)


rosvik (approved):
LGTM, I guess 😄 

If it's the animated header that's causing issues here, I guess one option is to remove the animation, and just have it as part of the page content, since that's being done as part of the next release anyway.

gorandalum:
> LGTM, I guess 😄
> 
> If it's the animated header that's causing issues here, I guess one option is to remove the animation, and just have it as part of the page content, since that's being done as part of the next release anyway.

I'll fall back to that solution if anything is found during testing :+1:

PR #3415, opened 2023-03-24
fix: duplicate favorite location bug

adriawh:
solution:  user can not add a favorite with same name or location as existing favorite.


jorelosorio (approved):

gorandalum (commented):
What if you update a favorite to be equal to another existing favorite? 🤔😁

jorelosorio:
> What if you update a favorite to be equal to another existing favorite? 🤔😁

adriawh:
> What if you update a favorite to be equal to another existing favorite? 🤔😁
@gorandalum @jorelosorio 

Made the same fix as with the add case. Should I give the user feedback, or just leave it like this?

PR #3418, opened 2023-03-24
build: update FRAM bundle ids

toretefre:
Setting new bundle ids for FRAM on format: `no.org.company.environment`

Also changing Apple app group to be `group.no.org.company`


adriansberg (approved):
Looks ~good~ encrypted to me! 😁

adriansberg:
`dev` vs `debug`. Is `debug` preferred?

PR #3420, opened 2023-03-27
refactor: remove onboarding of departures V2

mortennordseth:
Closes https://github.com/AtB-AS/kundevendt/issues/3587

FRAM would like to disable the departures v2 onboarding. The remote config for this exists, but it is not used. 

AtB currently has `enable_departures_v2_onboarding` set to `false` in production. This must be set to `true` to work as today with this change. I believe this can be updated now as it is currently not in use. 

NFK does not have any value set for `enable_departures_v2_onboarding` and the default value is `false`. To work as today NFK must add the config and set it to `true`.


toretefre:
Important to verify that AtB has received information that something needs to be added to Remote Config before releasing an app with this update.

rosvik:
Could we just remove the departures onboarding completely at this point? @ckbilstad @oyvindmikkelsenatb

toretefre:
> Could we just remove the departures onboarding completely at this point? @ckbilstad @oyvindmikkelsenatb

Think this should be alright with NFK as well

mortennordseth:
Updated the PR to remove the departures onboarding instead of disabling it 👍

toretefre (approved):
I guess it's impossible to make sure that the onboarding storage key is removed from the JSON structure? Other than that it should work nicely 🚀

gorandalum (approved):

PR #3431, opened 2023-03-29
refactor: Update styling for bike stations and scooters

KristianSelnas:
New icons (with the same names) are uploaded to Mapbox studio. This PR contains tweaks to accommodate these new icons.


src/components/map/components/mobility/Vehicles.tsx
```diff
@@ -82,11 +82,11 @@ export const Vehicles = ({mapCameraRef, vehicles, onPress}: Props) => {
         minZoomLevel={13.5}
         belowLayerID="clusterIcon"
         style={{
-          circleColor: '#920695',
-          circleStrokeColor: '#920695',
-          circleOpacity: 0.7,
-          circleStrokeOpacity: 0.2,
-          circleStrokeWidth: ['min', ['+', 2, ['get', 'point_count']], 12],
+          circleColor: '#4A753A',
+          circleStrokeColor: '#4A753A',
```
jorelosorio, line 86: Should this be on the design system instead? Or the remote config?

gorandalum: Might be something to take into consideration. It is hard to visualise how these colors are shown in the app just by doing the PR, but every color that a transport company might want to change should be in the design system.

gorandalum (approved):

KristianSelnas:
I think it's debatable whether the icon colors should be in the design system or not. They are currently matching the colors of icons from Mapbox Studio. Does each transport company have different icon sets in Mapbox Studio? If so, these colors should be in the design system.

gorandalum:
As far as I understand, each transport company has their own Mapbox account.. So the mobility color can probably be in the design system, and it is up to the transport companies to use the corresponding colors on the icons they add in Mapbox studio.

KristianSelnas:
> As far as I understand, each transport company has their own Mapbox account.. So the mobility color can probably be in the design system, and it is up to the transport companies to use the corresponding colors on the icons they add in Mapbox studio.

Ok. Mattis has currently used colors that don't exist in the design system. I'll discuss with Christian and Øyvind whether we can find suitable colors from the existing design system colors or whether we add these new colors to the design system.

PR #3434, opened 2023-03-29
build: Use automatic signing for local development

adriansberg:
These changes increase the DX for local development. Setting the signing to be automatic and setting the app group based on the value in the .env file should make it so that the developer does not have to do any manual clicks in XCode, given that the development team is set up as it should be.


mikaelbr:
I wonder if this should actually be MANUAL with correctly injected variables instead of doing it automatic. When being automatic we need all users to be added to Apple, but I don't think that is practical for all 🤔

ios/AtbAppIntent/AtbAppIntentDebug.entitlements
```diff
@@ -4,7 +4,7 @@
 <dict>
 	<key>com.apple.security.application-groups</key>
 	<array>
-		<string>group.no.mittatb</string>
+		<string>$(APP_GROUP_NAME)</string>
```
mikaelbr, line 7: So these changes are probably good

adriansberg:
Right, that makes sense. Looks like a provisioning profile has to be provided when it's set to "Manual". That might be tricky to inject?

adriansberg:
Discarding this as it's not as straight forward as I thought.

gorandalum:
I think on issues like this it would be beneficial to discuss with @jorelosorio, as he has quite a good understanding of the ios setup.

PR #3437, opened 2023-03-30
feat: Show live button if there is live data

ingvisim:
- Show live button on trip detail legs and departure details if there is live data
- Fetch live data from bff (https://github.com/AtB-AS/atb-bff/pull/243)
- Use new colors from upgraded theme on live button (https://github.com/AtB-AS/kundevendt/issues/3674)
- Disable buttons when using screenreader

Buttons are primary, not secondary, as they currently would not be visible as pill secondary button
Placing live button on departures temporarily, as it is intended to go into header component

<details>

Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-30 at 10 17 59
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-30 at 10 17 44
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-30 at 10 16 35
Simulator Screen Shot - iPhone 14 Pro - 2023-03-30 at 10 28 20
</details>

Close https://github.com/AtB-AS/kundevendt/issues/3406


src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -20,9 +20,11 @@ import {StyleSheet, View} from 'react-native';
 import MapLabel from './components/MapLabel';
 import MapRoute from './components/MapRoute';
 import {createMapLines, getMapBounds, pointOf} from './utils';
+import {VehiclePosition} from '@atb/api/types/generated/ServiceJourneyVehiclesQuery';
 
 export type TravelDetailsMapScreenParams = {
   legs: MapLeg[];
+  _vehiclePosition?: VehiclePosition;
```
gorandalum, line 27: Why the underscore prefix? I think can just be scrapped if it isn't used. The map will probably find the vehicle position through its own requests.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-get-service-journey-vehicles.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  useEffect(
+    () =>
+      setVehicleUpdates({
+        serviceJourneyVehicles: updatedVehicles,
+        error: error,
+      }),
+    [updatedVehicles, error],
+  );
```
gorandalum, line 36: Is this `useEffect` and the `useState` really needed? Can't we just use the state directly from `usePollableResource` without having this intermediate storage?

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -75,6 +80,19 @@ export const DepartureDetailsScreenComponent = ({
   ] = useDepartureData(activeItem, 20, !isFocused);
   const mapData = useMapData(activeItem);
 
+  const realtimeMapEnabled = useRealtimeMapEnabled();
+  const screenReaderEnabled = useIsScreenReaderEnabled();
+
+  const shouldShowLive = isDateInRangeFromNow(activeItem.serviceDate, 12 * 60);
```
gorandalum, line 86: Estimated calls should have a field `realtime` which maybe can be used for this instead, if I understand correctly what the purpose is. Could you look into that? By using a flag from Entur we might not need any custom logic on this.

src/travel-details-screens/components/Trip.tsx
```diff
@@ -51,6 +72,16 @@ const Trip: React.FC<TripProps> = ({
...
+                realtimePosition={tripVehiclePosition}
+                onExpand={() =>
+                  onPressDetailsMap({
+                    legs: tripPattern.legs,
+                    fromPlace: tripPattern.legs[0].fromPlace,
+                    toPlace:
+                      tripPattern.legs[tripPattern.legs.length - 1].toPlace,
+                    _vehiclePosition: tripVehiclePosition,
+                  })
+                }
```
gorandalum, line 84: Do we actually need two parameters here? If it is just for showing a button, then I think we can encapsulate this into a single parameter with something like:
`onExpand={tripVehiclePosition ? () => ... : undefined}`

Also I think the variable name should be a little more precise than `onExpand`. Maybe something like `onPressShowInMap` or something.

src/travel-details-screens/components/Trip.tsx
```diff
@@ -25,19 +28,37 @@ export type TripProps = {
...
+    : undefined;
+  const {vehiclePositions} = useGetServiceJourneyVehicles(ids);
+
   return (
     <View style={styles.container}>
       <TripMessages tripPattern={tripPattern} error={error} />
       <View style={styles.trip}>
         {tripPattern &&
           legs.map((leg, index) => {
+            const tripVehiclePosition = vehiclePositions?.find(
```
gorandalum, line 57: It is the position for the vehicle in a single leg, so I think `legVehiclePosition` is more precise.

gorandalum (changes requested):
Good work! Some comments to take a look at.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
ingvisim: The idea was to send an initial value to have data until we establish subscription, underscore is because it is not yet in use in the map page. The alternative is to send just the service journey id and provide no initial data, but the idea was that could be an easy way to show position instantly. Adding a todo-comment

src/travel-details-screens/components/Trip.tsx
```diff
@@ -25,19 +29,37 @@ export type TripProps = {
...
   );
+
+  const realtimeMapEnabled = useRealtimeMapEnabled();
+  const shouldShowLive =
+    !tripPattern.legs.find((a) => !a.realtime) && realtimeMapEnabled;
+
+  const ids = shouldShowLive
+    ? (tripPattern.legs
+        .map((leg) => leg.serviceJourney?.id)
+        .filter(Boolean) as string[])
```
gorandalum, line 48: Casting should be avoided if possible. If the type of `serviceJourney.id` changes (although unlikely) then this code may crash at runtime since it won't give any compile time error when casting to string.

You can take a look at the options here:
https://www.typescriptlang.org/docs/handbook/advanced-types.html

A more practical example is in this blog post:
https://dev.to/zirkelc/how-to-filter-values-type-safely-29jj

src/travel-details-screens/components/TripSection.tsx
```diff
@@ -220,6 +227,17 @@ const TripSection: React.FC<TripSectionProps> = ({
             />
           </TripRow>
         )}
+        {realtimeMapEnabled && onPressShowLive ? (
```
gorandalum, line 230: Now I am on a very detailed level. It is not necessary to change, but more just a food for thought:

I think the check on `realtimeMapEnabled` can just be skipped here. It is good practice to be consistent on which component (parent or child) has the responsibility of sanity checking if something should be possible. As implemented here it is better if the parent has responsibility for sanity checking, and this component show the button as long as `onPressShowLive` is declared.

gorandalum (approved):
Looks good! Some minor comments which you can take into consideration :)

src/travel-details-screens/components/Trip.tsx
ingvisim: Great! Will look at this

PR #3438, opened 2023-03-31
refactor: PurchaseTariffZonesSearchByMapScreen

adriawh:
To be able to use the tariffZonesSelector in the ticket assistant(bachelor) there was need for a refactor of the PurchaseTariffZonesSearchByMapScreen. 

I have taken the two main elements and made them into components. The buttons at the top (from/to) and the map with its buttons. The save button seen at the bottom is made optional, so that it the map can be used in the ticket assistant. 

Simulator Screen Shot - iPhone 14 Pro - 2023-04-03 at 20 11 31


src/tariff-zones-selector/navigation-types.ts
```diff
@@ -0,0 +1,21 @@
...
+    callerRouteParam: keyof Root_PurchaseTariffZonesSearchByMapScreenParams,
+  ) => void;
+};
+
+export type TariffZonesSelectorMapsProps = {
+  selectedZones: TariffZoneSelection;
+  isApplicableOnSingleZoneOnly: boolean;
+  setSelectedZones: (selectedZOnes: TariffZoneSelection) => void;
+  onSave?: () => void;
+};
```
gorandalum, line 21: Just inline these as `Props` if possible.

gorandalum (approved):

adriawh:
Found a bug where the from / to buttons dont switch between being bold

HanneLH:
> Found a bug where the from / to buttons dont switch between being bold

https://github.com/AtB-AS/mittatb-app/pull/3452 - Bug fixed in other PR.

PR #3439, opened 2023-04-03
feat: start and end date to global messsages

adriawh:
Acceptance criteria:
- [ ] Should show If no start or end date
- [ ] Should appear at given starDate if given 
- [ ] Should disappear at given endDate if given

### DIFF
--- a/src/global-messages/converters.ts
+++ b/src/global-messages/converters.ts
@@ -37,7 +37,6 @@ function mapToGlobalMessage(
   const startDate = result.startDate;
   const endDate = result.endDate;
 
-  console.log('Validated');
   if (!result.active) return;
   if (!isAppPlatformValid(platforms)) return;
   if (appVersionMin && appVersionMin > APP_VERSION) return;
--- a/src/global-messages/converters.ts
+++ b/src/global-messages/converters.ts
@@ -10,7 +10,6 @@ import {
   GlobalMessageType,
 } from '@atb/global-messages/types';
 import type {LanguageAndTextType} from '@atb/translations';
-
 export function mapToGlobalMessages(
   result: FirebaseFirestoreTypes.QueryDocumentSnapshot<GlobalMessageRaw>[],
 ): GlobalMessageType[] {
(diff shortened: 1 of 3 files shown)


src/global-messages/types.ts
```diff
@@ -21,6 +23,8 @@ export type GlobalMessageRaw = {
   appPlatforms: AppPlatformType[];
   appVersionMin: string;
   appVersionMax: string;
+  startDate?: Timestamp;
```
gorandalum, line 26: For better encapsulation of Firestore, I think it is better to have `number` here. I know we expose the Firestore `Timestamp` type on some ticket types as well, but in the spirit of easter: "Det er aldri for sent å snu".

src/global-messages/GlobalMessage.tsx
```diff
@@ -16,6 +17,8 @@ type Props = {
 
 const GlobalMessage = ({globalMessageContext, style}: Props) => {
   const {language} = useTranslation();
+  const [now, setNow] = useState<number>(Date.now());
+  useInterval(() => setNow(Date.now()), 2500);
```
gorandalum, line 21: Now there are several places where logic like this is implemented. Even though it is just a couple of lines, I think it is can be a utility function `useNow` under src/utils. It can take the milliseconds as a parameter. So it would just be `const now = useNow(2500)` here.

src/global-messages/converters.ts
```diff
@@ -54,6 +55,8 @@ function mapToGlobalMessage(
     body,
     title,
     isDismissable,
+    endDate,
+    startDate,
```
gorandalum, line 59: `endDate` before `startDate`!? Are you trying to trigger our OCD?

gorandalum (approved):
Good work! Some minor nitpicks, and some extremely minor nitpicks 😅

src/global-messages/types.ts
```diff
@@ -21,6 +21,8 @@ export type GlobalMessageRaw = {
   appPlatforms: AppPlatformType[];
   appVersionMin: string;
   appVersionMax: string;
+  startDate?: number;
+  endDate?: number;
```
gorandalum, line 25: For å gjøre dette helt korrekt så bør det være type `Timestamp` i `GlobalMessageRaw`, mens i `GlobalMessageType` så bør det være `number`. I fix slik at det blir merget nå før påske.

PR #3441, opened 2023-04-03
feat: changes to use shared fare product type config package, supporting new types for FRAM

mikaelbr:
- Adds support for single/multiple stop/zone by just defaulting to normal zone selector.
- Changes to use package for decoding fare product type config and search filters.

Updated `@atb-as/config-spec` to use union type for TransportMode and TransportSubmode and added mapping for this in app.

This PR limits use of `@atb-as/config-spec` for Fare Product Config Types and Search Filters. Will cover other mappings in different tasks (tasks TBA)

TODO

- [x] Given accept for suggested solution, do proper release of `@atb-as/config-spec`

Fixes https://github.com/AtB-AS/kundevendt/issues/3771
Fixes https://github.com/AtB-AS/kundevendt/issues/3770


gorandalum (commented):
I think this looks really good, and really cleans up the clients and reduces duplication between app and webshop.

mortennordseth (approved):
Looks good! Nice work 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types.ts
```diff
@@ -31,20 +31,13 @@ export type TransportIconModeType = Omit<
...
-  transportModes?: TransportModeFilterOptionType[];
-};
+export {
+  TravelSearchFiltersType,
+  TravelSearchTransport,
+} from '@atb-as/config-specs';
 
 export type TransportModeFilterOptionWithSelectionType =
-  TransportModeFilterOptionType & {selected: boolean};
+  TravelSearchTransport & {selected: boolean};
```
gorandalum, line 40: May look at the naming here. Now it is a maybe illogical "naming-jump" from `TravelSearchTransport` to `TransportModeFilterOption`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -364,3 +367,16 @@ function useJourneySearchModes(
...
+
+function transportModeToEnum(
+  modes: TravelSearchTransportModes[],
+): Types.TransportModes[] {
+  return modes.map((internal) => {
+    return {
+      transportMode: internal.transportMode as Types.TransportMode,
+      transportSubModes: internal.transportSubModes?.map(
+        (submode) => submode as Types.TransportSubmode,
+      ),
```
gorandalum, line 379: Possible to use helper `enumFromString` (which returns `undefined` if it doesn't exist) here, to have extra safety if at some point there is a mismatch between config-specs and the journeyplanner-types?

gorandalum (approved):
Good work! :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types.ts
mikaelbr: I was trying to name based on config settings instead of what it was named as in the app, but could also inverse that 🤔 But won't be able to rename in firestore though.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
mikaelbr: Hmm. Is there a mismatch now? 🤔 I thought it could be undefined. But I'll control check

mikaelbr: Ah, now I get what you meant... Fixed

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types.ts
gorandalum: I think just using `TravelSearchTransportWithSelectedType` is good. The main thing is that it is the same type name, just with the selected field.

gorandalum: Ahh, I see you have made the change the other way around. Just as good :+1:

mikaelbr: It should be named the same as before now FYI

PR #3443, opened 2023-04-11
feat: add large header component

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/3673

- Upgrades theme to 6.5
- Adds `LargeScreenHeader` component, based on `ContentWithDisappearingHeader`
	- Supports optional `title`
	- Supports generic content as `headerChildren`
	- Supports global messages
- Implements a first-iteration of new header for trip details and departure details.

<div>

</div>

https://user-images.githubusercontent.com/1774972/231142185-2b8af829-56ab-428b-a1dd-f6f8636ce3e4.mp4


ingvisim (approved):

src/components/screen-header/LargeScreenHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+   * For specifying the alert context for alerts that should be shown in this
+   * header. If no context is specified then no alerts are shown.
+   */
+  globalMessageContext?: GlobalMessageContextType;
+  style?: ViewStyle;
+  containerStyle?: ViewStyle;
+  color?: StaticColor;
+  setFocusOnLoad?: boolean;
+  headerChildren?: JSX.Element | JSX.Element[];
+  children?: JSX.Element | JSX.Element[];
```
gorandalum, line 26: Just be reading these parameters names I don't understand the difference. Should probably document it.

src/components/screen-header/LargeScreenHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+export type LargeScreenHeaderProps = {
+  title?: string;
+  titleA11yLabel?: string;
+  /**
+   * For specifying the alert context for alerts that should be shown in this
+   * header. If no context is specified then no alerts are shown.
+   */
+  globalMessageContext?: GlobalMessageContextType;
+  style?: ViewStyle;
+  containerStyle?: ViewStyle;
```
gorandalum, line 22: Will both these parameters be necessary? Can we skip one, or maybe even both? 🤔

src/components/screen-header/FullScreenHeader.tsx
```diff
@@ -12,8 +18,28 @@ export function FullScreenHeader(props: ScreenHeaderProps) {
...
+    <View style={{backgroundColor, paddingTop: top}}>
       <ScreenHeader {...props} />
     </View>
   );
 }
+
+export function LargeFullScreenHeader(
+  props: LargeScreenHeaderProps & {
+    buttonProps?: Omit<LargeHeaderButtonProps, 'color'>;
+  },
```
gorandalum, line 30: I think it is slightly better if all parameters are defined in the `LargeScreenHeaderProps`, especially when it is exported. Now the parameter `buttonProps` is kind of "hidden".

src/components/screen-header/HeaderButton.tsx
```diff
@@ -75,6 +77,34 @@ const BaseHeaderButton = ({
   </TouchableOpacity>
 );
 
+export type LargeHeaderButtonProps = Omit<HeaderButtonProps, 'type' | 'text'>;
```
gorandalum, line 80: I think the pattern with `type` and such is pretty good. And using it would probably work well with reusing the logic for the existing header?

src/components/screen-header/LargeScreenHeader.tsx
rosvik: I though it would be good to support most of the params from the current header, but looking at how that's used, it doesn't seem necessary. Removing them both!

src/components/screen-header/HeaderButton.tsx
rosvik: @gorandalum Right now `type` is defined as 

```ts
export type ButtonModes =
  | 'back'
  | 'cancel'
  | 'close'
  | 'chat'
  | 'skip'
  | 'status-disruption'
  | 'custom';
```

I don't think most of them makes sense in this context, maybe just `back`, `cancel` and `close`. And even then, I don't know when we should show the back arrow and not, since the new header concept in Figma isn't covering these use cases. Options to extend this are to add `backArrow` as a mode, it's own param `showBackArrow: boolean` or even a `leftIcon`, but I don't know what's best.

This is why I thought it would be ok to skip this complexity in this iteration, until we know what it will be used for.

rosvik:
@gorandalum Made a PR with your suggestions here: https://github.com/AtB-AS/mittatb-app/pull/3462

PR #3444, opened 2023-04-11
refactor: Rewrite from default export to named export

hanne-at-atb:
Rewrite from default export to named export to align with code guidelines


gorandalum (approved):
Great! Maybe also add [eslint rule](https://eslint.org/docs/latest/rules/no-restricted-exports) when you're at it!

gorandalum (approved):
Great! 🎉

PR #3451, opened 2023-04-12
Andershc/summary page

andershc:
Finished durations page and category picker. Also started on the summary page with connection to the API. Visually done, but missing some core functionality.


andershc:
The lint error is a variable ("navigation") that is not used, but it is going to be used later

PR #3452, opened 2023-04-12
fix: highlighting from/to when selecting zone in the map

adriawh:
Then highlighting of from/to was not switching when selecting zones in the map.

Skjermbilde 2023-04-12 kl  09 13 49

I also removed 

`onRegionDidChange={(region)=> {
               setRegionEvent({isMoving: false, region});
             }}       
onRegionWillChange={() => {
               setRegionEvent({isMoving: true, region: regionEvent?.region});
             }}`

as they are deprecated, and I didn't find any use of them.


rosvik (approved):
LGTM 👍

HanneLH:
Verifisert ok for Android and iOS ✅

PR #3453, opened 2023-04-12
fix: Error when compiling iOS from Xcode `Value not supported`

jorelosorio:
A recent version of Xcode is failing because of the following error:

Screenshot 2023-04-11 at 10 36 13


gorandalum (approved):

KristianSelnas (approved):

tormoseng:
Builds OK!

PR #3454, opened 2023-04-12
feat: Show city bike station names

KristianSelnas:


src/mobility/components/CityBikeStationBottomSheet.tsx
```diff
@@ -36,6 +42,7 @@ export const CityBikeStationSheet = ({station, close}: Props) => {
     appStoreUri,
     rentalAppUri,
   });
+  const stationName = getTextForLanguage(station.name.translation, language);
```
gorandalum, line 45: Just a note: Also possible to use `useTextForLanguage` here if you don't want to provide `language`.

gorandalum (approved):

src/mobility/components/CityBikeStationBottomSheet.tsx
KristianSelnas: Thank you, I wasn't aware of that one.

PR #3455, opened 2023-04-12
draft: configurable tariffzones

toretefre:
Just dumping what I've got before moving to creating FRAM app. Might be stuck here for a while


gorandalum:
Have you thought about having a boolean field `isDefault` (or something like it) on the actual tariff zones in reference data? By looking at the necessary code here I think it will be a slightly easier implementation, and also might be a more maintainable solution as the tariff zones logic isn't spread across more fields in firestore.

toretefre:
Solved by https://github.com/AtB-AS/mittatb-app/pull/3538

PR #3470, opened 2023-04-13
feat: Automatically reload vehicles in map every 10 seconds

KristianSelnas:


src/utils/use-previous.ts
```diff
@@ -0,0 +1,9 @@
+import {useEffect, useRef} from 'react';
+
+export const usePrevious = <T>(value: T): T | undefined => {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
```
gorandalum, line 9: I copied this from a SO post, but when looking at it I wondered why `value` wasn't a part of the dependency array in `useEffect`. When looking at other sources for similar custom hook, I see they all have `value` in dependency array.
https://usehooks.com/usePrevious/
https://blog.logrocket.com/accessing-previous-props-state-react-hooks/

gorandalum (approved):
Looks good! Might do a change to the `usePrevious` hook.

PR #3473, opened 2023-04-14
fix: Ensure accessible with focusRef

ingvisim:
Fix Android app crash on departure details screen and trip details screen


src/components/screen-view/ScreenWithLargeHeader.tsx
```diff
@@ -52,6 +52,7 @@ export const ScreenWithLargeHeader = ({
             {headerChildren && (
               <View
                 ref={!title ? focusRef : undefined}
+                accessible={!title}
```
gorandalum, line 55: A little unsure about if it is correct with `accessible: false` when `title` is present 🤔 But at least it is not crashing.

gorandalum (approved):

PR #3478, opened 2023-04-17
feat: Remote config for vehicles poll interval

KristianSelnas:


src/remote-config.ts
```diff
@@ -81,6 +82,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_new_travel_search: false,
   enable_from_travel_search_to_ticket: false,
   enable_vehicles_in_map: false,
+  vehicles_poll_interval: 20000,
```
gorandalum, line 85: Maybe have "ms" in the config name for extra clarity, just in case.

KristianSelnas: I saw that `favourite_departures_poll_interval` was in millieseconds, so I added `vehicles_poll_interval` in milliseconds as well, even though we actually need it in seconds. If the default for config params is milliseconds, I think it could be superflous?

gorandalum: It is no established or written down rule. I think it should be possible to find out without diving into the code, but we can also use the description field in Remote config.

KristianSelnas: I think we should either leave it as it is or rename both `favourite_departures_poll_interval` and `vehicles_poll_interval` to `favourite_departures_poll_interval_ms` and `vehicles_poll_interval_ms` respectively.

PR #3479, opened 2023-04-17
build: add verbose to xcodebuild args

toretefre:
Aiming to get proper error for failing FRAM xcodebuild

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -75,6 +75,7 @@ platform :ios do
       )
       match(
         type: ENV["FASTLANE_MATCH_TYPE"],
+        verbose: true,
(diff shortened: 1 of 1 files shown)


toretefre:
Asking for review in case any information that shouldn't become logged will become logged by adding verbose logging.

fastlane/Fastfile
```diff
@@ -172,9 +173,11 @@ platform :ios do
     cocoapods(podfile: './ios/Podfile')
     build_app(
       scheme: 'atb',
+      verbose: true,
```
toretefre, line 176: Haven't really found this as an option, will remove if causing to fail.

PR #3483, opened 2023-04-18
fix: Create theme and fix image colors for FRAM

adriansberg:
Add changes to theme creation such that a theme for FRAM is created. Updated the package `@atb-as/generate-assets` to get the updated images with the correct colors.

Fixes https://github.com/AtB-AS/kundevendt/issues/3773


mortennordseth (approved):
LGTM! 👍

PR #3487, opened 2023-04-19
fix: Change widget bundle id to match apple requirements

jorelosorio:
```
 ERROR: Asset validation failed (90347) Bad bundle identifier. The bundle identifier 'no.mittatb.store.departure.widget' of the application extension AtB.app/PlugIns/departureWidget.appex should start with the application's bundle identifier 'no.mittatb.store' and not contain more than one period "." after the application's bundle ID.
 ```


rosvik:
@jorelosorio Is there a reason why you're only updating the identifier for atb/store? Could it be changed to a similar name for staging, and for other orgs as well?

rosvik (approved):

jorelosorio:
> @jorelosorio Is there a reason why you're only updating the identifier for atb/store? Could it be changed to a similar name for staging, and for other orgs as well?

Seems like for other environments' apple do not complain about the name, only for the store there is requirement, so that is why I changed it, it can be changed later for others to keep the same naming convention but only if we want.

rosvik:
Think it would be easier to keep track of if the same naming convention is used for all environments. If it doesn't create any issues, I would just update them right away, otherwise it's most important to fix for store, then update for the rest later.

Is it easy enough to update all at the same time?

jorelosorio:
> 

Yes, it could be done now, but might need extra time, do I proceed?

toretefre:
NFK and FRAM can wait with updating our identifiers. Guess we won't have this issue until activating the widget. We currently have our hands full getting things uploaded already 😄

rosvik:
@jorelosorio I don't know, it's up to you, and what you think is best. Could be done after as well 😊

jorelosorio:
> @jorelosorio I don't know, it's up to you, and what you think is best. Could be done after as well 😊

Fixed for all the other envs ;) will merge

rosvik:
@jorelosorio FYI: I cherry-picked your changes to the `release/1.35` branch and make a release candidate now 🚀

tormoseng:
Builds OK!

PR #3490, opened 2023-04-21
feat: Follow live bus position

ingvisim:
Make map view follow bus
 - Decouple if user moves map
 - Start following again if user clicks bus icon

https://user-images.githubusercontent.com/85479566/233599976-5c9caa9d-ab99-474f-a3bf-07071d9667ac.mp4

Close https://github.com/AtB-AS/kundevendt/issues/3417
Close https://github.com/AtB-AS/kundevendt/issues/3418


src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -61,17 +71,33 @@ export const TravelDetailsMapScreenComponent = ({
       );
       if (vehicle && vehicle !== followVehicleMapPoint) {
         setFollowVehicleMapPoint(vehicle.location);
```
gorandalum, line 73: Shouldn't `setFollowVehicleMapPoint` in a `useEffect` which has `followVehicleMapPoint` in its dependency list.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -43,11 +46,18 @@ export const TravelDetailsMapScreenComponent = ({
...
+  const centerPosition = initialVehiclePosition?.location
+    ? [
+        initialVehiclePosition?.location?.longitude,
+        initialVehiclePosition?.location?.latitude,
+      ]
+    : undefined;
 
   const {t} = useTranslation();
   const controlStyles = useControlPositionsStyle();
   const vehicles = useGetLiveServiceJourneyVehicles(initialVehiclePosition);
```
gorandalum, line 59: The parameter name `initialVehiclePosition` was sort of confusing to me, as I thought it was actually just the vehicle position, and I couldn't understand how that could be the input for `useGetLiveServiceJourneyVehicles`. I think it would be quite clearer to name the type and param something like `VehicleWithPosition`.

gorandalum (changes requested):
Fun to see the bus moving in the map! :+1: 

Some comments, and I also get an error I haven't seen before:
```
ERROR Warning: Internal React error: Attempted to capture a commit phase error inside a detached tree. This indicates a bug in React. Likely causes include deleting the same fiber more than once, committing an already-finished tree, or an inconsistent return pointer.

Error message:

TypeError: Cannot read property '_state' of undefined
    in TravelDetailsMapScreenComponent (created by Departures_TravelDetailsMapScreen)
    in Departures_TravelDetailsMapScreen (created by SceneView)
    in StaticContainer
    in EnsureSingleNavigator (created by SceneView)
    in SceneView (created by CardContainer)
```

rosvik:
Updated to re-implement websockets in the app (without apollo). Now via the BFF.

hanne-at-atb (approved):
This looks good :)

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -43,35 +46,87 @@ export const TravelDetailsMapScreenComponent = ({
...
+  const [shouldTrack, setShouldTrack] = useState<boolean>(true);
+
+  const [followVehicleMapPoint, setFollowVehicleMapPoint] = useState<
+    Coordinates | undefined
+  >();
+
+  useEffect(() => {
+    const location = vehicle?.location;
+
+    if (vehicle === followVehicleMapPoint || !location) return;
```
gorandalum, line 77: Can `vehicle === followVehicleMapPoint` be true? Aren't they different types?

gorandalum, line 77: Generally all variables in a `useEffect` should be in the dependency array, and not having it there (or not being able to have it there) might signalise that something else is wrong. Maybe we can have a quick talk about these variables and what they are trying to achieve.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -153,19 +146,19 @@ const styles = StyleSheet.create({
...
 
 type VehicleIconProps = {
-  vehicles: Coordinates;
+  coordinates: Coordinates;
   setShouldTrack: React.Dispatch<React.SetStateAction<boolean>>;
 };
 
-const LiveVehicle = ({vehicles, setShouldTrack}: VehicleIconProps) => {
+const LiveVehicle = ({coordinates, setShouldTrack}: VehicleIconProps) => {
   const shapeSource = useRef<MapboxGL.ShapeSource>(null);
```
gorandalum, line 154: Not changed in this commit, but what is the purpose of this `shapeSource` ref? 🤔

gorandalum (approved):

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
rosvik: That does look unnecessary, yes 🤔 Is it ok to remove @ingvisim?

PR #3494, opened 2023-04-21
feat: Enable eslint import rule for src/components

gorandalum:
Enable rule to only allow imports from through index file, but not
from anywhere else in src/components.

This gave a total of 24 warnings which is fixed in this PR.


src/components/map/Map.tsx
```diff
@@ -16,8 +16,8 @@ import {useControlPositionsStyle} from './hooks/use-control-styles';
 import {MapCameraConfig, MapViewConfig} from './MapConfig';
 import {PositionArrow} from './components/PositionArrow';
 import {shadows} from './components/shadows';
-import * as Mobility from '@atb/components/map/components/mobility';
-import {MapFilter} from '@atb/components/map/components/filter/MapFilter';
+import * as Mobility from './components/mobility';
```
hanne-at-atb, line 19: Should we try to avoid wildcard imports with * ?

gorandalum: Maybe, do you want to avoid them? We don't have a rule on it as of now, and especially the section items is quite often imported with `import * as Sections from '@atb/components/sections'`.

hanne-at-atb: I've just learned that they are bad practice, at least in Java. And explicit imports don't import the whole module and  leads to clearer code :) 
Also found this, https://app.deepsource.com/directory/analyzers/javascript/issues/JS-C1003

gorandalum: Seems to quite easy to add by using this:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-namespace.md

You can create a PR adding the eslint-rule, and also add it to the [doc](https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md).

PR #3497, opened 2023-04-21
feat: Add Flexible transport message when searching within city zone

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/3850

Depend on: https://github.com/AtB-AS/design-system/pull/129

Example of the message.

Screenshot_1682343409


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-flexible-transport-direct-mode-enabled.ts
```diff
@@ -0,0 +1,19 @@
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {StorageModelKeysEnum} from '@atb/storage';
+import {useDebugOverride} from '@atb/debug';
+
+export const useFlexibleTransportDirectModeEnabled = () => {
+  const [debugOverride] = useFlexibleTransportDirectModeDebugOverride();
+  const {use_flexible_on_directMode: enabledInRemoteConfig} = useRemoteConfig();
+
+  if (debugOverride !== undefined) {
```
hanne-at-atb, line 9: I see the following code part is duplicated in 3 or more files. Could it go in its own hook?
`if (debugOverride !== undefined) {
    return debugOverride;
  }
  return enabledInRemoteConfig;`

Or alternatively combine the ones you have and return multiple things.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
hanne-at-atb, line 1: Note that according to our guidelines should a file with several hooks be named `hooks.ts` 🪝

hanne-at-atb (commented):
Just a few comments 🤸🏼‍♀️

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-flexible-transport-direct-mode-enabled.ts
jorelosorio: @hanne-at-atb I made a code refactor please take a look

hanne-at-atb: Nice!

hanne-at-atb (approved):
Looks good ⛵

src/reference-data/defaults/city-zones-message-texts.json
```diff
@@ -0,0 +1,20 @@
...
+  "multipleZones": {
+    "message": [
+      {
+        "lang": "nob",
+        "value": "Bestillingstransport tilgjengelig i {0} og {1}."
+      },
+      {"lang": "eng", "value": "Ordered transport is available in {0} and {1}."}
+    ]
+  }
+}
```
gorandalum, line 20: I feel these texts is more like every other "normal" text we have in the app, which is not specified in Firestore. Since we don't have "normal" texts in Firestore, it also means we don't have an existing generic solution for specifying texts outside of the app.

Here the reference data is used for this purpose, which might be ok for a quick fix, but I don't feel it is good for a long term solution. Also this implementation introduces the possibility to inject strings into the text, here with the placeholders `{0}` and `{1}` and supplementing logic in the method `getTextForLanguageWithFormat`:
```
for (let index = 0; index < values.length; index++) {
  text = text.replace(`{${index}}`, values[index]);
}
``` 

This also introduces a risk, since the template in firestore might come out of sync with the usage in the app. 

I'll also tag @tor

To summarise this comment into a question: Do we really need to have these texts in Firestore? Are they more important to be dynamic than other texts we have in the app?

I'll also add @tormoseng and @oyvindmikkelsenatb in case this have already been discussed with you.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
```diff
@@ -0,0 +1,110 @@
...
+export function useJourneyModes(
+  defaultValue: StreetMode = StreetMode.Foot,
+): Modes {
+  const flexibleTransportEnabled = useFlexibleTransportEnabled();
+  const flexibleTransportAccessModeEnabled =
+    useFlexibleTransportAccessModeEnabled();
+  const flexibleTransportDirectModeEnabled =
+    useFlexibleTransportDirectModeEnabled();
+  const flexibleTransportEgressModeEnabled =
+    useFlexibleTransportEgressModeEnabled();
```
gorandalum, line 59: It has been discussed and confirmed we need three different feature toggles for this? We end up adding quite a bit of toggles to Remote Config.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
```diff
@@ -363,6 +367,9 @@ export const Dashboard_TripSearchScreen: React.FC<RootProps> = ({
                   resetTransportModes={filtersState.resetTransportModes}
                 />
               )}
+              {isFlexibleTransportEnabledInRemoteConfig &&
```
gorandalum, line 370: If I remember correctly, wasn't it said that the city zones info box should have its own feature toggle? Haven't discussed it in quite some time, so not sure if that has been changed.

gorandalum (changes requested):
Great work! We are closing in! Some comments to take a look at.

src/reference-data/defaults/city-zones-message-texts.json
jorelosorio: @oyvindmikkelsenatb Base on the talk we had where the text were approved, is this still needed, be able to customize this form outside? With this, we wanted to have it open for later fixes and changes without make a new release @gorandalum, but everything has changed a lot. 

Regarding the function, it might be that from Firebase there is no ${0} or ${1} but the app will not crash since there will not be text to replace, the values are always passed by the app.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
jorelosorio: Yes it was when we started and now might be that is not needed @tormoseng because this will be always on but in the future. What do you think @tormoseng ?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
jorelosorio: @gorandalum yes is true and is made in the other PR https://github.com/AtB-AS/mittatb-app/pull/3518 about add the toggle, we were considering release without the toggle and this is why is that here.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const message = getTextForLanguageWithFormat(
+    messageTemplate.message,
+    language,
+    ...selectedCityZones.map((cityZone) => cityZone.name),
+  );
+
+  const openUrlForCityZone = (cityZone: CityZone) => {
+    const contactUrl = getTextForLanguage(cityZone.contactUrl, language);
+    if (contactUrl) {
+      return Linking.openURL(contactUrl);
```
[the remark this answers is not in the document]
jorelosorio: Fixed 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+  };
+
+  const messageActions = selectedCityZones.map(
+    (cityZone) =>
+      ({
+        text: cityZone.name,
+        onPress: () => openUrlForCityZone(cityZone),
+      } as ActionButton),
```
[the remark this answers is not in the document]
jorelosorio: Fixed 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const [isClosed, setClosed] = useState(false);
+
+  if (!selectedCityZones || isClosed) {
+    return null;
+  }
+
+  if (selectedCityZones.length === 0) {
+    return null;
+  }
```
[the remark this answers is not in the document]
jorelosorio: Fixed 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -0,0 +1,185 @@
...
+};
+
+type MessageBoxWithActionButtonsProps = {
+  icon: (props: SvgProps) => JSX.Element;
+  message: string;
+  onDismiss?: () => void;
+  actionButtons?: ActionButton[];
+};
+
+const MessageBoxWithActionButtons = ({
```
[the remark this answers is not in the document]
jorelosorio: Fixed 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
```diff
@@ -0,0 +1,110 @@
...
+  from: Location | undefined,
+  to: Location | undefined,
+): CityZone[] | undefined => {
+  const {cityZones} = useFirestoreConfiguration();
+  const fromCityZone = useFindCityZoneInLocation(from, cityZones);
+  const toCityZone = useFindCityZoneInLocation(to, cityZones);
+
+  const filteredCityZones = [fromCityZone, toCityZone].filter(
+    (cityZone) => cityZone && cityZone.enabled === true,
+  ) as CityZone[];
```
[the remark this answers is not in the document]
jorelosorio: Fixed 👍

src/reference-data/defaults/city-zones-message-texts.json
jorelosorio: @gorandalum We had a meeting, and we agree that it could be removed now. So I will do it.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
jorelosorio: @gorandalum this give us some flexibility for what is being considering for the future, we think is a good to have even now to test, however for future iterations we could consider that is not needed at all, so then we could remove it and will not cause any damage.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const filteredCityZones: CityZone[] = [];
+  fromCityZone &&
+    fromCityZone.enabled == true &&
+    filteredCityZones.push(fromCityZone);
+  toCityZone &&
+    toCityZone.enabled === true &&
+    filteredCityZones.push(toCityZone);
+
+  return filteredCityZones.filter(onlyUniquesBasedOnField('name'));
```
gorandalum, line 51: This works without casting. Another alternative similar to the old implementation is to use type guard (`cityZone is CityZone`) instead of casting. I'll just give some examples, but feel free to use the implementation that you think is best.

So instead of casting:
```ts
const filteredCityZones = [fromCityZone, toCityZone].filter(
  (cityZone) => cityZone && cityZone.enabled === true,
) as CityZone[]
```

it would be with a type guard: 
```ts
const filteredCityZones = [fromCityZone, toCityZone].filter(
  (cityZone): cityZone is CityZone => !!cityZone && cityZone.enabled === true,
)
```

Also possible here is an even shorter version:
```ts
return [fromCityZone, toCityZone]
  .filter((cityZone): cityZone is CityZone => !!cityZone?.enabled)
  .filter(onlyUniquesBasedOnField('name'));
```

gorandalum (approved):
LGTM! :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
jorelosorio: Great suggestion, I have implemented the shortest version.

PR #3498, opened 2023-04-24
fix: Load complete scooter details only when scooter is selected

KristianSelnas:


src/mobility/use-system.tsx
```diff
@@ -19,16 +19,16 @@ export const useSystem = <T extends {system: SystemFragment}>(
...
-        : entity.system.rentalApps?.android?.storeUri,
+        ? entity?.system.rentalApps?.ios?.storeUri
+        : entity?.system.rentalApps?.android?.storeUri,
     );
     setBrandLogoUrl(
       Appearance.getColorScheme() === 'dark'
-        ? entity.system.brandAssets?.brandImageUrlDark
-        : entity.system.brandAssets?.brandImageUrl,
+        ? entity?.system.brandAssets?.brandImageUrlDark
+        : entity?.system.brandAssets?.brandImageUrl,
```
gorandalum, line 28: A little unsure about using `getColorScheme` here as this `useEffect` will not fire when color scheme changes. I think using `useColorScheme` and having the color scheme in the dependency list is more correct.

Also think `useMemo` is more "correct" than using `useEffect` here as it is no syncing/fetching etc. from another service, and then the local states can be removed from this hook.

gorandalum (approved):
Looks good. Left a comment not related to this change, but that I saw now.

src/mobility/use-system.tsx
KristianSelnas: Does it actually make sense to use `useMemo` at all here? The contents are not "computationally heavy", and the computed value is a primitive.

gorandalum: I thought maybe more like this (pseudcode-ish, written by hand):
```tsx
return useMemo(() => {
  const appStoreUri = Platform.OS === 'ios'
        ? entity?.system.rentalApps?.ios?.storeUri
        : entity?.system.rentalApps?.android?.storeUri;
  const brandLogoUrl = themeName === 'dark'
        ? entity?.system.brandAssets?.brandImageUrlDark
        : entity?.system.brandAssets?.brandImageUrl;
  const operatorName = getTextForLanguage(operator?.translation, language) ??
      t(MobilityTexts.unknownOperator) ??
      ''
      
  return {
    appStoreUri,
    brandLogoUrl,
    operatorName,
  };
}, [entity])
```

But I agree, it is not computationally heave, so we can just skip both `useEffect` and `useMemo`.

PR #3505, opened 2023-04-26
fix: Fetch station on tap

KristianSelnas:
Same change as was done with scooters. Only data necessary to render the map is loaded by the map. Complete data for the station is loaded in a separate call when a station is selected.

Requires https://github.com/AtB-AS/atb-bff/pull/249 to be merged first.


src/mobility/utils.ts
```diff
@@ -15,17 +15,35 @@ import {getVisibleRange, toFeaturePoint} from '@atb/components/map';
...
+  (isStation(properties) &&
+    properties?.vehicleTypesAvailable?.some(
+      (types) => types.vehicleType.formFactor === FormFactor.Bicycle,
+    )) ??
+  false;
+
+export const getAvailableVehicles = (
+  types: VehicleTypeAvailabilityBasicFragment[] | undefined,
+) =>
+  types?.map((type) => type.count).reduce((sum, count) => sum + count, 0) ?? 0;
```
gorandalum, line 46: Should these also specify the FormFactor you want the count for?

gorandalum (approved):

src/mobility/utils.ts
KristianSelnas: Yes, absolutely. Good catch

PR #3506, opened 2023-04-26
refactor: added eslint rule no-restricted-syntax to restrict wildcard imports

hanne-at-atb:
I suggest we stop using wildcard import as they are considered an anti-pattern. The code becomes clearer when you explicitly import from a module, ref. https://app.deepsource.com/directory/analyzers/javascript/issues/JS-C1003 . 

Now introducing a new eslint rule to check for this and removing all wildcard imports in our project 👩🏼‍🚀

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,14 @@ module.exports = {
 
     'no-restricted-imports': ['warn', {patterns: ['@atb/components/*/']}],
 
+    'no-restricted-syntax': [
+      'error',
+      {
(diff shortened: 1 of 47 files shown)


rosvik (approved):
Looks fine by me 👍

jorelosorio (approved):
Good suggestion :) is more clear now what is used when importing.

gorandalum (approved):

HanneLH:
Regression test of App on Android and iOS ✅

PR #3515, opened 2023-04-28
feat: move map content, remove realtime from trip timestamps

hanne-at-atb:
Partly closes: https://github.com/AtB-AS/kundevendt/issues/3680 
* Moved map under travel details
* Removed real time icon from trip timestamps

<details>
<summary>Screenshot</summary>

image

</details>

### DIFF
--- a/src/components/divider/Divider.tsx
+++ b/src/components/divider/Divider.tsx
@@ -1,17 +1,19 @@
-import {View} from 'react-native';
 import React from 'react';
+import {View, ViewStyle} from 'react-native';
 import {useTheme} from '@atb/theme';
 
-export const Divider = () => {
+export const Divider = ({style}: {style: ViewStyle}) => {
   const {theme} = useTheme();
 
   return (
     <View
-      style={{
-        marginVertical: theme.spacings.medium,
-        borderTopWidth: theme.border.width.slim,
-        borderColor: theme.static.background.background_1.background,
-      }}
+      style={[
+        style,
+        {
+          borderTopWidth: theme.border.width.slim,
+          borderColor: theme.static.background.background_1.background,
+        },
+      ]}
     />
   );
 };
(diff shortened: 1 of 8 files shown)


src/components/divider/Divider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import React from 'react';
+import {useTheme} from '@atb/theme';
+
+export const Divider = () => {
+  const {theme} = useTheme();
+
+  return (
+    <View
+      style={{
+        marginVertical: theme.spacings.medium,
```
gorandalum, line 11: I don't think the shared `Divider` component should specify its own margin. Margins are situation dependent. I think it is better if it takes in a `style` parameter for the parent to specify margin.

src/travel-details-screens/components/Trip.tsx
```diff
@@ -51,6 +56,14 @@ export const Trip: React.FC<TripProps> = ({
...
   const {vehiclePositions} = useGetServiceJourneyVehicles(ids);
 
+  const tripPatternLegs = tripPattern?.legs.map((leg) => {
+    let mode: AnyMode = !!leg.bookingArrangements ? 'flex' : leg.mode;
+    return {
+      ...leg,
+      mode,
+    };
+  });
+
```
gorandalum, line 66: Now there are two places in the code doing this same mapping. Maybe take a look if it can either be done in a util-function, or maybe it is possible to do it earlier in the data fetching flow so it affects both the departure details and trip details screen.

gorandalum (approved):
Good work! A couple comments you may take a look at.

src/travel-details-screens/components/Trip.tsx
hanne-at-atb: As next part of this task the `PaginatedDetailsHeader` will be removed from `TripDetailsScreenComponent`. That will include removing the duplication, so will leave it for now ❇️

PR #3516, opened 2023-05-02
build: add correct FRAM url scheme to info.plist

toretefre:
Could you test this locally, @mortennordseth ?


toretefre:
Merging so you can test with a staging build tomorrow 👌

PR #3519, opened 2023-05-03
build: add fram to build-store workflows

mortennordseth:
This PR covers the necessary changes to add FRAM to the build-store workflows. I've used this branch to verify that everything works and the FRAM app is now published to Testflight and Google Play Internal. 

The FRAM app is built as an app bundle because the Google Play Console requires all new apps made after August 2021 to be published in the AAB format. Existing apps do not require this and can continue to use APK. (https://android-developers.googleblog.com/2020/11/new-android-app-bundle-and-target-api.html)

Closes https://github.com/AtB-AS/kundevendt/issues/3563


mortennordseth:
We have to do this for FRAM as the Google Play Console requires all new apps made after August 2021 to be published in the AAB format. Existing apps do not require this and can continue to use APK. (https://android-developers.googleblog.com/2020/11/new-android-app-bundle-and-target-api.html) 

There are a few benefits to using AAB over APK, such as app size optimization and [play feature delivery](https://developer.android.com/guide/playcore/feature-delivery). I am unsure if this is something we could take advantage of as the app is now, but it can be interesting to look into further. 

I think we should continue to use APK for AtB and NFK for now, and test AAB for FRAM. This way we can check if everything works as expected or if there are any changes that need to be done. What do you think, @gorandalum?

gorandalum (approved):
Looks ok. One possible suggestion would be to have `BUILD_AAB` or something in its own variable, and not be deduced from the org name, but I'm fine with whatever you think is best.

PR #3520, opened 2023-05-03
feat: add single zone flag on referenceData

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/3946

Makes three day ticket at NFK only selectable in a single zone using a `isApplicableOnSingleZoneOnly` flag that exists in the code from before (but not on the referenceData).

### DIFF
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -14,7 +14,7 @@ export type PreassignedFareProduct = {
   productAlias?: LanguageAndTextType[];
   distributionChannel: DistributionChannel[];
   alternativeNames: LanguageAndTextType[];
-  zoneSelectionMode: ZoneSelectionMode;
+  zoneSelectionMode?: ZoneSelectionMode;
   limitations: {
     userProfileRefs: string[];
     appVersionMin: string | undefined;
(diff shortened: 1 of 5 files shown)


gorandalum (dismissed):
The flag `isApplicableOnSingleZoneOnly` was used before we added the `zoneSelectionMode` option. I think we shouldn't reintroduce this flag, but rather make `zoneSelectionMode` work as intended. Just reach out if we should discuss how to make it work.

toretefre:
Hey @gorandalum, got a proposal from Mikael here: 

Could it be acceptable to introduce an optional `zoneSelectionMode` on the referenceData products that when present oversteers the `zoneSelectionMode` from the fareproductTypeConfig?

That way we are solving our need to set specific `zoneSelectionMode`s for given products within a group without introducing new types. I was thinking that the zoneSelectionMode type also could come from the shared fareproductTypeConfig-repo so that they will always be equal between the referenceData and the fareproductTypeConfig.

gorandalum:
@toretefre That is fully acceptable, and something I have thought we could introduce when necessary. I think everything under `configuration` in the `FareProductTypeConfig` can be overridden by each product if necessary (maybe except `requiresLogin`).

toretefre:
Maybe you could have a look as Gøran is unavailable today, @rosvik?

toretefre:
Great, can do!

PR #3521, opened 2023-05-04
feat: Update departure details screen content

ingvisim:
Update departure details screen to the new design

Still shows previous/next if screenreader is enabled, so user can choose departures even if from front page

<details>

Simulator Screen Shot - iPhone 14 Pro Max - 2023-05-04 at 13 51 17
Simulator Screen Shot - iPhone 14 Pro Max - 2023-05-04 at 13 51 05
</details>

Close https://github.com/AtB-AS/kundevendt/issues/3679


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -126,9 +126,9 @@ export const DepartureDetailsScreenComponent = ({
                 {title ?? t(DepartureDetailsTexts.header.notFound)}
               </ThemeText>
             </View>
-            {lastPassedStop || (vehiclePosition && mapData) ? (
+            {realtimeText || (vehiclePosition && mapData) ? (
               <View style={styles.headerSubSection}>
```
rosvik, line 130: I think we need to change the logic here when the map is removed. If there is no vehicle location data, there should still be a "Show in map" button. (Meaning, this section will always be visible)

PR #3522, opened 2023-05-05
fix: ui changes to contact sheet and correct notification icon on chat

adriawh:
Found that the contact sheet had no design in figma when looking into status disruption. Designed with øyvind in figma and applied the changes. Also fixed so that the chatIcon uses the correct notification. 

[Figma](https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/01.-Components?node-id=4632-15731&t=csb2PQqfPviugacR-4)

Simulator Screenshot - iPhone 14 - 2023-05-05 at 10 23 16


ingvisim (approved):
LGTM 🤠

PR #3523, opened 2023-05-05
fix: correct colors on line

adriawh:
The bottom line was fixed in another pr that is already merged in.

[Figma](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=10435-46556&t=XMbTjrHZhRsKTiFH-4)


rosvik (approved):

gorandalum (changes requested):
I think there is something wrong with the `Section`/`LocationInputSectionItem` if these styles are necessary. I would first look into why it is actually behaving this way, and not giving the "normal" delimiter color.

adriawh:
@oyvindmikkelsenatb

gorandalum (approved):

PR #3524, opened 2023-05-05
fix: wrong expiry date shown on payment card

adriawh:
When a cards shown expire date on the card is shown as 01.2024, it is valid thru the entire month of January and the expire date is technically 02.2024(this is how we store it).  I would say this is the correct way to store the expire date, and a simple fix is to do as I have done, subtract one month when getting the string to show.


rosvik (approved):

gorandalum (commented):
This will work both for winter and summer time?

adriawh:
> This will work both for winter and summer time?

I belive so. The expiry date will always be stored the same, one month ahead, and since the method is not using hours for any calculations it should. 

I also tried testing by setting the date to January and it seems to work correctly.

gorandalum (changes requested):
I have thought about this a little, and since the timestamp is exactly at midnight between two months, I think it is safer to subtract one day than one month.

gorandalum (approved):

PR #3526, opened 2023-05-08
feat: PostHog setup and instrumentation

KristianSelnas:


src/index.tsx
```diff
@@ -78,13 +79,15 @@ export const App = () => {
...
                                   <AppLanguageProvider>
                                     <GeolocationContextProvider>
-                                      <GlobalMessagesContextProvider>
-                                        <BottomSheetProvider>
-                                          <FeedbackQuestionsProvider>
-                                            <RootStack />
-                                          </FeedbackQuestionsProvider>
-                                        </BottomSheetProvider>
-                                      </GlobalMessagesContextProvider>
+                                      <AnalyticsContextProvider>
```
gorandalum, line 82: Is it intentional that the `AnalyticsContextProvider` is added both here and under `NavigationContainer`?

src/analytics/AnalyticsContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+      PostHog.initAsync(POSTHOG_API_KEY, {
+        host: POSTHOG_HOST,
+      }).then(setClient);
+    }
+  }, []);
+
+  // Since PostHog offers auto capture of navigation events,
+  // the PostHog provider expects to be setup inside a React Navigation object.
+  // In cases where the no navigation is available, e.g. the bottom sheet,
+  // the PostHog provider will throw an error.
```
rosvik, line 23: Does this mean we won't be able to have any analytics in the bottom sheet?

rosvik (approved):
LGTM 👍

src/index.tsx
KristianSelnas: Yes, It's necessary to get auto capture to work. Auto capture must be set up with navigation available.

PR #3535, opened 2023-05-10
feat: add live position icon styling

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/3686

simulator_screenshot_C074F0CF-ED63-4F6D-8E20-5E4957705E0D

### DIFF
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -175,7 +175,7 @@ const LiveVehicle = ({
   if (!coordinates) return null;
   return (
     <MapboxGL.ShapeSource
-      id="vehicle"
+      id="liveVehicle"
       shape={pointOf(coordinates)}
       cluster
       onPress={() => {
@@ -183,17 +183,17 @@ const LiveVehicle = ({
       }}
     >
       <MapboxGL.CircleLayer
-        id="icon"
-        minZoomLevel={1}
+        id="liveVehicleCircle"
+        minZoomLevel={4}
         style={{
           circleColor,
           circleRadius: 22,
         }}
       />
       <MapboxGL.SymbolLayer
-        id="transportIcon"
-        aboveLayerID="icon"
-        minZoomLevel={1}
+        id="liveVehicleIcon"
+        aboveLayerID="liveVehicleCircle"
+        minZoomLevel={4}
         style={{
           iconImage: {uri: iconName},
           iconAnchor: 'center',
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -203,6 +203,10 @@ const LiveVehicle = ({
   );
 };
 
+/**
+ * Get the name of the transportation mode icon which is stored under the
+ * "Images" section in Mapbox Studio.
+ */
 function vehicleIconName(mode?: AnyMode) {
   switch (mode) {
     case TransportMode.Bus:
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -184,7 +184,7 @@ const LiveVehicle = ({
     >
       <MapboxGL.CircleLayer
         id="liveVehicleCircle"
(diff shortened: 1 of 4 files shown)


src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -161,28 +182,42 @@ const LiveVehicle = ({coordinates, setShouldTrack}: VehicleIconProps) => {
         setShouldTrack(true);
       }}
     >
-      <MapboxGL.SymbolLayer
+      <MapboxGL.CircleLayer
         id="icon"
```
KristianSelnas, line 186: I'm not quite sure, but I believe these id's should be unique. If a different shape source adds a layer with this id 'icon' I think it may cause undesirable effects.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -161,28 +182,42 @@ const LiveVehicle = ({coordinates, setShouldTrack}: VehicleIconProps) => {
...
+          circleColor,
+          circleRadius: 22,
         }}
       />
       <MapboxGL.SymbolLayer
-        id="clusterCount"
-        minZoomLevel={1}
+        id="transportIcon"
         aboveLayerID="icon"
+        minZoomLevel={1}
```
KristianSelnas, line 196: Should this really be visible all the way to zoom level 1?

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -147,42 +158,66 @@ const styles = StyleSheet.create({
...
+        id="liveVehicleIcon"
+        aboveLayerID="liveVehicleCircle"
+        minZoomLevel={8}
         style={{
-          textField: 'BUS',
-          textColor: 'black',
-          textSize: 11,
-          textAnchor: 'center',
-          textAllowOverlap: true,
+          iconImage: {uri: iconName},
```
gorandalum, line 198: Could you elaborate a little about using images uploaded to MapBox instead of the transportation icons from the design system lib?

src/travel-details-screens/utils.ts
```diff
@@ -160,3 +161,14 @@ export function canSellCollabTicket(tripPattern: TripPattern) {
...
+/**
+ * Convert a `Mode` value to `TransportMode`. If the mode isn't a valid
+ * transport mode, return 'unknown'.
+ */
+export function modeToTransportMode(mode: Mode): TransportMode {
+  if (Object.values(TransportMode as any).includes(mode)) {
+    return mode as any as TransportMode;
+  }
+  return TransportMode.Unknown;
+}
```
gorandalum, line 174: Also possible to use `AnyMode` (and `AnySubMode`) in the component params, making this mapping perhaps not necessary, or at least done closer to where it is necessary to have it.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
rosvik: You can't inject svg data from the app (as far as I'm aware), so these images are manually uploaded in Mapbox Studio. `iconName` is a reference to this upload, not the design system. I can add a comment explaining this 👍 

I've only uploaded a subset of transport icons, since only a few is likely to appear with position data.

gorandalum: Aight. I looked at `SymbolLayer` doc now, and it seems pretty cumbersome to do it there. When reading the RN Mapbox docs it seemed like maybe [MarkerView](https://github.com/rnmapbox/maps/blob/main/docs/MarkerView.md) would be a more appropriate solution, but not necessary to go there now if we feel this works good enough.

gorandalum (approved):

KristianSelnas (approved):

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
rosvik: I think this works for now, but that might change as soon as we look at the direction arrow. Will look at MarkerView when I get to that 👍

gorandalum: Sounds as a good plan! Just by reading about MarkerView it seems like a good choice for this kind of feature.

PR #3536, opened 2023-05-10
fix: prevent crash when opening websocket connections

rosvik:
Related to feedback from testing of https://github.com/AtB-AS/kundevendt/issues/3418

Handles the different states of a websocket connection, and returns the subscription only when it is open. _Should_ prevent the app from crashing on slower iOS devices.

Also logs the status to Bugsnag if it doesn't follow the happy path.


gorandalum (changes requested):
- Even though the web socket is only used one place, I feel it can be refactored into it's own hook like `useWebSocket`.
- Could using the methods `onopen`, `onclose` and `onerror` remove the need for these "manual retries" and tracking the retry cound?
- Think a little about naming, the web socket changes name between `ws`, `openConnection` and `subscription` throughout the implementation, which makes it a little more confusing than it needs to be.

rosvik:
@gorandalum Re-did pretty much the entire implementation 😅 Decided to add a `useSubscription`, and named everything outside of it `subscription`, while the implementation uses `WebSocket`. Hopefully that makes more sense.

gorandalum:
> @gorandalum Re-did pretty much the entire implementation 😅 Decided to add a `useSubscription`, and named everything outside of it `subscription`, while the implementation uses `WebSocket`. Hopefully that makes more sense.

I think it looks much better :+1:

gorandalum (approved):

PR #3538, opened 2023-05-10
feat: configurable default tariff zone

mortennordseth:
This change allows us to configure the default tariff zone using the Firestore configuration. This has already been implemented in the webshop as NFK and FRAM do not want the default tariff zone to be the first from the list of tariff zones. 

The order for selecting a tariff zone is 
geolocation > default tariff zone from Firestore configuration > first tariff zone in the list of tariff zones.

Closes https://github.com/AtB-AS/kundevendt/issues/3659

### DIFF
--- a/src/configuration/FirestoreConfigurationContext.tsx
+++ b/src/configuration/FirestoreConfigurationContext.tsx
@@ -60,6 +60,7 @@ type ConfigurationContextState = {
   travelSearchFilters: TravelSearchFiltersType | undefined;
   appTexts: AppTexts | undefined;
   configurableLinks: ConfigurableLinks | undefined;
+  defaultTariffZone: TariffZone | undefined;
 };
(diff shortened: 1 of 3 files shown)


gorandalum (commented):
Throwing a suggestion out there:
Instead of using another field, it is possible to add `isDefault` or something on the actual tariff zones in reference data. I slightly prefer centralising tariff zone related data on the actual tariff zones over having it in multiple fields, but feel free to disagree 😊

toretefre (approved):
Looking good 👌

mortennordseth:
Good suggestion, @gorandalum! 👍  I've updated the PR with the suggested change and also made a [PR](https://github.com/AtB-AS/webshop2/pull/348) for the webshop so that we are consistent between app and webshop.

PR #3539, opened 2023-05-10
feat: Replace pagination in trip details screen with travel date

hanne-at-atb:
Closes one part of: https://github.com/AtB-AS/kundevendt/issues/3680

This PR removes the pagination between trip suggestions in details screen, it is replaced with the date of the trip when the trip is on a date later than today. 

image


gorandalum (changes requested):
I think there are some additional changes/simplifications that can be done when we no longer allow paging between trip patterns. The `TripDetailsScreenComponent` can probably just take in a single trip pattern and not an array of trip patterns, and then the logic in `useCurrentTripPatternWithUpdates` can be quite simplified, as it doesn't need to store updates on every trip pattern.

rosvik (approved):
Looking good to me 👍

gorandalum (approved):

PR #3544, opened 2023-05-11
feat: Add onboarding for mobile token without travelcard

adriansberg:
FRAM is not going to use or refer to travelcards in the app before we are all over on a new system and not using both an old and a new one. We are still going to utilize mobile token and the advantages that come with that, but we don't want to show information about travelcards. As a result of that I duplicated the existing onboarding for mobile token and called it MobileTokenWithoutTravelcardOnboarding, and removed references to travelcard. The SelectTravelTokenScreenComponent I did not duplicate, but I did a check for `disable_travelcard` (which is a new variable in remote config to control this) to see if the choice for travelcard should be hidden, in addition to using the correct translations.

Screenshots:

image
image
image
image
image
image
image

The illustrations are updated with correct colors in https://github.com/AtB-AS/design-system/pull/131

### DIFF
--- a/scripts/register-local-app-version.sh
+++ b/scripts/register-local-app-version.sh
@@ -7,7 +7,7 @@
 echo "Loading all env variables from .env file"
 export $(grep -v '^#' .env | gxargs -d '\n') > /dev/null 2>&1
 
-credentials=$(gcloud secrets versions access --project fram-staging-dc00 --secret=entur-client-credentials-publish latest)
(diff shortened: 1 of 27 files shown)


gorandalum (commented):
I think this can be solved in a more generic manner. Also it is possible for us in AtB to make some "sacrifices" to make the onboarding more generic so it works for every transport company with just slight modifications.

Maybe we can do a session where we evaluate what the options are? Just reach out.

adriansberg:
Yes, let's have a session where we can discuss needs and solutions. Do you have time today, e.g. 14:00?

PR #3545, opened 2023-05-12
fix: Remove overlays and show walking distance for city bike stations 

KristianSelnas:


src/components/map/hooks/use-decide-camera-focus-mode.tsx
```diff
@@ -47,6 +53,11 @@ export const useDecideCameraFocusMode = (
         return;
       }
 
+      if (mapSelectionAction.source === 'cluster-click') {
+        setCameraFocusMode(undefined);
```
gorandalum, line 57: I think it would be slightly more correct to set a camera focus mode which represents the cluster expansion zoom here, and to do it in the `useTriggerCameraMovement` hook like other zooms. But I know it is fragile to change it, so I'm okay with having it as it is now.

gorandalum (approved):

PR #3548, opened 2023-05-12
feat: Car sharing

KristianSelnas:


src/components/map/components/mobility/Stations.tsx
```diff
@@ -1,67 +1,67 @@
...
     })),
-  };
+  });
+
+  const bikeStations = stationsWithCount(
+    getFeaturesOfType(stations, FormFactor.Bicycle),
+  );
+  const carStations = stationsWithCount(
+    getFeaturesOfType(stations, FormFactor.Car),
+  );
```
gorandalum, line 59: I had to read through this code and the functions a couple of times to get a good grasp of what it was doing. Could it be better to combine `stationsWithCount` and `getFeaturesOfType` into a single method? I also have a slight preference for declaring functions outside of the component body if possible.

gorandalum (approved):

PR #3550, opened 2023-05-14
feat: Add onboarding for mobile token without travelcard, single onboarding

adriansberg:
FRAM is not going to use or refer to travelcards in the app before we are all over on a new system and not using both an old and a new one. We are still going to utilize mobile token and the advantages that come with that, but we don't want to show information about travelcards. As a result of that I added a new variable in remote config, `disable_travelcard`, to control which translations to be shown and whether to hide information about travelcards. 

Screenshots:

image
image
image
image
image
image
image

The illustrations are updated with correct colors in https://github.com/AtB-AS/design-system/pull/131

Closes https://github.com/AtB-AS/kundevendt/issues/3903


toretefre:
Looking good from my phone, but I’ll let Gøran have the final say here, following the meeting tomorrow.

Two thoughts:

1. Are there existing remote config values prefixed with `disable_`? I was under the impression that we mostly enabled things with remote config, not disabled, but disable might be preferred as AtB and NFK already have “enabled” travelcards.
2. Is the wording “mobil” used at AtB currently or should we ask @cflie whether “mobiltelefon” or “enheter” (in case someone should use their tablet or watch or whatever) could be preferred in the translations?

adriansberg:
1. Seeing as we should most definitely defaut this to false, I figured that naming it `disable_...` made more sense. But I see that many of the existing variables start with `enable_...` so we could easily switch over to use that instead and default it to `true`.
2. I don't know if the other translations has been run through with Christoffer. But since you've tagget him now, maybe he sees this? 👀

cflie:
2. @toretefre and @adriansberg I think "mobiltelefon" is clearer than just "mobil". However we should use the same wording as AtB and NFK. And "mobil" is common now or what? "Enhet" may confuse the user, will all users understand "enhet"? Is it possible to use the app on a smart watch (as a device)? Do we have any experience with anyone using the apps from AtB and NFK on a tablet to purchase tickets, and then in our case that they will bring the tablet on board to scan the QR code?

adriansberg:
Looks like "mobil" and "phone" is mostly used now.

Maybe @toretefre has some insights in the usage of tablets.

adriansberg:
Another question for you @cflie: the changes we are doing here are for the onboarding. The current onboarding refers to travelcards and the ability to switch between travelcard and phone as a ticket bearer. So the changes here are hiding the references to travelcards and showing translations without mentions of travelcard.

The question is whether we want to show this onboarding again when we introduce travelcards?

src/stacks-hierarchy/Root_MobileTokenOnboarding/components/InspectableTokenInfo.tsx
```diff
@@ -53,15 +59,22 @@ export function InspectableTokenInfo({
               color={themeColor}
               isMarkdown={true}
             >
-              {isTravelCardToken(inspectableToken)
+              {disable_travelcard
```
adriansberg, line 62: Not a huge fan of how this ternary is.

mortennordseth (approved):
LGTM! 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -345,7 +352,14 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
...
               </ThemeText>
-              <BetaTag style={style.betaTag} />
+              <View style={style.betaLabel}>
+                <ThemeText
+                  color="background_accent_3"
+                  style={style.betaLabelText}
+                >
+                  BETA
+                </ThemeText>
+              </View>
```
gorandalum, line 362: @adriansberg What's the reason for this change 🤔

adriansberg: Ehh. Must've snuck in there when rebasing. Fixing as soon as I have time.

PR #3554, opened 2023-05-16
fix: Fix an issue while migrating filters from old versions to new versions

jorelosorio:
There was a case where filter were not fully migrated, after review from @tormoseng 

Fixes: https://github.com/AtB-AS/kundevendt/issues/3857#issuecomment-1547365503

### DIFF
--- a/src/storage/StorageModel.tsx
+++ b/src/storage/StorageModel.tsx
@@ -50,7 +50,7 @@ const errorHandler = (error: any) => {
 };
 
 const leaveBreadCrumb = (
-  action: 'read-single' | 'save-single' | 'read-all',
+  action: 'read-single' | 'save-single' | 'remove-single' | 'read-all',
   key?: string,
   value?: string | null,
 ) => {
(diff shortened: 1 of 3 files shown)


gorandalum (changes requested):
I think this can be done some other way. Checking with `Array.isArray` makes the typing as `any[]` which is kinda imprecise. You should check the type of `oldStoredFilters`, I think that is the actual culprit.

jorelosorio:
> I think this can be done some other way. Checking with `Array.isArray` makes the typing as `any[]` which is kinda imprecise. You should check the type of `oldStoredFilters`, I think that is the actual culprit.

Inside getFilters there is a cast as the object is needed.

```
return (
      userFilters ? JSON.parse(userFilters) : {}
    ) as TravelSearchFiltersSelectionType;
```

gorandalum:
> Inside getFilters there is a cast as the object is needed.

Yes, but it is cast to `TravelSearchFiltersSelectionType` also for old filters, which is not the correct type.

jorelosorio:
> > Inside getFilters there is a cast as the object is needed.
> 
> Yes, but it is cast to `TravelSearchFiltersSelectionType` also for old filters, which is not the correct type.

Ok, I got it and made a refactor to it :)

gorandalum (changes requested):

gorandalum (approved):

PR #3558, opened 2023-05-16
fix: Show departures more than 3 hours in the future

ingvisim:
Fix issue with departures only showing up to 3 hours in advance. Also fix unexpected behaviour where we get more or less departures by switching between days.

Close https://github.com/AtB-AS/kundevendt/issues/3983


jorelosorio (approved):
Looks good, we save the result in a state

rosvik (approved):

PR #3559, opened 2023-05-16
feat: Draft for how local storage versioning could look

gorandalum:


src/storage/StorageModel.tsx
```diff
@@ -63,26 +64,39 @@ const leaveBreadCrumb = (
...
+  const {localStorageVersions: versions} = useFirestoreConfiguration();
+  return {
+    /** Necessary for communication between App and iOS Widget */
+    setAppGroupName: async (groupName?: string) => {
+      if (Platform.OS === 'ios') {
+        await AsyncStorage.setAppGroupName(groupName).catch(errorHandler);
+      }
+    },
+    get: async (key: string) => {
+      const keyAndVersion = appendVersion(key, versions);
```
adriansberg, line 77: ~When using a saved key to decide whether e.g. onboarding should be shown or not, will it only be checking that whatever is saved in local storage is different to the version is set in Firestore?~

~Not sure if I understood based on this how we'll know that the key saved is "older" than what we will currently be using.~

I think I understood it while I was writing the comment. If a version is set in Firestore, the new key won't exist in the local storage thus the getter will return false when checking the key.

PR #3564, opened 2023-05-22
fix: Correct filter outline color for NFK app

mikaelbr:
There were inconsistencies in the use of interactive colors on the filter menu, whereas one used interactive_0 and the other _1. This makes it consistent. It doesn't look like it affects AtB in any way.

Fixes https://github.com/AtB-AS/kundevendt/issues/3930

Screenshot 2023-05-22 at 07 58 10

Screenshot 2023-05-22 at 07 59 32


toretefre (approved):
Looks to be working great, also tried for NFK and seems just fine 👌

PR #3565, opened 2023-05-22
Fix: Add proper accessibility to the link button

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/3857


hanne-at-atb (approved):
Looking good!

PR #3566, opened 2023-05-22
fix: Correct focus when opening trip

gorandalum:
Fixes issue commented here:
https://github.com/AtB-AS/mittatb-app/pull/3534#issuecomment-1549397230


hanne-at-atb (approved):
Looks good! 🌻

PR #3567, opened 2023-05-23
bug: Accesibility fixes and remove a code warning

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/3857
Depend on: https://github.com/AtB-AS/firestore-configuration/pull/171


gorandalum (changes requested):
Shouldn't use index as key:
https://react.dev/learn/rendering-lists#why-does-react-need-keys

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -49,10 +50,13 @@ export const CityZoneMessage: React.FC<CityZoneMessageProps> = ({
   };
 
   const messageActions = selectedCityZones.map((cityZone) => ({
+    id: `${cityZone.id}_action`,
     text: cityZone.name,
     onPress: () => openUrlForCityZone(cityZone),
   }));
 
+  console.log(messageActions);
```
gorandalum, line 58: Should be removed

gorandalum (approved):
Looks good! Just a note that I think using `name` as key instead of adding `id` would work just as well.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
jorelosorio: Oops! My fault, removed now ;)

jorelosorio:
> Looks good! Just a note that I think using `name` as key instead of adding `id` would work just as well.

I thought the same, but the name might not be OK for an ID because it might contain special characters, spaces, etc. So it is nice to have a proper field for this.

gorandalum:
> I thought the same, but the name might not be OK for an ID because it might contain special characters, spaces, etc. So it is nice to have a proper field for this.

Just FYI React keys doesn't have limitations on what they can include, and can have spaces, special characters etc. The only rule is that they should be unique.

PR #3569, opened 2023-05-23
fix: Correct info text about trains on 24hr ticket

adriawh:
24hr tickets are to be used on trains in the same way as period tickets. It was also decided together with @oyvindmikkelsenatb and @Sebstorvik that the single tickets should get the same text


gorandalum (approved):

ingvisim (approved):
🅰️ ➕

PR #3574, opened 2023-05-24
feat: show info regarding departure from first stop in trip details

hanne-at-atb:
Closes: https://github.com/AtB-AS/kundevendt/issues/3680

Linked to changes in BFF: https://github.com/AtB-AS/atb-bff/pull/257

### DIFF
--- a/src/travel-details-screens/use-realtime-text.ts
+++ b/src/travel-details-screens/use-realtime-text.ts
@@ -1,15 +1,9 @@
 import {usePreferences} from '@atb/preferences';
 import {DepartureDetailsTexts, useTranslation} from '@atb/translations';
-import {EstimatedCallWithMetadata} from '@atb/travel-details-screens/use-departure-data';
 import {formatToClock, isInThePast} from '@atb/utils/date';
 import {getTimeRepresentationType} from '@atb/travel-details-screens/utils';
 
-export const useRealtimeText = (
-  estimatedCallsWithMetadata: EstimatedCallWithMetadata[],
-) => {
-  const lastPassedStop = estimatedCallsWithMetadata
-    .filter((a) => a.actualDepartureTime)
-    .pop();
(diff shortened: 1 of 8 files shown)


src/travel-details-screens/use-realtime-text.ts
```diff
@@ -1,21 +1,30 @@
...
 import {formatToClock, isInThePast} from '@atb/utils/date';
 import {getTimeRepresentationType} from '@atb/travel-details-screens/utils';
 
-export const useRealtimeText = (
-  estimatedCallsWithMetadata: EstimatedCallWithMetadata[],
-) => {
-  const lastPassedStop = estimatedCallsWithMetadata
-    .filter((a) => a.actualDepartureTime)
-    .pop();
+export type EstimatedCall = {
```
hanne-at-atb, line 6: I'm not sure about the naming here and whether this is a good approach for typing what the hook should take in

rosvik: Had to look at this locally, and noticed that we can simplify this quite a bit if we set `lastPassedStop` and `quay` to optionals here, then simplify usage:

DepartureDetailsScreenComponent:
```ts
const realtimeText = useRealtimeText({
  lastPassedStop,
  firstStop,
});
```

TripSection:
```ts
const realtimeText = useRealtimeText({
  lastPassedStop,
  firstStop: leg.serviceJourneyEstimatedCalls[0],
});
```

Would also rename this to something like `RealtimeTextParams` and removed the export when it's no longer used.

gorandalum (approved):

src/travel-details-screens/use-realtime-text.ts
hanne-at-atb: Thanks for the input @rosvik ! I had a look at this with @gorandalum, and we found a different solution that made fewer changes to the original implementation which we thought was a bit better :)

PR #3576, opened 2023-05-24
feature: Show message box only within zones

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/4061

Quick resume: The city box is only shown when the origin and destination are in the same city zone.

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
@@ -63,20 +63,16 @@ export const CityZoneMessage: React.FC<CityZoneMessageProps> = ({
     },
   ];
 
-  if (messageActions) {
-    return (
-      <Section style={style.cityZoneMessage}>
-        <CityZoneBox
-          message={t(CityBoxMessageTexts.message)}
-          icon={() => <FlexibleTransport />}
-          onDismiss={onDismiss}
-          actionButtons={messageActions}
-        />
-      </Section>
-    );
-  }
-
-  return null;
+  return (
+    <Section style={style.cityZoneMessage}>
+      <CityZoneBox
+        message={t(CityBoxMessageTexts.message)}
+        icon={() => <FlexibleTransport />}
+        onDismiss={onDismiss}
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -49,11 +55,13 @@ export const CityZoneMessage: React.FC<CityZoneMessageProps> = ({
...
-    text: cityZone.name,
-    onPress: () => openUrlForCityZone(cityZone),
-  }));
+  const messageActions = [
+    {
+      id: `${fromCityZone.id}_action`,
+      text: fromCityZone.name,
+      onPress: () => openUrlForCityZone(fromCityZone),
+    },
+  ];
```
rosvik, line 64: Will there ever be more than one messageAction now? If so, maybe we could simplify the logic here and in `CityZoneBox`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
```diff
@@ -49,11 +55,13 @@ export const CityZoneMessage: React.FC<CityZoneMessageProps> = ({
...
 
   if (messageActions) {
```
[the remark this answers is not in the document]
jorelosorio: Fixed :)

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/CityZoneMessage.tsx
jorelosorio: Not now, but I am working another ticket that requires more than one message action

gorandalum (commented):
The check on the `enabled` flag, is that no longer necessary?

jorelosorio:
> The check on the `enabled` flag, is that no longer necessary?

Yes it should be used, but is not included in the condition I missed it, I am working on a related task where I hope include it again using:

```
if (!fromCityZone?.enabled || !toCityZone?.enabled) {
    return null;
}
```

PR #3585, opened 2023-05-25
fix: remove train notices for FRAM

mortennordseth:
Currently the notice "Periodebilletter i sone A kan også brukes på tog i sone A" is shown for FRAM's "fylkesbilletter". This is because these travel rights don't have any zones and the `shouldShowValidOnTrainNotice`-check returns true. I've updated it to handle travel rights where the `tariffZoneRefs` is an empty array. 

Fixes https://github.com/AtB-AS/kundevendt/issues/4077


gorandalum (approved):
A classic! Not taking into consideration that `Array.every` returns `true` when the array is empty.

PR #3586, opened 2023-05-25
Fix: Remove comma from trip details time

marius-at-atb:
Force " " as conjunction char instead of default ","

No comma between "3 t" and "31 min":


gorandalum (approved):

rosvik (approved):
👍👍

PR #3587, opened 2023-05-25
fix: Do not read flex mode from booking data

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/4103


rosvik (approved):

gorandalum (commented):
If the format is '<company>:FlexibleLine:<number>' like 'AtB:FlexibleLine:2', I think the check should be more directed at that and check for 'FlexibleLine' instead of just 'flexible'.

Also @tormoseng will do an extra check with Entur about what is the optimal way to identify a flex line.

PR #3588, opened 2023-05-25
fix: Leselist problem stopping user from selecting a travel suggestion

adriawh:
#### Disappearing text and action
By testing with the text-panel enabled at the bottom, I belive this fixes the issue. However I am not able to test since I dont have a "leselist". 
Before this fix, the text disappeared when it was done reading. By adding the role the text stays there. 

#### Hard to read text
I also found the text to be built in a way that made it hard to read. It was read correctly but shown with large gaps. By joining the strings this way it is more readable and also easier to edit the texts later:

old: 

new:

Adding to this, I want to make a suggestion that the `screenReaderPause` should be a ". " instead of "\n" as this will make it easier to read if the text panel is adopted, but will not make a difference in how it is read by the screen reader. 

#### Margin moved
Lastly I moved a margin to so that the screen reader box marking what being read just covers the item, and not with the margin: 

Old:

new:


gorandalum (approved):
LGTM :+1:

PR #3589, opened 2023-05-25
feat: refund information link

mortennordseth:
Closes https://github.com/AtB-AS/kundevendt/issues/4004

FRAM would like to have a link to their refund form so I've extended the configurable links with `refundInfo`. 

image

### DIFF
--- a/src/configuration/FirestoreConfigurationContext.tsx
+++ b/src/configuration/FirestoreConfigurationContext.tsx
@@ -43,10 +43,10 @@ export type AppTexts = {
 };
 
 type ConfigurableLinks = {
-  ticketingInfo: LanguageAndTextType[];
-  termsInfo: LanguageAndTextType[];
-  inspectionInfo: LanguageAndTextType[];
-  refundInfo: LanguageAndTextType[];
+  ticketingInfo?: LanguageAndTextType[];
+  termsInfo?: LanguageAndTextType[];
+  inspectionInfo?: LanguageAndTextType[];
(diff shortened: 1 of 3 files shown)


toretefre (approved):

src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -398,12 +399,15 @@ function getConfigurableLinksFromSnapshot(
   const ticketingInfo = mapLanguageAndTextType(urls?.get('ticketingInfo'));
   const inspectionInfo = mapLanguageAndTextType(urls?.get('inspectionInfo'));
   const termsInfo = mapLanguageAndTextType(urls?.get('termsInfo'));
+  const refundInfo = mapLanguageAndTextType(urls?.get('refundInfo'));
 
-  if (!ticketingInfo || !inspectionInfo || !termsInfo) return undefined;
+  if (!ticketingInfo || !inspectionInfo || !termsInfo || !refundInfo)
+    return undefined;
```
gorandalum, line 405: Does this make it necessary for every transport company to have a refund info link?

mortennordseth: You're absolutely correct. I'll update the PR to fix this. Thanks!

gorandalum (approved):
I think when possible we should avoid checks on transport companies in the code. Maybe we should also use links for these kinds of info @Sebstorvik?

For now we will keep it like this.

PR #3591, opened 2023-05-26
fix: Ensures spacing consistency in title header

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3990

### DIFF
--- a/src/components/screen-header/ScreenHeader.tsx
+++ b/src/components/screen-header/ScreenHeader.tsx
@@ -129,7 +129,7 @@ const BaseHeader = ({
           css.headerTitle,
           {
             // Make space for absolute positioned buttons in case they are offset below title
-            marginBottom: buttonsTopOffset,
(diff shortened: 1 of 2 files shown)


src/components/screen-header/ScreenHeader.tsx
```diff
@@ -168,7 +168,8 @@ const BaseHeader = ({
 const useHeaderStyle = StyleSheet.createThemeHook((theme) => ({
   container: {
     paddingHorizontal: theme.spacings.medium,
-    paddingVertical: theme.spacings.large,
+    paddingTop: theme.spacings.large,
+    paddingBottom: theme.spacings.medium,
```
gorandalum, line 172: Did you check other pages without the global message box, if they are still the same?

marius-at-atb: Done now, fixed in the second commit

PR #3593, opened 2023-05-26
fix: Use line type and remove booking contact

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/4103
Depend on: https://github.com/AtB-AS/atb-bff/pull/260

Before:

image

After:

Screenshot_1685104708


src/travel-details-screens/components/Trip.tsx
```diff
@@ -157,7 +157,7 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
     paddingTop: theme.spacings.medium,
   },
   container: {
-    paddingBottom: theme.spacings.medium,
+    paddingVertical: theme.spacings.medium,
```
jorelosorio, line 160: Before this change:

image

After the change:

image

src/travel-details-screens/utils.ts
```diff
@@ -162,5 +162,5 @@ function someLegsAreByTrain(tripPattern: TripPattern): boolean {
 }
 
 export function isLegFlexibleTransport(leg: Leg): boolean {
-  return leg.line?.id.toLowerCase().includes('flexible') ?? false;
+  return !!leg.line?.flexibleLineType;
```
jorelosorio, line 165: At least not null?

src/travel-details-screens/components/Trip.tsx
jorelosorio: It was fixed in another branch

gorandalum (approved):
Looks good! Also remember to update the `journeyplanner-types_v3.ts` as the updates in the bff should be reflected in the app.

jorelosorio:
> Looks good! Also remember to update the `journeyplanner-types_v3.ts` as the updates in the bff should be reflected in the app.

I made a commit copying the generated file from BFF to the app, please double-check :)

src/travel-details-screens/utils.ts
gorandalum: Seems correct!

PR #3596, opened 2023-05-30
feat: Mobility operator filter

KristianSelnas:


src/api/mobility.ts
```diff
@@ -22,6 +32,7 @@ export const getVehicles = (
     lon,
     range: Math.ceil(range),
     formFactors: FormFactor.Scooter, //TODO: Read from variables
```
jorelosorio, line 34: Is this something to cover on this PR? inject formFactor from outside?

src/api/mobility.ts
```diff
@@ -42,3 +53,48 @@ export const getVehicle = (
...
+  id: string,
+  opts?: AxiosRequestConfig,
+): Promise<BikeStationFragment | undefined> => {
+  const url = '/bff/v2/mobility/station/bike';
+  const query = qs.stringify({ids: id});
+  return client
+    .get<GetBikeStationQuery>(stringifyUrl(url, query), {
+      ...opts,
+    })
+    .then((res) => (res.data.stations ? res.data.stations[0] : undefined));
```
jorelosorio, line 86: We could check if the `stations` if bigger than 0.

jorelosorio: Or `res.data.stations?.[0]`

src/api/mobility.ts
```diff
@@ -42,3 +53,48 @@ export const getVehicle = (
...
+export const getBikeStation = (
+  id: string,
+  opts?: AxiosRequestConfig,
+): Promise<BikeStationFragment | undefined> => {
+  const url = '/bff/v2/mobility/station/bike';
+  const query = qs.stringify({ids: id});
+  return client
+    .get<GetBikeStationQuery>(stringifyUrl(url, query), {
+      ...opts,
+    })
```
jorelosorio, line 85: `stringifyUrl(url, query), opts` might work as well. instead of `{...opts}`

src/api/mobility.ts
```diff
@@ -42,3 +53,48 @@ export const getVehicle = (
...
+  const query = qs.stringify({
+    lat,
+    lon,
+    range: Math.ceil(range),
+    availableFormFactors,
+    operators,
+  });
+  return client
+    .get<GetStationsQuery>(stringifyUrl(url, query), {
+      ...opts,
```
jorelosorio, line 71: See the comment below ;)

src/mobility/use-operators.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
+
+export const useOperators = () => {
+  const {mobilityOperators} = useFirestoreConfiguration();
+
+  return (types: undefined | FormFactor | FormFactor[]) => {
+    if (!types) {
+      return mobilityOperators ?? [];
+    }
+    types = Array.isArray(types) ? types : [types];
```
jorelosorio, line 11: Mmm, I wonder, If it will be at the end an array why not just (types?: FormFactor[])

jorelosorio (changes requested):

gorandalum (approved):

src/api/mobility.ts
KristianSelnas: It's an old TODO. It may appear to be added now, since this code has been moved from two files to one. It will be resolved as soon as other vehicles than scooters are added to the app. (bikes and cars are _stations_)

src/api/mobility.ts
KristianSelnas: Yes, this is strictly not necessary, but really convenient when you need to test against a local BFF an have to add `baseUrl`:
```
{
    ...opts,
    baseUrl: 'http://localhost:8080'
}
```
I'll remove it, nevertheless. All minor things like this will eventually add up

src/mobility/use-operators.tsx
KristianSelnas: It's just more convenient from where this function is called. E.g. I can now do like this `useOperators(FormFactor.Scooter)` rather than `useOperators([FormFactor.Scooter])`.

src/mobility/components/filter/MobilityFilters.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {MapFilterType, OperatorFilterType} from '@atb/components/map';
+import {StyleSheet} from '@atb/theme';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+type Props = {
+  initialFilter: MapFilterType;
+  onFilterChanged: (filter: MapFilterType) => void;
+};
+
+export const MobilityFilters = ({initialFilter, onFilterChanged}: Props) => {
```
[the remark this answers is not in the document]
KristianSelnas: Hmm... I see your point. Good spot 👍  React noob mistake.

Not sure if I see the best solution, though. Ideally, I don't want all details about mobility filter stuff to bleed all the way out to `MapFilters`.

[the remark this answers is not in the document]
KristianSelnas: Any suggestions?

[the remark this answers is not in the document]
KristianSelnas: I've made a change suggestion now (see my latest commit, https://github.com/AtB-AS/mittatb-app/pull/3596/commits/ef0b6069fc882fee3041abe55746af3e0c2c329b).

I don't think it's too bad. Updating the `MapFilterType` should probably be owned by `MapFilterSheet` anyways.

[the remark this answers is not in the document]
gorandalum: > I've made a change suggestion now

I think it looks good. A little more verbose, but it is more correctly implemented. 

Another strategy for cases like this is to split it into a React Component and hook. That would mean `MobilityFilters` only has the responsibility to show the toggles, and sync the toggle state to the parent. And then it would be a `useMobilityFilters` (or something like it) which takes in the toggle states and has the responsibility for applying them correctly and returning the state with the correct vehicles. Not necessarily better to do it like this here now, but just mentioning it.

> Any suggestions?

Just reach out on Slack if something should be discussed right away. I try to take a look at PRs at morning and after lunch, so will not always respond to comments immediately.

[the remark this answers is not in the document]
KristianSelnas: Ok, I'll leave it like this for now. 

I'll wait until after the release branch has been made to merge this. It requires both BFF and config changes, and it is strictly a part of next sprint anyways.

jorelosorio (approved):

PR #3598, opened 2023-05-30
fix: PostHog event logging for mobility

KristianSelnas:


src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -48,6 +49,21 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
     rentalAppUri,
   });
   const stationName = useTextForLanguage(station?.name.translation) ?? '';
+  const analytics = useAnalytics();
+
+  useEffect(() => {
+    analytics.logEvent('Mobility', 'Car sharing station selected', {
+      id: station?.id,
```
jorelosorio, line 56: Is `station?.id` the same as `stationId` in that case, we could use that value instead to avoid null values in case.

jorelosorio (changes requested):

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -48,6 +49,21 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+      name: stationName,
+      operator: {
+        id: station?.system.operator.id,
+        name: getTextForLanguage(
+          station?.system.operator.name.translation,
+          language,
+        ),
+      },
+    });
+  }, [station]);
```
gorandalum, line 66: There are some deviations here with regards to react best practices as there are multiple [reactive values](https://react.dev/learn/lifecycle-of-reactive-effects#all-variables-declared-in-the-component-body-are-reactive) used which are not declared in the dependency array, like `stationName`, `analytics` and `language`. 

By reading this code I'm guessing that changing the language of the app won't have an effect when the `useEffect` reruns as it uses the stale `language` value.

I think to do this more correctly logging the analytics can be done in [event handlers instead of in effects](https://react.dev/learn/separating-events-from-effects#choosing-between-event-handlers-and-effects).

src/mobility/components/CarSharingStationBottomSheet.tsx
KristianSelnas: It is true for stationId, yes. I'll change that.

It might be tempting to use `operatorName` from the above hook as well, but this is computed in a separate tick and if `operatorName` is included in the `useEffect` dependencies it will cause the event to be logged twice.

src/mobility/components/CarSharingStationBottomSheet.tsx
KristianSelnas: I agree, but this was an attempt to log the event with a bit more metadata than you have at the time of the event. When the event occurs you just have the `id` available, not stuff like operator, station name, etc. This requires the station to be looked up by id first.

I should probably not have used `stationName`, but rather compute it like it's done for `operator.name`. It's actually a point not to have any other dependencies than `station` in this `useEffect`, as we don't want duplicate event logging when other events happen (such as changing language). 

I'm not sure what the best solution here is. It's a tradeoff, no matter what. It's either accepting that we don't get all the data we ideally want in PostHog, or accept that logging is done in a less than ideal way in a `useEffect`. What do you think?

gorandalum: This really becomes quite easier when React introduces [effect events](https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event), but in the meantime I think we should adhere to the rules. To fix them here:
- Compute station name inside the effect
- Do we wan't to log the name based on the clients language? Maybe better to just hardcode the language to Norwegian inside the effect, thus removing the reactive value `language`.
- `analytics` should be in the dependency list. If this causes multiple log events, then `useMemo` should be used in `useAnalytics` so a new object isn't created every time.

KristianSelnas: Adding `analytics` to the dependencies array triggers multiple events, even when `useMemo` is used in `useAnalytics`.

KristianSelnas: I've pushed an update now

jorelosorio (approved):

PR #3602, opened 2023-05-30
fix: populate user count state with only option

toretefre:
Potentially closing https://github.com/AtB-AS/kundevendt/issues/4078

Fills user_state_count with the only possible option if `selectionMode` is none.


toretefre:
Closing following https://github.com/AtB-AS/mittatb-app/pull/3603

PR #3603, opened 2023-05-31
fix: proper support selectionMode none for TravellerSelection

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/4078

Adds support for selecting the first and best traveller when traveller selection mode is set to none. This is the same behaviour as in the webshop. Technically there could be more selectable / supported traveller types specified in Firestore, but I think it is reasonable to select the top most item.

Screenshot 2023-05-31 at 09 20 41

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
@@ -35,10 +35,22 @@ export function TravellerSelection({
       selectableUserProfiles.some((b) => a.id === b.id),
     );
   useEffect(() => {
-    const filteredSelection = userCountState.userProfilesWithCount.filter((u) =>
-      selectableUserProfiles.find((i) => i.id === u.id),
-    );
-    setTravellerSelection(filteredSelection);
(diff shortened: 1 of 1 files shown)


toretefre (approved):

toretefre:
Seems to be failing and giving BAD_RESPONSE with NFK youth ticket, trying to find out if related or not

gorandalum (approved):

toretefre:
Tested and works for FRAM fylkesbillett (similar ticket as NFK youth), so leaning towards something at Entur and nothing wrong here.

mikaelbr:
Was thinking on my way to lunch that there might be issues with overriding selectable when it happens in a useEffect like this and not derived state. But I guess this won't be an issue with selectmode none. Theoretically a bigger issue with existing code, so if that hasn't been a problem until this point I guess it's fine. Merging

gorandalum:
> Was thinking on my way to lunch that there might be issues with overriding selectable when it happens in a useEffect like this and not derived state. But I guess this won't be an issue with selectmode none. Theoretically a bigger issue with existing code, so if that hasn't been a problem until this point I guess it's fine. Merging

I think it should work ok, but I agree it is probably overkill. In general I think we may have been too eager to use effects and state around the app in places where it is not necessary.

PR #3604, opened 2023-05-31
fix: Set correct zoom level on map

ingvisim:
Set correct zoom level on opening the travel details map


gorandalum (approved):

HanneLH:
Verifisert ok for Android og iOS ✔️

PR #3606, opened 2023-05-31
fix: Upgrading mapbox version

jorelosorio:
It fixes an issue regarding rotation after discussion with @marius-at-atb 

Closes https://github.com/AtB-AS/kundevendt/issues/3688

### DIFF
--- a/src/components/map/MapConfig.ts
+++ b/src/components/map/MapConfig.ts
@@ -1,15 +1,15 @@
-//import {CameraStop, MapViewProps} from '@rnmapbox/maps';
 import {Platform} from 'react-native';
 import {MAPBOX_STOP_PLACES_STYLE_URL} from '@env';
 import {CameraStop} from '@rnmapbox/maps/lib/typescript/components/Camera';
 
 export const MapViewConfig = {
-  compassViewPosition: 1, // Upper right
   compassEnabled: true,
   scaleBarEnabled: false,
-  compassViewMargins: {
-    x: Platform.select({default: 10, android: 6}),
-    y: 60,
+  // `mapbox` (v10) Adds compass offset `compassViewMargins` is still supported but generates issues:
+  // Mapbox error fireEvent failed: <rnmapbox_maps.RCTMGLEvent: 0x6000028a0fe0>
+  compassPosition: {
+    top: 60,
+    right: Platform.select({default: 10, android: 6}),
   },
   attributionPosition: Platform.select({
     default: {bottom: 8, left: 95},
--- a/src/components/map/types.ts
+++ b/src/components/map/types.ts
@@ -38,7 +38,10 @@ export type MapRegion = {
   center: Position;
 };
 
-export type MappPadding = number | number[];
+export type MapPadding =
+  | number
+  | [number, number]
+  | [number, number, number, number];
 
 export type VehiclesState = {
   vehicles: FeatureCollection<GeoJSON.Point, VehicleFragment>;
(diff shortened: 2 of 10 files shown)


jorelosorio:
This change introduces another issues regarding `Mapbox"' has no exported member 'Padding'` and other, it will be on hold until have more time to migrate maps.

src/tariff-zones-selector/TariffZonesSelectorMap.tsx
```diff
@@ -182,7 +183,7 @@ const TariffZonesSelectorMap = ({
             </MapboxGL.ShapeSource>
             {featureCollection.features.map((f) => (
               <MapboxGL.ShapeSource
-                key={f.id}
+                key={typeof f.id === 'number' ? f.id?.toString() : f.id}
```
KristianSelnas, line 186: This could be simplified to `String(id)`. Not important at all, just a tip for later

src/components/map/types.ts
```diff
@@ -38,6 +38,8 @@ export type MapRegion = {
   center: Position;
 };
 
+export type MappPadding = number | number[];
```
KristianSelnas, line 41: `Mapp` -> `Map`

KristianSelnas, line 41: Why do you need this type? It seems a bit "dangerous" as the original type (MapboxGL.Padding) enforces the correct number of numbers in the array. This doesn't, and that could cause run-time errors that otherwise would have been discovered at compile time.

src/components/map/MapConfig.ts
```diff
@@ -1,8 +1,9 @@
-import {CameraStop, MapViewProps} from '@rnmapbox/maps';
+//import {CameraStop, MapViewProps} from '@rnmapbox/maps';
 import {Platform} from 'react-native';
 import {MAPBOX_STOP_PLACES_STYLE_URL} from '@env';
+import {CameraStop} from '@rnmapbox/maps/lib/typescript/components/Camera';
 
-export const MapViewConfig: MapViewProps = {
```
KristianSelnas, line 5: Why did you remove this?

src/components/map/MapConfig.ts
```diff
@@ -1,8 +1,9 @@
-import {CameraStop, MapViewProps} from '@rnmapbox/maps';
+//import {CameraStop, MapViewProps} from '@rnmapbox/maps';
```
KristianSelnas, line 1: If you have good reasons for removing the type for `MapViewConfig` you should remove his commented out import as well

src/components/map/types.ts
jorelosorio: Hi, that is how the parameter now is expected from the library before it was Padding but is removed:

image

I just created a MapPadding to do not repeat on all the places we use it. 

I will fix the typo ;) thanks

src/components/map/MapConfig.ts
jorelosorio: Thanks I will fix it good catch ;), I haven't properly double-checked it just a small draft to let @marius-at-atb test his changes.

src/components/map/types.ts
KristianSelnas: Ok. Maybe `MapPadding` should be more like the old `MapboxGL.Padding` then? I.e. `type MapPadding = number | [number, number] | [number, number, number, number]`? I assume that the padding must still consist of a 2 or 4 array?

jorelosorio: I made as you suggested ;)

PR #3610, opened 2023-05-31
fix: PostHog events for see live bus in map

KristianSelnas:


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -146,16 +148,27 @@ export const DepartureDetailsScreenComponent = ({
                         : DepartureDetailsTexts.map,
                     )}
                     interactiveColor="interactive_1"
-                    onPress={() =>
+                    onPress={() => {
+                      vehiclePosition &&
+                        analytics.logEvent(
+                          'Trip search',
```
rosvik, line 154: Not sure how this works, but where is this context/keyword defined? Sounds like we maybe should have one for departure details.

rosvik (approved):

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
KristianSelnas: I'm not sure what the difference between 'Trip details' and 'Departure details' is, but the contexts are defined in `src/analytics/analytics-event-context.ts`. I don't have an strong opinions on what these contexts should be

tormoseng: It would be nice to use `Departure details` when the user enters the live view from the departure details. If you want to view a departure you go to departure details. Then possibly to live busses. Trip details is from viewing a specific trip suggestion.

How to handle the definitions is up to you.

gorandalum:
> It would be nice to use Departure details when the user enters the live view from the departure details. If you want to view a departure you go to departure details. Then possibly to live busses. Trip details is from viewing a specific trip suggestion.

In the [screen wrapper](https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md#screen-reuse) `DepartureDetailsScreenComponent` we don't know in which context we are in when triggering the event. If we want correct context in the analytics event then we shouldn't log events in screen wrappers, but log them in the actual screen implementations, in this case in `Departures_DepartureDetailsScreen`, `Dashboard_DepartureDetailsScreen` and `Map_DepartureDetailsScreen`.

PR #3612, opened 2023-05-31
fix: add analytics to ticket assistant

adriawh:
https://github.com/AtB-AS/kundevendt/issues/3917#issuecomment-1568415456

> Inkludert Posthog analytics for se hvem som fullføre.

### DIFF
--- a/src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_SummaryScreen/TicketAssistant_SummaryScreen.tsx
+++ b/src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_SummaryScreen/TicketAssistant_SummaryScreen.tsx
@@ -14,6 +14,7 @@ import SvgInfo from '@atb/assets/svg/color/icons/status/Info';
 import {useAuthState} from '@atb/auth';
 import {Root_PurchaseConfirmationScreenParams} from '@atb/stacks-hierarchy/Root_PurchaseConfirmationScreen';
 import {useFocusOnLoad} from '@atb/utils/use-focus-on-load';
+import {useAnalytics} from '@atb/analytics';
 
 type SummaryProps = TicketAssistantScreenProps<'TicketAssistant_SummaryScreen'>;
 
@@ -21,6 +22,7 @@ export const TicketAssistant_SummaryScreen = ({navigation}: SummaryProps) => {
   const styles = useThemeStyles();
   const {t, language} = useTranslation();
   const {authenticationType} = useAuthState();
+  const analytics = useAnalytics();
   let {loading, inputParams, recommendedTicketSummary, error} =
     useTicketAssistantState();
   const focusRef = useFocusOnLoad();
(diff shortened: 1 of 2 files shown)


rosvik (approved):
LGTM 👍

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_SummaryScreen/TicketAssistant_SummaryScreen.tsx
```diff
@@ -177,6 +180,7 @@ export const TicketAssistant_SummaryScreen = ({navigation}: SummaryProps) => {
             mode="secondary"
             text={t(TicketAssistantTexts.closeButton)}
             onPress={() => {
+              analytics.logEvent('Ticketing', 'Ticket assistant closed');
```
KristianSelnas: If you create a new PostHog context for the ticketing assistant you should use that here as well.

src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
```diff
@@ -29,14 +30,23 @@ type Props = RootStackScreenProps<'Root_TicketAssistantStack'>;
...
     <TicketAssistantContextProvider>
       <FullScreenHeader
         leftButton={
           activeTab === 0
-            ? {type: 'close'}
+            ? {
+                type: 'close',
+                onPress: () => {
+                  analytics.logEvent(
+                    'Ticketing',
```
KristianSelnas, line 46: Create a 'Ticket assistant' context and use that here as well.

src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
```diff
@@ -29,14 +30,23 @@ type Props = RootStackScreenProps<'Root_TicketAssistantStack'>;
...
+                    'Ticket assistant closed from welcome screen',
+                  );
+                },
```
gorandalum: I would think this replaced the standard close action in the header, and would now only log the analytics event?

adriawh: Yes, saw at the same time as you. Should be fixed now

KristianSelnas (approved):

src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
```diff
@@ -29,14 +30,24 @@ type Props = RootStackScreenProps<'Root_TicketAssistantStack'>;
...
+            ? {
+                type: 'close',
+                onPress: () => {
+                  navigation.pop();
+                  analytics.logEvent(
+                    'Ticket assistant',
+                    'Closed from welcome screen',
+                  );
+                },
+              }
```
gorandalum, line 50: It is not optimal that we need to specify the navigation event when we wan't to just specify the log event. It is ment that we just say `type: 'close'`, and the navigation event is handled in the screen header.

I would rather that you may send an logging context in the header button props, something like this: 
```ts
{
  type: 'close',
  logEventContext: 'Ticket assistant'
}
```

And in the header button an event can be logged when clicked. Something like this:
```ts
if (logEventContext) {
  analytics.logEvent(
    logEventContext
    'Header button of type ${type} pressed',
  );
}
```

gorandalum (changes requested):

gorandalum (approved):

PR #3613, opened 2023-06-01
feat: new onboarding screens

mortennordseth:
FRAM would like to have some additional information in the app's onboarding to communicate the differences between the current app and the new one. I decided to use remote config to enable this. 

Before merging we need to release a new version of @atb-as/generate-assets and upgrade the dependency because of new illustrations (https://github.com/AtB-AS/design-system/pull/135). This will fix the tsc error.
<details>

image

image

</details>

Closes https://github.com/AtB-AS/kundevendt/issues/4066

### DIFF
--- a/src/translations/screens/Onboarding.ts
+++ b/src/translations/screens/Onboarding.ts
@@ -21,6 +21,22 @@ const OnboardingTexts = {
     ),
     mainButton: _('Neste', 'Next'),
   },
+  goodToKnow: {
+    title: _('Greit å vite', 'Good to know'),
+    description: _(
+      'Når du gjør et reisesøk får du opp forslag til hvordan du kan reise, og en korrespondanse som er garantert er merket med korrespondanse.',
+      'When you perform a travel search, you will receive suggestions on how to travel, and a guaranteed connection is marked with correspondance.',
+    ),
+    mainButton: _('Neste', 'Next'),
+  },
+  alsoGoodToKnow: {
+    title: _('Også greit å vite', 'Also good to know'),
+    description: _(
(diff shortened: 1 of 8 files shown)


src/translations/screens/Onboarding.ts
```diff
@@ -21,6 +21,22 @@ const OnboardingTexts = {
     ),
     mainButton: _('Neste', 'Next'),
   },
+  goodToKnow: {
+    title: _('Greit å vite', 'Good to know'),
+    description: _(
+      'Når du gjør et reisesøk får du opp forslag til hvordan du kan reise, og en korrespondanse som er garantert er merket med korrespondanse.',
```
adriansberg, line 27:
```suggestion
      'Når du gjør et reisesøk får du opp forslag til hvordan du kan reise, og en korrespondanse som garantert er merket med korrespondanse.',
```

adriansberg:
Looks good. I'll run the app when the newly added assets ase added.

adriansberg:
Different line thicknesses in the illustrations. Should they be the same or does it not matter?

image
vs
image

adriansberg (approved):
🚀

src/translations/screens/Onboarding.ts
mortennordseth: I think we should leave this as is or ask Christoffer for a new text if it is unclear. They want to state clearly that only guaranteed correspondences are marked with correspondence

adriansberg: Okay. Think I understood the sentence now after reading it six times. Suggested reword by AI: "Når du søker etter reisemuligheter, vil du få forslag til ulike reisealternativer, og korrespondansen som er bekreftet vil være tydelig merket."

mortennordseth:
> Different line thicknesses in the illustrations. Should they be the same or does it not matter?

That might be better, yes! I'll look into it 👍

mortennordseth:
A new illustration will be included in https://github.com/AtB-AS/kundevendt/issues/4141.

PR #3615, opened 2023-06-01
feat: add websocket reconnect

rosvik:


src/api/use-subscription.ts
```diff
@@ -25,40 +28,64 @@ export function useSubscription({
...
+
       webSocket.onclose = (event) => {
         setState(getSubscriptionStateCode(webSocket.readyState));
         Bugsnag.leaveBreadcrumb(`WebSocket closed with url: ${url}`);
+
+        // Reconnect immediately if close event is end of stream, otherwise use
+        // exponetial backoff to retry.
+        if (event.code === 1001 && reconnectCount < MAX_NUMBER_OF_RECONNECTS) {
+          setReconnectCount(reconnectCount + 1);
+        } else if (retryCount < MAX_NUMBER_OF_RETRIES) {
```
jorelosorio, line 62: It might be the intention but when the first `if` is evaluated the else is ignored, but it might be the case that both must be evaluated. Is that the case? Or is either reconnect Count and retry?

rosvik: That is as intended. We don't want to retry with exponential backoff if it's a regular reconnect (status code is 1001). So if `status code === 1001` then reconnect directly, otherwise run `exponentialBackoff`.

Maybe not that easy to read, so might rename the function to make it clearer 🤔

src/api/use-subscription.ts
```diff
@@ -25,40 +28,64 @@ export function useSubscription({
...
+
       webSocket.onmessage = (event) => {
         setState(getSubscriptionStateCode(webSocket.readyState));
         onMessage && onMessage(event);
       };
+
       webSocket.onerror = (event) => {
         setState(getSubscriptionStateCode(webSocket.readyState));
         Bugsnag.notify(`WebSocket error "${event.message}" with url: ${url}`);
         onError && onError(event);
```
jorelosorio, line 51: Do we need to try to reconnect here?

src/api/use-subscription.ts
jorelosorio: It might that is not clear the difference between reconnect and retry, but now I see why after explanation.

src/api/use-subscription.ts
rosvik: Looked into it now. It seem like when an `onerror` happens, an `onclose` event is also triggered, even if the connection hasn't been successfully opened in the first place. But an `onclose` event doesn't necessarily trigger `onerror`. If we add error handling here, we would need to add extra logic to make sure it isn't triggered twice, while the close event can handle both errors and "regular" close events quite easily.

So good question, but I think it's best to not add any reconnects here 😊

jorelosorio (approved):
Good work

src/api/use-subscription.ts
```diff
@@ -25,40 +28,66 @@ export function useSubscription({
...
     return () => {
+      if (retryTimeout) clearTimeout(retryTimeout);
       if (webSocket) {
         webSocket.onclose = null;
         onClose && onClose({code: 1000, reason: 'Cleanup'});
         webSocket.close();
       }
     };
-  }, [url]);
+  }, [url, reconnectCount, retryCount]);
```
gorandalum, line 90: Can't have an `useEffect` which has `reconnectCount` and `retryCount` in the dependency list, while using `setReconnectCount` and `setRetryCount` in the body. Need to refactor somehow. Maybe the counts can be refs instead of state if changing them isn't ment to cause a rerender.

rosvik: I realize there's a bit of implicit behavior here, that might be hard to spot when reading the code. 

`reconnectCount` and `retryCount` have two purposes:

1. They are used to keep track of how many times the app should try to retry/reconnect
2. They are used to trigger the retry/reconnnect itself, because updating the value also triggers the useEffect. Changing them to refs would defeat this purpose.

This seems to work very well in practice, but am I solving it in a completely roundabout-way? It's possible that I'm missing another good way to solve this, or should I update the name of the variables?

gorandalum: You can't have a state value in the dependency list of a `useEffect` which modifies that state value. Basic example

```ts
const [count, setCount] = useState(0);
// Infinite loop
useEffect(() => {
  setCount(count + 1)
}, [count]);
```

Even though some logic in this `useEffect` hinders the infinite lopp, the rule that setting state for a state variable that is in the dependency list is a clear no-no, and should be avoided.

Also note that having the `reconnectCount` and `retryCount` in the dependency list makes the cleanup function run every time one of those changes. So when everything works and `setRetryCount(0)` is run, then the cleanup triggers closing the connection and reconnecting again. (Not tested, just what I think will happen when reading the code)

gorandalum: You are right, I didn't find a well established source that setting state that is in the dependency list is strictly forbidden. As long as measures are done so it isn't causing a loop or unnecessary renders. I will soften my stance on it, but still be "on guard" when I see it happening since it can be the cause for some bugs :)

And good refactor! :+1:

PR #3619, opened 2023-06-05
fix: PostHog api key for dev

KristianSelnas:


gorandalum (commented):
Do we want to log events from dev environment to PostHog? For Bugsnag when running on dev we just log them in the console instead (logic in `bugsnagConfig.ts`) for shorter feedback loops when developing.

gorandalum (approved):

KristianSelnas:
> Do we want to log events from dev environment to PostHog? For Bugsnag when running on dev we just log them in the console instead (logic in `bugsnagConfig.ts`) for shorter feedback loops when developing.

It's to have a place to test PostHog logging, e.g. when you add a new event. It's also to avoid polluting the staging logs.

PR #3625, opened 2023-06-06
fix: Fix automatic local certificates and provisioning profiles for oms apps

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/3295


gorandalum:
To enlighten me, why wasn't this necessary before in local development for AtB?

jorelosorio:
> To enlighten me, why wasn't this necessary before in local development for AtB?

All the values for groups were prefilled for ATB only. This PR also provides an easy way to get certificates instead of doing it manually as before.

gorandalum (commented):
If `MATCH_PASSWORD` now is necessary for local development, should it be in `.env` files instead of GH Secrets?

jorelosorio:
> If `MATCH_PASSWORD` now is necessary for local development, should it be in `.env` files instead of GH Secrets?

We had that discussion yesterday with @toretefre, and we were thinking the same, but we decided to keep it secret, but we can change that, what do you think?

gorandalum:
I think it can be moved. The encrypted env files are there for exactly this, that local development can be done without passing around secrets (except git crypt key).

jorelosorio:
> I think it can be moved. The encrypted env files are there for exactly this, that local development can be done without passing around secrets (except git crypt key).

Done please @toretefre add yours in the dev envs :)

jorelosorio:
@gorandalum match is now on the .env files for each org

PR #3626, opened 2023-06-06
refactor: Login screens according to guidelines

gorandalum:
Moved the login screens under the stacks-hierarchy folder, and
renamed according to coding guidelines.

Also made a change to the main onboarding screens, so they are
loaded the same way as other onboarding screens.

### Acceptance criteria
- [ ] Login screens is triggered correctly and closed correctly from:
  - My profile
  - When buying period ticket while anonymous
  - When logging in during onboarding
  - When logging in during ticket assistant
- [ ] Onboarding is triggered correctly on new app install
- [ ] Mobiletokenonboarding is triggered correctly after login on new app install
- [ ] Logging in with Vipps works as before
- [ ] The warning "You have a ticket" is triggered when logging in if the anonymous account has one or more active tickets


gorandalum:
As always with these rafactorings it is quite a bit of changes, but there is mainly moving of code, so every codeline doesn't need to be reviewed.

hanne-at-atb (approved):
Well done ⭐

tormoseng:
@gorandalum 
- I can't seem to trigger the mobile token onboarding. Removed all AtB-apps and installed a new. No onboarding after log in. And nothing happens when I go into debug and set token onboarding to false either.
- Vipps login is currently not enabled. I think we can test this feature when we decide to release it.

gorandalum:
> I can't seem to trigger the mobile token onboarding. Removed all AtB-apps and installed a new. No onboarding after log in. And nothing happens when I go into debug and set token onboarding to false either. But if I close the app, and then re-open it, the token onboarding suddenly appears.

Fixed in https://github.com/AtB-AS/mittatb-app/pull/3650

> Vipps login is currently not enabled. I think we can test this feature when we decide to release it.

The reason I included Vipps in acceptance criteria is that it technically is "released" it is just not enabled. We just need to keep in mind that we want to do a round of testing before enabling it.

tormoseng:
Re-tested OK!

PR #3628, opened 2023-06-06
feat: override passenger category texts for special products

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/4174

Ensuring that Nordland can have separate age limits for Travel Pass Nordland that is going into sale with 1.37.

Code is checking if there's an override available for the given product type (only entered for travel-pass type), and returning descriptions as before if not.

### DIFF
--- a/src/translations/screens/subscreens/TicketTraveller.ts
+++ b/src/translations/screens/subscreens/TicketTraveller.ts
@@ -100,7 +100,7 @@ function specificUserProfileDescription(
   travellerType: string,
   ticketType: string | undefined,
 ) {
-  if (ticketType === undefined) return false;
+  if (ticketType === undefined) return _('', '');
   switch (ticketType) {
     case 'travel-pass':
       if (travellerType === TravellerType.adult) {
--- a/src/translations/screens/subscreens/TicketTraveller.ts
+++ b/src/translations/screens/subscreens/TicketTraveller.ts
@@ -1,9 +1,7 @@
-import {Language, translation as _} from '../../commons';
+import {translation as _} from '../../commons';
 import {APP_ORG} from '@env';
 import {orgSpecificTranslations} from '@atb/translations/orgSpecificTranslations';
 import {UserProfileWithCount} from '@atb/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Travellers/use-user-count-state';
-import {TFunc} from '@leile/lobo-t';
-import {getTextForLanguage} from '@atb/translations/utils';
 
 enum TravellerType {
   adult = 'ADULT',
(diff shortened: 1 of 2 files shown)


src/translations/screens/subscreens/TicketTraveller.ts
```diff
@@ -95,12 +96,40 @@ function generic(travellerType: string) {
...
+  }
+}
+
 const TicketTravellerTexts = {
   information: (travellerType: string, ticketType: string | undefined) => {
     return (
       specificOverrides(travellerType, ticketType) || generic(travellerType)
     );
   },
+  userProfileDescription: (
```
mortennordseth, line 123: Maybe we should rename this to `productTypeSpecificUserProfileDescription` or something similar to be clear that this is an override of user profile descriptions for some products? 🤔

src/translations/screens/subscreens/TicketTraveller.ts
```diff
@@ -95,12 +96,40 @@ function generic(travellerType: string) {
...
+    userProfile: UserProfileWithCount,
+    ticketType: string | undefined,
+  ) => {
+    const specificDescription = specificUserProfileDescription(
+      userProfile.userTypeString,
+      ticketType,
+    );
+    return specificDescription ? specificDescription : _('', '');
```
mortennordseth, line 131:
```suggestion
    return specificUserProfileDescription(
      userProfile.userTypeString,
      ticketType,
    );
```

src/translations/screens/subscreens/TicketTraveller.ts
```diff
@@ -95,12 +96,40 @@ function generic(travellerType: string) {
...
+  switch (ticketType) {
+    case 'travel-pass':
+      if (travellerType === TravellerType.adult) {
+        return _('Over 18 år', 'Age 18 or older');
+      } else if (travellerType === TravellerType.child) {
+        return _('Til og med 17 år', 'Age 17 or younger');
+      }
+      return false;
+    default:
+      return false;
```
mortennordseth, line 113:
```suggestion
      return _('', '');
    default:
      return _('', '');
```

src/translations/screens/subscreens/TicketTraveller.ts
toretefre: Better like this? 

https://github.com/AtB-AS/mittatb-app/pull/3628/commits/72d8c8dcea876e6e821dd74e0d265cf8f23ebb68

mortennordseth: Yes! Nice 👍

toretefre:
NFK will release a new release candidate when this is approved.

gorandalum (approved):
This is ok as a temporary solution if you need it fast. But hardcoded checks on the fare product type which isa  completely generic string is not the way to go forward, so we should aim to replace this pretty soon.

toretefre:
> This is ok as a temporary solution if you need it fast. But hardcoded checks on the fare product type which isa completely generic string is not the way to go forward, so we should aim to replace this pretty soon.

Yes, absolutely agree. On a longer term it should also be possible to override dynamically from Firestore. Do you think this is related to the rule engine, or is it more of a separate issue?

gorandalum:
It is at least not in the first iteration of the rule engine issue.

On a longer term we should look into how we can avoid these overrides in the app. I think the products must be aligned with what is supported by the Entur domain model. How will this separate age limits for travel pass work if inspecting with the control app?

PR #3629, opened 2023-06-06
feat: StopPlacesScreen use parallax scroll

gorandalum:
Instead of using the old 'SimpleDisappearingHeader' the stop places
screen(s) now uses the new 'FullScreenView' which has a parallax
scroll effect.

### Acceptance criteria
- [ ] Test the new scroll effect in these places:
  - Departures
  - Dashboard -> Add favorite
  - Profile -> Add favorite
 - [ ] Regression-test the places which already had parallax scroll effect:
   - TripSearch
   - TripDetails
   - DepartureDetails


src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
```diff
@@ -150,10 +151,12 @@ export const NearbyStopPlacesScreenComponent = ({
 
   return (
     <>
```
rosvik, line 153: This fragment isn't needed anymore. Could you remove it while you're at it?

rosvik (approved):
Looks good! Just departures v1 left before we can remove SimpleDisappearingHeader?

tormoseng:
Tested OK

PR #3635, opened 2023-06-08
fix: static QR-code readability for FRAM buses

mortennordseth:
I've made the static QR-code a bit smaller to make it easier to scan on FRAM buses.

Fixes https://github.com/AtB-AS/kundevendt/issues/4006

<details>

# Today

IMG_2006

# Smaller version

IMG_2005

</details>

### DIFF
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -222,7 +222,7 @@ const StaticQrCode = ({fc}: {fc: FareContract}) => {
   return (
     <GenericSectionItem>
       <View
-        style={styles.aztecCode}
+        style={[styles.aztecCode, styles.staticQrCode]}
         accessible={true}
         accessibilityLabel={t(FareContractTexts.details.barcodeA11yLabel)}
         testID="staticQRCode"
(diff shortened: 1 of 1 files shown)


src/fare-contracts/details/Barcode.tsx
```diff
@@ -240,4 +240,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     padding: theme.spacings.large,
     backgroundColor: '#FFFFFF',
   },
+  staticQrCode: {
+    aspectRatio: 1.25,
```
adriansberg, line 244: What's the benefit of increasing the aspect ratio vs reducing the width/height?

mortennordseth: Might be the wrong usage of aspect ratio here, but it centers things nicely. Might be better to use this?
```css
width: '75%',
marginLeft: 'auto',
marginRight: 'auto',
```

adriansberg: I see. The QR code will remain square because of the sizing set on L230. The aspect-ratio increase increases the height of the container it is in, thus making it smaller in width. Is the height affected in terms of scrolling?

Reducing the width and centering should also work. Margin can be reduced to one line if wanted I guess: `margin: '0 auto'`.

gorandalum (approved):

src/fare-contracts/details/Barcode.tsx
adriansberg: 🤪

PR #3636, opened 2023-06-09
feat: Shrink header on scroll and update title

ingvisim:
Show alternate title when scrolling past parallax content.
- Only show "reisedetaljer" and not line names, as they take up a lot of space and make the scrolling janky
- Start fade in at 50 percent of parallax content

<details>
<summary>Screenrecording</summary>

https://github.com/AtB-AS/mittatb-app/assets/85479566/d1f3bf87-932a-4f08-8784-b613816bf11e

https://github.com/AtB-AS/mittatb-app/assets/85479566/d56cbaac-b888-49be-aa97-de00ccc810f1

</details>

Close https://github.com/AtB-AS/kundevendt/issues/3675


rosvik (commented):
Great work 💯 Had one suggestion for refactoring

gorandalum (commented):
Really good work! Handling these native scroll events, on layout events, etc., is quite cumbersome, and leads to somewhat complicated code. But when handling animations it is necessary.

I think all the code here is necessary and well written, if I can suggest some improvement then it would be to try to encapsulate it more. Now the complexity is added both in `ParallaxScroll` and in `FullScreenView`, and especially `FullScreenView` has gotten more complex since it now needs to deal with scroll events.

Is it possible to encapsulate this more in the `ParallaxScroll` component? I am thinking it knows its own vertical offset (in the `scrollYRef` value) and it should also be able to know it's own height. Maybe the `ParallaxScroll` can do the calculations and just expose a value of where it is in the scroll (`scrollPercentage` or something?) up to `FullScreenView` instead?

I haven't tried this myself, so there might by some details that make it not possible. We can discuss it on monday.

rosvik (approved):

PR #3638, opened 2023-06-09
refactor: refactor useSubscription hook and remove retry limits

rosvik:
- Refactors the `useSubscription` hook, fixing the comment from https://github.com/AtB-AS/mittatb-app/pull/3615#discussion_r1221087098
- Removes limits to how many times the app will try to reconnect. Since there isn't a limit to this other places in the app, it was considered unnecessary here. 
- Exponential fallback on errors are capped at 10s, which means retries will happen with 1,2,3,4,8,10,10,10,10... seconds delay.
- Renames `SubscriptionState` to `SubscriptionStatus` to prevent confusion with states containing actual API data.

### Acceptance criteria

- [ ] On errors, retries will happen with 1,2,3,4,8,10,10,10,10... seconds delay. (Instead of 1,2,4,8,16,32 as before). Should also work after more than 1 minute disconnected, where the reconnect should happen within 10s.
- [ ] Reconnects (after timeouts that happen every 30s) should work "forever", longer than the 10 * 30s that were before.
- [ ] Otherwise bus in map should work as before.


src/api/use-subscription.ts
```diff
@@ -87,19 +81,40 @@ export function useSubscription({
         webSocket.close();
       }
     };
```
gorandalum, line 83: `webSocket` state variable is used in the clean up function, but it isn't in the dependency list, as it should be when [it is a reactive value](https://react.dev/learn/removing-effect-dependencies#to-remove-a-dependency-prove-that-its-not-a-dependency). However adding it to the dependency list would give a loop. 

One way to solve this is by also have the `webSocket` as a ref instead of a state value. Another option is to utilise the previous state in the `setWebSocket` function to retrieve the `webSocket` value:
```ts
// Cleanup
return () => {
  if (retryTimeout) clearTimeout(retryTimeout);
  setWebSocket(ws => {
    if (ws) {
      ws.onclose = null;
      onClose && onClose({code: 1000, reason: 'Cleanup'});
      ws.close();
    }
    return undefined;
  }
};
```

I am really focusing on the usage of `useEffect`s in PRs lately as I think missing dependencies and dependency loops have been, and still are, the cause of some hard-to-debug errors in the app. I have a goal of enabling the exhaustive deps linter rule to catch these situations.

gorandalum (changes requested):
I think it is much better! Just one more thing to fix with the `useEffect` dependency.

src/api/use-subscription.ts
rosvik: Aha, updated now. Sounds like a good idea to enable the rule 👍

tormoseng:
LGTM!

PR #3639, opened 2023-06-09
fix: Set minimum zoom level when clicking a scooter cluster

KristianSelnas:


src/components/map/components/mobility/Vehicles.tsx
```diff
@@ -33,10 +33,13 @@ export const Vehicles = ({mapCameraRef, vehicles, onClusterClick}: Props) => {
         maxZoomLevel={22}
         clusterMaxZoomLevel={21}
         onPress={async (e) => {
-          const [feature, ..._] = e.features;
+          const [feature, ,] = e.features;
```
gorandalum, line 36: Can have just `const [feature] = e.features`?

gorandalum (approved):

src/components/map/components/mobility/Vehicles.tsx
KristianSelnas: Maybe later, aligator

PR #3641, opened 2023-06-12
feat: disable retries on 429 responses

rosvik:
Doesn't look like this will actually change anything in practice, since the `config.retry` field defaults to undefined / falsy, and I don't think it is set to true anywhere at the moment. But disabling retries entirely on 429 (Too Many Requests) is nice to have for the future.

closes https://github.com/AtB-AS/kundevendt/issues/3978

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -33,14 +33,17 @@ declare module 'axios' {
 }
 
 function shouldRetry(error: AxiosError): boolean {
+  const isRateLimited = error.response?.status === 429;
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Good work :+1: And I think the retry logic is actually working, and if I remember correctly the config.retry value is set by the axios-retry librabry.

rosvik:
Aha, default is 3: https://github.com/softonic/axios-retry#options

PR #3645, opened 2023-06-14
fix: Live bus map error

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3688

Solves iOS stage build error by using onRegionIsChanging instead of onCameraChanged.
However still using onCameraChanged for Android, where onRegionIsChanging produces laggy behaviour.

Also fixes a minor issue where the live bus icon would be hidden until the first location change.

### DIFF
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -49,7 +49,7 @@ type Props = TravelDetailsMapScreenParams & {
 };
 
 const FOLLOW_ZOOM_LEVEL = 14.5;
-const FOLLOW_MIN_ZOOM_LEVEL = 8;
+const FOLLOW_MIN_ZOOM_LEVEL = 6; // keep below 6.32 to avoid mapbox bug
 const FOLLOW_ANIMATION_DURATION = 500;
 
 export const TravelDetailsMapScreenComponent = ({
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -128,16 +128,16 @@ export const TravelDetailsMapScreenComponent = ({
           }
         }} */
       >
-        {/*<MapboxGL.Camera
+        <MapboxGL.Camera
           ref={mapCameraRef}
           bounds={bounds}
           {...MapCameraConfig}
           zoomLevel={vehicleWithPosition ? FOLLOW_ZOOM_LEVEL : undefined}
           centerCoordinate={vehicleWithPosition ? centerPosition : undefined}
           animationDuration={0}
-        />*/}
-        {/*<MapboxGL.UserLocation showsUserHeadingIndicator />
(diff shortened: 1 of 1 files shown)


src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -48,7 +49,7 @@ type Props = TravelDetailsMapScreenParams & {
 };
 
 const FOLLOW_ZOOM_LEVEL = 14.5;
-const FOLLOW_MIN_ZOOM_LEVEL = 8;
+const FOLLOW_MIN_ZOOM_LEVEL = 6; // keep below 6.32 to avoid mapbox bug
```
jorelosorio, line 52: As we discussed it seems more like the values for Zoom are not so accurate when listening for camera changes, but for other places like by example `onMapIdle` are correct. There is a strange behaviour that shows 6.32 as map zoom as the first value (I think that must be the default value for Mapbox) but then it changes in a later state when a location is updated, so there is something else going on here. But yes, this might be the shortest solution for now, and take a look in further updates on MapBox.

jorelosorio (approved):
Looks good, good work!

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
gorandalum: I think we should really try to keep the state correct. With the solution now we "accept" that the faulty value 6.32 comes in the state, but it is circumvented by changing this `FOLLOW_MIN_ZOOM_LEVEL`. Decided together with Marius that we could look at if the value from `onMapIdle` is more correct.

jorelosorio: I also agree that `onMapIdle` is the better solution, separating the heading and zoom. Yesterday we tried this, but then @marius-at-atb had a talk with the designers and QA. So I also opt for this option!

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -90,14 +91,38 @@ export const TravelDetailsMapScreenComponent = ({
   });
 
   const [shouldTrack, setShouldTrack] = useState<boolean>(true);
+  const [zoomLevel, setZoomLevel] = useState<number>(FOLLOW_ZOOM_LEVEL);
   const [cameraState, setCameraState] = useState<{
```
jorelosorio, line 95: This could be renamed as [cameraHeading, setCameraHeading] for a proper name

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
marius-at-atb: Now using onMapIdle with the latest commit 👍

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
marius-at-atb: Good point, updated now

gorandalum (approved):

PR #3650, opened 2023-06-15
fix: Fix after login onboarding race condition

gorandalum:
After login in there was a race condition between the navigation
done by the login screen and the navigation to the mobile token
onboarding. This is fixed by only redirecting to the mobile token
onboarding when the TabNavigator is actually focused.


gorandalum:
Fix for https://github.com/AtB-AS/mittatb-app/pull/3626#issuecomment-1589303160

rosvik (approved):
Good catch! Looks good 👍

PR #3651, opened 2023-06-15
fix: mapbox env variables for FRAM

mortennordseth:
Updated Mapbox env variables for production as they were outdated (key and stop place style URL)


adriansberg (approved):

mikaelbr (approved):
Sure!

PR #3656, opened 2023-06-16
build: Fixed build id when registering Android app version

gorandalum:
When registering Android app versions the newest build overwrites
the old one. Meaning only one build, the newest, will be able to
create and renew mobile tokens.

In staging this is a problem since we have such frequent builds,
and it will be a chore to always ensure to be on the newest build
when testing something related to mobiletoken or ticketing in
general.

We asked Entur, and they said in staging it was ok to do the same
as is done in dev, that every Android build has the same value for
the build id. This means that all staging builds on the same app
version should keep working after new builds.


tormoseng (approved):
LGTM!

PR #3657, opened 2023-06-16
Move passenger category selection to bottom sheet

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3943

Created PassengerSelectionSheet to contain the options. Removed InfoToggle.
Also added Edit icon on the right side of "Select travellers" and "Select stop/zone(s)" fields.

@HanneLH @tormoseng Of course very important to test that all ticket purchases still work as expected.

Before:

<br /><br />
After:

    

<br /><br />

Technically the exact same component is used for single traveller selection! @ckbilstad @oyvindmikkelsenatb is this what you want?

Before:

After:


marius-at-atb:
Nice! Skal fikse "voksen" vs "1 voksen"

tormoseng:
@marius-at-atb Wait to merge until release branch for 1.38 is created

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,141 @@ export function TravellerSelection({
...
-        (u) => selectableUserProfiles.find((i) => i.id === u.id),
-      );
-      setTravellerSelection(filteredSelection);
-    }
-  }, [userCountState.userProfilesWithCount, fareProductType, selectionMode]);
+  const accessibility: AccessibilityProps = {
+    accessible: true,
+    accessibilityRole: 'button',
+    accessibilityLabel:
+      'Todo: Some label in the correct language here' + screenReaderPause,
```
jorelosorio, line 44: This todo must be fixed

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,141 @@ export function TravellerSelection({
...
+    setTravellerSelection(filteredSelection);
+  }, [fareProductType, selectionMode, selectableUserProfilesWithCount]);
 
   if (selectionMode === 'none') {
     return <></>;
   }
 
+  let numberOfPassengerCategoriesSelectedFrom = 0;
+  const totalPassengersCount = selectableUserProfilesWithCount.reduce(
+    (acc, sUPWC) => {
```
jorelosorio, line 64: I could suggest better names for those variables.

jorelosorio (dismissed):

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/PassengerSelectionSheet.tsx
```diff
@@ -0,0 +1,82 @@
...
+import {Confirm} from '@atb/assets/svg/mono-icons/actions';
+import {MultipleTravellersSelection} from './Travellers/MultipleTravellersSelection';
+import {SingleTravellerSelection} from './Travellers/SingleTravellerSelection';
+import {
+  UserProfileWithCount,
+  useUserCountState,
+} from './Travellers/use-user-count-state';
+
+type PassengerSelectionSheetProps = {
+  selectionMode: 'multiple' | 'single';
```
gorandalum: Can this type be `TravellerSelectionMode`?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: Renamed to activePassengerCategoryCount

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: Good catch! Done in the new commit.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,141 @@ export function TravellerSelection({
...
+      if (sUPWC.count > 0) {
+        numberOfPassengerCategoriesSelectedFrom += 1;
```
gorandalum: I don't like side effects in `reduce` or other functional concepts, since they easily go under the radar.

I would rather like something like this:
```ts
const numberOfPassengerCategoriesSelectedFrom = 
  selectableUserProfilesWithCount.filter(u => u.count).length
  ```

File: src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx [hunk 5]
```diff
@@ -27,65 +33,141 @@ export function TravellerSelection({
...
-    userCountState.updateSelectable(selectableUserProfiles);
-  }, [selectableUserProfiles]);
+    const filteredSelection = selectableUserProfilesWithCount.filter((u) =>
+      selectableUserProfiles.find((i) => i.id === u.id),
+    );
+    setTravellerSelection(filteredSelection);
+  }, [fareProductType, selectionMode, selectableUserProfilesWithCount]);
 
   if (selectionMode === 'none') {
     return <></>;
```
gorandalum, line 63: Not edited by you, but should be `return null`.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,145 @@ export function TravellerSelection({
...
+            if (acc !== '') {
+              acc += ', ';
+            }
+            return (
+              acc + sUPWC.count + ' ' + getReferenceDataName(sUPWC, language) // plural name values not ideal atm?
+            );
+          } else {
+            return acc;
+          }
+        }, '');
```
gorandalum: I think this can be refactored to be more readable. Maybe first get an array of selected user profiles with a `filter` function. 

Then it might be something like (handwritten/pseudocode)
```ts
selectedUserProfiles
  .map(up => `${up.count} ${getReferenceDataName(up, language)}`)
  .join(', ')
```

This `selectedUserProfiles` may also be reused to get the `activePassengerCategoryCount`.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,145 @@ export function TravellerSelection({
...
-          fareProductType={fareProductType}
-          {...userCountState}
-          userProfilesWithCount={selectableUserProfilesWithCount}
-        />
-      )}
+            : t(PurchaseOverviewTexts.travellerSelection.title_single)}
+        </ThemeText>
+      </View>
+      <Section {...accessibility}>
+        <GenericClickableSectionItem onPress={travellerSelectionOnPress}>
```
gorandalum, line 129: Some overflow when large text. Usually fixed by putting `flex: 1` somewhere.

gorandalum (changes requested):
Good work! Some minor comments :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/PassengerSelectionSheet.tsx
marius-at-atb: Yes

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: Fair point!

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: Indeed 👍

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: {flex: 1} did the trick

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -27,65 +33,138 @@ export function TravellerSelection({
...
+      : selectableUserProfilesWithCount
+          .filter((userProfile) => userProfile.count > 0)
+          .map(
+            (userProfile) =>
+              `${userProfile.count} ${getReferenceDataName(
+                userProfile,
+                language,
+              )}`,
+          )
+          .join(', ');
```
gorandalum, line 90: This is a comment not ment as critique, but more of a discussion point. You are free to not agree as everybody got their style, but I like these kinda comments/discussions and feel they are healthy for a team to have :)

I am a fan of verbose variable names to ensure greater readability :+1: However, it is also important to not add too many verbose variables names, as it ends up kinda cluttering to read. For example the variable `activeTravellerCategoryCount` I think isn't worth the cognitive load it demands for the user to read this code. Also there are some duplicate filtering on user count. 

By removing this the code can be reduced to something like this:
```ts
const selectedUserProfiles = selectableUserProfilesWithCount.filter((u) => u.count);
const totalTravellersCount = selectedUserProfiles.reduce((acc, {count}) => acc + count, 0);
const multipleTravellerCategoriesSelectedFrom = selectedUserProfiles.length > 1;
const multipleTravellersDetailsText = selectedUserProfiles
  .map((u) => `${u.count} ${getReferenceDataName(u, language)}`
  .join(', ') || t(PurchaseOverviewTexts.travellerSelection.no_travellers_selected);
```

Also when looking at it, is the variable `multipleTravellerCategoriesSelectedFrom` easier to understand compared to just using `selectedUserProfiles.length > 1` inline in the jsx? I'm not sure, just food for thought.

gorandalum (commented):
The code looks good! I had one comment on something, but it was more of a philosophical note 😅 

But I have a question, not directly regarding the code, but more of how it should work. In an earlier iteration of the app we had some similar logic like this, just that by opening the traveller selection the user was directed to a new screen instead of a bottom sheet. In that iteration the user couldn't save if he had selected no travellers. 

Is it by design that the user now can save his traveller selection even when no travellers are selected? If it isn't thought through then maybe you can discuss it with design? For my part I think it would be nice if it was ensured that a "valid" traveller selection was done in the bottom sheet, as that means easier code in the purchase overview screen.

marius-at-atb:
@gorandalum very good points. Following discussion with @ckbilstad, the "Confirm choice" button is now disabled in the bottom sheet if 0 travellers are selected, similar to how it was in the previous iteration.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
marius-at-atb: Nice touch with only one reduce iteration and destructuring :+1:

Regarding multipleTravellerCategoriesSelectedFrom I think it is good practice to keep it in its own variable since it is later used twice in a way where the conditions should stay in sync.

gorandalum (approved):

PR #3660, opened 2023-06-19
fix: Map filter: Show 'All' toggle when there are zero or more than one operator

KristianSelnas:
Fixes https://github.com/AtB-AS/kundevendt/issues/4029


gorandalum (approved):
Looks good. Remember to try to somehow connect these PRs to issues on the board. Either by adding issue dependency, or just linking to an issue comment or something.

PR #3667, opened 2023-06-21
fix: mobile token on android devices for FRAM

mortennordseth:
I noticed that we were using the upload key and not the signing key in prod for `ANDROID_SIGNING_CERTIFICATE_FINGERPRINT`. This should fix the issue with mobile token on Android devices for FRAM.


toretefre (approved):
:clap:

PR #3669, opened 2023-06-21
fix: inspection symbol for zone names with length greater than 3 characters

mortennordseth:
_Edited_
I've removed the zone names from the inspection symbol if the zone names have more than 3 characters. This is to ensure that we don't experience what is seen in the image below, and still keep it the as today for AtB. The designers will look into how we can make it consistent for all organizations. 

image
<details>
<summary>Original description</summary>
I've removed the zone names from the inspection symbol according to the new Figma sketch (See the issue for reasoning (https://github.com/AtB-AS/kundevendt/issues/4160)): 

image 

</details>
Fixes https://github.com/AtB-AS/kundevendt/issues/4160

### DIFF
--- a/src/fare-contracts/components/InspectionSymbol.tsx
+++ b/src/fare-contracts/components/InspectionSymbol.tsx
@@ -78,6 +78,17 @@ const InspectableContent = ({
   const InspectionSvg =
     fareProductTypeConfig?.illustration === 'night' ? Moon : Bus;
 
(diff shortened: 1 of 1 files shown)


mortennordseth:
I'll change this PR to draft because we have received feedback that validating tickets visually is hard and this would make it even harder. Awaiting new sketches.

toretefre (approved):

PR #3670, opened 2023-06-21
refactor: Restrict imports from stacks-hierarchy

gorandalum:
### Background
Components and modules outside of stacks-hierarchy shouldn't import
things from stacks-hierarchy. It should be the other way around,
that screens in stacks-hierachy imports components and modules from
the outside.

Also fixed the places where this rule broken.

One found exception to the rule is importing screen params, which
are necessary some places to enforce correct params on navigational
onPress handling. In these few cases the eslint rule may just be
disabled for the line.

### Acceptance criteria
Everything should work as before. No specific testing for this issue.


KristianSelnas (approved):
LGTM 👍

rosvik (approved):
Nice ✨

Did you look into restricting it for everything except imports of `*ScreenParams`? Not sure if that's easy to do, but that would avoid having to disable the rules in some cases.

gorandalum:
> Nice ✨
> 
> Did you look into restricting it for everything except imports of `*ScreenParams`? Not sure if that's easy to do, but that would avoid having to disable the rules in some cases.

Yes, I did try to make that work. I hoped `importNames` could be usable (something like `importNames: "!*ScreenParams"`), but I couldn't get it to work. Since it only was three places in the app I didn't invest any more time.

PR #3671, opened 2023-06-22
fix: ufo for bus in map

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/4238

Sometimes useGetServiceJourneyVehicles would return data before useDepartureData, and when tapping "See live" in between, this bug happened. Fixed by hiding the "See live" button until both have returned data.


ios/Podfile.lock
```diff
@@ -1693,7 +1693,7 @@ SPEC CHECKSUMS:
   RNSVG: d787d64ca06b9158e763ad2638a8c4edce00782a
   SocketRocket: fccef3f9c5cedea1353a9ef6ada904fde10d6608
   SwiftProtobuf: b02b5075dcf60c9f5f403000b3b0c202a11b6ae1
-  TimeReactNativeLib: 8c4e707e3fd89672913bc9265fbfd1235262ae58
+  TimeReactNativeLib: 88639629a9fa0849c13e465340b0de43a9ef6ef8
```
gorandalum, line 1696: Did you get this when doing pod install on master now? A little strange, as I don't get this change locally.

gorandalum (approved):
The main change looks good, but do a little test on whether the change in Podfile is needed.

ios/Podfile.lock
marius-at-atb: This was 2 days ago on a brand new branch from master (just before the "Update Podfile.lock" commit)

marius-at-atb: The exact same change happens on the current master with me after pod install

jorelosorio: The cause is because there was an update on some of the libraries in the package.json without update the Pod, so might not be part of this task but is ok to include the updates.

jorelosorio (approved):

ios/Podfile.lock
gorandalum: Seems like even though we use 1.1.14 of the mobiletoken sdk, the transitive dependency TimeReactNativeLib is however tagged to 1.1.14-SNAPSHOT. I'll talk with Entur about it.

PR #3673, opened 2023-06-22
feat: Outdated travel search, regarding bus in map

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/4214

Depend on: https://github.com/AtB-AS/atb-bff/pull/275

The proposed solution looks for `actualArrivalTime` in order to know if the travel has finished or not.


ingvisim (approved):

gorandalum (commented):
Will test this a little and see how it works.

gorandalum (changes requested):
Tried some testing locally, and it doesn't seem that this change has any effect. I think we need to check if the departure is finished or not somehow. Two main alternatives:
- Check the `actualArrivalTime` of the last stop and if the value is set, then not show the live button.
- Check the `lastUpdated` field on the vehicle position response (at the response time), and if that is greater than the stale threshold (`live_vehicle_stale_threshold` in RemoteConfig) then not show the live button.

Do you have some input on what you think would be better @rosvik @ingvisim ? Or maybe some other alternatives.

jorelosorio:
> Tried some testing locally, and it doesn't seem that this change has any effect. I think we need to check if the departure is finished or not somehow. Two main alternatives:
> 
> * Check the `actualArrivalTime` of the last stop and if the value is set, then not show the live button.
> * Check the `lastUpdated` field on the vehicle position response (at the response time), and if that is greater than the stale threshold (`live_vehicle_stale_threshold` in RemoteConfig) then not show the live button.
> 
> Do you have some input on what you think would be better @rosvik @ingvisim ? Or maybe some other alternatives.

When I moved to the state to bus details view I was more thinking on the position wasn't changing correctly for the view, but according to our tests we saw there are an approx 5 m time to refresh for the final end but still we go the same data, so the code I implemented do not make any changes because is the data that is not accurate (real-time at that point) as we want (some business logic might be related to work like that).

My suggestion is: Look for bus position is already in the position of the last quay (With a simple geo query) to know if reached the latest place without look for times, etc. But seems like `actualArrivalTime` does the same according to @gorandalum.

gorandalum (commented):
Great! Now that the field `vehicleStatus` is added in prod, I think it is appropriate to add it to this PR.

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -99,6 +100,10 @@ export const DepartureDetailsScreenComponent = ({
 
   const realtimeText = useRealtimeText(estimatedCallsWithMetadata);
 
+  const isJourneyFinished =
+    vehiclePosition?.vehicleStatus === VehicleStatusEnumeration.Completed ||
+    estimatedCallsWithMetadata.every((e) => e.actualArrivalTime);
```
rosvik, line 105: Do we need to keep the check on actualArrivalTime? I would hope vehicle status is reliable enough. Also it would help a bit on performance 🏃💨

rosvik (commented):
Nice! 

Maybe out of scope for this task, but could you look at handling when vehicleStatus becomes "completed" when viewing it in the map? Right now it just shows the spinner once it reaches the journey end.

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
jorelosorio: > Do we need to keep the check on actualArrivalTime? I would hope vehicle status is reliable enough. Also it would help a bit on performance 🏃💨

According to discussions with Entur, this `actualArrivalTime` is a fallback for the status, they are still working on an issue where the last quay is not updated with the final location.

rosvik: Aha, then I understand 👍

gorandalum (approved):

jorelosorio:
> Nice!
> 
> Maybe out of scope for this task, but could you look at handling when vehicleStatus becomes "completed" when viewing it in the map? Right now it just shows the spinner once it reaches the journey end.

Good suggestion @rosvik and I have a created a new task for it: https://github.com/AtB-AS/kundevendt/issues/4497

PR #3675, opened 2023-06-22
chore: add WS_API_BASE_URL env variable for FRAM prod

mortennordseth:
Add `WS_API_BASE_URL` for FRAM prod, as done in https://github.com/AtB-AS/mittatb-app/pull/3532.


toretefre (approved):
Genius

PR #3678, opened 2023-06-23
feat: Ticket bundling for mobility

KristianSelnas:
This is a 'quick fix' temporary solution. A more permanent solution must be made when it's more clear how this could be integrated more seamlessly with current ticket products.


gorandalum:
Looked through it quick and didn't find anything too crazy :+1: Some things to note, but I have a feeling it will change quite a bit when starting to use real data.

PR #3679, opened 2023-06-23
fix: disconnect websockets when out of focus

rosvik:


src/utils/use-is-focused-and-active.ts
```diff
@@ -0,0 +1,8 @@
+import {useIsFocused} from '@react-navigation/native';
+import {useAppStateStatus} from './use-app-state-status';
+
+export const useIsFocusedAndActive = (): boolean => {
+  const isFocused = useIsFocused();
+  const appStateStatus = useAppStateStatus();
+  return isFocused && appStateStatus === 'active';
```
KristianSelnas, line 7: Have you tested this on Android? I used the `useAppStateStatus` hook to send events to PostHog and experienced that Android sometimes reported these events a bit strangely.

KristianSelnas (approved):
Nice work! LGTM 👍

PR #3680, opened 2023-06-26
fix: improve contrast on login activity indicator

rosvik:
Noticed that the loading indicator spinner was black on the dark background color on the login screens. Updated to use the text color based on the pages `themeColor` setting.

Clarified with @oyvindmikkelsenatb 

## Before/after

Screenshot 2023-06-26 at 11 17 03
Screenshot 2023-06-26 at 11 18 17

## Acceptance criteria

- [ ] Loading indicator when submitting phone number is white
- [ ] Loading indicator when submitting confirmation code is white


jorelosorio (approved):
Looks good, I saw that the criteria applies only for that two screens, but I was wonder what about other places 🤔

HanneLH:
Verifisert ok for Android og iOS ✅

PR #3683, opened 2023-06-26
feat: product testing for debug users in production

mortennordseth:
I've added two new distribution channels: `debug-app` and `debug-web` which can be used to test products in production for users with debug enabled. 

The proposed solution has been discussed in this thread from [Slack](https://mittatb.slack.com/archives/C025NUF4WA0/p1682064525997789). From what I can see, there are no checks on the distribution channel that would create Bugsnag errors for previous app versions. 

Closes https://github.com/AtB-AS/kundevendt/issues/3897


gorandalum (approved):
Looks good! Remember to write acceptance criteria in the linked issue when merging, for the testers.

PR #3684, opened 2023-06-26
fix: refactor traveller selection state calculations

marius-at-atb:
fixes https://github.com/AtB-AS/kundevendt/issues/3943

Changes to the list of traveller types are now reflected correctly as you change duration for the period ticket.
Also handles cases with different kinds of default traveller options, and always ensures valid selection state.


src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-defaults.ts
```diff
@@ -90,20 +90,26 @@ const useTravellersWithPreselectedCounts = (
...
+      }));
+
+    if (
+      !mappedUserProfiles.some(({count}) => count) &&
+      mappedUserProfiles.length > 0 // how to handle if length 0?
+    ) {
+      mappedUserProfiles[0].count = 1;
+    }
+    //console.log('mappedUserProfiles', mappedUserProfiles);
+    return mappedUserProfiles;
```
gorandalum, line 111: Can clean up the comments here

gorandalum (approved):

PR #3686, opened 2023-06-27
fix: update websocket error handling to use onClose events

rosvik:
Updates websocket logic to rely on onClose events instead of onError, since they provide status codes, and doesn't of arbitrarily failing on reconnects on android. `useSubscription`s onclose now returns an `isError` field based on expected status codes from https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4

Resolves comment https://github.com/AtB-AS/kundevendt/issues/4213#issuecomment-1607173946

closes https://github.com/AtB-AS/kundevendt/issues/4236

### DIFF
--- a/src/api/use-subscription.ts
+++ b/src/api/use-subscription.ts
@@ -12,15 +12,13 @@ export type SubscriptionStatus =
 
 export type SubscriptionEventProps = {
   onMessage?: (event: WebSocketMessageEvent) => void;
-  onError?: (event: WebSocketErrorEvent) => void;
   onOpen?: () => void;
-  onClose?: (event: WebSocketCloseEvent) => void;
+  onClose?: (event: WebSocketCloseEvent & {isError: boolean}) => void;
 };
 
 export function useSubscription({
   url,
   onMessage,
-  onError,
   onClose,
   onOpen,
 }: {url: string | null} & SubscriptionEventProps) {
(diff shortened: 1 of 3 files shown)


KristianSelnas (approved):

src/api/use-subscription.ts
```diff
@@ -40,25 +38,29 @@ export function useSubscription({
         onMessage && onMessage(event);
       };
 
-      webSocket.onerror = (event) => {
-        setStatus(getSubscriptionStatus(webSocket.readyState));
-        Bugsnag.notify(`WebSocket error "${event.message}"`);
-        onError && onError(event);
-      };
```
gorandalum, line 47: Not sure about removing this whole `onerror` handling. Are we sure it won't happen in production?

rosvik: It does happen in production, but when it happens it seems to me like it's not actual errors that we want to handle. It seems to happen on reconnects on android, otherwise it seems to be on network errors, which we are handling with onClose events.

https://app.bugsnag.com/atb-as/mittatb/errors/648adc57dd45910007005d13?filters%5Bevent.since%5D=30d&filters%5Bsearch%5D=websocket

The `event` value is also completely different on iOS and Android. Android is usually `null`, and on iOS it's some text in Norwegian. This makes them hard to work with, and to me it sounds like it's not made to be handled. We could keep some bugsnag logging here, but I think it's better to not have the onError callback, that sounds like it solves some problem it in reality doesn't.

Makes sense?

rosvik: One solution would be to keep something like this:

```ts
webSocket.onerror = (event) => {
  setStatus(getSubscriptionStatus(webSocket.readyState));
  if (event.message !== null) {
    Bugsnag.notify(`WebSocket error "${event.message}"`);
  }
};
```

rosvik: @gorandalum This seems to be working well locally, so updated now. Does that look ok?

gorandalum: I'll read a little documentation first, so approval probably incoming soon :+1:

Just a note, I think the `useSubscription` hook shouldn't take in an `onError` if it isn't used.

rosvik: I already removed the `onError` param from `useSubscription`

gorandalum (approved):
Looks good! Based on the [WS specifiation](https://websockets.spec.whatwg.org/#closeWebSocket), it seems that sending error event is an optional step before sending the close event, and as such an error event should never occur without an accompanying close event.

Optional suggestion:
Based on these back-and-forth on the details of the implementation, we see that what we think is pretty self-explanatory is actually more complex, and might not behave exactly like we thought. Because of this I think it would be beneficial to reduce the scope of `useSubscription` to just cover the use-case we have today, so we don't have implemented code which is not actually "battle-tested" enough to say that it works just as intended. This means:
- Remove the `return` value from the hook.
- Remove `onOpen` as a parameter.
- Rebrand `onClose(ev, isError)` to `onError()`.

rosvik:
@gorandalum Aha, thanks for the clarification! And I think that makes sense, I'll make a new PR with a refactor 🕺

PR #3687, opened 2023-06-27
fix: Set customerAccountId on reserveOffer

gorandalum:
### Background
In march, when implementing sub accounts, it was added a field
´customer_account_id´ in the reserve offer endpoint, which also
was added to the reservations in Firestore.

It was added code in the app to check that a reservation has the
correct customer account id, however it was not added code to
set the customer account id when reserving a offer. As such no
reservations created from the app showed up in the app.

### Solution
Sending in customer account id when reserving offer.

### Acceptance criteria
- [ ] Reservation created and showing in app until fare contract is received, for both card and Vipps.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchasePaymentWithCreditCardScreen/use-terminal-state.ts
+++ b/src/stacks-hierarchy/Root_PurchasePaymentWithCreditCardScreen/use-terminal-state.ts
@@ -56,6 +56,7 @@ export function useTerminalState(
                 retry: true,
               },
               scaExemption: true,
+              customerAccountId: user?.uid!,
             })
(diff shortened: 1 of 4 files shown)


rosvik (approved):
Good catch!

KristianSelnas (approved):
Good job figuring out this so quickly 👍

tormoseng:
@gorandalum Buying tickets in the staging app now only gives "Betaling godkjent. Henter billett...". Related to this?

gorandalum:
Fixed here:
https://github.com/AtB-AS/mittatb-app/pull/3693

tormoseng:
Looks good!

PR #3689, opened 2023-06-27
fix: departure time in favorite departures

mortennordseth:
The issue where the next day's departure shows as "now" in favorite departures is most likely because the service journey ID is the same for these two and therefore the key for the component is the same. To solve this, I've added the aimed time to the key, as is done for the other components. 

An example of this issue can be seen in the following screenshot, where two of the departures marked as "now" is planned to leave the following two days. 
image

Fixes https://github.com/AtB-AS/kundevendt/issues/4277

### DIFF
--- a/src/departure-list/section-items/line.tsx
+++ b/src/departure-list/section-items/line.tsx
@@ -269,7 +269,7 @@ function DepartureTimeItem({
   }
   return (
     <Button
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Great! I think this should probably have given a duplicate react key error in Bugsnag, note to self to take a look at that.

PR #3692, opened 2023-06-27
fix: Skip purchase overview screen when selecting recent purchase

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/3766


gorandalum (changes requested):
- Should look into what flow we want. The recent ticket button went straight to confirmation screen earlier, but this was changed in september last year:
https://github.com/AtB-AS/kundevendt/issues/1774
- I think it should be the same flow whether the fare contract has tariff zones or not.

KristianSelnas:
Yes, we should discuss this a bit more. Currently, the startTime is not handled properly either.

PR #3693, opened 2023-06-28
fix: Use abt_it when reserving offer

gorandalum:


src/fare-contracts/PurchaseReservation.tsx
```diff
@@ -42,7 +42,7 @@ export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
   };
 
   // Filter out reservations for subaccounts
-  if (user?.uid && reservation.customerAccountId !== user?.uid) return null;
+  if (reservation.customerAccountId !== abtCustomerIdFull) return null;
```
jorelosorio, line 45: My only concern here is if `==!` is enough because is case-sensitive, and not sure if `ABT:CustomerAccount:` may also be `abt:customerAccount`

jorelosorio (approved):

src/fare-contracts/PurchaseReservation.tsx
mjenssen: The assumption is correct - these should be case sensitive. Either way, these values are just copied between systems, and only one system is creating them and that system create all accounts as `<codespace in caps>:CustomerAccount:<id>`. I've never seen codespaces in anything but all caps. The codespaces can be found in the table here: https://developer.entur.org/products-and-pricing-data

tormoseng:
Tested OK

PR #3694, opened 2023-06-28
fix: a11y focus for travellers and zones selected

marius-at-atb:
fixes https://github.com/AtB-AS/kundevendt/issues/3943

After selecting travellers or zone for a ticket, the focus should go back to that field when returning.


marius-at-atb:
Update: The issues in this PR have been resolved by https://github.com/AtB-AS/mittatb-app/pull/3702 and https://github.com/AtB-AS/mittatb-app/pull/3700 instead. Closing

PR #3695, opened 2023-06-28
fix: Remove the build file from being embedded in the Apple IPA

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/4243

A build was made on: https://github.com/AtB-AS/mittatb-app/actions/runs/5357274058/jobs/9717917636

```
Version 1.39 (1687529871)
Jun 23, 2023 at 17:44
18.19 MB
```

No files were exposes any more after `unzip` the `.ipa`, App seems to work as before, but must be fully tested to confirm this.


jorelosorio:
> Is the part that replaces BuildFile with (null) also necessary? When building locally this gets changed back for me.

Yes, this basically uncheck embed those files into the project, Xcode sets then null.

jorelosorio:
This how it looks before this change:

image

And how it looks after the change:

image

jorelosorio:
> This how it looks before this change:
> 
> image
> And how it looks after the change:
> 
> image

Debug, Release, Debug disappear from being copy into the bundle.

gorandalum (approved):

PR #3696, opened 2023-06-28
feat: bypass toggle restrictions on `MobileTokenRequired`

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/4135

Depends on: https://github.com/AtB-AS/abt-token-server/pull/49 & https://github.com/AtB-AS/abt-token-server/pull/51

### DIFF
--- a/src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
+++ b/src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
@@ -102,7 +102,7 @@ export const SelectTravelTokenScreenComponent = ({onAfterSave}: Props) => {
         return;
       }
       setSaveState({saving: true, error: false});
-      const success = await toggleToken(selectedToken.id);
+      const success = await toggleToken(selectedToken.id, false);
       if (success) {
         onAfterSave();
       } else {
--- a/src/stacks-hierarchy/Root_ActiveTokenOnPhoneRequiredForFareProductScreen.tsx
+++ b/src/stacks-hierarchy/Root_ActiveTokenOnPhoneRequiredForFareProductScreen.tsx
@@ -54,7 +54,7 @@ export const Root_ActiveTokenOnPhoneRequiredForFareProductScreen = ({
         return;
       }
       setSaveState({saving: true, error: false});
-      const success = await toggleToken(selectedToken.id);
+      const success = await toggleToken(selectedToken.id, true);
       if (success && nextScreen) {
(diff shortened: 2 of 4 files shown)


KristianSelnas:
How does this not bypass restrictions for all changes?

src/stacks-hierarchy/Root_ActiveTokenOnPhoneRequiredForFareProductScreen.tsx
```diff
@@ -54,7 +54,7 @@ export const Root_ActiveTokenOnPhoneRequiredForFareProductScreen = ({
         return;
       }
       setSaveState({saving: true, error: false});
-      const success = await toggleToken(selectedToken.id);
+      const success = await toggleToken(selectedToken.id, true);
```
jorelosorio, line 57: @KristianSelnas Bypass `true` is only sent while changing to mobile. Then it doesn't count to the user that change. Otherwise, will count the change.

src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
```diff
@@ -102,7 +102,7 @@ export const SelectTravelTokenScreenComponent = ({onAfterSave}: Props) => {
         return;
       }
       setSaveState({saving: true, error: false});
-      const success = await toggleToken(selectedToken.id);
+      const success = await toggleToken(selectedToken.id, false);
```
gorandalum, line 105: This sets `bypassRestrictions` to `false` also when changing to mobile through the `SelectTravelTokenScreen`. This may be what is preferred, but should be "formally" agreed upon so there are no surprises later on how the implementation works 😊

gorandalum (approved):
Looks good. But remember to update the issue when creating PRs, so the issue description and the solution is pretty much in sync. Now the solution (with sending a bypass parameter) is not the same as the one described in the issue (always allow toggle from travelcard to mobile).

src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
rosvik: Clarified with @Sebstorvik that this is how it should be 👍

rosvik (approved):

rosvik:
In case we remove the restriction bypass at some point, and the user is blocked from toggling, it would be good to have a more speciffic error messages than "Something went wrong". It's nice to have, so merging this now, but made a new issue for that here: https://github.com/AtB-AS/kundevendt/issues/5281

PR #3699, opened 2023-06-30
fix: updating the wrong departure in favorite departures with realtime data

mortennordseth:
To prevent us from updating the wrong departure in favorite departures with real-time data for another departure, I've added a check on `aimedDepartureTime`. This change was necessary because departures have the same `serviceJourneyId`, which caused the app to update the wrong departure in the user's favorite departures list. 

See https://github.com/AtB-AS/atb-bff/pull/279 for necessary changes to the bff. 

Closes https://github.com/AtB-AS/kundevendt/issues/4277

### DIFF
--- a/src/departure-list/utils.ts
+++ b/src/departure-list/utils.ts
@@ -7,13 +7,9 @@ import {
 } from '@atb/api/departures/types';
 import {EstimatedCall} from '@atb/api/types/departures';
 import {DepartureRealtimeData, DeparturesRealtimeData} from '@atb/sdk';
-import {
-  differenceInMinutesStrings,
-  isNumberOfMinutesInThePast,
-} from '@atb/utils/date';
+import {isNumberOfMinutesInThePast} from '@atb/utils/date';
 
 export const HIDE_AFTER_NUM_MINUTES = 1;
-const MAX_REALTIME_DELAY_MINUTES = 600;
 /***
  * Used to update all stops with new time from realtime mapping object returned
  * from the BFF. It also removes outdated departures which most likely have passed.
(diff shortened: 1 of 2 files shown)


rosvik:
Does this apply to cases like Ringen in Oslo as well, where the same journey passes the same stop several times?

Would a better solution be to return `aimedDepartureTime` for realtime data from the BFF, so that we always have an unique identifier for each departure?

mortennordseth:
If the journey has the same `serviceJourneyId` it does. 

That's probably a better solution. Maybe we could make a unique identifier using the service date and serviceJourneyId, `datedServiceJourneyId` (e.g. "2023‐06‐30_ATB:ServiceJourney:73_200329232360547_37_200109141539610")?

What are the consequences of adding a new field to the BFF, @rosvik? Do we need a new version (v3) of the endpoint?

rosvik:
Ringen does have the same ID, yes. It also passes the same stop twice in less than an hour, so just the date isn't enough to identify it. Other places, we have used the full `aimedDepartureTime + serviceJourneyId` to account for this, since this doesn't change with delays, and is unique no matter how frequent the departure is. See here: https://github.com/AtB-AS/mittatb-app/blob/master/src/place-screen/components/QuaySection.tsx#L175

It's safe to add new fields to BFF endpoints without making a new version 👍

mortennordseth:
Updated with a better solution that also works for "Ringen", thanks to @rosvik 🎉 PR for the changes to the BFF: https://github.com/AtB-AS/atb-bff/pull/279.

rosvik (approved):
Looks good 💯

PR #3700, opened 2023-06-30
chore: bottom sheet focus improvements

marius-at-atb:
Resolves https://github.com/AtB-AS/kundevendt/issues/4714

Return onOpenFocusRef and onCloseFocusRef in BottomSheetState, remove close from contentFunction.
Bonus: no longer using setTimeout for giveFocus.
Also provides onCloseFocusRef to GenericClickableSectionItem which fixes the specific case of returning from traveller selection.

### DIFF
--- a/src/components/bottom-sheet/BottomSheetContext.tsx
+++ b/src/components/bottom-sheet/BottomSheetContext.tsx
@@ -22,8 +22,8 @@ import {AnimatedBottomSheet} from './AnimatedBottomSheet';
 
 type BottomSheetContentFunction = (
   close: () => void,
-  /** Use focusRef to give a component accessibility focus when BottomSheet open. The component must be accessible! */
-  focusRef: RefObject<any>,
+  /** Use onOpenFocusRef to give a component accessibility focus when BottomSheet open. The component must be accessible! */
+  onOpenFocusRef: RefObject<any>,
 ) => ReactNode;
 
 type BottomSheetState = {
@@ -48,11 +48,11 @@ export const BottomSheetProvider: React.FC = ({children}) => {
   const [isOpen, setIsOpen] = useState(false);
   const [isBackdropEnabled, setBackdropEnabled] = useState(true);
   const [contentFunction, setContentFunction] = useState<
-    (close: () => void, focusRef: RefObject<any>) => ReactNode
+    (close: () => void, onOpenFocusRef: RefObject<any>) => ReactNode
   >(() => () => null);
 
   const animatedOffset = useMemo(() => new Animated.Value(0), []);
-  const focusRef = useFocusOnLoad();
+  const onOpenFocusRef = useFocusOnLoad();
   const [closeRef, setCloseRef] = useState<RefObject<any> | undefined>();
 
   useEffect(
@@ -75,7 +75,10 @@ export const BottomSheetProvider: React.FC = ({children}) => {
   };
 
   const open = (
-    contentFunction: (close: () => void, focusRef: RefObject<any>) => ReactNode,
+    contentFunction: (
+      close: () => void,
+      onOpenFocusRef: RefObject<any>,
+    ) => ReactNode,
     closeRef?: RefObject<any>,
     useBackdrop: boolean = true,
   ) => {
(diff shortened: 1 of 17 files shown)


src/components/bottom-sheet/BottomSheetContext.tsx
```diff
@@ -20,22 +21,20 @@ import {Backdrop} from './Backdrop';
...
-    closeRef?: RefObject<any>,
     useBackdrop?: boolean,
   ) => void;
   isOpen: () => boolean;
   close: () => void;
   height: number;
+  /** Use onOpenFocusRef to give a component accessibility focus when BottomSheet open. The component must be accessible! */
+  onOpenFocusRef?: RefObject<any>;
+  /** Optional ref to component which should be focused on sheet close */
+  onCloseFocusRef?: RefObject<any>;
```
gorandalum, line 37: Does these need to be typed as optional?

src/components/bottom-sheet/BottomSheetContext.tsx
```diff
@@ -69,18 +68,14 @@ export const BottomSheetProvider: React.FC = ({children}) => {
   const close = () => {
     setContentFunction(() => () => null);
     setIsOpen(false);
-    if (closeRef) {
-      setTimeout(() => giveFocus(closeRef), 200);
-    }
+    onCloseFocusRef && giveFocus(onCloseFocusRef);
```
gorandalum, line 71: `onCloseFocusRef` should always be set, it is the `current` value which is optional.

src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
```diff
@@ -76,7 +76,7 @@ export const useUpdateBottomSheetWhenSelectedEntityChanges = (
               }}
             />
           ),
-          undefined,
+          //undefined,
```
gorandalum, line 79: Can remove these instead of commenting them out.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-travel-search-filters-state.tsx
```diff
@@ -17,7 +17,7 @@ type TravelSearchFiltersState =
       anyFiltersApplied: boolean;
       resetTransportModes: () => void;
       disableFlexibleTransport: () => void;
-      closeRef: React.Ref<any>;
+      onCloseFocusRef: React.RefObject<any> | undefined;
```
gorandalum, line 20: I think this doesn't need to be optional.

src/components/bottom-sheet/BottomSheetContext.tsx
gorandalum, line 71: No longer necessary to wrap `giveFocus` with `setTimeout`?

gorandalum (changes requested):
Very good! Some smaller comments :)

src/components/bottom-sheet/BottomSheetContext.tsx
marius-at-atb: Not needed I think, removed now

src/components/bottom-sheet/BottomSheetContext.tsx
marius-at-atb: true!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-travel-search-filters-state.tsx
marius-at-atb: No longer optional 👍

src/components/bottom-sheet/BottomSheetContext.tsx
marius-at-atb: No longer necessary it seems

gorandalum (approved):

PR #3701, opened 2023-06-30
fix: Do a `pod install` to keep update Xcode project with the fixes made on security fix at https://github.com/AtB-AS/mittatb-app/pull/3695

jorelosorio:
A part of: https://github.com/AtB-AS/kundevendt/issues/4243

NOTE: This is not including the Debug, Release and Debug config files to be embedded again, it is a result of a pod install after these files were removed.


gorandalum (approved):
:+1:

A note that the commit message subject should be short and informative. There is a convention saying it should be max 50 characters, but I think that is a little strict, so I don't follow that rigorously myself.

PR #3708, opened 2023-07-04
fix: spacing issue between long stack titles and buttons

mortennordseth:
Today we render the buttons in a header on a new line in case of long titles or large accessibility font sizes to avoid overlap. According to [Figma](https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/01.-Components?type=design&node-id=4383-8566&mode=design&t=oXpM43wQHj9EROUk-0), we should wrap the text instead and have the same space on both sides of the title. 

I've tried a few possible solutions but found this the best. This way, we use a defined width for the title, ensuring that the title is centered, not overlapping with the buttons, and rendered correctly at first render. 

Another possible solution I tried was not to use absolute positioned buttons, calculate the widest button width and use this as the width for both the left and right buttons (even though there is only one button). This caused the title to move after the first render because of changes to the button widths (in the dashboard it would move some pixels to the right). A similar problem can be found in the app today, where the chat icon on the dashboard is first rendered outside the screen, then popped back into the correct position. I've solved this for titles that are on one line by setting the default button height to match the line height of the font.

<details>

image

image

image

image

</details>

Fixes https://github.com/AtB-AS/kundevendt/issues/4224

### DIFF
--- a/src/components/screen-header/ScreenHeader.tsx
+++ b/src/components/screen-header/ScreenHeader.tsx
@@ -116,7 +116,7 @@ const BaseHeader = ({
   const themeColor = color ?? 'background_accent_0';
   const focusRef = useFocusOnLoad(setFocusOnLoad);
 
-  const {buttonsHeight, buttonsTopOffset, setLayoutFor} = useHeaderLayouts();
+  const {buttonsHeight, setLayoutFor} = useHeaderLayouts();
 
   const backgroundColor = getStaticColor(themeName, themeColor).background;
 
@@ -127,13 +127,7 @@ const BaseHeader = ({
         accessible={!!title && !!textOpacity}
         importantForAccessibility={!!title ? 'yes' : 'no-hide-descendants'}
         accessibilityRole="header"
-        style={[
-          css.headerTitle,
-          {
-            // Make space for absolute positioned buttons in case they are offset below title
-            marginBottom: theme.spacings.medium + buttonsTopOffset,
-          },
-        ]}
+        style={css.headerTitle}
         onLayout={setLayoutFor('container')}
         ref={focusRef}
       >
@@ -144,9 +138,9 @@ const BaseHeader = ({
         >
           <ThemeText
             accessible={false}
-            onLayout={setLayoutFor('title')}
             type="body__primary--bold"
             color={themeColor}
+            style={{textAlign: 'center'}}
           >
             {title && textOpacity > 0 ? title : '\u00a0'}
           </ThemeText>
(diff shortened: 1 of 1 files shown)


src/components/screen-header/ScreenHeader.tsx
```diff
@@ -180,13 +173,16 @@ const useHeaderStyle = StyleSheet.createThemeHook((theme) => ({
     paddingBottom: theme.spacings.medium,
     borderTopLeftRadius: theme.border.radius.circle,
     borderTopRightRadius: theme.border.radius.circle,
+    marginBottom: theme.spacings.medium,
+    alignItems: 'center',
   },
-  headerTitle: {alignItems: 'center'},
+  headerTitle: {alignItems: 'center', width: '66%'},
```
toretefre, line 179: How is 66% decided?

src/components/screen-header/ScreenHeader.tsx
```diff
@@ -215,28 +206,15 @@ const useHeaderLayouts = () => {
...
-    const leftButtonVisibleWidth = leftButton.height ? leftButton.width : 0;
-    const rightButtonVisibleWidth = rightButton.height ? rightButton.width : 0;
-    const widestButtonWidth = Math.max(
-      leftButtonVisibleWidth,
-      rightButtonVisibleWidth,
-    );
-    const buttonsAndTitleWidth = title.width + widestButtonWidth * 2;
-    const containerWidth = container.width;
-    return buttonsAndTitleWidth > containerWidth - 10;
+    return headerLayouts.container?.height || 0;
```
toretefre, line 214: Is setting `0` here the most sensible default?

toretefre (approved):
Looking good, also a solid simplification of how things are done 🙌 

Gøran is on vacation, so I'll approve this unless @rosvik has any comments?

src/components/screen-header/ScreenHeader.tsx
mortennordseth: This way we can check if the value is falsy and use the line height if so (i.e. the expression on line 154). Another approach is to make the hook have access to `theme` and return the line height instead of 0 🤔

src/components/screen-header/ScreenHeader.tsx
mortennordseth: I did some testing and figured that 2/3 of the screen made sense for the title. Might be a better value to use here, but it has to be a percentage so that zooming works nicely. Any suggestions for a better width or how to find one?

toretefre: Nope, just curious! As long as it works with 200% zoom on devices of various widths it should be alright.

rosvik:
I know @ingvisim looked at a similar solution a few weeks ago (when working on https://github.com/AtB-AS/mittatb-app/pull/3636), but realized it might make more problems than it solved 😅 I don't know exactly what was decided, but I don't think a simple solution like this will work in all cases. 

I tested locally a bit, and got issues with overlapping text when using 200% text size, so I do think we need a fallback to text above or below buttons in some cases. But maybe it's worth talking to Ingvild or Gøran when their back from vacation. 

<div>



</div>

mortennordseth:
Yup 😄 
> but realized it might make more problems than it solved 😅

A fallback solution sounds like a good idea to add, but it might be just better to leave it as is today and don't wrap the title 🤔 Either way, I'll wait for Ingvild/Gøran and hear what they decided when working on this.

PR #3712, opened 2023-07-06
feat: Non transit travel search

KristianSelnas:


src/components/button/Button.tsx
```diff
@@ -227,3 +223,9 @@ const useButtonStyle = StyleSheet.createThemeHook((theme: Theme) => ({
     borderWidth: theme.border.width.medium,
   },
 }));
+
+function getTextType(mode: string, type: string) {
+  if (mode === 'tertiary') return 'body__primary';
+  if (type === 'pill') return 'body__secondary';
```
KristianSelnas, line 229: After discussions with the designers, we decided to change the default font size for `pill` to secondary. The designers always intended this button to have the secondary font size, and we believe it is currently a bug that it's primary. If you oppose this, please let me know.

I've checked every instance of this button type in the app, and the designers and I believe secondary works best in every instance.

marius-at-atb:
Update: see https://github.com/AtB-AS/mittatb-app/pull/3728

PR #3710, opened 2023-07-04
feat: new FareProductTile design with new illustrations

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3363

Before:

<br /><br />

After:

PR #3713, opened 2023-07-06
feat: update ticket assistant tile design

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3363

This updates the design of the TicketAssistantTile.

Also TicketingTile.tsx was created, shared by both FareProductTile and TicketAssistantTile, as [suggested](https://github.com/AtB-AS/mittatb-app/pull/3710#issuecomment-1623166177).
Turned FareProductIllustration into TicketingTileIllustration.

Also updates the BetaTag design.

Before:

<br /><br />

After:

<br />

### DIFF
--- a/src/components/beta-tag/BetaTag.tsx
+++ b/src/components/beta-tag/BetaTag.tsx
@@ -1,7 +1,7 @@
-import {StyleProp, View, ViewStyle} from 'react-native';
-import {ThemeText} from '@atb/components/text';
+import {StyleProp, ViewStyle} from 'react-native';
 import {StyleSheet} from '@atb/theme';
 import React from 'react';
+import {InfoChip} from '../info-chip';
 
 type Props = {
   style?: StyleProp<ViewStyle>;
@@ -9,11 +9,11 @@ type Props = {
 export const BetaTag = ({style}: Props) => {
   const styles = useStyles();
   return (
-    <View style={[styles.betaLabel, style]}>
-      <ThemeText type={'body__tertiary'} color="background_accent_2">
-        Beta
-      </ThemeText>
-    </View>
+    <InfoChip
+      text="Beta"
+      interactiveColor="interactive_0"
+      style={[styles.betaLabel, style]}
+    />
   );
 };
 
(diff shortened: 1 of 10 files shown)


marius-at-atb:
@rosvik @KristianSelnas I just pushed another commit after requesting reviews, having just realised that the whole TicketingTile should be clickable. Ready now.

src/components/beta-tag/BetaTag.tsx
```diff
@@ -10,7 +10,7 @@ export const BetaTag = ({style}: Props) => {
   const styles = useStyles();
   return (
     <View style={[styles.betaLabel, style]}>
```
KristianSelnas, line 12: Could the `InfoChip` component have been used here? I know this is not your code originally, but when it's revisited I think it's a good idea to take the chance to improve the code. Generally, creating one-off UI elements (as is effectively done here) should not be done without really clear reasons.

marius-at-atb: Yes, InfoChip can be used here as implemented in the new commit

src/components/beta-tag/BetaTag.tsx
```diff
@@ -1,27 +1,26 @@
...
 };
 export const BetaTag = ({style}: Props) => {
   const styles = useStyles();
   return (
-    <View style={[styles.betaLabel, style]}>
-      <ThemeText type={'body__tertiary'} color="background_accent_2">
-        Beta
-      </ThemeText>
-    </View>
+    <InfoChip
```
KristianSelnas, line 12: I believe this still kinda defeats the purpose. It's still a one-off UI component (has explicitly set backgroundColor and borderRadius)

marius-at-atb: Explicitly setting backgroundColor and borderRadius is currently needed to match the exact Figma drawings as far as I can tell?

KristianSelnas: Well, then it's the Figma drawings that are incorrect. I sincerely don't believe the designers intended to introduce a new UI element just for this single purpose. If for some reason this is the case, the `BetaTag` component should be moved to the `components` folder.

KristianSelnas: I suggest you check this with the designers.

marius-at-atb: Following discussion with the designers, it should be kept as is and moved to the components folder. Perhaps it should then also receive text as input, and be called InfoTag instead?

marius-at-atb: @KristianSelnas done

KristianSelnas (approved):

PR #3720, opened 2023-07-12
feat: ticketing valid duration info

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3830

Add text line that informs about ticket time duration in Purchase Confirmation Screen.

<br />

<br />

### DIFF
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -13,6 +13,8 @@ export type PreassignedFareProduct = {
   type: string;
   productAlias?: LanguageAndTextType[];
   distributionChannel: DistributionChannel[];
+  durationDays?: number;
+  durationMinutes?: number;
   alternativeNames: LanguageAndTextType[];
   zoneSelectionMode?: ZoneSelectionMode;
   limitations: {
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -13,8 +13,6 @@ export type PreassignedFareProduct = {
   type: string;
   productAlias?: LanguageAndTextType[];
   distributionChannel: DistributionChannel[];
-  durationDays?: number;
-  durationMinutes?: number;
   alternativeNames: LanguageAndTextType[];
   zoneSelectionMode?: ZoneSelectionMode;
   limitations: {
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -151,8 +151,6 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
     }),
   );
 
-  const validDuration = secondsToDuration(validDurationSeconds || 0, language);
-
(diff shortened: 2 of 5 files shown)


toretefre:
Does this handle 3 hour single ticket at AtB in any way?

marius-at-atb:
No, it is not supported as I wasn't aware of these variations in single tickets. Will look into it now, moving back to draft :sweat_smile:

toretefre:
Alright, no worries! 

For the 3 hour validity time I would've looked at the implementation of schedules for the night tickets, maybe it can be solved in a recognizable way?

And for the additional time for each added zone, it might be possible to specify a field either on the product (referenceData) or in the fareProductTypeConfig (single), since it will differ for the different OMS partners. (Checked now, AtB has 60 additional minutes, while I think FRAM has 30 additional minutes). 

However, counting zones isn't necessarily the easiest task, since I'm not sure if the offer includes information on all zones included in the ticket (doubt it really). Might want to ask Sebastian what to do in that case 😅

marius-at-atb:
@mjenssen mentioned that the offers have from and to time data that should be correct. Here you can see it in the webshop:

webshopFromToData

Seems like a good solution to use this data

toretefre:
Agree! Always good to use existing data instead of guessing, didn't know 🚀

marius-at-atb:
@toretefre New approach in place now 😄

src/reference-data/types.ts
```diff
@@ -13,6 +13,8 @@ export type PreassignedFareProduct = {
   type: string;
   productAlias?: LanguageAndTextType[];
   distributionChannel: DistributionChannel[];
+  durationDays?: number;
+  durationMinutes?: number;
```
marius-at-atb, line 17: Note: Will revert this change

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -149,6 +151,8 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
     }),
   );
 
+  const validDuration = secondsToDuration(validDurationSeconds || 0, language);
```
toretefre, line 154: You could consider calling this one `validDurationString` to be in line with the strings declared just below, if they do the same thing?

Might also be unnecessary to declare the variable at all if `validDurationSeconds` is falsy, which is part of a check at line 314? Just thinking out loud here, not really seeing any other way to do it, since I like the approach with declaring the variable outside of the function to make it more readable, and I guess running with the default 0 won't cause any issues 👌

toretefre (approved):
Looking good, really nice addition to the app that many users will appreciate! 🎉 

Added a small nit-picky comment that you can feel free to ignore if you like 🚀

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
marius-at-atb: Good point! By not declaring the variable at all, and moving it inside the validDurationSeconds check, we also don't need the "|| 0" :+1:

PR #3725, opened 2023-07-20
chore: Update preassigned fare products for boat tickets

ingvisim:
Update reference data to match firestore data that fixes bug where you cannot pick duration on period boat tickets

https://github.com/AtB-AS/firestore-configuration/pull/224

https://github.com/AtB-AS/kundevendt/issues/4044


marius-at-atb (approved):
Lgtm!

PR #3729, opened 2023-07-21
feat: Update ticket confirmation screen

ingvisim:
Update texts on confirmation screen for boat tickets, and add boat tickets to default modes we sell tickets for

Simulator Screen Shot - iPhone 14 Pro Max - 2023-07-21 at 12 18 58

Close https://github.com/AtB-AS/kundevendt/issues/4044

### DIFF
--- a/src/configuration/defaults.ts
+++ b/src/configuration/defaults.ts
@@ -7,6 +7,8 @@ export const defaultModesWeSellTicketsFor: string[] = [
   'localTram',
   'regionalBus',
   'shuttleBus',
+  'highSpeedPassengerService',
+  'highSpeedVehicleService',
 ];
 export const defaultPaymentTypes: PaymentType[] = [
   PaymentType.Vipps,
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -127,6 +127,13 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
   const {travellerSelectionMode, zoneSelectionMode} =
     fareProductTypeConfig.configuration;
 
+  const offerEndpoint =
+    zoneSelectionMode === 'none'
+      ? 'authority'
+      : zoneSelectionMode === 'multiple-stop-harbor'
+      ? 'stop-places'
+      : 'zones';
+
   const {
     offerSearchTime,
     isSearchingOffer,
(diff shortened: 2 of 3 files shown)


src/configuration/defaults.ts
```diff
@@ -7,6 +7,8 @@ export const defaultModesWeSellTicketsFor: string[] = [
   'localTram',
   'regionalBus',
   'shuttleBus',
+  'highSpeedPassengerService',
+  'highSpeedVehicleService',
```
marius-at-atb, line 11: defaultModesWeSellTicketsFor should also be reused here:

https://github.com/AtB-AS/mittatb-app/blob/11e160b6bb6f6731eb6b87285c6219a686593dfd/src/travel-details-screens/TripDetailsScreenComponent.tsx#L194-L201

However, since all the elements in the list are actually submodes, this variable is incorrectly named. Not a part of your changes, but I think it is a good opportunity to correct it.
Should rename modesWeSellTicketsFor in FirestoreConfigurationContext as well.

marius-at-atb (changes requested):
Looks good :tada:
Just one minor change requested.

src/configuration/defaults.ts
marius-at-atb: Suggested renaming:
defaultModesWeSellTicketsFor -> defaultSubModesWeSellTicketsFor
modesWeSellTicketsFor -> subModesWeSellTicketsFor

ingvisim: defaultModesWeSellTicketsFor should not be used there, as is mentioned in the comments above:
https://github.com/AtB-AS/mittatb-app/blob/11e160b6bb6f6731eb6b87285c6219a686593dfd/src/travel-details-screens/TripDetailsScreenComponent.tsx#L192-L193

marius-at-atb: Good catch! However still add 'highSpeedPassengerService', 'highSpeedVehicleService'

ingvisim: > Suggested renaming:
> defaultModesWeSellTicketsFor -> defaultSubModesWeSellTicketsFor
> modesWeSellTicketsFor -> subModesWeSellTicketsFor

This would have to be changed in firestore as well, and would be a bigger change

marius-at-atb: I guess it can be part of this issue, or a new one. Your call :+1:

ingvisim: > Good catch! However still add 'highSpeedPassengerService', 'highSpeedVehicleService'

We do not currently support puchasing boat tickets from trip details, and as of now only supports regular single tickets

marius-at-atb: Fair enough :+1:

marius-at-atb (approved):

PR #3736, opened 2023-07-28
feat: flexible transport trip details

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/3856

This PR handles flexible transport in trip details and adds a bottom sheet component with more info about it.
See comments down below for screenshots.

Notes:
~~- Padding bottom is not showing correctly in the simulator, see comment about this issue below~~ -> now fixed
- The text in the message links is dynamically created by the app instead of being retrieved from Firestore
- Phone booking option shows independently of online booking option
- "Remove the top info message" -> this one was already removed it seems

<br />
<br />

Relevant values in Remote Config:
- new: flex_booking_number_of_days_available
- existing: use_flexible_on_accessMode, use_flexible_on_directMode, use_flexible_on_egressMode 

In Firestore: new document configuration.urls

All of these values need to be followed up on when going into prod.


marius-at-atb:
When < 1 hour left to booking deadline:
<section>


</section>
<br />

Dark Mode:
<section>


</section>

marius-at-atb:
When > 1 hour left to booking deadline, while booking is available:
<section>


</section>
<br />

Dark Mode:
<section>


</section>

marius-at-atb:
Some more message cases while booking is available:
<section>




</section>

marius-at-atb:
What the messages look like when too early to book:
<section>




</section>

marius-at-atb:
Regarding the padding issue mentioned above, here it is shown with red background:

On device, correct &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; In simulator, too small

src/travel-details-screens/utils.ts
```diff
@@ -179,3 +181,112 @@ function someLegsAreByTrain(tripPattern: TripPattern): boolean {
...
+  secondsRemainingToAvailable: Infinity,
+  latestBookingDate: new Date(Number.MAX_VALUE),
+  earliestBookingDate: new Date(Number.MAX_VALUE),
+};
+
+export function getBookingRequirementForLeg(
+  leg: Leg | undefined,
+  now: number,
+  flex_booking_number_of_days_available: number, // would be better to get earliestBookingTime as ISO string from Entur's API.
+): BookingRequirement {
```
gorandalum, line 224: I would prefer not mapping this in the `Results` component, creating a new domain model for `TripPattern` (`AvailableTripPattern`) which is passed through the app. Generally we use the domain models from Entur/BFF throughout the app, and it is the business-specific lead nodes that utilise the data as needed.

This means for example to have a util function `isNecessaryBookingToEarly` which may be used in the leaf nodes, rather than passing around the value `isTooEarly` in a custom domain model. I feel this create less coupling between the different components in the app.

gorandalum (changes requested):

src/travel-details-screens/utils.ts
marius-at-atb: Updated now as requested.
Even though the implementation was more concise originally, I can see why you would want this pattern implemented.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -303,24 +309,35 @@ export const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
             <DestinationIcon style={styles.iconContainer} />
             <View style={styles.departureTimes}>
               <ThemeText type="body__tertiary" color="primary" testID="endTime">
-                {formatToClock(tripPattern.expectedEndTime, language, 'ceil')}
+                {(legs.length && isLegFlexibleTransport(legs[legs.length - 1])
```
jorelosorio, line 312: I suggest to check for `legs.length > 0` instead, to avoid the case that legs has length but 0.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -679,9 +701,16 @@ const tripSummary = (
     ),
   );
 
-  const requiresBookingText = tripPattern.bookingRequirement?.requiresBooking
-    ? t(TripSearchTexts.results.resultItem.footer.requiresBooking)
-    : undefined;
+  const bookingsRequiredCount =
+    getTripPatternBookingsRequiredCount(tripPattern);
```
jorelosorio, line 705: Seems like this function is only used here, consider move it to an internal function, instead of in `utils/`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -303,24 +309,35 @@ export const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
-}> = ({bookingRequirement}) => {
+  tripPattern: TripPattern;
+}> = ({tripPattern}) => {
   const styles = useThemeStyles();
   const {t} = useTranslation();
 
-  const {requiresBooking, requiresBookingUrgently} = bookingRequirement;
+  const now = useNow(2500);
+
+  const requiresBookingUrgently = getTripPatternRequiresBookingUrgently(
```
jorelosorio, line 334: Seems like this function is only used here, consider move it to an internal function, instead of in `utils/`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -94,34 +95,36 @@ export const Results: React.FC<Props> = ({
...
-          ) : (
-            <ResultItemOld
-              tripPattern={tripPattern}
-              onDetailsPressed={() => {
-                onDetailsPressed(tripPattern, i);
-              }}
-              searchTime={searchTime}
-              testID={'tripSearchSearchResult' + i}
+      {tripPatterns
+        .filter((tp) => !getIsTooLateToBookTripPattern(tp, now))
```
jorelosorio, line 99: Seems like this function is only used here, consider move it to an internal function, instead of in `utils/.`, and consider renaming it as `isLateToBook`

src/translations/dictionary.ts
```diff
@@ -74,6 +74,23 @@ const dictionary = {
...
+        case -1:
+          return _('i går', 'yesterday', 'i går');
+        case 0:
+          return _('i dag', 'today', 'i dag');
+        case 1:
+          return _('i morgen', 'tomorrow', 'i morgon');
+        case 2:
+          return _('i overmorgen', 'the day after tomorrow', 'i overmorgon');
+        default:
+          return _('', '', '');
```
jorelosorio, line 90: Shouldn't be better to show here like `in 3 days`, `in 5 days`, `20 days ago`?

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,201 @@
...
+      />
+      <TouchableOpacity style={style.closeIconContainer} onPress={close}>
+        <CloseCircle height={theme.icon.size.large} />
+      </TouchableOpacity>
+
+      <ScrollView
+        style={[
+          style.contentContainer,
+          {
+            marginBottom: Math.max(safeAreaBottom, theme.spacings.medium),
```
jorelosorio, line 79: Not sure about these calculations in the UI, what does this fix?

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,201 @@
...
+            <FlexibleTransport height={107} />
+          </View>
+        )}
+
+        <ThemeText type="heading__title" style={style.title}>
+          {t(TripDetailsTexts.flexibleTransport.contentTitle(publicCode))}
+        </ThemeText>
+        <View style={style.steps} accessibilityRole="list">
+          {
+            // eslint-disable-next-line rulesdir/translations-warning
```
jorelosorio, line 103: Is it not possible to remove? the Lint here, it is because of the `${i + 1}. `

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+        <TouchableOpacity
+          style={style.readMoreAbout}
+          onPress={() => {
+            const flexTransportInfoUrl =
+              getTextForLanguage(
+                configurableLinks?.flexTransportInfo,
+                language,
+              ) || '';
+            flexTransportInfoUrl && Linking.openURL(flexTransportInfoUrl);
```
jorelosorio, line 133: I think no matter what, here `Linking.openURL(flexTransportInfoUrl)` will be called because of the `|| ''`

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,201 @@
...
+            {t(TripDetailsTexts.flexibleTransport.readMoreAbout(publicCode))}
+          </ThemeText>
+        </TouchableOpacity>
+
+        {bookingIsAvailable && <FlexibleTransportBookingOptions leg={leg} />}
+
+        {Platform.OS === 'android' && (
+          <View // scroll bugfix
+            style={{height: theme.spacings.xLarge}}
+          />
```
jorelosorio, line 151: Looks hacky 😅

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,201 @@
...
+};
+
+const useStyle = StyleSheet.createThemeHook((theme) => ({
+  title: {
+    paddingHorizontal: theme.spacings.small,
+    paddingTop: theme.spacings.small,
+  },
+  closeIconContainer: {
+    position: 'absolute',
+    top: theme.spacings.large / 2 + theme.spacings.small,
```
jorelosorio, line 165: What does this solve? why not then another value like `theme.spacing.small` or another value from the spacing values in the theme? Instead of `theme.spacings.large / 2`.

src/travel-details-screens/components/FlexibleTransportBookingOptions.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const bookingMethods = bookingArrangements?.bookingMethods;
+
+  const showBookOnlineOption =
+    bookingUrl && bookingMethods?.some((bm) => bm === BookingMethod.Online);
+  const showBookByPhoneOption =
+    bookingPhone &&
+    bookingMethods?.some((bm) => bm === BookingMethod.CallOffice);
+
+  if (showBookOnlineOption || showBookByPhoneOption) {
```
jorelosorio, line 37: This condition might not be needed.

src/travel-details-screens/components/FlexibleTransportBookingOptions.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </>
+    );
+  } else {
+    return null;
+  }
+};
+
+const useStyle = StyleSheet.createThemeHook((theme) => ({
+  flexBookingOption: {
+    paddingVertical: theme.spacings.medium / 2,
```
jorelosorio, line 88: Same as before, why is need to be / 2? use a small value from the spacings.

src/travel-details-screens/components/Time.tsx
```diff
@@ -10,22 +10,26 @@ import {usePreferences} from '@atb/preferences';
...
+
   const {aimedTime, expectedTime} = timeValues;
   const representationType = getTimeRepresentationType(timeValues);
-  const scheduled = formatToClock(
-    aimedTime,
-    language,
-    roundingMethod,
-    debugShowSeconds,
-  );
+  const scheduled =
```
jorelosorio, line 26: I saw a translation to handle this, try to reuse if possible.

src/travel-details-screens/components/TripSection.tsx
```diff
@@ -87,6 +97,7 @@ export const TripSection: React.FC<TripSectionProps> = ({
   const isWalkSection = leg.mode === Mode.Foot;
   const isBikeSection = leg.mode === Mode.Bicycle;
   const isFlexible = isLegFlexibleTransport(leg);
+  const timesAreApproximations = isFlexible;
```
jorelosorio, line 100: Might not be needed `timesAreApproximations` use isFlexible instead directly.

src/travel-details-screens/components/TripSection.tsx
```diff
@@ -108,6 +119,27 @@ export const TripSection: React.FC<TripSectionProps> = ({
...
+  const now = useNow(2500);
+  const {flex_booking_number_of_days_available} = useRemoteConfig();
+  const bookingIsAvailable = getLegBookingIsAvailable(
+    leg,
+    now,
+    flex_booking_number_of_days_available,
+  );
+  const requiresBookingUrgently = getLegRequiresBookingUrgently(leg, now);
+
+  const atbAuthorityId = 'ATB:Authority:2';
```
jorelosorio, line 133: I could consider to have this value somewhere in a way if this change. For instance: Another company has support for Flexible booking. So, moving it to a remote config or something gives that flexibility, is just a suggestion we can ask other team members what do they think.

src/utils/date.ts
```diff
@@ -17,18 +17,34 @@ import {
...
 } from 'date-fns';
 import en from 'date-fns/locale/en-GB';
 import nb from 'date-fns/locale/nb';
 import humanizeDuration from 'humanize-duration';
-import {DEFAULT_LANGUAGE, Language} from '@atb/translations';
+
+import {
+  parse as parseIso8601Duration,
+  toSeconds as toSecondsIso8601Duration,
+} from 'iso8601-duration';
```
jorelosorio, line 29: It might not need to have another library, please check the library we are using now: https://date-fns.org/v2.30.0/docs/formatISO and test :)

jorelosorio (changes requested):
Good work :), there were a lot of changes here. As general comment, please review that functions are used only locally belongs to that context only. I saw many were defined in utils in another place but were just internal usage in one Component, so I. suggest moving it to that context and remove from utils.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
marius-at-atb: I don't think there is a difference? https://stackoverflow.com/questions/32911424/array-length-vs-array-length-0

src/translations/dictionary.ts
marius-at-atb: Good point :+1:

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
marius-at-atb: It is part of the bugfix for https://github.com/AtB-AS/mittatb-app/pull/3736#issuecomment-1655766138

src/travel-details-screens/components/FlexibleTransportBookingOptions.tsx
marius-at-atb: Currently needed to avoid rendering this when no options show:
```
<View // 200% zoom bugfix for FlexibleTransportBookingDetails
   style={{height: Math.max(safeAreaBottom, theme.spacings.medium)}}
/>
```

src/travel-details-screens/components/FlexibleTransportBookingOptions.tsx
marius-at-atb: The total spacing between two flexBookingOptions should be medium. Half of medium is 6, for which there is no other spacing value.

src/travel-details-screens/components/TripSection.tsx
marius-at-atb: Good to keep it like this I think. Both for readability and in case the condition doesn't remain exactly like this.

src/utils/date.ts
marius-at-atb: Not supported yet, but will be in V3 it seems:
https://github.com/date-fns/date-fns/issues/3150

src/travel-details-screens/components/FlexibleTransportBookingOptions.tsx
marius-at-atb: Update: following a change in the bugfix, it is indeed no longer needed :+1:

src/travel-details-screens/components/TripSection.tsx
marius-at-atb: Yeah it could be a good idea. @gorandalum any thoughts about this?

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
marius-at-atb: It adds this gray area at the bottom, and sets it to whatever is biggest of safeAreaBottom and theme.spacings.medium:
safeArea

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
marius-at-atb: Update: fixed

src/travel-details-screens/components/TripSection.tsx
marius-at-atb: Update: following discussion with @gorandalum, leave it as is for now.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
jorelosorio: We agreed to have it on utils for future reusability.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
jorelosorio: We agreed to have it on utils for future reusability.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/hooks.ts
```diff
@@ -11,9 +11,7 @@ import {StorageModelKeysEnum} from '@atb/storage';
 import {useDebugOverride} from '@atb/debug';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import {RemoteConfigKeys} from '@atb/remote-config';
-import {useNow} from '@atb/utils/use-now';
-import {isLegFlexibleTransport} from '@atb/travel-details-screens/utils';
-import {AvailableTripPattern, TripPatternWithKey} from '../types';
+
```
jorelosorio, line 14: It might be not needed to add a new line.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -8,17 +8,16 @@ import {DateString, SearchTime} from '@atb/journey-date-picker';
 import {TripSearchPreferences, usePreferences} from '@atb/preferences';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import {useSearchHistory} from '@atb/search-history';
-import {
-  SearchStateType,
-  TripPatternWithKey,
-} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types';
+import {SearchStateType} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/types';
+
```
jorelosorio, line 12: It might be not needed to add a new line.

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,182 @@
...
+              <FlexibleTransport height={107} />
+            </View>
+          )}
+
+          <ThemeText type="heading__title" style={style.title}>
+            {t(TripDetailsTexts.flexibleTransport.contentTitle(publicCode))}
+          </ThemeText>
+          <View style={style.steps} accessibilityRole="list">
+            {
+              // eslint-disable-next-line rulesdir/translations-warning
```
jorelosorio, line 95: We could create a task to fix this lint and remove this in the future, not for this task.

src/travel-details-screens/components/TripSection.tsx
jorelosorio: Consider creating a task to fix this in the future when needed.

src/utils/date.ts
jorelosorio: Please consider creating a task to change when version is released, fewer dependencies better :)

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
marius-at-atb: That would be nice :+1:

src/utils/date.ts
marius-at-atb: https://github.com/AtB-AS/kundevendt/issues/7795

src/travel-details-screens/components/FlexibleTransportBookingDetails.tsx
```diff
@@ -0,0 +1,182 @@
...
+              TripDetailsTexts.flexibleTransport.steps.map((step, i) => {
+                const stepNumberText = `${i + 1}. `;
+                const stepInstructionText = t(step);
+                return (
+                  <View
+                    key={i}
```
gorandalum, line 101: Shouldntt use index as key. Might not give unintended side effects here, but generally it is good to not use the anti-pattern. Can just use the text as key.

src/travel-details-screens/components/TripSection.tsx
```diff
@@ -177,11 +231,45 @@ export const TripSection: React.FC<TripSectionProps> = ({
             <SituationMessageBox noStatusIcon={true} situation={situation} />
           </TripRow>
         ))}
-        {notices.map((notice) => (
-          <TripRow rowLabel={<ThemeIcon svg={Info} />}>
+        {notices.map((notice, i) => (
+          <TripRow key={i} rowLabel={<ThemeIcon svg={Info} />}>
```
gorandalum, line 235: Here using index as key might actually introduce bugs. Also using index is the default fallback that React uses, so adding it makes no difference other than removing the warning. Does the notice have an id that can be used?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -49,7 +50,7 @@ export const Results: React.FC<Props> = ({
   const [errorMessage, setErrorMessage] = useState<string>('');
   const {t} = useTranslation();
 
-  const availableTripPatterns = useAvailableTripPatterns(tripPatterns);
+  const now = useNow(2500);
```
gorandalum, line 53: Since we now need to evaluate the time in the `Results` component, this will introduce quite a bit of rerenders of what might be a long list. I think it is worth to handle this better, as we have seen some performance issues in the app lately.

Possible options are:
- Use `React.memo` so only the result items that actually needs to rerender are rerendered.
- Skip using `useNow`, which makes the booking time only evaluated at search time. Which might be good enough?

gorandalum (commented):
I think it looks really good, and I at least like this usage of utility methods where the data is needed, instead of creating new domain models.

Some comments about using index as key, and one comment regarding performance.

PR #3739, opened 2023-07-31
fix: prevent immediate triggers if useInterval is disabled

rosvik:
There was an bug in `useInterval` which would cause it to run when deps changed at the same time as it was disabled. This happened in departure details, which would `triggerImmediately` on loss of focus, while also being disabled by the same focus change.

This changes the hook to always abort if `disabled = true`.


gorandalum (approved):
Good catch! :+1:

PR #3742, opened 2023-08-01
fix: Remove "Buy ticket" button from bike journeys

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/7276


gorandalum (commented):
Could you use a little time to look into why the button appears in the first place? For me it seems really strange that it would appear, and it probably signals that there is something illogical elsewhere in the code 🤔

KristianSelnas:
@gorandalum After our discussion, should we just merge this as it is now, and revisit the entire logic when boat ticket is handled?

gorandalum (approved):

PR #3749, opened 2023-08-02
fix: Add duration to bicycle leg

KristianSelnas:
Adds duration, e.g. "Sykle i 3 minutter" to a bicycle leg, similar to what's done for walking legs.

From findings during testing of https://github.com/AtB-AS/kundevendt/issues/7198


rosvik (approved):
LGTM 👍

PR #3751, opened 2023-08-02
refactor: Harbor names from Entur in ticket purchase flow

gorandalum:
Reuse the harbors query to fetch harbor names from Entur, instead
of using Firestore.


jorelosorio (approved):
LGTM

rosvik (approved):
Looks good 👍

PR #3755, opened 2023-08-07
feat: Park'n Ride in map

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/4093

Screen shots can be seen here: https://github.com/AtB-AS/kundevendt/issues/4093#issuecomment-1669156106


rosvik (approved):
Nice! Noticed two visual nitpicks from the screenshots:

- The spacing between "Reis fra/til" buttons and "Bilparkering"/"Plasser totalt" is inconsistent.
- Is the spacing above and below the walking distance component correct? Looks like there may be more below than above.

KristianSelnas:
> Nice! Noticed two visual nitpicks from the screenshots:
> 
> * The spacing between "Reis fra/til" buttons and "Bilparkering"/"Plasser totalt" is inconsistent.
> * Is the spacing above and below the walking distance component correct? Looks like there may be more below than above.

Thanks! I'll have a look

gorandalum (approved):
LGTM :+1:

PR #3759, opened 2023-08-09
feat: nynorsk duration humanizer

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/5989.

Restructuring use of [humanize-duration](https://www.npmjs.com/package/humanize-duration) to support Nynorsk. 

Features:
- Initializing humanizer object with nynorsk languages provided
- Removing duplicate wrongly defined Bokmål and English (if these had been used, single units would have been plural e.g. 1 days / 1 minutter)
- Using initialized humanizer instead of calling "humanizeDuration" directly

Also considering adding a PR to the library so we can remove all the nynorsk specific stuff, but guess that will take some time to have approved.

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -35,32 +35,8 @@ const nynorskHumanizerObj = {
   ms: () => 'millisekund',
 };
 
-const completeLanguageObject = {
-  en: {
-    y: () => 'years',
-    mo: () => 'months',
-    w: () => 'weeks',
-    d: () => 'days',
-    h: () => 'hours',
-    m: () => 'minutes',
-    s: () => 'seconds',
-    ms: () => 'milliseconds',
-  },
-  nn: nynorskHumanizerObj,
-  no: {
-    y: () => 'år',
-    mo: () => 'måneder',
-    w: () => 'uker',
-    d: () => 'dager',
-    h: () => 'timer',
-    m: () => 'minutter',
-    s: () => 'sekunder',
-    ms: () => 'millisekunder',
-  },
-};
-
 const humanizer = humanizeDuration.humanizer({
-  languages: completeLanguageObject,
+  languages: {nn: nynorskHumanizerObj},
 });
 
 function parseIfNeeded(a: string | Date): Date {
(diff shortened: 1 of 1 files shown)


src/utils/date.ts
```diff
@@ -64,13 +78,26 @@ export function secondsToMinutesLong(
...
+const languageToHumanizerLanguage = (language: Language): HumanizerLanguage => {
+  switch (language) {
+    case Language.Norwegian:
+      return 'no';
+    case Language.English:
+      return 'en';
+    case Language.Nynorsk:
+      return 'nn';
+  }
+};
```
rosvik, line 92: Is this strictly necessary? Looks like as long as our `Language` enum resolves to a sensible two-letter language code ([ISO 639-1 to be exact!](https://github.com/EvanHahn/HumanizeDuration.js/tree/main#language)), we could pass the raw language value to the humanizer below. 

Or is there any downsides to that?

rosvik:
Also: +1 for making a PR with the nynorsk translation to the library 😊

src/utils/date.ts
toretefre: I tried passing the language prop directly in `secondsToDuration` and `getHumanizer`, it ended with `No language found` error. 

I think it might be because language in our case might be defined as `nb`, while [the library](https://github.com/EvanHahn/HumanizeDuration.js/blob/dcd80f80a380f3bd3bb026927a7a414ceefc76f3/humanize-duration.js#L902) and ISO 639-1 seems to expect `no` for bokmål. Potentially an obstacle for having a PR to the library 😄 

IMG_5739

rosvik: Aha, "macrolanguages" are of course a thing 😆 

Screenshot 2023-08-09 at 15 12 16

rosvik (approved):

src/utils/date.ts
toretefre: Of course, this is why we can't have nice things! 😄

HanneLH:
> @HanneLH Should we include this in the release? ☝️

Faktisk litt usikker, er det noe vi må teste for AtB da?
Pr. nå skal vi ikke slå på nynorsk for AtB, så vist dette er bare noe i bakgrunnen så tenker jeg ja.

rosvik:
@HanneLH Jepp, vi bør kjapt dobbelsjekke at tidspunkt fungerer som før: f.eks. "12 minutter" på bokmål og "12 minutes" på engelsk.

HanneLH:
@toretefre / @rosvik 
Høres ut som en enkel jobb, bare kjør på i 1.40 branchen 😄

PR #3761, opened 2023-08-09
feat: Handle cost free tickets

jorelosorio:
NOTE: Before review this, please have a look at this one first: https://github.com/AtB-AS/fare-product-type-configs/pull/20

Closes: https://github.com/AtB-AS/kundevendt/issues/7194

Screenshot_1691581471


src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
```diff
@@ -142,6 +143,11 @@ export function useOfferState(
 
   const updateOffer = useCallback(
     async function (cancelToken?: CancelToken) {
+      if (skip) {
```
jorelosorio, line 146: This is basically is to skip request for prices when we know that is already free. 
A better solution? :) is welcome.

src/api/types/generated/fragments/stop-places.ts
```diff
@@ -3,4 +3,5 @@ export type StopPlaceFragment = {
   name: string;
   latitude?: number;
   longitude?: number;
+  isFree?: boolean;
```
gorandalum, line 6: These fragments shouldn't be manually edited, as they are one-to-one with the generated fragments in the BFF.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
gorandalum: I think it is ok. Many other hooks have a flag named `disabled`, would that also be appropriate naming here?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -34,6 +35,12 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
 }) => {
   const styles = useStyles();
   const {t, language} = useTranslation();
+  const harborConnections = useHarbors(params.fromPlace?.id);
+
+  const toPlaceConnection = harborConnections.data.find(
+    (harbor) => harbor.id === params.toPlace?.id,
+  );
+  const isFree = !!toPlaceConnection?.isFree;
```
gorandalum, line 43: The `isFree` flag isn't already on the `params.toPlace`? If not, and this is necessary, I think it would be appropriate to move it out into a `useIsFree` hook or something.

src/api/types/generated/fragments/stop-places.ts
jorelosorio: I made a fix for this, good observation :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
jorelosorio: It was moved to a useIsFree hook

gorandalum: I don't fully understand why this part is necessary, since harbor connections returned from `useHarbors` should now already have the `isFree` flag 🤔 Could you take an extra look, and just contact if we should do a video session.

jorelosorio: Hi, let's have a talk ;)

jorelosorio: I made a refactor to the code after a talk in Slack ;)

gorandalum (approved):
:+1:

PR #3763, opened 2023-08-11
feat: add dynamic global messages to trip details

rosvik:
follow up to https://github.com/AtB-AS/kundevendt/issues/4054

Adds global messages to trip details, and replaces two messages:

> With a single or periodic ticket for zone A from AtB, you can travel by train within zone A.

> This journey requires a ticket that is not available from this app, or must be purchased from a provider other than AtB.

- Added logic to fetch which Zones the whole trip is within, which is then used by rule in global messages to show one or the other of the above messages. 

I hoped I could remove the `canSellCollabTicket` utility function, but it's still being used by the from travel search to ticketing feature.


rosvik:
## Acceptance criteria

- [ ] Two global messages have been added to staging in firestore: "CollabTicketTrips" and "UnavailableTicketTrips". These should work just like the hard coded messages they replaced in trip details. If they work, they can be added to production for v1.41.

gorandalum (approved):

tormoseng:
@rosvik Is checking the `appMinVersion` that didn't work for me

rosvik:
@tormoseng Fixed it 🙈 https://github.com/AtB-AS/docs-private/commit/6ac72eb16fe021e1efd1867d4b51092d342f8138

tormoseng:
Tested OK!

PR #3769, opened 2023-08-14
feat: Upgrade TouchableOpacity to Pressables

ingvisim:
Upgrade TouchableOpacity in app to use Pressable, as it is more future proof, preferred by React Native, and removes warning about "Excessive number of pending callbacks".

From React native docs on TouchableOpacity:

> If you're looking for a more extensive and future-proof way to handle touch-based input, check out the [Pressable](https://reactnative.dev/docs/pressable) API.

https://reactnative.dev/docs/touchableopacity


rosvik (commented):
Did some testing locally now, and noticed two issues:


1. The "Select in map" page is broken:



2. The hitbox for the tabs the ticketing screen is smaller than before ("Buy" / "Active"):

gorandalum (changes requested):
Looks generally good! Just take a look at all added `containerStyle` parameters on shared components. I think we should try to avoid those, and make due with just `style` like before.

gorandalum (approved):
Looks good! Remember to add labels and write acceptance criteria :+1:

PR #3771, opened 2023-08-16
feat: added repeat purchase of boat tickets

hanne-at-atb:
Added pointToPointValidity for to/from harbors to recentFareContract, allowing repeat purchase of boat tickets including pre-filling harbors.
Added orderId to recentFareContract to be able to get directionality of boat trip.

Closes: https://github.com/AtB-AS/kundevendt/issues/7398

image


jorelosorio (approved):
Good work 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
```diff
@@ -117,18 +122,33 @@ export const TicketTabNav_PurchaseTabScreen = ({navigation}: Props) => {
     analytics.logEvent('Ticketing', 'Recently used fare product selected', {
       type: fareProductTypeConfig.type,
     });
+    const getPlace = (
+      harbor: string | undefined,
```
gorandalum, line 126: Regarding naming, I think it is too early to name it `harbor` here. There is nothing in the domain model that specifies that `fromPlace`/`toPlace` on `pointToPointValidity` are harbors.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/types.ts
```diff
@@ -10,5 +14,7 @@ export type RecentFareContract = {
   preassignedFareProduct: PreassignedFareProduct;
   fromTariffZone?: TariffZone;
   toTariffZone?: TariffZone;
+  orderId?: string;
```
gorandalum, line 17: Is `orderId` an optional field? 🤔

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/Components/RecentFareContracts/RecentFareContractComponent.tsx
```diff
@@ -74,26 +85,55 @@ export const RecentFareContractComponent = ({
...
+            )
+          : t(
+              RecentFareContractsTexts.a11yPreLabels.harbors.oneWayTrip(
+                fromName,
+                toName,
+              ),
+            );
+      }
+      return '';
+    };
```
gorandalum, line 122: In recent times I have become less of a fan of creating these functions during the component render, since that means it is recreated on every render. Creating functions during render (without `useCallback`) can lead to errors if used incorrectly, like in a `useEffect`.

I think it is better to either:
- Take it out of the component render as it's own function someplace else in the file.
- Just calculate the `harborInfo` during render instead of a reactive function for it.
- If the calculation of `harborInfo` is heavy and might give performance issues when done on every render, then wrap it with `useMemo`.

And after writing this and reading the code one more time I see that the `harborInfo` function is actually inside the `returnAccessibilityLabel` function. Then my comments on `harborInfo` doesn't really apply, but I still leave them here as they still apply for the `returnAccessibilityLabel` function. Not something you need to fix now.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/use-recent-fare-contracts.ts
```diff
@@ -125,17 +125,33 @@ const mapBackendRecentFareContracts = (
     recentFareContract.zones.slice(-1)[0],
   );
 
+  const pointToPointValidity = recentFareContract.point_to_point_validity;
+
+  const orderId = recentFareContract.order_id;
+
+  const fromId =
+    pointToPointValidity?.fromPlace !== undefined
```
gorandalum, line 133: Maybe `fromId = pointToPointValidity?.fromPlace || fromTariffZone?.id`. Or is there a reason for the explicit check on `undefined`?

gorandalum (approved):
Looks good! Some comments/personal opinions on technical stuff, but no blockers for merge 😊

PR #3774, opened 2023-08-17
feat: Fare product groups

KristianSelnas:
Splits fare contracts into categories (groups), as specified in https://github.com/AtB-AS/kundevendt/issues/4306

Groups are defined in config, see https://github.com/AtB-AS/firestore-configuration/pull/248

If no groups are defined in config, all products are shown as before, with "Alle billetter" as heading. Products not explicitly added to a group will appear under the heading "Andre billetter".


src/components/text/ThemeText.tsx
```diff
@@ -12,7 +12,7 @@ import {
 } from '@atb/theme/colors';
 import {ContrastColor} from '@atb-as/theme';
 
-type ColorType = TextColor | StaticColor | ContrastColor;
+export type ColorType = TextColor | StaticColor | ContrastColor;
```
gorandalum, line 15: Maybe not export this, and just use `TextColor` directly in `TransportModes`? The `ColorType` is probably broader than what we want `TransportModes` to support, and includes stuff like background themes (like `background_accent_1`) and message box status colors (like `info` and `warning`).

src/components/transportation-modes/TransportModes.tsx
```diff
@@ -40,28 +42,42 @@ export const getTransportModeText = (
...
   if (modesCount > modesDisplayLimit) {
     return t(FareContractTexts.transportModes.multipleTravelModes);
   }
-  return modes
-    .map((tm) => t(FareContractTexts.transportMode(tm.mode, tm.subMode)))
-    .filter(removeDuplicateStringsFilter)
-    .join('/');
+  if (unknownModeText && modes.map((m) => m.mode).includes('unknown')) {
+    return unknownModeText;
+  }
```
gorandalum, line 55: Is this handling of `unknownModeText` intuitive? I kinda feel if we somehow get `rail` and `unknown` as transport modes, we would kinda like to still show the train icon and train text? I know it makes sense in the context it is used now, but maybe not if we think of `TransportModes` as a shared generic component?

src/components/transportation-modes/TransportModes.tsx
```diff
@@ -40,28 +42,42 @@ export const getTransportModeText = (
...
 };
 
 export const TransportModes = ({
   modes,
   iconSize,
   disabled,
+  textType,
+  textColor,
+  unknownModeText,
+  useUnknownIcon = true,
```
gorandalum, line 71: Could it be an alternative to just not set any `transportModes` on the group that should be "Andre billettyper", and have this handling outside of `TransportModes` component? 

Then it would be something like this:
```ts
fareProductGroup.transportModes ? <TransportModes ...> : <ThemeText>Andre billettyper<ThemeText/>
```
in the `FareProductGroup` component.

It would reduce the complexity a bit in the shared `TransportModes` component, and we probably wouldn't need this `unknownModeText` and `useUnknownIcon` params.

gorandalum (approved):
Looks good! I have some comments which is more on a detailed level, pick and choose what you think makes sense and not 😊

KristianSelnas:
@gorandalum I've fixed the stuff you commented on in your previous review. Do you want to look through the changes?

gorandalum (approved):

PR #3775, opened 2023-08-18
fix: boat ticket screen reader focus fix

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/7367

The ref used to focus the screen reader when returning from a boat harbor selection, is now set according to which selection you just set (from or to).

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/HarborSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/HarborSelection.tsx
@@ -2,7 +2,7 @@ import {ThemeText} from '@atb/components/text';
 import {FareProductTypeConfig} from '@atb/configuration';
 import {StyleSheet} from '@atb/theme';
 import {PurchaseOverviewTexts, useTranslation} from '@atb/translations';
-import React, {ForwardedRef, forwardRef, useRef} from 'react';
+import React, {forwardRef, useRef} from 'react';
 import {StyleProp, TouchableOpacity, View, ViewStyle} from 'react-native';
 import {GenericClickableSectionItem, Section} from '@atb/components/sections';
 import {PreassignedFareProduct} from '@atb/reference-data/types';
@@ -17,6 +17,9 @@ type StopPlaceSelectionProps = {
   onSelect: (params: Root_PurchaseHarborSearchScreenParams) => void;
   style?: StyleProp<ViewStyle>;
 };
+
+const FROM = 'from';
+const TO = 'to';
 
 export const HarborSelection = forwardRef<
   TouchableOpacity,
@@ -53,12 +56,12 @@ export const HarborSelection = forwardRef<
         </ThemeText>
 
         <Section accessible={false}>
-          {(['from', 'to'] as ('from' | 'to')[]).map((fromOrTo) => {
-            const harbor = fromOrTo === 'from' ? fromHarbor : toHarbor;
+          {([FROM, TO] as (typeof FROM | typeof TO)[]).map((fromOrTo) => {
+            const harbor = fromOrTo === FROM ? fromHarbor : toHarbor;
             const harborStyle =
-              fromOrTo === 'from' ? styles.fromHarbor : styles.toHarbor;
-            const disabled = fromOrTo === 'to' && !fromHarbor;
-            const ref = fromOrTo == 'from' ? fromRef : toRef;
+              fromOrTo === FROM ? styles.fromHarbor : styles.toHarbor;
+            const disabled = fromOrTo === TO && !fromHarbor;
+            const ref = fromOrTo == FROM ? fromRef : toRef;
             return (
               <GenericClickableSectionItem
                 key={fromOrTo}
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/HarborSelection.tsx
```diff
@@ -48,78 +54,65 @@ export const HarborSelection = forwardRef<
...
-              })
-            }
-            testID="selectHarborsButton"
-          >
-            <View style={styles.toHarbor}>
-              <ThemeText
-                color={!!fromHarbor ? 'secondary' : 'disabled'}
-                type="body__secondary"
-                style={styles.toFromLabel}
+          {([FROM, TO] as (typeof FROM | typeof TO)[]).map((fromOrTo) => {
```
jorelosorio, line 59: As alternative, you could create an internal component called `HarborSectionItem` that get as parameter a style, and if is disabled or not. The call it is twice like: 
```
<HarborSectionItem style={styles.fromHarbor}>
<HarborSectionItem style={styles.toHarbor} enabled={!fromHarbor}>
```

it could be internally to do not lost the context.

jorelosorio (changes requested):

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/HarborSelection.tsx
marius-at-atb: So if we are to introduce HarborSelectionItem, I think it should still do some internal logic regarding FROM and TO. Then it would be much more similar to the content of the .map function, though without the array to map

marius-at-atb: Here is something in between. Perhaps the most readable solution:


```

<HarborSelectionItem
  ref={harborInputSectionItemRef}
  fromOrTo={FROM}
  harbor={fromHarbor}
  disabled={false}
  onPress={() => {
    onSelect({
      fareProductTypeConfig,
      preassignedFareProduct,
    });
  }}
/>
<HarborSelectionItem
  ref={harborInputSectionItemRef}
  fromOrTo={TO}
  harbor={toHarbor}
  disabled={!fromHarbor}
  onPress={() => {
    fromHarbor &&
      onSelect({
        fromHarbor,
        fareProductTypeConfig,
        preassignedFareProduct,
      });
  }}
/>
          
          
          
--------
          
type HarborSelectionItemProps = {
  harbor?: StopPlaceFragment;
  disabled: boolean;
  onPress: () => void;
  fromOrTo: typeof FROM | typeof TO;
};

const HarborSelectionItem = forwardRef<
  TouchableOpacity,
  HarborSelectionItemProps
>(
  (
    {harbor, onPress, disabled, fromOrTo}: HarborSelectionItemProps,
    harborInputSectionItemRef,
  ) => {
    const itemRef = useRef<TouchableOpacity>(null);
    const {t} = useTranslation();
    const styles = useStyles();

    return (
      <GenericClickableSectionItem
        ref={itemRef}
        accessibilityState={{disabled}}
        accessible={true}
        accessibilityRole="button"
        accessibilityLabel={t(
          PurchaseOverviewTexts.stopPlaces.harborSelection[fromOrTo].a11yLabel(
            harbor?.name,
          ),
        )}
        accessibilityHint={
          disabled
            ? undefined
            : t(
                PurchaseOverviewTexts.stopPlaces.harborSelection[fromOrTo]
                  .a11yHint,
              )
        }
        onPress={() => {
          onPress();
          if (
            harborInputSectionItemRef &&
            typeof harborInputSectionItemRef !== 'function'
          ) {
            harborInputSectionItemRef.current = itemRef.current;
          }
        }}
        testID="selectHarborsButton"
      >
        <>
          <View style={fromOrTo === FROM ? styles.fromHarbor : styles.toHarbor}>
            <ThemeText
              color={disabled ? 'disabled' : 'secondary'}
              type="body__secondary"
              style={styles.toFromLabel}
            >
              {t(PurchaseOverviewTexts.fromToLabel[fromOrTo])}
            </ThemeText>
            <HarborLabel harbor={harbor} disabled={disabled} />
          </View>
        </>
      </GenericClickableSectionItem>
    );
  },
);
          ```

**Review by jorelosorio — APPROVED** [review]
Good work :)

**Comment by marius-at-atb** [comment]
Thanks :)

File: src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/HarborSelection.tsx [hunk 2]
```diff
@@ -18,6 +18,9 @@ type StopPlaceSelectionProps = {
   style?: StyleProp<ViewStyle>;
 };
 
+const FROM = 'from';
+const TO = 'to';
```
gorandalum, line 22: In situations like this, if we need a type, these should rather be `type fromOrTo = 'from' | 'to'` or maybe `type direction = 'from' | 'to'`.

In this case where it is used only one placed, I think I would just inline it in the props to the component, like `fromOrTo: 'from' | 'to'`.

gorandalum (changes requested):
I think we are mixing strategies for how to give focus, and that makes it more tricky to follow the flow.

When implementing the focus flow some weeks ago we said that it would be controlled by a parameter from the harbor selection screen, and not done by local state/logic inside purchase overview screen. This was to make them more decoupled from eachother.

But now we are introducing local state/logic further down in the stack to remember which component to focus. I think it would be preferable to just let the parameter decide it, instead of mixing this parameter-driven and internal-logic-driven ways of solving it.

marius-at-atb:
@gorandalum True. See https://github.com/AtB-AS/mittatb-app/pull/3779

PR #3778, opened 2023-08-21
feat: Mobility ticket bundling

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/3869 and https://github.com/AtB-AS/kundevendt/issues/3976

When a user is eligible for a mobility operator benefit (free unlock, free use etc), this will be shown in the corresponding bottom sheet and the action button should perform the action set up in Firestore config for that benefit. Such action (url) could potentially include a value code, e.g. app://voi.com?valueCode=123

Simulator Screenshot - iPhone 14 - 2023-08-21 at 15 13 42


src/mobility/use-benefits.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<Error>();
+
+  useEffect(() => {
+    if (operatorId) {
+      setLoading(true);
+      Promise.all([
+        getBenefits()
+          .then((benefits) => benefits.find((b) => b.operator === operatorId))
+          .then((benefits) => setUserBenefits(benefits?.benefits ?? [])),
```
gorandalum, line 23: I'm a little confused by the naming here. We got an `benefits` array, and after finding one item in the array it is still called `benefits`, and then `benefits.benefits` is called. So that makes it three layers of `benefits`? 😅 Is it possible to differentiate about what they are, or what level they are on?

src/mobility/use-operators.tsx
```diff
@@ -4,7 +4,7 @@ import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
 export const useOperators = () => {
   const {mobilityOperators} = useFirestoreConfiguration();
 
-  return (types: undefined | FormFactor | FormFactor[]) => {
+  const byFormFactor = (types: undefined | FormFactor | FormFactor[]) => {
```
gorandalum, line 7: I think we might want to start to always wrap methods returned in hooks with `useCallback`. After the "freeze-research" I have become more aware of how many renders we have and how easy it is to do things wrong, even though the current usage doesn't demand `useCallback`.

This was more of a general comment and not really necessary to change now, as we probably have quite a lot of hooks not enforcing this.

src/mobility/components/PricingPlan.tsx
```diff
@@ -69,6 +82,11 @@ const PriceInfo = ({price, pricingSegment, unit}: PriceInfoProps) => {
     <VehicleStat
       primaryStat={`${formatPrice(pricingSegment, language)} kr/${unit}`}
       secondaryStat={t(ScooterTexts.pricingPlan.price(price))}
+      secondaryStatStyle={
+        price > 0 && eligibleBenefits?.includes('free-unlock')
+          ? {textDecorationLine: 'line-through'}
+          : undefined
+      }
```
gorandalum, line 89: My slight preference is to encapsulate the lower level styling into the component, and rather use a variable to switch between styles. Would be something like `isFree` in this case. I feel it is easier to enforce the design system in the code that way.

src/mobility/components/ScooterSheet.tsx
```diff
@@ -30,18 +37,39 @@ type Props = {
...
     rentalAppUri,
   });
 
+  // The data model handles multiple benefits per operator,
+  // but we currently know there is only one,
+  // and the UI has to change anyway to support an undetermined number of benefits.
+  const operatorBenefit =
+    operatorBenefits && operatorBenefits.length > 0
+      ? operatorBenefits[0]
+      : undefined;
```
gorandalum, line 71: Can just be `operatorBenefits?.[0]`.

gorandalum (approved):
Looks good! Hope we can get bundling up and running soon!

Some comments which are more sort of stating my preferences and views on some stuff, which you are completely free to not feel the same about 😄

PR #3779, opened 2023-08-21
fix: refactor screen reader focus logic

marius-at-atb:
Refactoring the [original implementation](https://github.com/AtB-AS/mittatb-app/pull/3775) to use logic with params instead of ref overwriting.

Discussion on useImperativeHandle vs forwarding multiple refs:
https://stackoverflow.com/questions/53561913/react-forwarding-multiple-refs

closes https://github.com/AtB-AS/kundevendt/issues/7367


src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -94,11 +100,15 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
...
   useEffect(() => {
     if (params.onFocusElement === 'from-to-selection') {
-      giveFocus(fromToInputSectionItemRef);
+      giveFocus(fromToInputSectionItemRef.current?.fromToRef);
+    } else if (params.onFocusElement === 'from-selection') {
+      giveFocus(fromToInputSectionItemRef.current?.fromRef);
+    } else if (params.onFocusElement === 'to-selection') {
+      giveFocus(fromToInputSectionItemRef.current?.toRef);
     }
   }, [params.onFocusElement]);
```
jorelosorio, line 113: This new way of tracking ref in a better way, loosk great, however if this is gonna be used on different places, I will suggest then encapsulate it in a hook, like by example:

```ts
type FocusElement = {
  focus: string;
  callback: () => void;
}

const useFocusHandler = (onFocus: string) => {
  const [elements, setElements] = useState<FocusElement[]>([]);
  const addFocus = (focus: string, callback: () => void) => {
    elements.push({focus, callback});
  };

  useEffect(() => {
      const element = elements.find((e) => e.focus === onFocus);
      element?.callback();
  }, [onFocus]);

  return {
    addFocus,
  }
}
```

It is just an example, the main idea could be also having the references for the ref and make it reusable as much as possible.

marius-at-atb: Extracting it to a reusable hook is a very interesting idea! I'll look into it when I'm back on Friday

marius-at-atb: The hook can be as simple as this:
```
export type FocusRefsType = {
  [key: string]: RefObject<TouchableOpacity>;
};

export function useFocusRefs(focusOnElementName: string | undefined) {
  const ref = useRef<FocusRefsType>(null);

  useEffect(
    () => giveFocus(ref.current?.[focusOnElementName + 'Ref' || '']),
    [focusOnElementName],
  );

  return ref;
}
```

marius-at-atb: Usage would then be simplified from
```
const fromToInputSectionItemRef = useRef<FromToRefType>(null);

useEffect(() => {
  if (params.onFocusElement === 'from-to-selection') {
    giveFocus(fromToInputSectionItemRef.current?.zonesRef);
  } else if (params.onFocusElement === 'from-selection') {
    giveFocus(fromToInputSectionItemRef.current?.fromHarborRef);
  } else if (params.onFocusElement === 'to-selection') {
    giveFocus(fromToInputSectionItemRef.current?.toHarborRef);
  }
}, [params.onFocusElement]);
```
to just 
`const focusRefs = useFocusRefs(params.onFocusElement);`

gorandalum (approved):

marius-at-atb:
@gorandalum @jorelosorio useFocusRefs hook introduced

jorelosorio (approved):
Nice and clean :) good work!

PR #3781, opened 2023-08-22
fix: Fix translations for FRAM

adriansberg:
- Fix a couple of words in some translations
- Don't render summary if the text is empty. This is to not show the text regarding a ticket being valid for bus/tram.

Closes https://github.com/AtB-AS/kundevendt/issues/7804


mortennordseth (approved):
Looks good 👍

PR #3782, opened 2023-08-23
chore: Fix flipper in CI

gorandalum:
Solution based on:
https://reactnative.dev/docs/next/speeding-ci-builds#:~:text=Flipper%20is%20a%20debugging%20tool,built%20in%20the%20CI%20environment.

https://dev.to/retyui/disable-flipper-on-ci-builds-for-react-native-complete-guide-bnd

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -30,7 +30,9 @@ target 'app' do
     #
     # Note that if you have use_frameworks! enabled, Flipper will not work and
     # you should disable the next line.
-    :flipper_configuration => FlipperConfiguration.enabled,
+    :flipper_configuration => ENV['CI'] ?
+        FlipperConfiguration.disabled :
+        FlipperConfiguration.enabled,
     # An absolute path to your application root.
     :app_path => "#{Pod::Config.instance.installation_root}/.."
   )
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -4,4 +4,9 @@ module.exports = {
     android: {},
   },
   assets: ['./assets/fonts'],
+  dependencies: {
+    ...(process.env.CI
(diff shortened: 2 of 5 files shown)


react-native.config.js
```diff
@@ -4,4 +4,9 @@ module.exports = {
     android: {},
   },
   assets: ['./assets/fonts'],
+  dependencies: {
+    ...(process.env.CI
+      ? {'react-native-flipper': {platforms: {ios: null}}}
```
rosvik, line 9: `ios: null` means it's disabled for iOS and enabled on android?

rosvik (commented):
LGTM 👍

rosvik (approved):

react-native.config.js
jorelosorio: Seems like yes, but would be nice to have also on iOS as well.

ios/Podfile
```diff
@@ -30,7 +30,9 @@ target 'app' do
     #
     # Note that if you have use_frameworks! enabled, Flipper will not work and
     # you should disable the next line.
-    :flipper_configuration => FlipperConfiguration.enabled,
+    :flipper_configuration => ENV['CI'] ?
+        FlipperConfiguration.disabled :
+        FlipperConfiguration.enabled,
```
jorelosorio, line 35: It could be also configured to enabled on staging app? as tests? But for that we need to enable  `FB_SONARKIT_ENABLED`

jorelosorio (approved):

react-native.config.js
gorandalum: Yeah. As I've understood it is disabled in Android as default.

PR #3783, opened 2023-08-23
feat: Apple Pay interface suppression

mortennordseth:
Closes https://github.com/AtB-AS/kundevendt/issues/7215

This PR adds new functionality to suppress the Apple Pay interface when the fare contract details screen is shown. To use this functionality, permission from Apple must be granted, which currently only applies to fram-debug and fram-production. To prevent any errors for environments without permission, a Fastlane command has been created to update the entitlements file with the entitlement. This command is used only for FRAM in the build-store-ios workflow. Although this capability could be added to the staging workflow, we will have to wait for permission. If you want to test locally, you can manually add the entitlement to the entitlements file:

```
<key>com.apple.developer.passkit.pass-presentation-suppression</key>
<true/>
```
Adding the entitlement to the entitlements file without having permission to use it from Apple will result in a build error. 

### Acceptance criteria
- [ ] The app signing process works as before for provisioning profiles that do or do not have permission to use the entitlement
- [ ] Apple Pay interface is suppressed when QR-code is shown for FRAM in production
- [ ] Apple Pay interface suppression ends when fare contract details screen is closed


src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
```diff
@@ -26,6 +27,13 @@ export function Root_FareContractDetailsScreen({navigation, route}: Props) {
   const firstTravelRight = fc?.travelRights[0];
   const {t} = useTranslation();
 
+  useEffect(() => {
+    RCTPassPresentation.requestAutomaticPassPresentationSuppression();
+    return () => {
+      RCTPassPresentation.endAutomaticPassPresentationSuppression();
+    };
+  }, []);
```
mikaelbr, line 35: I think this could be defined as a custom hook and placed inside the `suppress-pass-presentation` module.

mikaelbr (changes requested):
Fantastic solution. Only comment is abstracting the suppression hook some more. But other than that I think this is brilliant!

.github/workflows/build-store-ios.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       - name: Disable widget
         if: ${{ env.ENABLE_WIDGET != 'true' }}
         run: bundle exec configure_extensions remove ios/atb.xcodeproj app departureWidget AtbAppIntent
+      - name: Enable apple pass suppression
+        if: matrix.org == 'fram'
```
jorelosorio, line 45: Here as a small improvement we could do the same as `if: ${{ env.ENABLE_WIDGET != 'true' }}` defining a .env var with this is more dynamic in a way if in ATB we want to disable this in the future.

jorelosorio (approved):
Good work :), small comment for future improvements

.github/workflows/build-store-ios.yml
mortennordseth: I agree! I've added `ENABLE_APPLE_PASS_SUPPRESSION` as an environment variable. I sat it to false for all environments except FRAM production.

src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
mortennordseth: Good catch 👍

jorelosorio (approved):
Good work :)

gorandalum (approved):

mikaelbr (approved):

PR #3784, opened 2023-08-23
chore: Upgrades React Native to 0.72.4

mikaelbr:
Upgrades to React Native 0.72.4 and thus Hermes engine. This also removes unnecessary dependency react-native-reanimated which blocked upgrading.

- Upgrades dependencies for React Native
- Upgrades build tooling to support the latest React Native
- Does not do general dependency upgrades other than making it work with the latest RN.
- Removes react-native-reanimated

Fixes https://github.com/AtB-AS/kundevendt/issues/7802
Fixes https://github.com/AtB-AS/kundevendt/issues/7772

## Suggested review strategy

- [ ] Test locally running both Android and iOS
- [ ] Try clean builds


package.json
```diff
@@ -130,35 +132,38 @@
...
     "react-native-dotenv": "^3.4.7",
     "react-native-flipper-performance-plugin": "^0.4.0",
     "react-test-renderer": "^18.2.0",
-    "ts-jest": "^29.0.5",
-    "typescript": "^4.9.4"
+    "ts-jest": "^29.1.1",
+    "typescript": "^5.1.6"
+  },
+  "engines": {
+    "node": ">=16"
```
gorandalum, line 167: We changed to node 18 on the builds, should it be the same here?

ios/Podfile
```diff
@@ -1,8 +1,29 @@
...
+# If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.
+# because `react-native-flipper` depends on (FlipperKit,...) that will be excluded
+#
+# To fix this you can also exclude `react-native-flipper` using a `react-native.config.js`
+# ```js
+# module.exports = {
+#   dependencies: {
+#     ...(process.env.NO_FLIPPER ? { 'react-native-flipper': { platforms: { ios: null } } } : {}),
+# ```
+flipper_config = ENV['NO_FLIPPER'] == "1" ? FlipperConfiguration.disabled : FlipperConfiguration.enabled
```
gorandalum, line 20: When changing this logic to be based on the env variable `NO_FLIPPER` instead of `CI`, I think our build script need to set `NO_FLIPPER=1` before running pod install.

Also the logic in `react-native.config.js` (mentioned some line above here) should be changed to be dependant on the `NO_FLIPPER` variable instead of `CI`.

gorandalum (commented):
A couple comments while I test some locally.

android/app/src/debug/java/no/mittatb/ReactNativeFlipper.java
```diff
@@ -26,13 +26,16 @@
...
+ * Class responsible of loading Flipper inside your React Native application. This is the debug
+ * flavor of it. Here you can add your own plugins and customize the Flipper setup.
+ */
 public class ReactNativeFlipper {
   public static void initializeFlipper(Context context, ReactInstanceManager reactInstanceManager) {
     if (FlipperUtils.shouldEnableFlipper(context)) {
       final FlipperClient client = AndroidFlipperClient.getInstance(context);
 
       client.addPlugin(new InspectorFlipperPlugin(context, DescriptorMapping.withDefaults()));
-      client.addPlugin(new ReactFlipperPlugin());
```
jorelosorio, line 35: This one was added by @gorandalum if I am not wrong.

android/build.gradle
```diff
@@ -2,30 +2,24 @@
...
 
-        if (System.properties['os.arch'] == "aarch64") {
-            // For M1 Users we need to use the NDK 24 which added support for aarch64
-            ndkVersion = "24.0.8215888"
-        } else {
-            // Otherwise we default to the side-by-side NDK version from AGP.
-            ndkVersion = "21.4.7075529"
-        }
+        // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
+        ndkVersion = "23.1.7779620"
```
jorelosorio, line 14: Why not then the latest?

jorelosorio (approved):

android/build.gradle
mikaelbr: Good question! This is copied from RN migration. I don't know how tightly coupled ffi and support here is and if using e.g. `25.2.9519653` has some negative impact 🤔

android/app/src/debug/java/no/mittatb/ReactNativeFlipper.java
mikaelbr: I think this is moved to be handled by the flipper client base, so it's not necessary to include anymore 🤔

mikaelbr:
Hmm. The gemfile look relevant at least 🤔 Don't know why I don't get them though. Can add that after the fact if more get them also.

gorandalum (commented):
Looks good here! Tried on iOS simulator, iOS device, Android emulator and Android device. Also works with Flipper on both.

PR #3786, opened 2023-08-23
feat: Add toggle button for product description

adriansberg:
Added a toggle button, same (similar) as the one used for traveller selection previously (removed in https://github.com/AtB-AS/mittatb-app/pull/3657). The button is added only for products with `productSelectionMode: product`, which is only a few of the products. This is mainly affecting FRAM which has a couple of products of this kind, AtB/NFK mostly has the night ticket.

image

2023-08-23 at 22 55 41@2x

Fixes https://github.com/AtB-AS/kundevendt/issues/4223


mortennordseth (approved):
A nice improvement for FRAM 🎉 This looks good to me.

PR #3787, opened 2023-08-24
feat: Harbor connection override

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/7365

This PR uses `harborConnectionOverrides` from Firestore to add additional connection between harbors to those returned by EnTur.

We search for harbor connections using a Trip serarch, and that does not give us connections across multiple lines. These additional connections need to be added to Firestore.

Example: 
From Sætervika Hurtigbåtkai (NSR:StopPlace:42867) an EnTur trip search returns two possible connections:
- NSR:StopPlace:42003: Sandviksberget hurtigbåtkai
- NSR:StopPlace:41641: Ramsøyaunet hurtigbåtkai

From Sætervika it should also be possible to buy tickets to a few other stop places, including
- NSR:StopPlace:41640: Ramsøy hurtigbåtkai
- NSR:StopPlace:41639: Hepsøy hurtigbåtkai

Getting these harbors as options can be achieved by adding the additional stops to Firestore:
- `{ from: "NSR:StopPlace:42867", to: "NSR:StopPlace:41640" }`
- `{ from: "NSR:StopPlace:42867", to: "NSR:StopPlace:41639" }`


src/harbors/use-harbors.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [isSuccess, setIsSuccess] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [error, setError] = useState<unknown>();
+  const [data, setData] = useState<StopPlaceFragment[]>([]);
+
+  const harbors = useHarborsQuery();
+  const connections = useHarborsQuery(fromHarborId);
+  const overrides = useHarborConnectionOverrides(fromHarborId);
+
+  useEffect(() => {
```
jorelosorio, line 20: I have seen this conditions before, outside an useEffect like by example: https://github.com/AtB-AS/mittatb-app/blob/3e23d045775b8cc76a6cba92a87a7f18b08d2263/src/components/map/components/DeparturesDialogSheet.tsx#L71 In this way we can reduce the use of effects?

src/harbors/use-harbors.tsx
```diff
@@ -0,0 +1,76 @@
...
+import _ from 'lodash';
+
+export const useHarbors = (fromHarborId?: string) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [isSuccess, setIsSuccess] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [error, setError] = useState<unknown>();
+  const [data, setData] = useState<StopPlaceFragment[]>([]);
+
+  const harbors = useHarborsQuery();
```
jorelosorio, line 16: Why not just have all in our remote config instead of add overrides if something is missing? I harder to maintain to know what is missed on Entur and add those in another place.

MHO

jorelosorio (approved):
Good work, I just left a couple of comments that might improve the code as it is now, In general looks good. I do not have strong opinions on how to use the `useEffects` but as I mentioned there are other places that `useEffect` is used less.

src/harbors/use-harbors.tsx
KristianSelnas: Well, that's the idea here. The connections defined in Firestore are added to the ones we receive from Entur

gorandalum: I can expand a little on this. In general we have three options:
- **Always use overrides just from Firestore and not query Entur**
If a transport company (other than AtB) has a more "logical" boat-ticket setup, and they don't need tickets to span across different boat lines, then I think they would appriciate now having to hassle with adding the possible overrides to Firestore. For them Entur is good enough.

- **Use overrides in Firestore if they exist, if else use connections from Entur**
Could be done, and would maybe be more clear-cut logic. 

- **Use overrides in Firestore if they exist, and merge them with connections from Entur**
A little more complex logic, with the added benefit that if new harbors are added to Entur, it will still work even though somebody forgot to notify the development team.

I discussed this quickly with Kristian, and from my view I think both of the two last options are good, and don't really have a personal preference.

gorandalum (changes requested):
Some comments to clean up the `useEffect` usage.

I also have a feeling that splitting the `useHarborsQuery`  into something like `useAllHarborsQuery` and `useHarborConnectionsQuery`, and then do the overrides-specific code in a wrapper around just `useHarborConnectionsQuery` would be slightly cleaner.

src/harbors/use-harbors.tsx
KristianSelnas: It's the latter that's done in this PR on line 48 in `use-harbors.tsx`

KristianSelnas:
> Some comments to clean up the `useEffect` usage.
> 
> I also have a feeling that splitting the `useHarborsQuery` into something like `useAllHarborsQuery` and `useHarborConnectionsQuery`, and then do the overrides-specific code in a wrapper around just `useHarborConnectionsQuery` would be slightly cleaner.

I agree that could be cleaner. I'll have a look and see if I find the refactoring worth it

src/harbors/use-harbors.tsx
jorelosorio: I will go for option 2, **Use overrides in Firestore if they exist only**, because it also will help in the boat free ticket which is not done by entur, they do not have that information yet, having it on our system and depends only on that for now should be fine. So that means that we need to add a new attribute here, to mark it as free between two stop places.

KristianSelnas: I discussed this with @gorandalum, and we agreed to keep it as is, i.e. option 3 above. We believe that this will work well with free lines as well.

KristianSelnas:
@gorandalum As we found out yesterday, refactoring this to become more elegant requires a little bit of effort. I've made some minor adjustments. It would be nice if you could sanity check the latest commit.

gorandalum (changes requested):

gorandalum (approved):

PR #3788, opened 2023-08-24
refactor: Auth flow

gorandalum:
Not really changed much in the auth flow, just refactored so the
details are more optional to read into. Previously looking into
the AuthContext could feel a little overwhelming.

The main change of this refactor is that the AuthContext has an
AuthStatus which may be 'loading', 'creating-account, 'finished',
or 'error'.

### DIFF
--- a/src/auth/auth-utils.ts
+++ b/src/auth/auth-utils.ts
@@ -1,0 +1,89 @@
+import {Dispatch} from 'react';
+import auth, {FirebaseAuthTypes} from '@react-native-firebase/auth';
+import Bugsnag from '@bugsnag/react-native';
+import {AuthReducerAction} from './types';
+
+const ERROR_INVALID_PHONE_NUMBER = 'auth/invalid-phone-number';
+const ERROR_INVALID_CONFIRMATION_CODE = 'auth/invalid-verification-code';
+const ERROR_SESSION_EXPIRED = 'auth/session-expired';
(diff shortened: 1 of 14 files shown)


src/auth/auth-utils.ts
rosvik, line 1: Not sure I understand the difference between utils.ts and auth-utils.ts. Should they be the same file?

rosvik (approved):
Looks really good 💯 Had a couple of minor questions.

jorelosorio (approved):
Looks good in general, more clear refactor, nice to see the improvement. I haven't tested it yet properly.

src/auth/auth-utils.ts
gorandalum: Nice catch, I didn't have a good feeling abouth the `auth-utils.ts` and you poked straight into that 😅 

I originally had them in the same file, and ended up splitting them since normally when I think of utils I think of side-effect free utility functions, but the `auth-utils.ts` has functions which are connected to the Firebase auth and didn't really fit into what I would consider "normal" utility functions.

One option can be to introduce a concept of helpers, where we say that the difference between util functions and helper functions is that helper functions can have side effects or state. Then it would be named something like `auth-helpers.ts`. However I didn't wan't to introduce it in the coding guidelines now just for this case, but we might introduced it later if we some more places where we have util functions with side-effects.

For now I think `auth-utils.ts` is slightly better than putting them in `utils.ts`.

PR #3789, opened 2023-08-24
feat: Remove Remote Config force refresh

gorandalum:
The force refresh of RemoteConfig should no longer be necessary, as
it was mainly for development before we got the button for
refreshing it in the debug menu.

In addition the remote config force refresh created a full remount
on every app startup, creating significant wait time with just
a blank screen.

### DIFF
--- a/src/RemoteConfigContext.tsx
+++ b/src/RemoteConfigContext.tsx
@@ -1,9 +1,14 @@
-import React, {createContext, useContext, useEffect, useState} from 'react';
+import React, {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 import remoteConfig from '@react-native-firebase/remote-config';
 import {defaultRemoteConfig, getConfig, RemoteConfig} from './remote-config';
 import Bugsnag from '@bugsnag/react-native';
 
-import {useAppState} from './AppContext';
 import {FeedbackConfiguration} from '@atb/components/feedback';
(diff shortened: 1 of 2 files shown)


rosvik (approved):
LGTM!

src/RemoteConfigContext.tsx
```diff
@@ -83,49 +86,35 @@ export const RemoteConfigContextProvider: React.FC = ({children}) => {
...
   );
 
-  async function refresh() {
+  const refresh = useCallback(async () => {
     const configApi = remoteConfig();
     const {minimumFetchIntervalMillis} = configApi.settings;
-    await configApi.setConfigSettings({
-      minimumFetchIntervalMillis: 0,
-    });
+    await configApi.setConfigSettings({minimumFetchIntervalMillis: 0});
```
jorelosorio, line 113: This will refresh immediately the data, when refresh is called. So the idea is not refresh it, right? Remote config suggest the following patterns https://firebase.google.com/docs/remote-config/loading and one of them is just refresh at the start of the app.

gorandalum: The guide suggest using `fetchAndActivate` which we already use at startup. But in addition to that we had the force refresh code which triggered on every build number change. I think the force refresh code is not needed anymore, and we can just rely on `fetchAndActivate` as mentioned by the guide.

The guide also suggest using `addOnConfigUpdateListener`, which I think would be nice. However, I think that would be its own issue to not bloat this small change.

PR #3793, opened 2023-08-24
chore: LogBox ignore new NativeEventEmitter warnings

gorandalum:
In debug, we get the warnings "`new NativeEventEmitter()` was
called with a non-null argument without the required
`removeListeners` method.

Doing this will make those errors not pop up as warning boxes in
the app. They will still appear in the console.


KristianSelnas (approved):

src/index.tsx
```diff
@@ -38,6 +38,8 @@ MapboxGL.setAccessToken(MAPBOX_API_TOKEN);
 
 trackAppState();
 
+LogBox.ignoreLogs(['new NativeEventEmitter']);
```
jorelosorio, line 41: Mm, I know is a warning, but it is possible to fix it? or is this a temp solution?

gorandalum: It is at least not code we use, but libraries that use some deprecated stuff. Hopefully the libraries should fix it themselves. It might be possible to contribute with PRs to those libraries if we wan't to speed up the process.

PR #3803, opened 2023-08-25
feat: Remove i18n flag

gorandalum:
Removing some unnecessary RemoteConfig flags in the codebase. This also has the benefit as removing it removes a full app rerender at app startup.


mikaelbr (approved):
I think it no longer makes sense to have i18n flag for anyone, so this looks reasonable.

rosvik (approved):

tormoseng:
✅

PR #3808, opened 2023-08-25
feat: Remove enable login RemoteConfig flag

gorandalum:


mikaelbr:
This is the one I'm most unsure about if it removes actual value or not 🤔 But right now I don't think it's important for any orgs to start without login. So if that's the case at some point, we'll reintroduce it.

gorandalum:
> This is the one I'm most unsure about if it removes actual value or not 🤔 But right now I don't think it's important for any orgs to start without login. So if that's the case at some point, we'll reintroduce it.

This flag didn't work now either way. It only disabled the login button in the profile screen, but didn't affect the more newly introduced ways to login, like through the onboarding flow and through purchase of ticket which requires login. Note to team that the feature flags that we choose to support should be maintained so that they really work.

rosvik (approved):

tormoseng:
✅

PR #3816, opened 2023-08-30
fix: resetting disconnect icon for live vehicles

mikaelbr:
There is some newly introduced issue in later React Native that forces Android Websocket connection to close after 30 sec (1006 error). Currently can't find the cause of this issue, but we are able to immediately reconnect. This issue becomes very apparent to the client by the fact that there are stale functions used in `useSubscription` causing the disconnected indicator in the live map _not_ to be updated.

There are apparently cases from before where the client times out by design. I don't think this is that, but as it is already by the design the negative effect of Java WebSocket bug isn't as critical and the workaround with reconnecting is fine I think. Not ideal, but fine.

This PR:

- Fixes error indicator not updating after reconnect for `useSubscription`
- Introduces a delay on the error indicator so that it is only updated after 3 seconds. I think this is a good practice in any case as if it's not connected in only a few seconds the user are non the wiser and get's more confused by why the icon is flashing red.

Related: https://github.com/AtB-AS/kundevendt/issues/7802 https://github.com/AtB-AS/mittatb-app/pull/3784

### DIFF
--- a/src/api/use-subscription.ts
+++ b/src/api/use-subscription.ts
@@ -27,7 +27,7 @@ export function useSubscription({
       const ws = new WebSocket(url);
 
       ws.onmessage = (event) => {
-        onMessage && onMessage(event);
+        onMessage?.(event);
       };
 
       ws.onclose = (event) => {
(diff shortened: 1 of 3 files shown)


rosvik (commented):
Looks good!

One detail: I tried to stay away from mixing in the "realtime" term in this context, since having the GPS location of a vehicle is different from having a "realtime" estimated departure time. Could we call it `liveVehicle` instead in TravelDetailsMapScreenComponent?

gorandalum (approved):

KristianSelnas (approved):

mikaelbr:
@rosvik Yeah, I was back and forth on this, but ultimately decided on naming it something other than the component just in camelCase. But I'm happy as long as it is something other than `vehicle` and `isError`

PR #3818, opened 2023-08-30
perf: improve EstimatedCallItem performance

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/7580

Attempt at improving the performance of the EstimatedCallItem / QuaySection in departures v2. There are quite a bit of code changes, mostly related to moving logic out of EstimatedCallItem, but in general:

- Makes `EstimatedCallItem` a [memo](https://react.dev/reference/react/memo) component
- The renderItem function of QuaySection is no longer inline.
- Reduces the windowSize of the QuaySection flatlist to 5 (from the default 21). This means there will be on average 2 times the screen height rendered above and below the viewport.

### Remaining work

There are still more optimizations to be done after this PR, but it should be a good start.
- An EstimatedCallItem renders whenever `now` changes (which is every tick / 10s). This could be optimized by calculating the departure time texts in advance, and passing the texts, instead of the date, to EstimatedCallItem.
- `QuayView` and `StopPlaceView` can be simplified by not using FlatList (as mentioned in https://github.com/AtB-AS/kundevendt/issues/7580)


src/components/map/components/DeparturesDialogSheet.tsx
```diff
@@ -101,7 +101,7 @@ export const DeparturesDialogSheet = ({
             showOnlyFavorites={false}
             setShowOnlyFavorites={(_) => {}}
             testID="departuresContentView"
-            allowFavouriteSelection={false}
+            showFavorites={false}
```
jorelosorio, line 104: Minor thing. Seems like we are using Favourites word.

src/place-screen/components/EstimatedCallItem.tsx
```diff
@@ -154,49 +116,66 @@ export function EstimatedCallItem({
...
+      return false;
+    if (prev.accessibilityHint !== next.accessibilityHint) return false;
+    if (prev.onPress !== next.onPress) return false;
+    if (prev.showFavorite !== next.showFavorite) return false;
+    if (prev.existingFavorite?.id !== next.existingFavorite?.id) return false;
+    if (prev.favoriteAccessibilityLabel !== next.favoriteAccessibilityLabel)
+      return false;
+    if (prev.onPressFavorite !== next.onPressFavorite) return false;
+    if (prev.showNotices !== next.showNotices) return false;
+    return true;
```
jorelosorio, line 161: It would be nice if this can be refactored or moved it to his own checker.

src/departure-list/section-items/line.tsx
```diff
@@ -70,15 +70,15 @@ export function LineItem({
...
-      stop,
-    );
+    useOnMarkFavouriteDepartures(quay, stop);
 
   if (hasNoDeparturesOnGroup(group)) {
     return null;
   }
+  const favouriteDepartureLine = {
+    ...group.lineInfo,
+    id: group.lineInfo?.lineId ?? '',
```
jorelosorio, line 80: I can't see here the full picture here, but are there any implications if the ID is '' or would be best to just ignore if undefined?

gorandalum:
I added a commit now. Mainly:
- It seemed like every EstimatedCall item still rerendered when the parent rerendered, so this was fixed
- Since we are updating the realtime data every 10 seconds we don't need `useNow` or anything for extra rerenders
- date-fns tools like format and parseIso are surprisingly slow on slow phones. This makes it really hard to have good performance on slow phones.
- Set window-size to 5, which means less items are kept in memory. Will give cheaper rerenders, but more costly scrolling.

rosvik:
> Set window-size to 5, which means less items are kept in memory. Will give cheaper rerenders, but more costly scrolling.

I had a commit for that locally, that I just hadn't pushed 🤦 Thanks!

rosvik (commented):
@gorandalum A bit surprised that every EstimatedCallItem rerendered for you before 🤔 It looked good to me locally, but I might be missing something.

Either way, your refactor looks good, and I think it makes more sense than before 👍 Tested a bit locally, and it feels performant. Not getting any warnings about slow renders.

src/components/map/components/DeparturesDialogSheet.tsx
rosvik: We're consistently very inconsistent here 😅 When searching for "favourite" in the code, I get around 200 results, while "favorite" is around 1100. We should probably take the time to fix that some day... For now, I think we should try to stick to US English (favorite).

src/departure-list/section-items/line.tsx
rosvik: I didn't spend much time on this because it only affects departures v1, which is soon to be removed anyway. I don't think it will happen very often, if ever, and if it does it looks like it will make the favorite-star not work properly.

src/place-screen/components/EstimatedCallItem.tsx
rosvik: Since there are quite a few things that have changed since you suggested this, do you still think it should be refactored? We could always move the `(prev, next) => {}` block to it's own function, but to me this is just as readable 🤷

gorandalum (approved):

PR #3820, opened 2023-08-30
refactor: Small auth changes

gorandalum:
I found it was better to change the auth status to retry the
account creation check. By doing it this way the clients doesn't
need to handle clearing timeouts and such.


rosvik (approved):
LGTM 👍

PR #3822, opened 2023-08-30
chore: flexible transport screenreader

marius-at-atb:
Resolves screen reader issues for flex transport noted [here](https://github.com/AtB-AS/kundevendt/issues/3856#issuecomment-1691341542)

It also addresses the last comments on the previous PR regarding flexible transport [here](https://github.com/AtB-AS/mittatb-app/pull/3736#pullrequestreview-1595303686)

closes https://github.com/AtB-AS/kundevendt/issues/3856

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
@@ -47,7 +47,7 @@ export const Results: React.FC<Props> = ({
   const [errorMessage, setErrorMessage] = useState<string>('');
   const {t} = useTranslation();
 
-  const now = useNow(2500);
+  const now = useNow(30000);
 
   useEffect(() => {
     if (errorType) {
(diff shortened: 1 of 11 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -100,11 +100,9 @@ export const Results: React.FC<Props> = ({
               departureTime={tripPattern.expectedStartTime}
               previousDepartureTime={tripPatterns[i - 1]?.expectedStartTime}
             />
-            <ResultItem
+            <MemoizedResultItem
               tripPattern={tripPattern}
-              onDetailsPressed={() => {
-                onDetailsPressed(tripPattern, i);
-              }}
+              onDetailsPressed={() => onDetailsPressed(tripPattern, i)}
```
gorandalum, line 105: When `React.memo` is used you can't inline function creation like done for `onDetailsPressed` here. This will recreate the function for every render, thus props to `MemoizedResultItem` is changed for every render, and as such `React.memo` has no effect.

src/travel-details-screens/components/TripSection.tsx
```diff
@@ -255,6 +255,9 @@ export const TripSection: React.FC<TripSectionProps> = ({
                           publicCode,
                         ),
                       ),
+                      a11yLabelAppendText: t(
+                        TripDetailsTexts.flexibleTransport.isButtonHint,
+                      ),
```
gorandalum, line 260: This should not be necessary. Should rather be done with `accessibilityRole`.

gorandalum (changes requested):
A couple comments.

Also, even though we use `React.memo`, do we really need to update `now` every 2.5 seconds? I think something like once every minute is more than enough for checking whether the booking time has passed.

marius-at-atb:
@gorandalum The yellow/warning functionality is there to indicate urgency when needed. I don't think it's good to reduce the refresh rate too much either. Updated to 30s now.

gorandalum (approved):

PR #3823, opened 2023-08-31
fix: implement useInterval with requestAnimationFrame

marius-at-atb:
closes https://github.com/AtB-AS/kundevendt/issues/8115

Wait until js thread is ready before executing repeated callbacks


marius-at-atb:
Solved by https://github.com/AtB-AS/mittatb-app/pull/3825 instead

PR #3824, opened 2023-08-31
feat: Loading screen on startup and while logging in

gorandalum:
Shows a loading screen while the auth connection is established and
the account is created. If this takes more than 10 seconds, the user
will be shown an error screen with some extra information, and the
possibility to try again.


jorelosorio (approved):
Looks good

PR #3825, opened 2023-09-01
fix: rewrites use-interval to use setTimout

mikaelbr:
This removes useInterval to remove cases
where useInterval is called infinitely. Rewriting
to setTimeout will only call recursively and not
fill the message queue with as many awaiting
invocations in some cases.

Still not a perfect solution, but will be an improvement.

---

I have tested some locally and started writing some tests for this, but having some issues with fake timers and invocations. We should experiment with the implementations. But feeling the lack of good unit tests in cases like this. I haven't checked out if Detox tests are fixed.

This might also just be reference implementation to show a more elaborate example from what we discussed on Slack @gorandalum and @marius-at-atb .

Trying to debug some I added some logging. There are A LOT of time loops running in the app 😅. Many of the cases is either `useNow` or `setNow()` inside `useInterval`. All these should be replaced with a global clock timer I think.

### DIFF
--- a/src/utils/__tests__/use-interval.test.ts
+++ b/src/utils/__tests__/use-interval.test.ts
@@ -65,6 +65,44 @@ describe('useInterval', () => {
     await runPendingTimers();
     expect(callback).toBeCalledTimes(2);
   });
+
+  it('should be able to disable useInterval', async () => {
+    const hook = renderHook(
+      ({disabled}) => useInterval(callback, 200, [], disabled),
+      {
+        initialProps: {
+          disabled: false,
+        },
+      },
+    );
+
+    // Start with it enabled, should call callback
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(1);
+
+    hook.rerender({disabled: true});
+
+    // Should not call callback any time as it is disabled
+    // even over multiple timer ticks
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(1);
+
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(1);
+
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(1);
+
+    // Re-enable interval,
+    hook.rerender({disabled: false});
+
+    // Callback called on each timer tick.
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(2);
+
+    await nextTimerTick();
+    expect(callback).toBeCalledTimes(3);
+  });
 });
 
 async function advanceByTimer(time: number) {
--- a/src/utils/__tests__/use-interval.test.ts
+++ b/src/utils/__tests__/use-interval.test.ts
@@ -19,18 +19,23 @@ describe('useInterval', () => {
   });
 
   it('should invoke with specified delay', () => {
-    const {result} = renderHook(() => useInterval(callback, 300));
+    const delay = 300;
+    const {result} = renderHook(() => useInterval(callback, delay));
+    expect(callback).toHaveBeenCalledTimes(0);
 
+    jest.advanceTimersByTime(delay);
     expect(result.current).toBeUndefined();
     expect(callback).toHaveBeenCalledTimes(1);
   });
 
-  it.only('should invoke with specified delay and immediate if set', () => {
+  it('should invoke with specified delay and immediate if set', () => {
+    const delay = 300;
     const {result} = renderHook(
       // () => useInterval(callback, 300),
-      () => useInterval(callback, 300, [], false, true),
+      () => useInterval(callback, delay, [], false, true),
     );
-    jest.runOnlyPendingTimers();
+
+    jest.advanceTimersByTime(delay);
 
     expect(result.current).toBeUndefined();
     expect(callback).toHaveBeenCalledTimes(2);
--- a/src/utils/__tests__/use-interval.test.ts
+++ b/src/utils/__tests__/use-interval.test.ts
@@ -1,0 +1,47 @@
+import {renderHook} from '@testing-library/react-hooks';
+import {useInterval} from '../use-interval';
+
+describe('useInterval', () => {
+  let callback = jest.fn();
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+    jest.spyOn(global, 'setTimeout');
+  });
+
+  afterEach(() => {
+    callback.mockRestore();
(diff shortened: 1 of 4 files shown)


marius-at-atb:
Awesome! Works very well and passes the test with extremely short interval time that froze Android before.

mikaelbr:
That is somewhat cheating by me, though. As I've set an explicit minimums time of 100ms 😅

src/utils/use-interval.ts
```diff
@@ -17,15 +18,32 @@ export function useInterval(
...
     function tick() {
-      savedCallback.current();
-    }
-    if (triggerImmediately) {
-      tick();
+      return setTimeout(async function innerCallback() {
+        await savedCallback.current();
+
+        if (!disabled) {
+          tick();
```
marius-at-atb, line 30: As an extra safe guard, could it be considered to use `requestAnimationFrame(() => tick());`?

marius-at-atb:
I replaced it with this to test:
`const actualDelay = delay / 1000; // Math.max(delay ?? 0, 100);`

src/utils/use-interval.ts
mikaelbr: I really, really don't think `requestAnimationFrame` should be used as a general timing function outside animations that need to hook into the animation frame. The degree we call useInterval in the codebase, I think, will increase layout paint overhead significantly and I think it will have the opposite effect rather than improving performance.

But this is just my understanding of the effects of these and if everyone else disagrees I won't stand in your way but you'll have to add it as a separate PR to easier revert it after the fact 😂

mikaelbr:
Ah, yeah, I see. Yes. This won't periodically retrigger interval callback and stack it up if blocked js thread. But good to do a sanity check! 🙌

src/utils/use-interval.ts
```diff
@@ -17,15 +18,32 @@ export function useInterval(
...
+      if (triggerImmediately) {
+        await savedCallback.current();
+      }
+      id = tick();
     }
+
+    invoke();
+    return () => {
+      if (id) {
+        clearTimeout(id as number);
```
gorandalum, line 45: Was a little unsure about if this would clear recursively created `setTimout` invocations, so I tried a quick test in the console, and it seems like it doesn't. Maybe create an `timeoutIdRef` or something?

Screenshot 2023-09-01 at 12 10 20

mikaelbr: Hmm. Yeah, thats a good point 🤔 might refer to different value there

gorandalum: Or just set `id = tick()` in the recursive call as well🤦:sweat_smile:

mikaelbr: Haha, yes. That would work better 😅

gorandalum (commented):
Would it be worth it with a test for this? Maybe with something like [sinon](https://sinonjs.org/#faking-time).

mikaelbr:
Like mentioned in the description: I started, and some of the tests works as expected, but hit a snag in the immediate test case and didn't have the time to debug properly. I can add the test case here and someone else look at it if interesting. I have to pack in for the day (at the SAS lounge and have to board soon)

src/utils/use-interval.ts
marius-at-atb: Not going to insist 😛

src/utils/use-interval.ts
```diff
@@ -1,13 +1,14 @@
 import {useRef, useEffect} from 'react';
```
marius-at-atb, line 2: Should e.g. https://developer.mozilla.org/en-US/docs/Web/API/setInterval#ensure_that_execution_duration_is_shorter_than_interval_frequency be added in a comment for future reference to explain why setInterval isn't used instead?

src/utils/__tests__/use-interval.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  afterEach(() => {
+    callback.mockRestore();
+    jest.clearAllTimers();
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
+
+  it('should invoke with specified delay', () => {
```
marius-at-atb, line 21:
```
it('should invoke with specified delay', () => {
  const delay = 300;
  const {result} = renderHook(() => useInterval(callback, delay));
  expect(callback).toHaveBeenCalledTimes(0);
  jest.advanceTimersByTime(delay);
  expect(result.current).toBeUndefined();
  expect(callback).toHaveBeenCalledTimes(1);
});
```

src/utils/__tests__/use-interval.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    );
+    jest.runOnlyPendingTimers();
+
+    expect(result.current).toBeUndefined();
+    expect(callback).toHaveBeenCalledTimes(2);
+  });
+
+  it('should not have less than 100ms delay', () => {
+    const {result} = renderHook(() => useInterval(callback, 0));
+    jest.advanceTimersByTime(100);
```
marius-at-atb, line 41:
```
jest.advanceTimersByTime(99);
expect(callback).toHaveBeenCalledTimes(0);
jest.advanceTimersByTime(1);
```

src/utils/__tests__/use-interval.test.ts
mikaelbr: These two tests seems to be working as intended with existing code.

Had troubles with `should invoke with specified delay and immediate if set` however. But looking more into this right now.

marius-at-atb: Hmm, when I comment out 'should invoke with specified delay and immediate if set', then 'should invoke with specified delay' fails as well for me.

mikaelbr: Might be some setup/teardown timer issue making it non-deterministic behaviour.

mikaelbr: Testing some more it seems like faked testers in jest don't play nice with async code. I added some workarounds and it seems to be working as expected now. Might look into adding some more test cases soon.

marius-at-atb: Very nice :+1:

marius-at-atb (approved):
Lgtm :+1:

tormoseng:
Cannot find any drawbacks with this issue. Approved!

PR #3826, opened 2023-09-01
chore: upgrades entur tokens dependency

mikaelbr:
Upgrades to the latest dependencies for token handling.

Doesn't seem to require any changes, but I have some local network issues so I'm not able to verify properly and do real end-2-end tests.

Entur notet that upgrading token dependency helped some with the locking js thread.

## Acceptance

- [ ] Verify use of tokens on Android and iOS


mikaelbr:
I did a podinstall, it changed some hermes settings which I don't quite see how it is relevant. But someone else can run pod install and see if it changes the local lock file also

mikaelbr:
I also see I forgot the most relevant package in the list of all packages that requires update. Looks like there are some breaking API-changes on a minor release here.... Checking them out.

gorandalum (approved):

tormoseng:
Looks good!

PR #3830, opened 2023-09-04
feat: Add one second loading after user change

gorandalum:
This is to let the app "settle" before rendering, by letting the
different user data be fully loaded before we start rendering the
app.


jorelosorio (approved):
Looks good ;)

PR #3832, opened 2023-09-06
chore: fix jest test preset to work with react native

mikaelbr:
Fixes existing tests that use some ENVs and React Native specific imports by setting react-native as jest preset. Also adds running the tests to the CI on changes to source files.


gorandalum (approved):
Great!

PR #3834, opened 2023-09-07
feat: add support for custom fareproductgroup heading

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/8114

Also fixing that "Other tickets" title had different font size than the transportmode texts, due to being regular ThemeText. 

Please let me know if this is intentional, they seem to have been designed with same size according to [Figma sketches](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=17165-67089&mode=design&t=Z25Fo7tOiaaAc7DH-0).

<details>
<summary>Screenshots - current title difference</summary>

Notice larger font size of "Andre billetter" title in first image

IMG_8099

IMG_8100

</details>

<details>
<summary>Screenshots - new behaviour</summary>

IMG_8098

IMG_8097

IMG_8094

IMG_8093

</details>

### DIFF
--- a/src/components/transportation-modes/TransportModes.tsx
+++ b/src/components/transportation-modes/TransportModes.tsx
@@ -115,7 +115,12 @@ export const TransportModes = ({
         type={textType ?? 'label__uppercase'}
         color={textColor ?? 'secondary'}
         accessibilityLabel={t(
-          FareContractTexts.transportModes.a11yLabel(transportModeText),
+          customTransportModeText
+            ? FareContractTexts.transportModes.a11yLabelWithCustomText(
+                transportModeText,
+                customTransportModeText,
+              )
+            : FareContractTexts.transportModes.a11yLabel(transportModeText),
         )}
       >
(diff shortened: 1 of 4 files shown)


gorandalum (approved):
Looks good!

Just a food for thought: One other alternative if we want the `TransportModes` to be "cleaner" is to split `TransportModes` into `TransportModeIcons` and `TransportModeText`, where the code would be something like this (pseudocode):
```tsx
<TransportModesIcons
  modes={transportModes}
  iconSize={'small'}
/>
{ heading
    ? <ThemeText>{heading}</ThemeText>
    : <TransportModesText modes={transportModes}/> }
```
Not something that should be done now, since we then should also think about solving "Andre billettyper" in the same manner.

toretefre:
Thanks! I agree on the things you mentioned, so I created a separate issue that can be prioritized later. Also added improved a11y hint after testing it a bit. Merging

PR #3839, opened 2023-09-11
fix: Onboarding triggered at startup

gorandalum:
For some reason triggering the onboarding didn't work after adding
the loading screen. It seems to be a race condition of some sort,
and wrapping it in the IntercationManager.runAfterInteractions
seems to work.


rosvik (approved):
Haven't managed to reproduce the bug after the change, so seems like this fixes it! There is an extra animation where the onboarding slides in on start, but that's probably not worth spending much time on removing.

PR #3841, opened 2023-09-12
fix: Adjustments to ticket assistant

adriawh:
[Figma for the assistant](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=16086-67186&mode=design&t=FQJHhx9OhcQOBN4l-4)

Changes:
Moved the text, and changed it according to figma:

Added a custom text to the child ticket:

Changed spacings between between texts(right is the updated):
  

Small changes to the summary, and made short ticket period info to a message box.

### DIFF
--- a/src/components/info-tag/InfoTag.tsx
+++ b/src/components/info-tag/InfoTag.tsx
@@ -2,16 +2,25 @@ import {StyleProp, ViewStyle} from 'react-native';
 import {StyleSheet} from '@atb/theme';
 import React from 'react';
 import {InfoChip} from '../info-chip';
+import {InfoTagTexts, useTranslation} from '@atb/translations';
 
 type Props = {
-  text: string;
+  text: 'beta' | 'new' | string; // Restricting text to either 'beta', 'new' or any other string
   style?: StyleProp<ViewStyle>;
 };
 export const InfoTag = ({text, style}: Props) => {
   const styles = useStyles();
+  const {t} = useTranslation();
+
+  const getLocalizedText = () => {
+    if (text === 'beta') return 'Beta'; // beta is same in all languages
+    if (text === 'new') return t(InfoTagTexts.newText);
+    return text; // If neither 'beta' nor 'new', return original text
+  };
+
   return (
     <InfoChip
-      text={text}
+      text={getLocalizedText()}
       interactiveColor="interactive_0"
       style={[styles.infoLabel, style]}
     />
--- a/src/components/info-tag/InfoTag.tsx
+++ b/src/components/info-tag/InfoTag.tsx
@@ -5,17 +5,18 @@ import {InfoChip} from '../info-chip';
 import {InfoTagTexts, useTranslation} from '@atb/translations';
 
 type Props = {
-  text: 'beta' | 'new' | string; // Restricting text to either 'beta', 'new' or any other string
+  mode: 'beta' | 'new';
   style?: StyleProp<ViewStyle>;
 };
-export const InfoTag = ({text, style}: Props) => {
+export const InfoTag = ({mode, style}: Props) => {
   const styles = useStyles();
   const {t} = useTranslation();
 
   const getLocalizedText = () => {
-    if (text === 'beta') return 'Beta'; // beta is same in all languages
-    if (text === 'new') return t(InfoTagTexts.newText);
-    return text; // If neither 'beta' nor 'new', return original text
+    if (mode === 'new') return t(InfoTagTexts.newText);
+    else {
+      return 'Beta';
+    }
   };
 
   return (
--- a/src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
+++ b/src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
@@ -48,7 +48,7 @@ export const Root_TicketAssistantStack = ({navigation}: Props) => {
                 },
               }
         }
-        rightButton={{type: 'chat'}}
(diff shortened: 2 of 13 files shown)


src/components/info-tag/InfoTag.tsx
```diff
@@ -2,16 +2,25 @@ import {StyleProp, ViewStyle} from 'react-native';
 import {StyleSheet} from '@atb/theme';
 import React from 'react';
 import {InfoChip} from '../info-chip';
+import {InfoTagTexts, useTranslation} from '@atb/translations';
 
 type Props = {
-  text: string;
+  text: 'beta' | 'new' | string; // Restricting text to either 'beta', 'new' or any other string
```
gorandalum, line 8: Not a fan of mixing texts and "modes" this way. Maybe look at some other way of doing it.

Maybe the `InfoTag` component can have predefined modes like `new` and `beta`, and the `InfoChip` can still just take in `string`?

src/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack.tsx
```diff
@@ -48,7 +48,9 @@ export const Root_TicketAssistantStack = ({navigation}: Props) => {
                 },
               }
         }
-        rightButton={{type: 'chat'}}
+        rightButton={
+          activeTab !== 5 && activeTab !== 0 ? {type: 'close'} : undefined
```
gorandalum, line 52: Is there some way to find out how many tabs there is? When just using 5 "magically" in the code like this, it will most likely be forgotten to change if a tab is added or removed.

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_CategoryPickerScreen.tsx
```diff
@@ -64,6 +65,51 @@ export const TicketAssistant_CategoryPickerScreen = ({
...
 
+  function getAdditionalTitleText(userTypeString: string) {
+    switch (userTypeString) {
+      case 'CHILD':
+        return t(_('/ungdom', '/youth', '/ungdom'));
+      default:
+        return '';
+    }
+  }
+  function AdditionalDescription(userTypeString: string) {
```
gorandalum, line 76: Functions should start with lower case. Also is there a reason why the text is inline in the function instead of in the texts files?

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_FrequencyScreen.tsx
```diff
@@ -185,6 +185,7 @@ const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
   },
   mainView: {
     flex: 1,
+    gap: theme.spacings.xLarge,
```
gorandalum, line 188: Nice! I wasn't aware that gap was added in RN 0.71

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_SummaryScreen/TicketAssistant_SummaryScreen.tsx
```diff
@@ -143,6 +142,16 @@ export const TicketAssistant_SummaryScreen = ({navigation}: SummaryProps) => {
...
             <TicketSummary />
+            {doesTicketCoverEntirePeriod && (
+              <MessageBox
+                style={styles.infoBox}
+                type={'info'}
+                title={t(TicketAssistantTexts.summary.durationNotice.title)}
+                message={t(
+                  TicketAssistantTexts.summary.durationNotice.description,
+                )}
+              ></MessageBox>
```
gorandalum, line 153: Doesn't need to have `</MessageBox>` when there is no children. Just end the tag with `/>`

gorandalum (changes requested):
Nice work! I commented some nitpick here and there.

src/components/info-tag/InfoTag.tsx
```diff
@@ -2,16 +2,26 @@ import {StyleProp, ViewStyle} from 'react-native';
...
+export const InfoTag = ({mode, style}: Props) => {
   const styles = useStyles();
+  const {t} = useTranslation();
+
+  const getLocalizedText = () => {
+    if (mode === 'new') return t(InfoTagTexts.newText);
+    else {
+      return 'Beta';
+    }
+  };
```
gorandalum, line 20: Not something that needs to be fixed, just a FYI:

Creating the `getLocalizedText` function like this is not preferred, as the function is recreated every render for no real reason. It would be more efficient, and cleaner, to write this as:
```tsx
const text = mode === 'new' ? t(InfoTagTexts.newText) : 'Beta';
```
However, when handling union values like this (`'beta' | 'new'`) it is preferred to use `switch` statement, as it gives more type-safety when adding new `mode` values. But a switch statement doesn't really look good inline in the render function, so then it is appropriate to create a function for it. Note then that it is preferred to create it outside of the render function, so it isn't recreated on every render. Something like this:
```tsx
const getLocalizedText = (mode: Props['mode']) => {
  switch (mode) {
    case 'new':
      return t(InfoTagTexts.newText);
    case 'beta':
      return 'Beta';
  }
}
```

Doing it like this will give typescript error during compiling if a new `mode` is added, without being handled by the `switch` statement. Just doing `if`/`else` will not give that type of compiler help.

gorandalum (approved):

src/components/info-tag/InfoTag.tsx
adriawh: Yes that definitely makes sense💡

PR #3844, opened 2023-09-12
fix: Onboarding triggered multiple times

gorandalum:
### Background
Tor Kjetil found a bug where the onboarding was triggered at app startup even though the user had finished the onboarding.

The reason was that the storage of the "has_onboarded" flag did not always happen, as
the `useFinishOnboarding` hook had unmounted. This was now changed
to a more simple logic:
- Set onboarded to true
- Then fire the request for geolocation


jorelosorio (approved):
Looks good

rosvik (approved):

tormoseng:
Tested OK!

PR #3845, opened 2023-09-12
refactor: Move LoadingBoundery

gorandalum:
We see some strange effects of rendering the NavigationContainer
when there is no navigators to render. This was changed by moving
the LoadingScreenBoundary outside of the NavigationContainer.

### Acceptance criteria
- [ ] My position should be filled on Dashboard root screen, even after login
- [ ] Loading boundary should work as before
- [ ] Should do some random samples to check that the loading screen works as before both on iOS and Android.


rosvik (approved):

tormoseng:
Tested OK on
- [x] Android
- [ ] iOS

PR #3847, opened 2023-09-12
fix: Favorite chips margins

KristianSelnas:
Fixes an issue found during testing of https://github.com/AtB-AS/kundevendt/issues/8088

See https://github.com/AtB-AS/kundevendt/issues/8088#issuecomment-1710018616

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/Dashboard_RootScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/Dashboard_RootScreen.tsx
@@ -189,8 +189,8 @@ export const Dashboard_RootScreen: React.FC<RootProps> = ({
         contentContainerStyle={style.scrollView}
         testID="dashboardScrollView"
       >
-        <View style={[style.contentSection, style.searchHeader]}>
-          <Section>
+        <View style={style.searchHeader}>
+          <Section style={style.contentSection}>
             <LocationInputSectionItem
               accessibilityLabel={
                 t(TripSearchTexts.location.departurePicker.a11yLabel) +
@@ -435,6 +435,7 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
   },
   favoriteChips: {
     marginTop: theme.spacings.medium,
+    paddingHorizontal: theme.spacings.medium,
   },
   searchHeader: {
     marginTop: 0,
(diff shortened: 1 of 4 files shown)


rosvik:
Screenshot 2023-09-12 at 15 04 20

Looks a lot better, but still doesn't line up on the right edge when scrolling all the way to the right. Should be 12px off the screen edge I think.

KristianSelnas:
> Screenshot 2023-09-12 at 15 04 20
> 
> Looks a lot better, but still doesn't line up on the right edge when scrolling all the way to the right. Should be 12px off the screen edge I think.

Strange. I cannot figure out why, and I don't see any immediate fixes. There is only 12px (`theme.spacing.medium`) padding on the right edge. I have no idea why it appears to be more. Personally, I don't think it's a big deal now. It is a whole lot better to get this fix in before the release with this minor detail, than not having it this fix at all.

gorandalum:
> Strange. I cannot figure out why, and I don't see any immediate fixes. There is only 12px (`theme.spacing.medium`) padding on the right edge. I have no idea why it appears to be more. Personally, I don't think it's a big deal now. It is a whole lot better to get this fix in before the release with this minor detail, than not having it this fix at all.

It's the `marginRight` stuff in `FavoriteChips` that is acting up. The add button should have no margin right. But it was here before, so no blocker.

gorandalum (approved):

rosvik:
simulator_screenshot_1DF1B26D-D4A4-4857-BD57-8D6A6006B186

Looks like it's happening because all buttons have `small` margin on the right, so what we're seeing is 8+12px padding on the screen edge. I think there has been something like `marginRight: medium - small` before.

But you're right, this is a lot better 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/Dashboard_RootScreen.tsx
```diff
@@ -435,6 +435,7 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
   },
   favoriteChips: {
     marginTop: theme.spacings.medium,
+    paddingHorizontal: theme.spacings.medium,
```
rosvik, line 438: This seems to line up for me. Not the most beautiful code, but more beautiful app 😄 

```suggestion
    paddingLeft: theme.spacings.medium,
    paddingRight: theme.spacings.medium - theme.spacings.small,
```

PR #3848, opened 2023-09-12
feat: Add analytics for loading boundary error

gorandalum:
### Background
[This comment](https://github.com/AtB-AS/kundevendt/issues/4157#issuecomment-1708190735) from Tor:
> @gorandalum Would it be possible to add an event to BugSnag or PostHog when a user enters the Try again screen? And possibly also on the Try again button itself? @Sebstorvik would really like to know how many users are affected by the missing claims.

### Acceptance criteria
- [ ] PostHog events for when the user got the error screen, when they pressed try again, and when they pressed contact customer service.

### DIFF
--- a/src/analytics/types.ts
+++ b/src/analytics/types.ts
@@ -3,6 +3,7 @@ export type AnalyticsEventContext =
   | 'Departure details'
   | 'Favorites'
   | 'Flexible transport'
+  | 'Loading boundary'
   | 'Map'
   | 'Mobility'
   | 'Receipt'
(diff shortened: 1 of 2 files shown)


rosvik (approved):
Looks good! 

Will we be able to determine "x% of startups goes to the try again screen" based on this, or do we also need to add logging for successful startups?

gorandalum:
> Looks good!
> 
> Will we be able to determine "x% of startups goes to the try again screen" based on this, or do we also need to add logging for successful startups?

I hope we will be able to deduce that. I think we can't add logging for successful startup just for this metric as it will be quite a lot of events and quite costly.

tormoseng:
> Will we be able to determine "x% of startups goes to the try again screen" based on this, or do we also need to add logging for successful startups?

Since we have removed automatic events like `Background` and `Active`, we now have an event `Identify` that is sent when the app has received the `abtCustomerId`, i.e. I guess it will be sent _after_ the claims are received. But anyhow, I think the absolute number also gives value if we compare it to active users (from either PostHog or Intercom).

tormoseng:
Tested OK!

PR #3850, opened 2023-09-13
fix: Add active flag to announcements

KristianSelnas:
Adds a property 'active' to announcements. This makes it easier to toggle announcements on and off without deleting them.


gorandalum (changes requested):
Versions from master are a part of this PR. This should be cleaned up.

src/announcements/types.ts
```diff
@@ -3,6 +3,7 @@ import {FirebaseFirestoreTypes} from '@react-native-firebase/firestore';
 
 export type AnnouncementRaw = {
   id: string;
+  active: boolean;
```
gorandalum, line 6: Do we need the field here since we use `active == true` in the query?

gorandalum (approved):
LGTM :+1:

src/announcements/types.ts
KristianSelnas: Belte og bukseseler, er det ikke det de kaller det?

PR #3851, opened 2023-09-13
fix: Accessibility and language fixes for Announcements

KristianSelnas:
Fixes bugs found during testing of https://github.com/AtB-AS/kundevendt/issues/7907

See https://github.com/AtB-AS/kundevendt/issues/7907#issuecomment-1717246455

### DIFF
--- a/src/translations/screens/Dashboard.ts
+++ b/src/translations/screens/Dashboard.ts
@@ -8,6 +8,16 @@ const DashboardTexts = {
     },
   },
   buyButton: _('Kjøp billetter', 'Buy tickets', 'Kjøp billettar'),
+  announcemens: {
+    header: _('Aktuelt', 'Announcements', 'Aktuelt'),
+    button: {
+      accessibility: _(
+        'Trykk for å lese mer',
+        'Press to read more',
+        'Trykk for å lese meir',
+      ),
+    },
+  },
 };
 
 export default DashboardTexts;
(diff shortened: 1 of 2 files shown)


src/translations/screens/Dashboard.ts
```diff
@@ -8,6 +8,16 @@ const DashboardTexts = {
...
   },
   buyButton: _('Kjøp billetter', 'Buy tickets', 'Kjøp billettar'),
+  announcemens: {
+    header: _('Aktuelt', 'Announcements', 'Aktuelt'),
+    button: {
+      accessibility: _(
+        'Trykk for å lese mer',
+        'Press to read more',
+        'Trykk for å lese meir',
+      ),
```
gorandalum, line 18: Hints like these should say "activate to …" instead of "press to …". Accessibility users may for example use voice activation instead of pressing to activate.

gorandalum (changes requested):

src/translations/screens/Dashboard.ts
KristianSelnas: Ok, it's not consistent throughout the app. What would that be in Norwegian? "Aktiver ..."?

gorandalum: Yeah, "Aktiver for"/"Activate to" is the correct phrasing. I made a quick search in the app now and think we seem fairly consistent, I just found a few recently added Nynorsk-texts with the wrong phrasing.

gorandalum (approved):

PR #3853, opened 2023-09-13
feat: add latest activation date option to datepicker

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/8388

Adding option to provide limitation to a product in the referenceData, where `latestActivationDate` is provided as a string given by epoch time. Tested with FRAM Buss Vaksen, having latest day of activation 31st of October 23:59:59. 

Will be part of FRAM 1.41 on a separate release branch and AtB 1.42. (No plans to use for AtB AFAIK)

### DIFF
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -19,6 +19,7 @@ export type PreassignedFareProduct = {
     userProfileRefs: string[];
     appVersionMin: string | undefined;
     appVersionMax: string | undefined;
+    latestActivationDate?: string;
   };
 };
 
--- a/src/reference-data/types.ts
+++ b/src/reference-data/types.ts
@@ -19,7 +19,7 @@ export type PreassignedFareProduct = {
     userProfileRefs: string[];
     appVersionMin: string | undefined;
     appVersionMax: string | undefined;
-    latestActivationDate?: string;
+    latestActivationDate?: number;
   };
 };
 
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
@@ -86,9 +86,7 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
 
   const maximumDateObjectIfExisting = preassignedFareProduct.limitations
     ?.latestActivationDate
-    ? new Date(
-        Number(preassignedFareProduct.limitations.latestActivationDate) * 1000,
-      )
+    ? new Date(preassignedFareProduct.limitations.latestActivationDate * 1000)
     : undefined;
 
   const hasSelection =
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
@@ -83,6 +83,14 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
     travellerSelection,
     travelDate,
   );
+
+  const maximumDateObjectIfExisting = preassignedFareProduct.limitations
+    ?.latestActivationDate
+    ? new Date(
+        Number(preassignedFareProduct.limitations.latestActivationDate) * 1000,
+      )
+    : undefined;
+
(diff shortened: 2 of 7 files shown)


mortennordseth (approved):
I think this looks good 👍 I thought about suggesting changing `maximumDate` to `latestActivationDate` to have consistency, but I see that `minimumDate` is already used so maybe that's better.

toretefre:
> I think this looks good 👍 I thought about suggesting changing `maximumDate` to `latestActivationDate` to have consistency, but I see that `minimumDate` is already used so maybe that's better.

Yes, I considered it, but as you say minimumDate and maximumDate is already defined as a prop to the DatePicker, so went for parts of code that potentially can be reused for other purposes, and `latestActivationDate` where the code is specific for this issue. 

Merging soon 👌

src/reference-data/types.ts
```diff
@@ -19,6 +19,7 @@ export type PreassignedFareProduct = {
     userProfileRefs: string[];
     appVersionMin: string | undefined;
     appVersionMax: string | undefined;
+    latestActivationDate?: string;
```
gorandalum, line 22: I think this should be `number`? Don't understand why it is a `string` 🤔

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
```diff
@@ -61,8 +67,23 @@ export const TravelDateSheet = forwardRef<ScrollView, Props>(
...
         >
+          {maximumDate && (
+            <MessageBox
+              type="info"
+              message={t(
+                TravelDateTexts.latestActivationDate.warning(
+                  formatToVerboseFullDate(maximumDate, language),
+                ),
+              )}
+            />
```
gorandalum, line 78: Note: Now possible to use the `subtle` flag if we don't want such a "screaming" info box.

gorandalum (approved):
Looks good. A couple minor comments. Also this should be done for the webshop, as it is kinda risky to have fields in the reference data that is supported by app, but not webshop.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
toretefre: Good idea, seems to be even more suitable 💡 

Also wrapped the MessageBox in a section to have some padding to the pickers.

src/reference-data/types.ts
toretefre: I guess number would be more correct, and we'll save a conversion later. 

Good catch, typically defaulting to strings as they are not as easily changed in case they start with zeroes etc., but shouldn't be an issue here.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -51,6 +53,10 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
     navigation.setParams({
       preassignedFareProduct: fp,
     });
+    if (fp.limitations.latestActivationDate && travelDate) {
+      if (isAfter(travelDate, formatISO(fp.limitations.latestActivationDate)))
```
gorandalum, line 57: I would use `new Date` instead of `formatIso`, since in `isAfter` the iso is straightly parsed to a Date object. From performance measuring we see that handling date (like `parseIso`, `format`, etc.) is some of the cost that significantly adds up during add render.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
```diff
@@ -61,8 +67,26 @@ export const TravelDateSheet = forwardRef<ScrollView, Props>(
           ref={focusRef}
           centerContent={true}
         >
+          {maximumDate && (
+            <Section withBottomPadding>
```
gorandalum, line 71: `Section` should only be used with section item components. Should rather just pass a `style` to the message box if we want some bottom margin.

toretefre:
Good feedback @gorandalum , I've addressed it in https://github.com/AtB-AS/mittatb-app/pull/3856, thank you! 🙌

PR #3855, opened 2023-09-14
fix: Remove spinner stuck fix

gorandalum:
Adding/removing RefreshControl based on focus gives odd behaviour
on Android where the scroll offset is lost after navigating.

The adding/removing of RefreshControl was added to fix a stuck
spinner on iOS issue, but this doesn't seem to still happen after
updating react-native versions. So the fix for now is to remove
the stuck spinner fix.

The original fix was implemented here:
https://github.com/AtB-AS/mittatb-app/pull/3762

Based on the behaviour in this comment:
https://mittatb.slack.com/archives/C0116FMPX4Y/p1691563608147469


marius-at-atb (approved):
Lgtm! The issue where the pull to refresh indicator is hidden behind the header remains, but the scroll height issue on return works now :+1:

KristianSelnas (approved):

PR #3860, opened 2023-09-15
chore: Update get-release dep

gorandalum:
Hopefully it this fixes the build error:
https://github.com/AtB-AS/mittatb-app/actions/runs/6186469796/job/16803423932#step:10:54


mikaelbr (approved):
No idea if you test this in CI when it is registered as a PR, but approving in any case. Usually OK to upgrade in any case

gorandalum:
I had 5 minutes before standup to fix this, and wanted to run the CI during standup 🤠

PR #3864, opened 2023-09-15
chore: Remove travel search filter onboarding

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/8542


rosvik:
@KristianSelnas Could you update @atb-as/generate-assets to 9.0.0 as well? That will remove illustrations that are not in use (https://github.com/AtB-AS/design-system/pull/143)

PR #3868, opened 2023-09-15
feat: add app-purchase-overview global messages

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/7512

- Adds a new globalmessage context `app-purchase-overview`
- Removes most "Summary" texts, and makes them global messages in firestore (more details in https://github.com/AtB-AS/kundevendt/issues/7512)
- If isFree is true, no other messages are shown (added after input from @oyvindmikkelsenatb)
- Removes `enable_nfk_nightbus_warning` and `enable_nfk_hour24_warning`


src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
```diff
@@ -86,6 +89,17 @@ export function Root_FareContractDetailsScreen({navigation, route}: Props) {
...
 
+export function getValidOnTrainNoticeText(
+  t: TranslateFunction,
+  fareProductType?: string,
+) {
+  if (fareProductType === 'single')
+    return t(FareContractTexts.samarbeidsbillettenInfo.single);
+  if (fareProductType === 'hour24')
+    return t(FareContractTexts.samarbeidsbillettenInfo.hour24);
+  return t(FareContractTexts.samarbeidsbillettenInfo.period);
```
gorandalum, line 100: We may have talked about this, but are these texts planned to be removed? They are using logic based on fare product type, which shouldn't be done.

rosvik: Talked to @oyvindmikkelsenatb about it now, and I think it might make sense to add another global message context on the fare product page, since we might want to show info about things like city bikes, and other org specific exceptions. I think it can be solved in it's own task.

rosvik: After that is removed, the only code remaining related to the collab ticket is some "from travel search to ticketing" - code.

gorandalum: I think it would be nice to do it in the context of https://github.com/AtB-AS/kundevendt/issues/7512, since it kinda belongs there, and creating a new issue makes it necessary to get it through sprint planning again.

gorandalum (approved):
Nice work!

PR #3869, opened 2023-09-18
test: add test-ids to departure details

tormoseng:
Adding test-ids to departure details.


gorandalum (approved):
Looks ok, but some tsc errors.

PR #3871, opened 2023-09-19
fix: Change loading screen texts for FRAM

adriansberg:
Change loading screen texts for FRAM

Closes https://github.com/AtB-AS/kundevendt/issues/8785


toretefre (approved):
Nice! Also need to change the url, not sure where it is done

adriansberg:
Same case as with the ticket page. The URL is in Remote Config, so it won't be fetched in offline mode.

mortennordseth:
> Same case as with the ticket page. The URL is in Remote Config, so it won't be fetched in offline mode.

Maybe we could move the default customer service URL to the env-file, similar to what is done with the privacy policy URL?

adriansberg:
Great idea. I'll do that. Do any of you have access to the other tenants and get their customer service url?

mortennordseth:
I got these from the remote config: 

NFK: https://www.reisnordland.no/kontaktskjema
AtB: https://www.atb.no/kontakt/
FRAM: https://frammr.no/kontakt-oss

adriansberg:
Different for the english pages, right?

toretefre:
https://www.reisnordland.com/contact-us
https://www.atb.no/en/customer-service/
https://en.frammr.no/contact-us

mortennordseth:
They do exist in English, but afaik that's not supported by the app today?

adriansberg:
Right. Used to the webshop having both.

adriansberg:
Added it now.

toretefre:
So, we are now trying Firestore first, and then defaulting to env url, right?

adriansberg:
As I understand it the default is now set through the .env, and then Remote Config have precedence over the default that is set.

adriansberg:
https://github.com/AtB-AS/mittatb-app/pull/3871/commits/3ed3956a380a568f493b51f8157529cb0970c24c should fix the ticketing page being shown in offline mode.

toretefre (approved):

PR #3874, opened 2023-09-19
fix: various header fixes

rosvik:
I've noticed there are several places where the header takes up more space than it should according to the sketches. Tried to clean up that here:

- Reduces header margin to 12px/medium instead of large, as that's what's used in [the sketches](https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components?node-id=4383%3A8566&mode=dev) 
	- This also adds a bit of bottom sheet specific code to give it more padding, that will probably be replaced by https://github.com/AtB-AS/kundevendt/issues/4211
- Makes the status bar on android transparent, fixing some contrast issues, and utilizing more of the screen.
- Removes a bit of padding above the StopPlace/Quay selector on departures

### DIFF
--- a/src/components/screen-header/ScreenHeader.tsx
+++ b/src/components/screen-header/ScreenHeader.tsx
@@ -67,7 +67,7 @@ export const ScreenHeader: React.FC<ScreenHeaderProps> = (props) => {
   return <BaseHeader leftIcon={leftIcon} rightIcon={rightIcon} {...props} />;
 };
 
-type ScreenHeaderWithoutNavigationProps = ScreenHeaderProps & {
+type ScreenHeaderWithoutNavigationProps = Omit<ScreenHeaderProps, 'style'> & {
   leftButton?: HeaderButtonWithoutNavigationProps;
   rightButton?: HeaderButtonWithoutNavigationProps;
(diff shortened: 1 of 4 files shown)


gorandalum:
Not sure what the purpose is here, but would it be just as quick to just do this?
https://github.com/AtB-AS/kundevendt/issues/4211

In that issue I'm guessing we will create a new header-component just for bottom-sheet, and we can clean up the screen header component by removing all that without navigation stuff, and probably then don't need all the base header stuff either.

rosvik:
@gorandalum Updated the description now, so hopefully it makes more sense. Didn't want to spend a lot of time on this, since it's just some in-between-time before I'm unblocked on other tasks, so I just added a TODO for the bottom sheet padding hack.

gorandalum (approved):

rosvik:
### Acceptance criteria:

- [ ] The header has less padding and works well in general
- [ ] Bottom sheets works like before
- [ ] Android devices has no/transparent header on the map tab, but the status bar (clock, wifi-icon) is visible. 
	- A good idea to test different devices here with different notches and front camera placements

PR #3875, opened 2023-09-20
feat: add app-fare-contract-details global message context

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/7512

Adds a `app-fare-contract-details` global message context, replacing the last hard coded collab ticket messages in the app 🎉  (But some "from travel search to ticketing" logic still remains)

### DIFF
--- a/src/fare-contracts/details/DetailsContent.tsx
+++ b/src/fare-contracts/details/DetailsContent.tsx
@@ -21,6 +21,13 @@ import {
   LinkSectionItem,
   Section,
 } from '@atb/components/sections';
+import {
+  GlobalMessage,
+  GlobalMessageContextEnum,
+  useGlobalMessagesState,
+} from '@atb/global-messages';
+import {View} from 'react-native';
+import {StyleSheet} from '@atb/theme';
 
 type Props = {
   fareContract: FareContract;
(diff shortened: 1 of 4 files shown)


gorandalum (approved):
🎉

src/fare-contracts/details/DetailsContent.tsx
```diff
@@ -60,6 +69,18 @@ export const DetailsContent: React.FC<Props> = ({
     );
 
     const validityStatus = getValidityStatus(now, validFrom, validTo, fc.state);
+
+    const globalMessageRuleVariables = {
+      fareProductType: preassignedFareProduct?.type ?? 'unknown',
+      firstTravelRightType: firstTravelRight.type,
+      validityStatus: validityStatus,
+      tariffZones: firstTravelRight.tariffZoneRefs ?? [],
```
marius-at-atb, line 77: When I read the name "tariffZoneRefs", I first thought about react refs. After looking into it, I found that it is a list of ids referencing object id fields in referenceData. Perhaps it should be considered to rename it to tariffZoneRefIds, to both be precise about what it references, and also to avoid confusion with react refs in naming?
I see there are several other variables with the same naming convention, which should also be updated in case we change this..

marius-at-atb (commented):
Very nice!

src/fare-contracts/details/DetailsContent.tsx
rosvik: I agree that it's not very intuitive naming, but this naming stems from the firestore (and probably entur) travelrights.

Screenshot 2023-09-21 at 10 49 53

There is probably _some_ code that can be renamed, but I don't think we can get completely rid of it.

marius-at-atb: Following discussion with @gorandalum and @mjenssen, it turns out that the naming convention is defined on our side. However it would require quite a bit of work with versioning for backward compatibility to change, so not worth it at this time.

PR #3877, opened 2023-09-20
feat: Save payment card outside of purchase process

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/3533

### DIFF
--- a/src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
+++ b/src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
@@ -1,0 +1,94 @@
+import {FullScreenHeader} from '@atb/components/screen-header';
+import React, {useEffect, useState} from 'react';
+import {View} from 'react-native';
+import {StyleSheet} from '@atb/theme';
+import {useAddPaymentMethod} from '@atb/stacks-hierarchy/Root_AddPaymentMethodScreen/use-add-payment-method';
+import {Processing} from '@atb/stacks-hierarchy/Root_PurchasePaymentWithCreditCardScreen/Processing';
+import {useTranslation} from '@atb/translations';
+import {MessageBox} from '@atb/components/message-box';
+import {Button} from '@atb/components/button';
+import AddPaymentMethodTexts from '@atb/translations/screens/subscreens/AddPaymentMethodTexts';
+import WebView from 'react-native-webview';
+import {RootStackScreenProps} from '@atb/stacks-hierarchy';
+
+type Props = RootStackScreenProps<'Root_AddPaymentMethodScreen'>;
+
+export const Root_AddPaymentMethodScreen = ({navigation}: Props) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+  const [showWebView, setShowWebView] = useState<boolean>(true);
+  const {
+    terminalUrl,
+    onWebViewLoadStart,
+    onWebViewNavigationChange,
+    onWebViewLoadEnd,
+    onWebViewError,
+    isLoading,
+    error,
+  } = useAddPaymentMethod();
+
+  useEffect(
+    () => navigation.addListener('blur', () => setShowWebView(false)),
+    [navigation],
+  );
+
+  const navigateBack = () => navigation.pop();
+
+  return (
+    <View style={styles.container}>
+      <FullScreenHeader
+        title={t(AddPaymentMethodTexts.header)}
+        leftButton={{
+          type: 'cancel',
+          onPress: () => navigateBack(),
+        }}
+      />
+      {isLoading && (
+        <View style={styles.center}>
+          <Processing message={t(AddPaymentMethodTexts.loading)} />
+        </View>
+      )}
+      {!!error && (
+        <View style={styles.center}>
+          <MessageBox
+            message={t(AddPaymentMethodTexts.genericError)}
+            type="error"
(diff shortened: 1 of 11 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentOptionsScreen.tsx
```diff
@@ -49,10 +57,40 @@ export const Profile_PaymentOptionsScreen = ({}: PaymentOptionsProps) => {
   }
 
   useEffect(() => {
-    if (isFocused) {
-      refreshCards();
-    }
-  }, [isFocused]);
+    refreshCards();
```
KristianSelnas, line 60: isFocused is removed to prevent reloading the cards when you return from the add card process, and corresponding race conditions in setting `isFocused`. The only (minor) consequence is that if you navigate to this page first, then purchase a ticket, and finally navigate back here again, the card you just saved won't appear until you pull down to refresh. That's considered acceptable.

src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <FullScreenHeader
+        title={t(AddPaymentMethodTexts.header)}
+        leftButton={{
+          type: 'cancel',
+          onPress: () => navigateBack(),
+        }}
+      />
+      {isLoading && (
+        <View style={styles.center}>
+          <Processing message={t(AddPaymentMethodTexts.loading)} />
```
gorandalum, line 47: Can use the loading text from the dictionary instead.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentOptionsScreen.tsx
```diff
@@ -49,10 +57,40 @@ export const Profile_PaymentOptionsScreen = ({}: PaymentOptionsProps) => {
...
+
+  useEffect(() => {
+    const addPaymentMethodCallbackHandler = async ({url}: {url: string}) => {
+      if (
+        url.includes('response_code') &&
+        url.includes('recurring_payment_id')
+      ) {
+        // Timeout required to make loading spinner appear,
+        // see https://github.com/facebook/react-native/issues/37308
+        setTimeout(() => setIsLoading(true), 20);
```
gorandalum, line 71: Could this also be solved by setting the initial state of `isLoading` as `false`, and to set it to `true` at start of loading, both here and in the `useEffect` on mount..

gorandalum (approved):
Nice!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentOptionsScreen.tsx
marius-at-atb: Now there are two useEffects with empty dependency arrays. Any particular reason why not just one, or can refreshCards(); be moved inside the other one?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentOptionsScreen.tsx
```diff
@@ -84,25 +127,31 @@ export const Profile_PaymentOptionsScreen = ({}: PaymentOptionsProps) => {
...
+      {showError && <GenericError />}
+      <ScrollView
+        style={style.contentContainer}
+        refreshControl={
+          <RefreshControl refreshing={isLoading} onRefresh={refreshCards} />
+        }
+      >
+        {storedCards.length > 0 && (
+          <Section>
+            {storedCards.map((item) => (
```
marius-at-atb, line 139: Just my taste to call each item `storedCard` or `sc` instead of the generic `item` in mappings like this

marius-at-atb (approved):
Great work!

src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const navigateBack = () => navigation.pop();
+
+  return (
+    <View style={styles.container}>
+      <FullScreenHeader
+        title={t(AddPaymentMethodTexts.header)}
+        leftButton={{
+          type: 'cancel',
+          onPress: () => navigateBack(),
```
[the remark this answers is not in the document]
KristianSelnas: No idea. Just inspired by somewhere else, I guess?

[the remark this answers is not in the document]
gorandalum: Can just try to remove it. The header sets up `navigation.goBack()` as default handler when type cancel.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentOptionsScreen.tsx
KristianSelnas: No unfortunately, that doesn't work

PR #3878, opened 2023-09-20
chore: Distribute APK directly to Google Play

gorandalum:
We are having trouble with the AppCenter integration:
https://mittatb.slack.com/archives/C02EEG7D8EL/p1695204736271979?thread_ts=1695200240.568439&cid=C02EEG7D8EL

But for the store build AppCenter is an unnecessary middle step, so we'll
try sending it directly to Google Play like NFK and Fram does instead.


marius-at-atb (approved):
Lgtm

PR #3880, opened 2023-09-21
feat: Beacons feature flag

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/8617

### DIFF
--- a/src/beacons/use-is-beacons-enabled.tsx
+++ b/src/beacons/use-is-beacons-enabled.tsx
@@ -1,0 +1,13 @@
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {useDebugOverride} from '@atb/debug';
+import {StorageModelKeysEnum} from '@atb/storage';
+
+export const useIsBeaconsEnabled = () => {
+  const {enable_beacons} = useRemoteConfig();
+  const [debugOverride] = useBeaconsEnabledDebugOverride();
+  return debugOverride !== undefined ? debugOverride : enable_beacons;
+};
+
+export const useBeaconsEnabledDebugOverride = () => {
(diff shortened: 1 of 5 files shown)


gorandalum (approved):

src/beacons/use-is-beacons-enabled.tsx
```diff
@@ -0,0 +1,13 @@
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {useDebugOverride} from '@atb/debug';
+import {StorageModelKeysEnum} from '@atb/storage';
+
+export const useIsBeaconsEnabled = () => {
+  const {enable_beacons} = useRemoteConfig();
+  const [debugOverride] = useBeaconsEnabledDebugOverride();
```
marius-at-atb, line 7: If useIsBeaconsEnabled is going to be used reactively, it is fine as is.
However if useIsBeaconsEnabled is going to be used as a one-time app start init call for beacon activation, you also want to retrieve and return debugOverrideReady (see example in the useFlexibleTransportEnabled hook). This is because until debugOverrideReady is true, enable_beacons will be set to its default value, not a value read from storage.

marius-at-atb (commented):
Looks good! Just one comment to keep in mind for future usage of this hook

src/beacons/use-is-beacons-enabled.tsx
KristianSelnas: Good point! I'll create a new pr for this

KristianSelnas: https://github.com/AtB-AS/mittatb-app/pull/3884

PR #3883, opened 2023-09-21
fix: add numberOfZones as fare contract details ruleVariable

rosvik:
Since the collab ticket only applies to tickets that are just in zone A, and not e.g. from A to B1, we need to check that numberOfZones == 1 when showing the message.

fixes https://github.com/AtB-AS/kundevendt/issues/7512#issuecomment-1729336551


gorandalum (commented):
Could it be an alternative to instead support `equalTo` for arrays as well?
So it would be more like `tariffZones equalTo [AtB:TariffZone:A]` in FireStore?

gorandalum:
> Could it be an alternative to instead support `equalTo` for arrays as well? So it would be more like `tariffZones equalTo [AtB:TariffZone:A]` in FireStore?

Discussed with Johannes, and we won't do this for now.

PR #3884, opened 2023-09-21
fix: Retun debugOverrideReady from useIsBeaconsEnabled

KristianSelnas:
Fix based on comment in https://github.com/AtB-AS/mittatb-app/pull/3880


marius-at-atb (approved):
Nice!

PR #3885, opened 2023-09-21
feat: added support for errorText

hanne-at-atb:
As part of https://github.com/AtB-AS/kundevendt/issues/4275, we need to support error state for TextInputSectionItems.

Implemented to look like this:
image

### DIFF
--- a/src/components/sections/items/TextInputSectionItem.tsx
+++ b/src/components/sections/items/TextInputSectionItem.tsx
@@ -1,4 +1,4 @@
-import React, {forwardRef, useRef, useState} from 'react';
+import React, {forwardRef, useEffect, useRef, useState} from 'react';
 import {
   AccessibilityInfo,
   NativeSyntheticEvent,
@@ -15,11 +15,11 @@ import {ThemeText, MAX_FONT_SCALE} from '@atb/components/text';
 import {ThemeIcon} from '@atb/components/theme-icon';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
-import {SectionTexts, useTranslation} from '@atb/translations';
+import {dictionary, SectionTexts, useTranslation} from '@atb/translations';
 import composeRefs from '@seznam/compose-react-refs';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
 import {Error} from '@atb/assets/svg/color/icons/status';
-import dictionary from '@atb/translations/dictionary';
+import {giveFocus} from '@atb/utils/use-focus-on-load';
 
 type FocusEvent = NativeSyntheticEvent<TextInputFocusEventData>;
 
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):
Lgtm! :)

src/components/sections/items/TextInputSectionItem.tsx
```diff
@@ -102,40 +115,56 @@ export const TextInputSectionItem = forwardRef<InternalTextInput, TextProps>(
...
+        {errorText !== undefined && (
+          <View
+            style={styles.error}
+            accessibilityLiveRegion={'polite'}
+            accessibilityLabel={`${t(
+              dictionary.messageTypes.error,
+            )}, ${errorText}`}
+          >
+            <ThemeIcon svg={Error} />
+            <ThemeText type={'body__secondary'} style={styles.errorMessage}>
```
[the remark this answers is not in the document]
hanne-at-atb: Fixed both places :)

src/components/sections/items/TextInputSectionItem.tsx
```diff
@@ -102,40 +115,56 @@ export const TextInputSectionItem = forwardRef<InternalTextInput, TextProps>(
...
+              >
+                <ThemeIcon svg={Close} />
+              </PressableOpacity>
+            </View>
+          ) : null}
+        </View>
+        {errorText !== undefined && (
+          <View
+            style={styles.error}
+            accessibilityLiveRegion="polite"
```
gorandalum, line 157: Nice solution! I took a look at some aria doc to see if a specific role also would be appropriate, and think maybe role "alert" would be nice here. Some cliffs from the doc:
> The alert role is used to communicate an important and usually time-sensitive message to the user. When this role is added to an element, the browser will send out an accessible alert event to assistive technology products which can then notify the user.
>
> The alert role should only be used for information that requires the user's immediate attention, for example:
>
>- An invalid value was entered into a form field

Maybe check if when adding accessibility role `alert`, is the announcements done even without adding live region?

gorandalum (approved):
Very good!

src/components/sections/items/TextInputSectionItem.tsx
hanne-at-atb: Added `role: 'alert'` now. Does however look like accessibilityLiveRegion is only supported by Android, so exchanged it with focusRef that in turn will trigger the screen reader to read out error message when it appears for both android and ios.

src/components/sections/items/TextInputSectionItem.tsx
```diff
@@ -183,6 +184,7 @@ const useInputStyle = StyleSheet.createTheme((theme) => ({
     borderColor: theme.static.background.background_0.background,
   },
   containerInline: {
+    backgroundColor: 'red',
```
marius-at-atb, line 187: Remember to remove this one :)

PR #3886, opened 2023-09-21
minimal bottomsheet example

toretefre:
Trying to close https://github.com/AtB-AS/kundevendt/issues/8886.

Just a minimal example of what I'm trying to do now, not ready for merge.

### DIFF
--- a/src/components/sections/items/date-input/DateInputSectionItem.ios.tsx
+++ b/src/components/sections/items/date-input/DateInputSectionItem.ios.tsx
@@ -39,7 +39,7 @@ export function DateInputSectionItem(props: DateInputSectionItemProps) {
         display="compact"
         testID="dateInput"
         minimumDate={new Date()}
-        maximumDate={maximumDate}
+        maximumDate={undefined}
         onChange={(_, date) => {
           if (date) onChange(date.toISOString());
         }}
--- a/src/components/sections/items/date-input/DateInputSectionItem.ios.tsx
+++ b/src/components/sections/items/date-input/DateInputSectionItem.ios.tsx
@@ -9,7 +9,7 @@ import {DateInputSectionItemProps} from './utils';
 import {useLocaleContext} from '@atb/LocaleProvider';
 
 export function DateInputSectionItem(props: DateInputSectionItemProps) {
-  const {value, onChange, maximumDate, ...innerprops} = props;
+  const {value, onChange, ...innerprops} = props;
   const {t} = useTranslation();
   const locale = useLocaleContext();
   const {theme} = useTheme();
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
@@ -1,5 +1,9 @@
 import React from 'react';
-import {PurchaseOverviewTexts, useTranslation} from '@atb/translations';
+import {
+  PurchaseOverviewTexts,
+  TravelDateTexts,
+  useTranslation,
+} from '@atb/translations';
 import {ThemeText} from '@atb/components/text';
 import {InteractiveColor} from '@atb/theme/colors';
 import {StyleProp, View, ViewStyle} from 'react-native';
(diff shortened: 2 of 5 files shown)


src/components/sections/items/date-input/DateInputSectionItem.ios.tsx
```diff
@@ -39,7 +39,7 @@ export function DateInputSectionItem(props: DateInputSectionItemProps) {
         display="compact"
         testID="dateInput"
         minimumDate={new Date()}
-        maximumDate={maximumDate}
```
toretefre, line 42: This is set to mimic Android behaviour on iOS.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
```diff
@@ -89,6 +100,16 @@ export function StartTimeSelection({
           },
         ]}
       />
+      {showActivationDateWarning && (
```
toretefre, line 103: This one **is** shown immediately on date change

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
```diff
@@ -83,11 +116,21 @@ export const TravelDateSheet = forwardRef<ScrollView, Props>(
           <Section>
             <DateInputSectionItem
               value={dateString}
-              onChange={setDate}
+              onChange={onSetDate}
               maximumDate={maximumDate}
             />
             <TimeInputSectionItem value={timeString} onChange={setTime} />
           </Section>
+          {showActivationDateWarning && (
```
toretefre, line 124: This one is **not** showing immediately on date change, only when re-mounting the bottom sheet.

toretefre:
Hey @gorandalum (or code owner @marius-at-atb), 

Do you have any idea why state changes in parent root stack does not propagate in realtime to the `TravelDateSheet`? 

In this example, the yellow warning box in the `TravelDateSheet` should appear at the same time as the one in the background (`Root_PurchaseOverviewScreen.tsx`), but it does not appear until I re-open the `TravelDateSheet`. 

My hypothesis is that the `TravelDateSheet` in some way is disconnected from state updates until re-opening, but I can't see why.

See video and notice the yellow `MessageBox` appearing immediately behind the bottom sheet.

https://github.com/AtB-AS/mittatb-app/assets/21310942/69b1ecad-fd62-4478-90dc-7ad8f5196805

Let me know if I can clarify something. If you want to test with AtB tickets you could do so by setting `latestActivationDate` in limitations of a product in staging as done for [fram-staging Vaksen Buss](https://github.com/AtB-AS/firestore-configuration/blob/d7de8eae9f61d3bf91abdec1ddf92fa46504dbdd/documents/fram-staging/referenceData.json#L785C31-L785C31).

marius-at-atb:
@toretefre I think the props for TravelDateSheet are set when calling openTravelDateSheet. TravelDateSheet then updates its internal state while "alive", but openTravelDateSheet is not called reactively when the props have changed.
Not sure if we have an existing way to handle this @gorandalum?

If there is no good way to update the props from outside, I guess it is possible to create an internal state variable that mirrors showActivationDateWarning inside TravelDateSheet 😬

toretefre:
Good idea @marius-at-atb, I was thinking about something similar, but decided it might not be best practice.

There is a PR for this solution here, if possible it would be great to have a review before the weekend so we could go ahead with FRAM release: https://github.com/AtB-AS/mittatb-app/pull/3889

toretefre:
Closed in favour of https://github.com/AtB-AS/mittatb-app/pull/3889, thanks for good help!

gorandalum:
> @toretefre I think the props for TravelDateSheet are set when calling openTravelDateSheet. TravelDateSheet then updates its internal state while "alive", but openTravelDateSheet is not called reactively when the props have changed. Not sure if we have an existing way to handle this @gorandalum?
> 
> If there is no good way to update the props from outside, I guess it is possible to create an internal state variable that mirrors showActivationDateWarning inside TravelDateSheet 😬

In general we update the local state internally in the bottom sheet continuously, and update the external state when pressing save. This is the same as done in the `TravelSearchFiltersBottomSheet`.

PR #3888, opened 2023-09-22
fix: accessibility improvements ticket assistant

adriawh:
### PR targeting these comments:
 [Comment](https://github.com/AtB-AS/kundevendt/issues/4240#issuecomment-1727277990)
>Comments form testing.
Should the "header" billettveileder be in the app, is missing now compared with figma sketches.
Missing a space between dot and next sentence.
Is it a reason for both Back and close bottoms?
200% zoom the part with "hvor ofte" and "hvor lenge" gets ugly on 200% text size, for the numbers and calender.

 [Additional comment on screen reader focus](https://github.com/AtB-AS/kundevendt/issues/4240#issuecomment-1727465075). 
 >Skjermeleser kommentar.
Det har vært optimalt at fokuset til skjermleseren går opp til overskriften når man går til neste siden, nå er det litt bingo i hva som skjer når man kommer til neste side. En gang går den til overskriften, en gang er den nede neste knappen.

Regarding this comment:
> Is it a reason for both Back and close bottoms?

This was added after a decision from @Sebstorvik and marked.

Changes done to useFocusOnLoad discussed on slack with @marius-at-atb and @gorandalum 

### Changes: 
<details>
<summary>Added header</summary>

Simulator

</details>
<details>
<summary>Fix for 200% zoom</summary>

Simulator Screenshot - iPhone 15 Pro - 2023-09-22 at 11 27 54

</details>

- Fixed space after dot in child ticket
- With screen reader, focus now on header after navigating


src/stacks-hierarchy/Root_TicketAssistantStack/use-naviagtion-focus.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+export const useNavigationFocus = ({
+  navigation,
+  screenName,
+}: UseNavigationFocusProps) => {
+  const focusRef = useRef(null);
+
+  useEffect(() => {
+    if (
+      navigation.getState().routeNames[navigation.getState().index] ==
```
marius-at-atb, line 17: Whether the array lookup is a safe operation depends on the navigation state.
I'm not sure if it is always valid, and would add a safety check just for peace of mind ;)

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_CategoryPickerScreen.tsx
```diff
@@ -182,38 +189,38 @@ export const TicketAssistant_CategoryPickerScreen = ({
...
-                        style={styles.expandedContent}
-                        isMarkdown={true}
-                      >
-                        {description}
-                      </ThemeText>
-                    </View>
-                  </PressableOpacity>
-                </View>
+                <PressableOpacity
+                  key={index}
```
marius-at-atb, line 193: key should probably be `u.id`

marius-at-atb (changes requested):
Nice! Some initial comments

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_CategoryPickerScreen.tsx
```diff
@@ -20,8 +20,9 @@ import {useAccessibilityContext} from '@atb/AccessibilityContext';
 import {Traveller} from '@atb/stacks-hierarchy/Root_TicketAssistantStack/types';
 import {useTicketAssistantState} from '@atb/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistantContext';
 import {ExpandableSectionItem, Section} from '@atb/components/sections';
-import {useFocusOnLoad} from '@atb/utils/use-focus-on-load';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
+import {useNavigationFocus} from '@atb/stacks-hierarchy/Root_TicketAssistantStack/use-navigation-focus';
+import {TICKET_ASSISTANT_CATEGORY_PICKER_SCREEN} from '@atb/stacks-hierarchy/Root_TicketAssistantStack/Root_TicketAssistantStack';
```
jorelosorio, line 25: We use this kind of upper-cased words for constant, not sure if this is the case.

jorelosorio (approved):
Looks good.

src/stacks-hierarchy/Root_TicketAssistantStack/use-navigation-focus.ts
gorandalum, line 1: Could you write some about why this `useNavigationFocus` is needed, and what differs it from `useFocusOnLoad`? In general it is not beneficial for the app to have similar hooks that does almost the same thing. Maybe it can instead be a param to `useFocusOnLoad`?

gorandalum (changes requested):

src/stacks-hierarchy/Root_TicketAssistantStack/use-navigation-focus.ts
marius-at-atb: Regarding `useFocusOnLoad`, maybe we should use `useFocusEffect` instead?
https://reactnavigation.org/docs/use-focus-effect/#how-is-usefocuseffect-different-from-adding-a-listener-for-focus-event

As `useFocusOnLoad` is currently implemented, `giveFocus` is triggered when `focusRef` is set, without checking if the component is part of a page that is in focus. So e.g. if a component is created in the DOM, but is part of a different tab than the one in focus, it will still trigger `giveFocus` (unless manually updating and overriding with `setFocusOnLoad`).

However if we skip `useFocusEffect`, and instead update `useFocusOnLoad` with a param as suggested by @gorandalum, then I guess that integration can also fix the issue I just described at the same time.

marius-at-atb (commented):
Some details and a topic of discussion regarding the hook

marius-at-atb (approved):
Very nice!

gorandalum (approved):

PR #3889, opened 2023-09-22
fix: enforce latestActivationDate also on Android

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/8886

Based on suggestion from https://github.com/AtB-AS/mittatb-app/pull/3886#issuecomment-1731197343

Enforces `latestActivationDate` also on Android where the `maximumDate` prop is not respected in our datepicker.

If a date later than the latest activation date is entered, a yellow warning `MessageBox` will appear and the submit button will be disabled. These measures are reverted when a valid date before the latest activation date is set, or when a different category / product is selected.

https://github.com/AtB-AS/mittatb-app/assets/21310942/ad92f541-c38d-4419-892b-6bc7c3b97ba0

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
@@ -1,5 +1,9 @@
 import React from 'react';
-import {PurchaseOverviewTexts, useTranslation} from '@atb/translations';
+import {
+  PurchaseOverviewTexts,
+  TravelDateTexts,
+  useTranslation,
+} from '@atb/translations';
 import {ThemeText} from '@atb/components/text';
 import {InteractiveColor} from '@atb/theme/colors';
 import {StyleProp, View, ViewStyle} from 'react-native';
@@ -12,6 +16,7 @@ import {useBottomSheet} from '@atb/components/bottom-sheet';
 import {TravelDateSheet} from '@atb/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet';
 import {RadioSegments} from '@atb/components/radio';
 import {TimeSelectionMode} from '@atb/configuration';
+import {MessageBox} from '@atb/components/message-box';
 
 type StartTimeSelectionProps = {
   color: InteractiveColor;
@@ -95,12 +100,25 @@ export function StartTimeSelection({
           },
         ]}
       />
+      {showActivationDateWarning && (
+        <MessageBox
+          style={styles.warningBox}
+          type={'warning'}
+          message={t(
+            TravelDateTexts.latestActivationDate
+              .selectedDateShouldBeEarlierWarning,
+          )}
+        />
+      )}
     </View>
   );
 }
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   radioSegments: {
+    marginTop: theme.spacings.medium,
+  },
+  warningBox: {
     marginTop: theme.spacings.medium,
   },
 }));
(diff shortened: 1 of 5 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
```diff
@@ -97,4 +103,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
   radioSegments: {
     marginTop: theme.spacings.medium,
   },
+  warningBox: {
```
marius-at-atb, line 106: Is this being used?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
```diff
@@ -29,15 +30,53 @@ type Props = {
...
+  ) => {
     const {t, language} = useTranslation();
     const styles = useStyles();
 
     const defaultDate = travelDate ?? new Date().toISOString();
     const [dateString, setDate] = useState(defaultDate);
+    const [
+      replicatedShowActivationDateWarning,
+      setReplicatedShowActivationDateWarning,
+    ] = useState<boolean>(false);
```
marius-at-atb, line 57: Should init to showActivationDateWarning I think

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
```diff
@@ -29,15 +30,53 @@ type Props = {
...
+    };
+
+    const onSetDate = (date: string) => {
+      if (!maximumDate) setDate(date);
+      else {
+        if (isAfter(date, maximumDate)) {
+          setInternalAndExternalWarningState(true);
+        } else if (
+          replicatedShowActivationDateWarning ||
+          showActivationDateWarning
```
marius-at-atb, line 71: When replicatedShowActivationDateWarning is initted to showActivationDateWarning, `|| showActivationDateWarning` should not be needed here

marius-at-atb (approved):
Very nice! The comments are just some nitpicks.
I agree with you [here](https://github.com/AtB-AS/mittatb-app/pull/3886#issuecomment-1731341612) that the state replication is probably not best practice. At the very least maybe make it in a generalised reusable way for all components opened with openBottomSheet. However, for now it is an easy enough solution that works I guess..

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/StartTimeSelection.tsx
toretefre: Nope, there was a message for testing, but removed it. Good catch 😄

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravelDate/TravelDateSheet.tsx
toretefre: Smart!

toretefre:
> Very nice! The comments are just some nitpicks. I agree with you [here](https://github.com/AtB-AS/mittatb-app/pull/3886#issuecomment-1731341612) that the state replication is probably not best practice. At the very least maybe make it in a generalised reusable way for all components opened with openBottomSheet. However, for now it is an easy enough solution that works I guess..

Yep, but I think it makes more sense to make it general when we have another use case. 

For now, merging as-is. Thanks for the help! 🙌

gorandalum (commented):
Just a general comment from me, and not something that needs to be changed in a PR :)

My intuitive thought is that the `showActivationDateWarning` is really derived state, based on the selected date and the latest activation date, and as such doesn't need to be its own state, neither in the sheet nor in the purchase overview.

Quote from [the doc](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state):
> When something can be calculated from the existing props or state, [don’t put it in state.](https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state) Instead, calculate it during rendering.

And I also know that there is a difference between theory and practice, so it might still be needed here, even though I don't see it on first glance 😊

PR #3890, opened 2023-09-22
fix: Error while encoding secure container on iOS

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/5196

Upgrade Token SDK Version to `1.3.3` that includes a fix.

## Context

This Fix was made on the Entur and released on the current version.

There was a concern with the method `DCAppAttestService.shared.generateAssertion` being invoked too frequently, which could interrupt the encoding process. Additionally, the [React Native documentation](https://reactnative.dev/docs/native-modules-android.html#callbacks) suggests that callbacks should be invoked only once but can be stored for future use.

### Solution Overview

**Singleton Design:** We transformed the `TokenEncoder` class into a singleton pattern. This ensures that there's only a single shared instance of the encoder throughout the app's lifecycle, ensuring centralized handling of all encoding tasks.

**Request Queue Mechanism:** We introduced a queuing mechanism to manage incoming encoding requests.
> When a new encoding request arrives, it's added to the `pendingRequests queue`.
If the queue was empty prior to the addition, the request processing starts immediately.
Subsequent requests wait in the queue until the ongoing request completes.

**Sequential Processing:** The requests are processed sequentially. After finishing one request, the next in line is picked from the queue and processed. This ensures that `DCAppAttestService.shared.generateAssertion` is not invoked concurrently, avoiding potential disruptions in the encoding.

**Thread Safety:** All interactions with the `pendingRequests` queue occur within a dedicated `DispatchQueue`, ensuring thread safety. This means multiple threads won't interfere or disrupt the order or processing of the tasks.

**Adherence to Callback Constraints:** By using this queue mechanism, we ensure that each callback associated with a request is called exactly once, in line with React Native's guidelines. This design not only satisfies the callback constraint but also ensures that callbacks are executed in the order their associated requests arrived.


hanne-at-atb (approved):
This looks good :)

gorandalum (approved):
Looks good! To be in sync with the token-sdk versions, could you also create a PR for the token-server? Just connect it to the same issue.

PR #3892, opened 2023-09-22
fix: travel search cleanup and animation 

adriawh:
** Updated after comment from @gorandalum **

### Changes:
- Moved favorites on the same line as location and map buttons
- Only show search results when typing
- Animation between states using `animateNextChange`
- Moved info box within scrollview to simplify logic

Video:
https://github.com/AtB-AS/mittatb-app/assets/70323886/7d2ffed3-ceaa-4a67-bc50-092ec16db546


gorandalum (changes requested):
I think this solution may be too complex for what it is trying to solve. It introduces animations, timeouts, listening to chip layout changes, and some more. I know it might be necessary to introduce those complexities to fully implement the design as described, but we should take a foot in the ground and think if it is really worth it. From experience we know animations is hard to both write and read, and also they can be buggy and brittle.

My first test would be how will it look without any custom animations. If we instead simply just not render the chip-components when the user starts typing, in combination with adding the `animateNextChange`. I would like to see a video of how that looks :)

marius-at-atb:
Even though animations can be brittle and introduce more complexity, I think it is a challenge we should accept.
The best apps tend to be very good at subtle animations that produce a super fluent feel.

adriawh:
@marius-at-atb 

Were able to create animations with a  "cheaper" method, and also simplified the logic in my latest changes

gorandalum (approved):
Looks good!

src/favorites/FavoriteChips.tsx
```diff
@@ -39,52 +39,51 @@ export const FavoriteChips: React.FC<Props> = ({
...
         keyboardShouldPersistTaps="handled"
-        contentContainerStyle={
+        contentContainerStyle={[
           (activeType('favorites') && favorites.length > 0) ||
           activeType('add-favorite')
             ? style
-            : undefined
-        }
+            : undefined,
+          style,
```
marius-at-atb, line 59: If adding style anyway, there is no need for the conditional check above. Or did you mean to remove line 59?

marius-at-atb (approved):
I like it!
Bonus point if you can get the travel search results section above the buttons:
Simulator Screenshot - iPhone 14 - 2023-09-26 at 14 47 52

PR #3896, opened 2023-09-26
feat: Add Kettle SDK

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/8614

Changes made on this PR:

- The Android Manifest needs a new attribute to be added (KETTLE_API_KEY) but it might be only needed by AtB, so in order to handle this, the manifest was moved to the env/org folder and is copied to the android/src/main and append the API if is present in the .env files.
> All this happens automatically in the yarn setup command.
- Add SDK, for iOS if the API KEY is not found, then it does not include.


gorandalum (changes requested):
Have you looked at how we add Bugsnag key to AndroidManifest? If we could do it the same, we don't need to create new files for each env. It is in the step "Adding Bugsnag API key and release stage to AndroidManifest.xml" in `scripts/android/override-config-files.sh`.

jorelosorio:
> Have you looked at how we add Bugsnag key to AndroidManifest? If we could do it the same, then we wouldn't need to create new files for each env. It is in the step "Adding Bugsnag API key and release stage to AndroidManifest.xml" in `scripts/android/override-config-files.sh`.

Hi, yes but that is for staging / production, I had to move it to another place because we also have dev and prod keys.

jorelosorio:
> Have you looked at how we add Bugsnag key to AndroidManifest? If we could do it the same, then we wouldn't need to create new files for each env. It is in the step "Adding Bugsnag API key and release stage to AndroidManifest.xml" in `scripts/android/override-config-files.sh`.

Hi, @gorandalum I made fixes for this, and it is no longer needed to have a manifest for each org, we are using .env inside the Manifest.xml file instead.

ios/atb/Info.plist
```diff
@@ -85,6 +87,8 @@
 	<string>Ved å tillate deling av posisjon kan du finne nærmeste holdeplass og planlegge reisen fra din lokasjon</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Ved å tillate deling av posisjon kan du finne nærmeste holdeplass og planlegge reisen fra din lokasjon</string>
+	<key>NSMotionUsageDescription</key>
+	<string>Du gir AtB tillatelse til å bruke bevegelsesdata til analyse og større større forståelse av reisemønster, f.eks. for å se forskjell på sykling, gåing og bilkjøring.</string>
```
gorandalum, line 91: Will this affect the presentation in App Store for all transport companies?

android/app/proguard-rules.pro
```diff
@@ -12,4 +12,5 @@
 -keep class com.facebook.jni.** { *; }
 -keep class com.swmansion.reanimated.** { *; }
 -keep class com.facebook.react.turbomodule.** { *; }
-
+# For more details, see the requirements setting for Kettle at https://developer.kogenta.com/docs/react-native/add-kettle#android
+-keepclassmembers class com.kogenta.kettle.core.model.** { *; }
```
gorandalum, line 16: Will this actually have any effect? We don't have enabled proguard in production build now, as far as I know.

jorelosorio: I wasn't sure if is in use or not, but there are also another stuff set there, or Mapbox and Facebook, so we might be using it.

ios/atb/Info.plist
jorelosorio: The attribute is there but... if is not requested in the permission prompt, then it never appears, but AtB is the only one that uses it.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DebugInfoScreen.tsx
```diff
@@ -572,6 +593,22 @@ export const Profile_DebugInfoScreen = () => {
             }
           />
         </Section>
+
+        {isBeaconsEnabled && KETTLE_API_KEY && (
```
marius-at-atb, line 597: Should be `!!KETTLE_API_KEY` to ensure a boolean value I think

jorelosorio: KETTLE_API_KEY is a string but might be undefined, this condition just to make sure is not undefined.

android/app/proguard-rules.pro
gorandalum: At least we have `def enableProguardInReleaseBuilds = false` in build.gradle now. If Kogenta has a requirementet that the app should be obfuscated with proguard, we would need to look into enabling it.

It was originally planned in this issue, but not done yet:
https://github.com/AtB-AS/kundevendt/issues/3744

jorelosorio: Removed for now then ;)

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DebugInfoScreen.tsx
jorelosorio: Fixed!!

marius-at-atb: Nice!!

android/app/proguard-rules.pro
gorandalum: Aight. My comment wasn't mainly on whether this should be included or not, just that we need to be clear in communications with Kogenta that our app is not obfuscated on Android, and whether they have that requirement or not.

jorelosorio: In their SDK, they have mentioned: if Proguard is used, then add those rules. Doesn't sound like it is a requirement.

gorandalum: Ok. Maybe check it up to be sure.

ios/atb/Info.plist
gorandalum: This is what we have in App Store for the AtB app today. I wondered if adding this would change something there.
Screenshot 2023-10-02 at 13 57 01

marius-at-atb (approved):
Lgtm! Just two minor comments

ios/atb/Info.plist
jorelosorio: I don't think so, there is no one related to that motion tracking however is attached to the location as the documentation mention: https://developer.apple.com/documentation/coremotion/cmmotionactivitymanager for Apple and Kettle https://developer.kogenta.com/docs/react-native/usage#ios I do not know how they manage this.

gorandalum (approved):

marius-at-atb (approved):
Nice!

tormoseng:
@jorelosorio How are we using the SDK currently? Are we sure that the SDK is not exposed in our open-source repo? It's a worry that Olga brought up.

android/app/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,7 @@
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
     <!-- API level higher than 28, you need to explicitly request Manifest.permission.ACTIVITY_RECOGNITION during onboarding. If this doesn't work, you may have to explicitly request com.google.android.gms.permission.ACTIVITY_RECOGNITION instead. -->
     <!-- For more details: https://developer.kogenta.com/docs/react-native/usage#kettle-motion-tracking -->
-    <uses-permission android:name="com.google.android.gms.permission.ACTIVITY_RECOGNITION"/>
+    <uses-permission android:name="com.google.android.gms.permission.ACTIVITY_RECOGNITION" android:enabled="${ENABLE_ACTIVITY_RECOGNITION}" />
```
marius-at-atb, line 9:
```suggestion
    <uses-permission android:name="com.google.android.gms.permission.ACTIVITY_RECOGNITION" tools:node="${ACTIVITY_RECOGNITION_NODE_VALUE}" />
```

android/app/build.gradle
```diff
@@ -80,11 +80,17 @@ android {
...
+                ENABLE_ACTIVITY_RECOGNITION: "true"
+            ]
         } else {
-            // If this value is not set, an exception will be raised at runtime. Refer to the documentation at https://developer.kogenta.com/docs/react-native/add-kettle#android for more information.
+            // If `KETTLE_API_KEY` value is not set, an exception will be raised at runtime. Refer to the documentation at https://developer.kogenta.com/docs/react-native/add-kettle#android for more information.
             // NOTE: To avoid breaking the app and to prevent the SDK from initializing, you can use a value other than an API key.
-            manifestPlaceholders = [KETTLE_API_KEY: "NoKettleIdNeeded"]
+            manifestPlaceholders = [
+                KETTLE_API_KEY: "NoKettleIdNeeded",
+                ENABLE_ACTIVITY_RECOGNITION: "false"
```
marius-at-atb, line 92:
```suggestion
                ACTIVITY_RECOGNITION_NODE_VALUE: activityRecognitionBool ? "merge" : "remove"
```

And also add this further up:
```
def activityRecognitionEnv = System.getenv("ENABLE_ACTIVITY_RECOGNITION")
def activityRecognitionBool = activityRecognitionEnv != null && activityRecognitionEnv.equalsIgnoreCase("true")
```

And then set `ENABLE_ACTIVITY_RECOGNITION` in the env files

marius-at-atb (commented):
I'm not sure if `android:enabled` will work on uses-permission tags?
Solution with tools:node suggested if needed.

android/app/build.gradle
jorelosorio: Is true is by using tool:node, I made the fix. We do not need an env var for this, we just use if kettle is set or not.

jorelosorio: It is removed when the Kettle SDK Key is not present, which is the desired. is Inside a condition. Not sure if I understood you, if is something else.

marius-at-atb: Sorry yes it should be working now. Somehow I didn't notice the System.getenv()["KETTLE_API_KEY"] check earlier..

gorandalum (approved):

KristianSelnas:
@jorelosorio @gorandalum  Ooops, sorry guys. I merged this PR by mistake. I had too many tabs open...

jorelosorio:
I will create a revert hehe is broken now

PR #3897, opened 2023-09-26
fix: TariffZone map recenters on location change

gorandalum:
- Calculating starting coordinates only at mount, so they aren't
  affected by geolocation change.
- Added UserLocation pin on the map
- Pressing the "My location" button no longer auto-selects the zone
  your position is in.


marius-at-atb (approved):
Lgtm!

PR #3900, opened 2023-09-28
feat: adds timeout notification for token and fallback to static aztec

mikaelbr:
I've added a new RemoteConfig prop defaulting to 0 so we can opt-in on the timeout behavior: `token_timeout_in_seconds`. If this is 0 timeout will always be false.

This feature adds a timeout flag to the token context. It does not stop the execution of the code, as I think that is hard to stop the execution flow like this and I don't think it is favorable either. So the logic here is to have a timeout mechanism that marks the loading as timed out and if it proceeds afterwards it will unmark it as timed out making it proper again.

I've also added a BugSnag notification when the timeout handler is invoked. This will allow us to see how often this happens and all the breadcrumbs that the token SDK cases will get posted to BugSnag. Hopefully, we'll get more context and information on what is going on.

I have only now checked for timeout in the Barcode component. Better error messages to the user could be added to other places that uses token context also, but trying to minimize development time here.

NB: I haven't verified this myself yet, but adding as PR while I do so to save some time and get more eyes on this.

## Verification of PR

- [ ] Run device and let token take a long time. Should trigger timeout mechanism.

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -37,6 +37,7 @@ type MobileTokenContextState = {
   remoteTokens?: RemoteToken[];
   deviceIsInspectable: boolean;
   isLoading: boolean;
+  isTimedout: boolean;
   isError: boolean;
   getSignedToken: () => Promise<string | undefined>;
   toggleToken: (
@@ -61,6 +62,7 @@ const MobileTokenContext = createContext<MobileTokenContextState | undefined>(
 export const MobileTokenContextProvider: React.FC = ({children}) => {
   const {abtCustomerId, authStatus} = useAuthState();
 
+  const {token_timeout_in_seconds} = useRemoteConfig();
   const hasEnabledMobileToken = useHasEnabledMobileToken();
 
   const {enable_token_fallback: fallbackEnabled} = useRemoteConfig();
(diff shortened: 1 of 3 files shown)


gorandalum (approved):
Think this looks ok as a quick fix for this :+1:

mikaelbr:
I've tested this locally. It works kind of like intended. Compact viewings and overviews of tickets show "Not inspectable" and bugsnag gets triggered, if the token is fetched after it updates and fixes it self. I think this is a good step 1 and then work on better user messaging like showing fallback solution on every component where tokens are used.

src/mobile-token/MobileTokenContext.tsx
```diff
@@ -224,10 +252,14 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
...
+        setTimeoutFlag(false);
+
         setIsLoading(false);
       }
     }
-  }, [enabled, loadNativeToken, loadRemoteTokens]);
+  }, [enabled, loadNativeToken, loadRemoteTokens, token_timeout_in_seconds]);
 
   useEffect(() => {
     load();
```
jorelosorio, line 265: not sure if we should return in this useEffect something like

```
const cancelTimeoutHandler = load();
return () => cancelTimeoutHandler();
```

mikaelbr: Hmm. I think that either its cancelled in the `finally` block, or it is executed in a sense. I think that in any sense (even if the useEffect block changes dependency) the current execution flow will finish and not stop "mid execution" if that makes sense

PR #3901, opened 2023-09-28
refactor: Always log PostHog events on loading timeout

gorandalum:
- Introduced a new RemoteConfig variable 'enable_loading_error_screen'
  to specify whether error screen should be shown or not after loading
  screen. If set to false, the user will get "normal app" even if
  loading error/timeout.
- PostHog events add auth status and is app state loading metadata.
- Logic regarding timout was moved from AuthContext to LoadingBoundary.


rosvik (approved):

tormoseng:
Looks OK on the staging build for 1.41.1 (Android). Needs to be tested on iOS.
- [ ] PostHog event
- [ ] Error screen remote config

A small observation is that if a user logs in with missing claims and enter the app, no customer number is shown (log out option is there). If I fix the claims ([here](https://atb-staging.internal.mittatb.no/user-claims/?uid=MmDIsjz423WjeRWhKHsM8VPzkNE2)) and close/re-open the app, I enter the app with an anonymous user.

tormoseng:
Checked out the behaviour above. Only happens in my scenario.

PR #3907, opened 2023-09-29
fix: specify kotlin version, enable force build option

marius-at-atb:
Enable a force build option for Android in staging.

Also, when building the Android app, the warning "Detected multiple Kotlin daemon sessions at build/kotlin/sessions" is fixed by specifying a kotlin version. See https://github.com/facebook/react-native/issues/33488

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,6 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "33.0.0"
+        kotlin_version = '1.8.0'
         minSdkVersion = 23
         compileSdkVersion = 33
         targetSdkVersion = 33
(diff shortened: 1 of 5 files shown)


android/build.gradle
```diff
@@ -3,6 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "33.0.0"
+        kotlin_version = '1.8.0'
```
jorelosorio, line 6: Does it matter if we use `1.9.10`? :)

marius-at-atb: I got the version number from here:
https://github.com/AtB-AS/mittatb-app/blob/13d7e7f05292966e4faf28b4ebdd63fbe3bcb820/android/app/build.gradle#L141-L142
I think this is correct?
Or do you mean if we should upgrade?

jorelosorio (approved):
Good work!

PR #3909, opened 2023-09-29
 feat: Retry auth resubscribes to Firebase auth

gorandalum:


src/auth/use-subscribe-to-auth-user-change.ts
```diff
@@ -25,5 +29,9 @@ export const useSubscribeToAuthUserChange = (
       }
     });
     return () => unsubscribe();
-  }, []);
+  }, [resubscribeToggle]);
+
+  return {
+    resubscribe: useCallback(() => setResubscribeToggle((s) => !s), []),
```
jorelosorio, line 35: We could just send back directly a function without a use Callback, or it is a function that runs once only?

gorandalum: In general it is good practice to use `useCallback` on returned functions and `useMemo` on returned objects from hooks, so it is not recreated on every render. I think this is a good principle even though the current usage doesn't necessarily use it in a hook (like `useEffect`), since it diminishes the risk for later introducing an error.

And, actually your comment made me realise that the `resubscribe` function is not in the dependency list of the `useCallback` where it is used, which it should be. I'll fix that :)

PR #3916, opened 2023-10-04
feat: empty states template

marius-at-atb:
Add component `EmptyState` which is used by the new `NoActiveTickets` for when there are no tickets:

Using `EmptyState` for Travel Search and Departures will be in another PR.

Closes https://github.com/AtB-AS/kundevendt/issues/7302


src/fare-contracts/NoActiveTickets.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <EmptyState
+      title={t(
+        TicketingTexts.activeFareProductsAndReservationsTab.noActiveItemsTitle,
+      )}
+      details={t(
+        TicketingTexts.activeFareProductsAndReservationsTab
+          .noActiveItemsDetails,
+      )}
+      illustrationComponent={
+        <TicketTilted height={theme.icon.size.large * 4} />
```
rosvik, line 20: I don't think it makes sense to couple this size to icon sizes: If we change the large icon size, I don't think we want this illustration to change size. But maybe we should add sizes for illustrations in the design system at some point.

For now it might be better to hard code a size here?

rosvik (commented):
Good work! Nice idea to have a dedicated empty state component! Is there any other places in the app we can use it going forward?

Noticed the same component, with the same texts, is used for the active fare contract list and ticket history. Will it not be misleading to say "No active tickets" on an empty ticket history page?

src/fare-contracts/NoActiveTickets.tsx
marius-at-atb: Good point, I just used the biggest icon size I could find, but it wasn't enough 😛
Will fix :+1:

marius-at-atb:
@rosvik  The `EmptyState` component will be used in 3 places to begin with. PRs for the other two will follow:
Skjermbilde 2023-10-04 kl  14 38 35

marius-at-atb:
@oyvindmikkelsenatb See [the comment by @rosvik](https://github.com/AtB-AS/mittatb-app/pull/3916#pullrequestreview-1657265925) above.
Should the text "No active tickets" say something else when you have never purchased a ticket before?

rosvik:
> @oyvindmikkelsenatb See [the comment by @rosvik](https://github.com/AtB-AS/mittatb-app/pull/3916#pullrequestreview-1657265925) above. Should the text "No active tickets" say something else when you have never purchased a ticket before?


@marius-at-atb I was thinking of the Ticket History page on My Profile:

marius-at-atb:
aha, my bad!
Yes, with the current changes, it looks like this:


So change "No active tickets" to e.g. "No tickets purchased yet"?

marius-at-atb:
Update: good thing you noticed @rosvik!
With the latest changes, the screens look like this:

rosvik (approved):
Smooth!

PR #3917, opened 2023-10-04
fix: update section separator color

rosvik:
One for the pixel-peepers 🧐:

Since most places where sections are used are on a background_2 color, I'm updating the SectionSeparator to look more correct in the places we use it today, and with slightly more contrast.

### Before/after
Screenshot 2023-10-04 at 10 06 31
Screenshot 2023-10-04 at 10 06 25


marius-at-atb (approved):
:+1:

PR #3923, opened 2023-10-10
feat: adds device brightness adjustment on ticket details sreen

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/9404

This _SHOULD_ adjust brightness when opening the screen for control to read the QR code better. In many cases the screen is too dark to read properly, and instead of having users adjust it manually we temporary try to adjust brightness when opening screen for inspection.

## Important

- This is impossible to test on simulator and emulator. We need to find a way to test this on the device and see if it works and if it is feasible. 
- The dependency used here seems old and not very well maintained. However, there is no other alternative and for now the implementation seems so simple and straightforward that our own implementation would be the same. So reducing administrative overhead and resisting the urge of "not invented here"-isms. I decided to propose this package.
- We MIGHT need permissions on Android. It hard to say without testing as some documentations differ.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1188,6 +1188,8 @@ PODS:
   - React-jsinspector (0.72.4)
   - React-logger (0.72.4):
     - glog
+  - react-native-device-brightness (1.2.7):
+    - React
   - react-native-flipper (0.191.0):
     - React-Core
   - react-native-flipper-performance-plugin (0.4.0):
@@ -1324,7 +1326,7 @@ PODS:
     - React-perflogger (= 0.72.4)
   - RNBootSplash (4.4.1):
     - React-Core
-  - RNCAsyncStorage (1.17.11):
+  - RNCAsyncStorage (1.19.3):
     - React-Core
   - RNCClipboard (1.11.1):
     - React-Core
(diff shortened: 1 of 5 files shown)


mikaelbr:
@jorelosorio Adding you here as I think you had something to do with RAAsyncStorage? For some reason my podlock file includes change for new version here. I've tried clearing stuff so it seems that it should be there and there is an issue with existing podlock file?

gorandalum (approved):
Looks good. Could also contemplate to use the hook in the actual QR code component instead of the Details screen.

Also, would it be benficial to use the hook `useIsFocusedAndActive` and use the value in the dependency list? This could improve handling when the component/screen is unfocused or the app is backgrounded? I haven't tested this myself, so this might already be handled by the library for all I know.

mortennordseth (approved):
I've tested this on my iPhone and the screen brightness adjustment works 🎉 The screen brightness is set to 100% when I open ticket details. When I close ticket details it's set to the initial brightness level. If I close the app or put it in the background while the ticket details screen is open the brightness stays at 100% (as expected)

mikaelbr:
I've added resetting on inactive now. I don't think it's a perfect solution and might cause the reset to run several times, but I don't think it has any real side-effects here, so no need for a complicated locking mechanism - just allow the reset to run multiple times.

PR #3924, opened 2023-10-10
feat: add edit profile screen to my profile

hanne-at-atb:
Presenting a new page with profile details!
- Includes possibility to edit firstName, surname and email
- Validation of email adresses
- Status on submission
- Uses endpoints from profile service to get and patch profile data. A minor refactor to add them along with profile delete in api/profile.ts
- Showing phone number and customer number form AuthContext
- Entry point in Root_ProfileScreen has been changed from delete profile to edit profile. Existing logic / screen entry for deleting profile has been moved into edit profile page and can be accessed through a button.

Updated:
- Using new api endpoints with TanStack Query
- Added support for handling upstreamError is axios client 
- Using upsteamErrors to show error message on email input regarding emails already in use.

Closes: https://github.com/AtB-AS/kundevendt/issues/4275


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,324 @@
...
+  };
+
+  const fetchProfile = useCallback(async () => {
+    try {
+      const response = await getProfile();
+      setCustomerProfile(response);
+
+      prefillPersonalia(response.email, setEmail);
+      prefillPersonalia(response.firstName, setFirstName);
+      prefillPersonalia(response.surname, setSurname);
```
marius-at-atb, line 111:
```suggestion
      const nonUnderscoreString = (str: string) => (str === '_' ? '' : str);
      setEmail(nonUnderscoreString(response.email));
      setFirstName(nonUnderscoreString(response.firstName));
      setSurname(nonUnderscoreString(response.surname));
```
You can consider if you like this solution. It is a bit shorter because it allows `prefillPersonalia` to be removed.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,324 @@
...
+      ) : (
+        <>
+          <View style={styles.personalDetails}>
+            {profileState === 'loading' && <ActivityIndicator size="large" />}
+            <ThemeText accessibilityRole="header" color="secondary">
+              {t(EditProfileTexts.personalDetails.header)}
+            </ThemeText>
+          </View>
+          {profileState === 'error' ? (
+            <MessageBox
```
marius-at-atb, line 158: This box should probably also be wrapped with a `<Section withPadding>` I think

marius-at-atb (approved):
Very smooth with clear and easy to read code :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,324 @@
...
+      )}
+    </FullScreenView>
+  );
+};
+
+function isValidEmail(email: string) {
+  // Just a really simple check to avoid the most obvious wrong ones.
+  // More validation is done on the server
+  return /^\S+@\S+\.\S+$/.test(email);
+}
```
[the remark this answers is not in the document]
rosvik: I guess this is relevant here: https://mittatb.slack.com/archives/C0116FMPX4Y/p1700468560795789

We have an util function, though it isn't perfect.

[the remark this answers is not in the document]
marius-at-atb: Relevant indeed. Should probably use [the existing util function](https://github.com/AtB-AS/mittatb-app/blob/master/src/utils/validation.ts) then.
If it has issues, it could be worth checking whether those are solved using the suggestion from StackOverflow above (it has quite a lot of upvotes).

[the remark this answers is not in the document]
hanne-at-atb: This was a raw copy of the email validation we have in Webshop for updating profile. 
Ref. https://github.com/AtB-AS/webshop2/blob/793c5b0ca114454961f70f5846f399d3570a4637/src/pages/profile/profile-info.tsx#L152 
Used it here as we stated in the issue (https://github.com/AtB-AS/kundevendt/issues/4275) to use the same input validation in edit profile screen in the app as in the existing one in webshop. 
Do you still think we should change it to what you suggest @marius-at-atb ? If so, what about webshop? :)

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,324 @@
...
+        } catch {
+          setSubmissionStatus('submission-error');
+        }
+      } else {
+        setSubmissionStatus('unavailable-email');
+      }
+    } else {
+      setSubmissionStatus('invalid-email');
+    }
+  };
```
gorandalum, line 79: We should use react-query now that we have added it to the app. And I think this will also be a little cleaner when using it, as loading states and such are handled by the lib.

gorandalum (changes requested):
I think it looks good! However I think we shouldn't introduce new api endpoints now without using react-query.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
gorandalum: `Section` should only be used around section items (`ButtonSectionItem`, `LinkSectionItem`) etc.

marius-at-atb: Right. Maybe just wrap it with a `View` then, to get that horizontal padding

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
hanne-at-atb: I nominate react query then for a team talk!
Will look at changing it here then 🌜

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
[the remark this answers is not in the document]
tormoseng: Have tested with some weird, but valid, addresses - and both the util function and the Stackoverflow seems to handle them. However, even though there are few (if any?) top level domains ending with a digit, I guess it's actually allowed as I interpret [RFC 3696](https://datatracker.ietf.org/doc/html/rfc3696#section-2):

```
a domain name must consist of only the ASCII alphabetic and numeric characters, plus the hyphen....
There is an additional rule that essentially requires that top-level domain names not be all-numeric.
```

Either way, I guess both works for all practical reasons. 

When it comes to testing, we could add a unit test for this case. There is also a possibility to add component tests through the e2e-framework Cypress to test the email input field in isolation. Will look into such things a bit later.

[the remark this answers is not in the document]
hanne-at-atb: Concluded with using the same validation as Entur, so swapped it with their Regex 📧

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
hanne-at-atb: Updated the code to use TanStack Query now 🚀

src/queries/update-profile-query.ts
```diff
@@ -1,12 +1,17 @@
...
 export const useProfileUpdateMutation = () => {
+  const queryClient = useQueryClient();
+
   return useMutation({
     mutationKey: ['patchProfile'],
     mutationFn: (profile: CustomerProfileUpdate) => {
       return updateProfile(profile);
     },
+    onSuccess: () => {
+      return queryClient.invalidateQueries({queryKey: ['getProfile']});
```
gorandalum, line 14: I think the `getProfile` string should be as a constant somewhere, maybe exported from the `get-profile-query`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,296 @@
...
+      return t(EditProfileTexts.personalDetails.email.unavailableError);
+    } else if (invalidEmail) {
+      return t(EditProfileTexts.personalDetails.email.formattingError);
+    }
+  };
+
+  useEffect(() => {
+    if (customerProfile) {
+      const profile: CustomerProfile = customerProfile;
+      const nonUnderscoreString = (str: string) => (str === '_' ? '' : str);
```
gorandalum, line 80: Some doc here? As it is kinda hard to understand why it is being done when people doesn't know that this underscore stuff is a "hack" at Entur.

gorandalum (approved):
Good work! :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_EditProfileScreen.tsx
```diff
@@ -0,0 +1,296 @@
...
+    } else {
+      setInvalidEmail(true);
+    }
+  };
+
+  const getEmailErrorText = (
+    invalidEmail: boolean,
+    errorOnUpdate: any,
+  ): string | undefined => {
+    if (errorOnUpdate?.status === 602) {
```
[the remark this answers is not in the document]
hanne-at-atb: Martin did mention exposing this error from Entur, and also it's the exact same check that is done in Webshop: https://github.com/AtB-AS/webshop2/blob/793c5b0ca114454961f70f5846f399d3570a4637/src/server-api-service/utils.ts#L215

PR #3926, opened 2023-10-10
feat: support via parameter

marius-at-atb:
Support the via parameter in destinationDisplay from Entur data.
Resolves https://github.com/AtB-AS/kundevendt/issues/2732
Related BFF-PR: https://github.com/AtB-AS/atb-bff/pull/301

Previously only frontText was used to show lineName.
Now it supports joining together frontText and destinations in via. This is done in many places with the `getDestinationLineName` function.

Also for favourites, instead of using lineName as the identifier, it now uses destinationDisplay.
Locally stored favourites with the old format, are migrated into the new format with `mapLegacyLineNameToDestinationDisplay`

Example:
```
destinationDisplay: {
   frontText: "Servicelinjen",
   via: ["Sentrum"]
}
```

| Before | After |
|--------|-------|
|  |  |


src/travel-details-screens/utils.ts
```diff
@@ -106,9 +108,61 @@ export function significantWaitTime(seconds: number) {
   return seconds > MIN_SIGNIFICANT_WAIT_IN_SECONDS;
 }
 
-export function getLineName(leg: Leg) {
+/* NB this is the same function as in the bff. Keep in sync! */
+export function destinationDisplaysAreEqual(
```
gorandalum, line 112: Doh! We needed this here as well. The strategy that was chosen was kinda based on that there was no such code in the app. Let's have a talk about this.

gorandalum (commented):
In general it looks good, but lets discuss the implications of having the comparison code in the app as well as in the bff.

src/travel-details-screens/utils.ts
```diff
@@ -108,6 +111,91 @@ export function significantWaitTime(seconds: number) {
...
+            storedFavDep,
+            favDepMigrationPair,
+          )
+        ) {
+          aFavoriteDepartureWasMigrated = true;
+
+          upToDateFavDep = {
+            ...storedFavDep,
+            destinationDisplay: favDepMigrationPair?.destinationDisplay,
+          };
```
marius-at-atb, line 186: Note to self:
```suggestion
upToDateFavDep.destinationDisplay =
            favDepMigrationPair.destinationDisplay;
```

src/travel-details-screens/utils.ts
marius-at-atb: I don't think the `destinationDisplaysAreEqual` function in itself is the issue, as it is basically just a deep equals for destinationDisplay objects. The challenge is the input we give it.

marius-at-atb:
Thinking about it more closely, all the migration info we need is available when the bff sends lineName and destinationDisplay together. The PR has now been updated in a way that utilises this.

Should probably add some tests to these new functions.
The manual tests in my simulator with the expected changing data seem to work at least.

src/travel-details-screens/utils.ts
```diff
@@ -106,9 +111,146 @@ export function significantWaitTime(seconds: number) {
...
+  const storedFrontTextIncludesVia = !!storedFrontText?.includes(' via ');
+  const storedFavDepDDViaIsEmpty = (storedFavDepDD?.via?.length || 0) === 0;
+  const destinationDisplayViaIsNotEmpty =
+    (destinationDisplay?.via?.length || 0) > 0;
+  return (
+    storedFrontTextEqualsLineName &&
+    storedFrontTextIncludesVia &&
+    storedFavDepDDViaIsEmpty &&
+    destinationDisplayViaIsNotEmpty
+  );
```
gorandalum, line 157: What happens if the user has two favorites, "Lade via Strindheim" and "Lade via Tunga"? 🤔 Let's talk next week.

marius-at-atb: `storedFrontTextEqualsLineName` will ensure there is no cross-matching in that case. `lineName` comes from the bff and contains the whole string

gorandalum (commented):
I think I see the big picture now, and that this seems like a solution that would work in most cases. There is something to be fixed about using functions in reducer actions, but generally it looks good 👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/use-favorite-departure-data.ts
```diff
@@ -64,7 +65,9 @@ const initialState: Omit<
 type DepartureDataActions =
   | {
       type: 'LOAD_INITIAL_DEPARTURES';
-      favoriteDepartures?: UserFavoriteDepartures;
+      dashboardFavoriteDepartures: UserFavoriteDepartures;
+      favoriteDepartures: UserFavoriteDepartures;
+      setFavoriteDepartures(favorite: UserFavoriteDepartures): Promise<void>;
```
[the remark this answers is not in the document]
marius-at-atb: I'm not sure about that either 😅
Will look into another solution for this

marius-at-atb:
Updates since your last review @gorandalum:
- improved naming
- added tests for `shouldDestinationDisplayBeMigrated`
- `setFavoriteDepartures` and `favoriteDepartures` removed from LOAD_INITIAL_DEPARTURES  action object
- merge conflicts fixed

src/travel-details-screens/utils.ts
```diff
@@ -106,9 +111,145 @@ export function significantWaitTime(seconds: number) {
...
+    (migrationDestinationDisplay?.via?.length || 0) > 0;
+  return (
+    frontTextEqualsMigrationLineName &&
+    frontTextIncludesViaString &&
+    viaIsEmpty &&
+    migrationViaIsNotEmpty
+  );
+};
+
+export const getUpToDateFavoriteDepartures = (
```
gorandalum, line 162: Do this belong more in the favourites module where the migration happens?

src/travel-details-screens/utils.ts
```diff
@@ -106,9 +111,145 @@ export function significantWaitTime(seconds: number) {
...
+    viaNames =
+      via.slice(0, -1).join(', ') +
+      ` ${t(dictionary.listConcatWord)} ` +
+      via[via.length - 1];
+  }
+
+  return frontText + ` ${t(dictionary.via)} ` + viaNames;
+}
+
+export function mapLegacyLineNameToDestinationDisplay(
```
gorandalum, line 238: Do this belong more in the favourites module where the migration happens?

src/travel-details-screens/__tests__/shouldDestinationDisplayBeMigrated.test.ts
gorandalum, line 1: Great!

gorandalum (approved):
Nice work!

PR #3928, opened 2023-10-10
ci: run tsc check when changes in package.json and yarn.lock

hanne-at-atb:


.github/workflows/tsc.yml
```diff
@@ -7,6 +7,8 @@ on:
       - '**.ts'
       - '**.tsx'
       - '**/tsc.yml'
+      - 'package.json'
+      - 'yarn.lock'
```
jorelosorio, line 11: Maybe `**.lock`?

hanne-at-atb: That will also trigger Gemfile.lock and Podfile.lock, is that something we want? :)

PR #3930, opened 2023-10-10
chore: add a11y support for pull to refresh

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/10085

Following a discussion, it was decided that instead of [adding an extra button](https://github.com/AtB-AS/kundevendt/issues/3327), the screen reader should read more specific instructions for how to activate the natively supported pull to refresh functionality. 

So for example on this screen on iOS, the screen reader users will hear "Pull down with 3 fingers to refresh." instead of "Pull down to refresh if you can't find your ticket."

And similar for this screen as well:

### DIFF
--- a/src/translations/dictionary.ts
+++ b/src/translations/dictionary.ts
@@ -24,9 +24,9 @@ const dictionary = {
   },
   missingRealTimePrefix: _('ca. ', 'ca. ', `ca. `),
   a11yPullToRefresh: _(
-    `Dra ned med ${a11yPullToRefreshFingers} fingre for å oppdatere`,
-    `Pull down with ${a11yPullToRefreshFingers} fingers to refresh`,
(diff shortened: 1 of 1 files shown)


gorandalum (commented):
I think we shouldn't mention the specific gestures that the user should use, as they are configurable by the users. It is the same line of thought we use when we say "activate" instead of "press" in accessibility hints, since how the user activates is configurable.

I think just having the "pull down to refresh" is preferable to this.

marius-at-atb:
@Sebstorvik @tormoseng @oyvindmikkelsenatb @Maiaandressen See [this comment above](https://github.com/AtB-AS/mittatb-app/pull/3930#pullrequestreview-1667733678). Thoughts?

oyvindmikkelsenatb:
Love the solution 👌😀

Sebstorvik:
Same!

marius-at-atb:
Closing this following a change of decision

PR #3931, opened 2023-10-10
docs: add chore as a changelog type

hanne-at-atb:
Based on observations of "chore: " being used as a type in changelogs, I suggest including it in our file for contributions. It is even part of the example below in the file 🙈 
This is based on the description of conventional commits: https://github.com/commitizen/conventional-commit-types/blob/master/index.json


rosvik (approved):

marius-at-atb (approved):
Good thinking :+1:

PR #3933, opened 2023-10-11
feat: remove label new from token toggle in my pofile

hanne-at-atb:
As we had the "New" label for token toggle has been around for more than a year, let's remove it 👷🏼‍♀️ 
Also added preview of labels in design system 🌟 

Closes: https://github.com/AtB-AS/kundevendt/issues/3928


marius-at-atb (approved):
Nice to also add it in the DS screen for future reference!

gorandalum (approved):

PR #3935, opened 2023-10-12
test: fix tsc error

tormoseng:
The E2E-test GH Action job runs a tsc check on the e2e folder. The errors are fixed here.

### DIFF
--- a/.github/workflows/test-performance-android.yml
+++ b/.github/workflows/test-performance-android.yml
@@ -19,7 +19,7 @@ on:
         required: true
         default: 'master'
   schedule:
-    - cron: '0 6 * * *'
+    - cron: '0 7 * * *'
 
 jobs:
   run_tests:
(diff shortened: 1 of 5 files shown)


e2e/test/specs/departure.e2e.ts
```diff
@@ -106,8 +106,7 @@ describe('Departure', () => {
       // Departure details
       const linePublicCode = await DepartureOverviewPage.getLinePublicCode();
       const lineName = await DepartureOverviewPage.getLineName();
-      const departure = await DepartureOverviewPage.getDeparture();
-      await departure.click();
```
tormoseng, line 110: This wasn't allowed according to tsc.

HanneLH (approved):
🌸

.github/workflows/test-performance-android.yml
```diff
@@ -19,7 +19,7 @@ on:
         required: true
         default: 'master'
   schedule:
-    - cron: '0 6 * * *'
```
tormoseng, line 22: Avoid starting both e2e jobs at the same time. Will change to manual start after seeing that it works.

rosvik (approved):

PR #3938, opened 2023-10-12
test: better waiting for trip results

tormoseng:
Seems like the tests don't wait long enough for travel search results ([latest test](https://github.com/AtB-AS/mittatb-app/actions/runs/6494641867)). Added some more waiting.

And removed the check for onboarding in the map.


HanneLH (approved):
🌸

PR #3943, opened 2023-10-12
chore: Eslint rule prefer const

gorandalum:


hanne-at-atb (approved):
Const is best, over let, no need to fret. 🚀 This change I've met, it's a solid bet. Approved, you get!

tormoseng:
App opens and looks good. Nothing more to test. QA is enough.

PR #3945, opened 2023-10-13
fix: fix consistent use of favourite

mikaelbr:
This will not change domain language/code language to be consistent with UK English in regards to favorite vs favourite. And we should discuss if translation language convention is the same as code language. It isn't automatically the same.

This just fixes 3 specific places in translations where it still said favorite.

Fixes https://github.com/AtB-AS/kundevendt/issues/9424


marius-at-atb (approved):
👍

PR #3919, opened 2023-10-05
refactor: refactor FareContract components

rosvik:
Refactors the fare contract components: `DetailsContent`, `PreActivatedFareContractInfo` and `CarnetDetails`, and their sub components. Also adds a new Design System -> Fare Contracts page that shows several mocked fare contracts.

Every "section" in a farecontract is now a `Section`, instead of a components with `SectionSeparators` as part of them. This fixes various padding issues due to inconsistent styling across components.

The Fare Contracts page includes an mock implementation of `FirestoreTimestamp`, which can be removed by not relying on firestore-specific classes in the long run.

## Before and after

https://github.com/AtB-AS/mittatb-app/assets/1774972/9de6cd2e-299c-4ea3-ad57-5aeb00a6d311

https://github.com/AtB-AS/mittatb-app/assets/1774972/a4271f7c-dea2-40e4-8036-f958fef6e19b

PR #3949, opened 2023-10-13
fix: increase qr code size

mikaelbr:
See the description and problem at https://github.com/AtB-AS/kundevendt/issues/12080

This only changes the size of Static QR (which FRAM also uses for all QRs as of now)

Don't know if there is a specific reason as to why it is small (other than taking a lot of visual space ofc).

## Changes from
Screenshot 2023-10-13 at 15 39 26

## To
Screenshot 2023-10-13 at 15 39 18

Fixes https://github.com/AtB-AS/kundevendt/issues/12080

### DIFF
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -258,7 +258,7 @@ const StaticQrCode = ({fc}: {fc: FareContract}) => {
   if (!qrCodeSvg) return null;
 
   return (
-    <GenericSectionItem>
+    <GenericSectionItem type="compact">
       <View
         style={[styles.aztecCode, styles.staticQrCode]}
         accessible={true}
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Surprise to me that it was so small. Not on purpose I guess.

mikaelbr:
I think this was an adjustment from @mortennordseth after actually testing on a test rig and seeing it required some distance. But hearing from users with bad QR code scanners it looks like we should try to increase it. This + brightness should help.

rosvik:
After #3919, I think we can remove all `GenericSectionItem`s from Barcode.tsx, as `Barcode` is inside Sections in DetailsContent instead, so it's done twice. My mistake, but it's related to this, so do you want to fix that in this PR?

mikaelbr:
So just remove Generic Sections? Sure. Fixing

rosvik (approved):
👌

mikaelbr:
I hear reports that it CAN get too big also 😅 So testing like this. If it get's too big we can test making it interactive / clickable

rosvik:
Maybe we actually need a `max-width: [X]px` 😱

mikaelbr:
Might be max-width is a good solution. But also reduce clutter around QR code / isolate it better so the camera in the FLV+ won't be confused 🤷  Trying out this in QA now and testing it on FLV+ to see how well it functions and if any other measures needs to be taken

PR #3951, opened 2023-10-16
fix: shows warning when logging in with active tickets

mikaelbr:
When logging in with active tickets we don't show a warning if we navigate using the warning box on the ticket view. This is causing some issues for FRAM where people don't know they'll lose their tickets.

This solves it in the same way as it is solved for login button on Profile page. It doesn't solve the same direct navigation as before if vipps is not activated. Don't know if this is needed?

Also, not an ideal solution as such, you create some nested stacks here that might not seem like an optimal flow. Might also add the warning on the step before showing why you should log in, but this is also kind of odd 🤔

This is just a quick proposal for me and we can discuss if this is the best way to solve it or not.

<details>
<summary>Screen recordings</summary>

https://github.com/AtB-AS/mittatb-app/assets/606374/cf552361-87d9-4796-9451-f644d5135b75

https://github.com/AtB-AS/mittatb-app/assets/606374/3f861b3d-9ce4-4648-9a2d-ad65416a2c8c

</details>

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseAsAnonymousConsequencesScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseAsAnonymousConsequencesScreen.tsx
@@ -1,7 +1,11 @@
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import React from 'react';
 import {AnonymousPurchaseConsequencesScreenComponent} from '@atb/anonymous-purchase-consequences-screen';
-import {RootStackScreenProps} from './navigation-types';
+import {RootStackParamList, RootStackScreenProps} from './navigation-types';
+import {
+  filterActiveOrCanBeUsedFareContracts,
+  useTicketingState,
+} from '@atb/ticketing';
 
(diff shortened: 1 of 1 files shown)


gorandalum:
Yeah, the quick fix here is to add the same logic in anonymous-purchase-consequences screen as is present in the profile screen and in the login-required-for-this-ticket screen. I think the logic should also handle if Vipps is not enabled, and also will probably need the `afterLogin` param to be correct.

That said it is pretty cumbersome to remember this when adding new ways to login. Can just do the quick fix now, and at the same time create an issue to improve this somehow, either in code and/or design.

mikaelbr:
Looks like `afterLogin` isn't passed on through `Root_LoginActiveFareContractWarningScreen`. Will look at it.

gorandalum:
Coming to think of it, with the loading screen change a couple sprints ago, the after login param probably doesn't work. After login the user is back to the dashboard screen. Was ok-ish for designer/PO since the login doesn't really happen often.

mikaelbr:
So it is an existing problem? I updated code here so it sends it in as before at least.

gorandalum (approved):

tormoseng:
Looks good on iOS and Android

PR #3952, opened 2023-10-17
feat: adds bottom sheet with isolated barcode on barcode press

mikaelbr:
After more insights into troubles with reading barcodes using the FLV+, I think the problem isn't the size itself, but also that the device isn't able to separate the QR code from the rest on the page. We have reports even that in some cases the barcode is too big.

Also there is problem with the FLV+ on dark mode. So for the bottom sheet here, in addition to increasing screen brightness I set the entire background to white no matter what theme you use.

Users of the existing app for FRAM are used to being able to press the QR code to see it bigger, even though this is kind of a hidden feature here. But there is existing work on improving visual inspection and restructuring this page where the QR code is smaller, so I think this is OK until that feature gets worked on.

I only added the bottom sheet for static codes, as the bottom sheet doesn't respond well to changes and for FRAM using barcode validation we won't be using mobile token before the entire screen is redesigned.

Fixes https://github.com/AtB-AS/kundevendt/issues/12080 (again)

See demo video:

https://github.com/AtB-AS/mittatb-app/assets/606374/88718c0f-64c2-4512-8db9-2021ccb3d72a

### DIFF
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -241,7 +241,11 @@ const StaticAztec = ({fc}: {fc: FareContract}) => {
       accessibilityLabel={t(FareContractTexts.details.barcodeA11yLabel)}
       testID="staticBarcode"
     >
-      <PressableOpacity onPress={onOpenBarcodePress}>
+      <PressableOpacity
+        onPress={onOpenBarcodePress}
+        accessibilityRole="button"
+        accessibilityHint={t(FareContractTexts.details.barcodeButtonA11yLabel)}
+      >
         <SvgXml xml={aztecXml} width="100%" height="100%" />
       </PressableOpacity>
     </View>
@@ -269,7 +273,11 @@ const StaticQrCode = ({fc}: {fc: FareContract}) => {
       accessibilityLabel={t(FareContractTexts.details.barcodeA11yLabel)}
       testID="staticQRCode"
     >
-      <PressableOpacity onPress={onOpenBarcodePress}>
+      <PressableOpacity
+        onPress={onOpenBarcodePress}
+        accessibilityRole="button"
+        accessibilityHint={t(FareContractTexts.details.barcodeButtonA11yLabel)}
+      >
         <SvgXml xml={qrCodeSvg} width="100%" height="100%" />
       </PressableOpacity>
     </View>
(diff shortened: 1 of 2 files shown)


src/fare-contracts/details/Barcode.tsx
```diff
@@ -250,12 +264,14 @@ const StaticQrCode = ({fc}: {fc: FareContract}) => {
...
-      style={[styles.aztecCode, styles.staticQrCode]}
+      style={[styles.aztecCode, styles.staticQrCode, styles.staticQrCodeSmall]}
       accessible={true}
       accessibilityLabel={t(FareContractTexts.details.barcodeA11yLabel)}
       testID="staticQRCode"
     >
-      <SvgXml xml={qrCodeSvg} width="100%" height="100%" />
+      <PressableOpacity onPress={onOpenBarcodePress}>
+        <SvgXml xml={qrCodeSvg} width="100%" height="100%" />
+      </PressableOpacity>
```
gorandalum, line 274: Maybe add some `accessibilityRole` and `accessibilityHint`?

gorandalum (approved):

tormoseng:
@mikaelbr On iOS with VoiceOver I don't get any hints about showing the barcode separately. Focus reads _Barkode. Vis frem denne koden i billettkontroll_.

toretefre:
Issue probably being that parent component is accessible=true. https://github.com/AtB-AS/mittatb-app/blob/765826e2aa089d73864ca50b7fdeab9abb94d34b/src/fare-contracts/details/Barcode.tsx#L240

toretefre:
We've just now merged a fix with https://github.com/AtB-AS/mittatb-app/pull/3976 @tormoseng 

Also includes activate (or tap) giantQR™ in bottom-sheet to close bottom-sheet. And changed testID for giantQR™ to something else than the small one.

tormoseng:
Re-test:
- [x] iOS
- [ ] Android

toretefre:
Oops, of course. We changed the original label. Will create a separate one.

tormoseng:
Re-tested OK https://github.com/AtB-AS/mittatb-app/pull/3952#issuecomment-1766362259

Will retest https://github.com/AtB-AS/mittatb-app/pull/3952#issuecomment-1779257983 after https://github.com/AtB-AS/mittatb-app/pull/3981

toretefre:
https://github.com/AtB-AS/mittatb-app/pull/3981 merged 💥

tormoseng:
Tested OK! 🎉

PR #3955, opened 2023-10-17
feat: Log storage value in Bugsnag breadcrumbs

gorandalum:
When debugging errors it is some times helpful to see the values
which were retrieved from local storage.

Also delete some unused storage keys.


jorelosorio (approved):

tormoseng:
Tested OK! Size now up to 150 KB in my tests, but we have room up to 1 MB.

PR #3959, opened 2023-10-19
feat: add option for fallback on mobile token timeout

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/11351

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -46,7 +46,7 @@ type MobileTokenContextState = {
   ) => Promise<boolean>;
   retry: () => void;
   wipeToken: () => Promise<void>;
-  fallbackEnabled: boolean;
+  fallbackActive: boolean;
   // For debugging
   createToken: () => void;
   validateToken: () => void;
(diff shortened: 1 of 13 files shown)


src/mobile-token/MobileTokenContext.tsx
```diff
@@ -65,7 +65,10 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
   const {token_timeout_in_seconds} = useRemoteConfig();
   const hasEnabledMobileToken = useHasEnabledMobileToken();
 
-  const {enable_token_fallback: fallbackEnabled} = useRemoteConfig();
+  const {
+    enable_token_fallback: fallbackOnErrorEnabled,
```
rosvik, line 69: This remote config flag isn't very precisely named now, I would have called it `enable_token_fallback_on_error` if I created it now. But not changing it since (1) the effect of the flag shouldn't have changed at all and (2) we would need to maintain two fields in firebase for backwards compatibility.

gorandalum (approved):
I think this looks pretty good. Also [my issue regarding handling timeout in loading screen](https://github.com/AtB-AS/kundevendt/issues/10087) touches some of this code, so it may be refactored here and there.

PR #3960, opened 2023-10-20
feat: share travel habits screen

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/10523

Note: permission utils will be handled in https://github.com/AtB-AS/kundevendt/issues/8615, just placeholders for now.

Adds a new screen that shows after 10 sessions on iOS and 0 on Android (as specified in remote config):

Update: this PR also includes a fix for pausing location watching when the app is in the background.


src/GeolocationContext.tsx
```diff
@@ -143,39 +152,46 @@ export const GeolocationContextProvider: React.FC = ({children}) => {
...
+  }, [hasPermission, geoLocationName, dispatch]);
+
+  const stopLocationWatcher = () => Geolocation.clearWatch(watchIdRef.current);
+
+  useEffect(() => {
     startLocationWatcher();
-    return () => Geolocation.clearWatch(watchId);
-  }, [hasPermission]);
+    return stopLocationWatcher;
+  }, []);
```
gorandalum, line 191: `startLocationWatcher` must be in dependency list so changes in `hasPermission` and `getLocationName` have effect?

src/GeolocationContext.tsx
```diff
@@ -18,6 +25,8 @@ import {updateMetadata as updateChatUserMetadata} from './chat/metadata';
 import {useAppStateStatus} from './utils/use-app-state-status';
 import {GeoLocation} from '@atb/favorites';
 import {dictionary, useTranslation} from '@atb/translations';
+import {useOnResume} from './utils/use-on-resume';
+import {useOnPause} from './utils/use-on-pause';
```
gorandalum, line 29: Other places we have used the term `active`, like in the hook `useIsFocusedAndActive`. Is it possible to align the terms used in our app?

src/utils/use-on-pause.ts
```diff
@@ -0,0 +1,18 @@
+import {useRef, useEffect} from 'react';
+import {AppState} from 'react-native';
+
+export function useOnPause(runOnPauseFunc: () => void) {
```
gorandalum, line 4: Is `backgrounded` more intuitive than `pause`?

src/utils/use-on-pause.ts
```diff
@@ -0,0 +1,18 @@
...
+  const appState = useRef(AppState.currentState);
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener('change', (nextAppState) => {
+      if (appState.current.match(/active/) && nextAppState === 'background') {
```
gorandalum, line 9: What if the state change goes 'active' -> 'inactive' -> 'background'?

src/beacons/use-on-beacons-session-count.tsx
```diff
@@ -0,0 +1,78 @@
...
+  runOnCountFunc: () => void,
+  keyAppendix: string, // appendix to ensure unique storage keys
+) => {
+  const beaconsSessionCount = useBeaconsSessionCounter(count + 1, keyAppendix);
+
+  useEffect(() => {
+    if (beaconsSessionCount !== undefined && beaconsSessionCount === count) {
+      runOnCountFunc();
+    }
+  }, [beaconsSessionCount]);
```
gorandalum, line 17: Needs `runOnCountFunc` in dependency list

src/beacons/use-on-beacons-session-count.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+enum storeKey {
+  beaconsSessionCountKey = '@ATB_session_count_',
```
gorandalum, line 21: Is this ment to end with `_ `?

src/beacons/use-on-beacons-session-count.tsx
```diff
@@ -0,0 +1,78 @@
...
+      );
+      const beaconsSessionCount = parseInt(beaconsSessionCountString ?? '0');
+      if (beaconsSessionCount < stopCountingOn) {
+        setStoredBeaconsSessionCount(beaconsSessionCount + 1);
+      } else {
+        setBeaconsSessionCount(beaconsSessionCount);
+      }
+    };
+    loadStoredBeaconsSessionCount();
+  }, []); // load and conditionally increase count on app open
```
gorandalum, line 56: Missing dependency `setStoredBeaconsSessionCount`.

src/beacons/use-on-beacons-session-count.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  useEffect(() => {
+    if (
+      !isBeaconsEnabled &&
+      isBeaconsEnabledDebugOverrideReady &&
+      beaconsSessionCount !== 0
+    ) {
+      setStoredBeaconsSessionCount(0); // restart count when beacons disabled
+    }
+  }, [isBeaconsEnabled, isBeaconsEnabledDebugOverrideReady]);
```
gorandalum, line 75: Inline creation of functions is not optimal.

src/GeolocationContext.tsx
```diff
@@ -143,39 +152,46 @@ export const GeolocationContextProvider: React.FC = ({children}) => {
...
+
+  useEffect(() => {
     startLocationWatcher();
-    return () => Geolocation.clearWatch(watchId);
-  }, [hasPermission]);
+    return stopLocationWatcher;
+  }, []);
+
+  useOnResume(startLocationWatcher);
+  useOnPause(stopLocationWatcher);
```
gorandalum, line 194: In general there is quite a bit more risk when doing it like this than just doing it all in one effect, like this (pseudocode):
```tsx
const isFocusedAndActive = useIsFocuesAndActive();
useEffect(() => {
  if (isFocusedAndActive) {
    const id = setupWatcher():
    return () => clearWatch(id)
  }
}, [isFocusedAndActive])
```

We can take a quick discussion regarding risk/benefit.

gorandalum (changes requested):
Some comments here and there.

Also could you write up a summary regarding why this start watch/clear watch stuff is necessary now, while it wasn't before. Is it something with new permissions that we ask for that makes it necessary?

src/GeolocationContext.tsx
marius-at-atb: Yes, good catch!

src/utils/use-on-pause.ts
marius-at-atb: `inactive` will also be matched by `.match(/active/)`, but I guess it is better to be more explicit about it :+1:

src/beacons/use-on-beacons-session-count.tsx
marius-at-atb: Yes. As `keyAppendix` is set to `"share_travel_habits"`, the full key will be `"@ATB_session_count_ share_travel_habits"` in this case.

src/GeolocationContext.tsx
marius-at-atb: Solved by using appStatus instead of useOnResume and useOnPause hooks

src/GeolocationContext.tsx
marius-at-atb: Currently refactored like this:

```ts
  useEffect(() => {
    if (appStatus === 'active') {
      if (!hasPermission) {
        updateLocation(null, geoLocationName);
      } else {
        const watchId = Geolocation.watchPosition(
          (position) => updateLocation(position, geoLocationName),
          handleLocationError,
          config,
        );
        return () => Geolocation.clearWatch(watchId);
      }
    }
  }, [geoLocationName, hasPermission, appStatus]);
```

src/beacons/use-on-beacons-session-count.tsx
gorandalum, line 1: There is something a little iffy about this file. With this many effects it is hard to follow the logic, and see the consequences. An example of something I am unsure of is handled correct:
- Beacons turned of and count set to 0
- But seems like it still counts even though beacons turned off?
- Beacons enabled again, which will then also count app opens during the beacons was disabled?

I will experiment a little with this code today, and we can talk about it tomorrow.

gorandalum (changes requested):
In general it looks good, just something about `use-on-beacons-session-count` and some comments on encapsulation.

src/beacons/use-on-beacons-session-count.tsx
```diff
@@ -0,0 +1,105 @@
...
+import {useIsBeaconsEnabled} from './use-is-beacons-enabled';
+import {storage} from '@atb/storage';
+
+import {useAppStateStatus} from '@atb/utils/use-app-state-status';
+import {usePrevious} from '@atb/utils/use-previous';
+
+export const useOnBeaconsSessionCount = (
+  runOnCount: number,
+  runOnCountFunc: () => void,
+  keyAppendix: string, // appendix to ensure unique storage keys
```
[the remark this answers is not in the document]
marius-at-atb: Yes, and after no longer accepting count or storage key appendix as parameters, the hook usage now looks like this:

```ts
const showShareTravelHabitsScreen = useCallback(() => {
    InteractionManager.runAfterInteractions(() =>
      navigation.navigate('Root_ShareTravelHabitsScreen'),
    );
  }, [navigation]);

  useMaybeShowShareTravelHabitsScreen(showShareTravelHabitsScreen);
```

gorandalum (approved):
I think this looks quite alright now, and more encapsulated :+1:

PR #3961, opened 2023-10-20
fix: Feature toggle for parking violations reporting

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/13276

This pr seems a bit bigger than it is, since I had to move some of the stuff from the stacks folder to a separate feature folder due to import constraints


marius-at-atb (approved):
Yeah kind of hard to read the changes for ContactSheet.tsx, but looks good I think 😄

KristianSelnas:
The ContactSheet changes are a bit strange. Prettier insists on reformatting this file when I touch it. Gøran and I looked at it, but we cannot really see why it happens. It doesn't really matter, except it makes this pr a bit harder to read. I should probably have commented this in the description.

gorandalum (approved):

PR #3962, opened 2023-10-20
feat: resubscribe to firestore config data on retry in loading boundary

hanne-at-atb:
This issue allows us to resubscribe to firestore config data if we don't have it in the cache. In case we lack the firestore config data the loading screen will show followed by the error boundary screen in case it times out (given that the error boundary screen is turned on). When clicking on the retry button in the error boundary screen we will resubscribe to firestore config data.

This issue also includes:
* Delete reference data which was used as default state in firestore config
* Moved types and utils from reference-data folder to src/configuration
* Updated export from index-file in src/configuration, and thus cleaned up the relevant imports
* Added and updated tests for loading screen

Closes: https://github.com/AtB-AS/kundevendt/issues/8271 and https://github.com/AtB-AS/kundevendt/issues/10086


src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -205,6 +184,10 @@ export const FirestoreConfigurationContextProvider: React.FC = ({children}) => {
           );
         },
       );
+  };
+
+  useEffect(() => {
+    subscribeFirestore();
   }, []);
```
gorandalum, line 191: `subscribeFirestore` must be in dependency list.

src/configuration/FirestoreConfigurationContext.tsx
```diff
@@ -239,10 +223,16 @@ export const FirestoreConfigurationContextProvider: React.FC = ({children}) => {
...
     harborConnectionOverrides,
+    firestoreConfigStatus,
   ]);
 
   return (
-    <FirestoreConfigurationContext.Provider value={memoizedState}>
+    <FirestoreConfigurationContext.Provider
+      value={{
+        ...memoizedState,
+        resubscribeFirestoreConfig: useCallback(() => subscribeFirestore(), []),
```
gorandalum, line 233: Same here, needs dependency in dependency list. 

Also, should the previous subscription be unsubscribed before new subscriptions? 

And I think we can be slightly more explicit regarding resubscription:
- Is resubscription only allowed if there is no data? If so it should be a check on that here.
- Or is resubscription also allowed even if there is data? if so the state should cleared and the status should be set to `loading` on resubscription.

gorandalum (changes requested):
I think this looks really good! A couple comments on some minor details.

src/configuration/FirestoreConfigurationContext.tsx
hanne-at-atb: Good point ☝🏼 Went for the solution with clearing the state and setting to loading.

rosvik (approved):
Good work 💯 LGTM

gorandalum (approved):

PR #3964, opened 2023-10-24
fix: added case to ticket assistant

adriawh:
This PR is dependent on https://github.com/AtB-AS/atb-ticket/pull/196. 

If the recommended ticket is very close to the price of single tickets, we still recommend the periodic one as it allows for more flexibility. Was discussed with @Sebstorvik


marius-at-atb (approved):
Lgtm!

PR #3965, opened 2023-10-24
fix: Not use quay public code as leg public code

gorandalum:
### Background
https://mittatb.slack.com/archives/C0116FMPX4Y/p1698127130815159

The quay public code (like P2) should not be used as leg public
code, as it represents the quay number in a stop place with
numbered quays.

### Solution
When finding public code for the leg only use the line public code, and not the quay public code.

### Acceptance criteria
- [ ] All interchange texts should reference the line public code, and not quay public codes.


marius-at-atb (approved):

HanneLH:
Tested on iOS and Android, and different lines! 🚌

PR #3967, opened 2023-10-24
fix: Misc parking violations reporting fixes

KristianSelnas:
Fixes found during testing of https://github.com/AtB-AS/kundevendt/issues/7908

Corresponding BFF pr: https://github.com/AtB-AS/atb-bff/pull/307

### DIFF
--- a/src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
+++ b/src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
@@ -141,7 +141,7 @@ export const Root_ParkingViolationsQr = ({
     <ScreenContainer
       title={t(ParkingViolationTexts.qr.title)}
       secondaryText={t(ParkingViolationTexts.qr.instructions)}
-      leftHeaderButton={isLoading ? {type: 'none'} : {type: 'back'}}
+      leftHeaderButton={isLoading ? undefined : {type: 'back', withIcon: true}}
       buttons={
         <Button
           disabled={isError}
--- a/src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
+++ b/src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
@@ -39,7 +39,7 @@ export const Root_ParkingViolationsQr = ({
     }
   }, [isFocused]);
 
-  const submitReport = async (providerId: number) => {
+  const submitReport = async (providerId?: number) => {
     setIsLoading(true);
     closeBottomSheet();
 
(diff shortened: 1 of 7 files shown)


src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
```diff
@@ -138,6 +141,7 @@ export const Root_ParkingViolationsQr = ({
     <ScreenContainer
       title={t(ParkingViolationTexts.qr.title)}
       secondaryText={t(ParkingViolationTexts.qr.instructions)}
+      leftHeaderButton={isLoading ? {type: 'none'} : {type: 'back'}}
```
gorandalum, line 144: Why not just `isLoading ? undefined : {type: 'back'}`. I think it is slightly preferable to have a single method for specifying that it should be no header button, and not both `undefined` and `{type: none}`.

src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
```diff
@@ -113,7 +113,10 @@ export const Root_ParkingViolationsQr = ({
   const selectProvider = (qrScanFailed?: boolean) => {
     openBottomSheet(() => (
       <SelectProviderBottomSheet
-        providers={providers}
+        providers={[
+          ...providers,
+          {name: t(ParkingViolationTexts.selectProvider.unknownProvider)},
+        ]}
```
gorandalum, line 119: I think this should be done with `useMemo`. If you have a hook which has `providers` in the dependency list this would fire on every render now, as the array is recreated on every render.

src/stacks-hierarchy/Root_ParkingViolationsReporting/components/ProviderLogo.tsx
```diff
@@ -17,13 +20,36 @@ export const ProviderLogo = ({provider, size = 50, style}: Props) => {
...
+};
+
+const UnknownProvider = ({height, width}: {height: number; width: number}) => {
+  const {theme} = useTheme();
+  return (
+    <View
+      style={{
+        backgroundColor:
+          theme.static.background['background_accent_0'].background,
+        //backgroundColor: 'red',
```
gorandalum, line 45: Remove?

gorandalum (approved):
Nice. Some comments which aren't strictly blockers.

src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
KristianSelnas: Because of the default in `ScreenContainer`. The alternative is to remove that default and add it explicitly to the other screens using `ScreenContainer`.

src/stacks-hierarchy/Root_ParkingViolationsReporting/components/ProviderLogo.tsx
KristianSelnas: Absolutely

src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ParkingViolationsQr.tsx
KristianSelnas: I agree and decided to remove the default and rather add it explicitly to the other screens.

PR #3968, opened 2023-10-24
feat: use interactive color on token toggle screen

hanne-at-atb:
Changed from using static colors to interactive colors for default and active in RadioBoxes, providing a prop to set the interactiveColor. 

Looks like this after current changes:
image

Closes: https://github.com/AtB-AS/kundevendt/issues/8652

### DIFF
--- a/src/components/radio/RadioBox.tsx
+++ b/src/components/radio/RadioBox.tsx
@@ -1,6 +1,5 @@
 import {View, ViewStyle} from 'react-native';
 import {StyleSheet, Theme, useTheme} from '@atb/theme';
-import {StaticColorByType} from '@atb/theme/colors';
 import {ThemeText} from '@atb/components/text';
 import {RadioIcon} from './RadioIcon';
 import React from 'react';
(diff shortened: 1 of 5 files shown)


src/components/radio/RadioBox.tsx
```diff
@@ -65,16 +67,14 @@ export function RadioBox({
     >
       <ThemeText
         type="heading__title"
-        color={themeColor}
-        style={{...styles.title, marginBottom: spacing}}
+        style={[{...styles.title, marginBottom: spacing}, styleText]}
```
gorandalum, line 70: Note that it is slightly preferable to send in a `ContrastColor` in the `color` param, as using styles is mainly used for overrides that is "outside" of the theme colors.

In this situation it would be something like:
```tsx
const themedColor = theme.interactive[interactiveColor][selected ? 'active' : 'default'];`
<ThemeText color={themedColor} />
```

gorandalum (approved):
Very nice! It is good that more of our components starts using interactive color correctly :+1:

PR #3969, opened 2023-10-24
feat: Add Kogenta SDK to collect data in the debug mode

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/8614


src/utils/use-beacons.ts
gorandalum, line 1: I think this hook should be under src/beacons, as it is to complex to be a general utility hook.

src/utils/use-beacons.ts
```diff
@@ -0,0 +1,133 @@
...
+    async function checkKettleInfo() {
+      const status = await Kettle.isStarted();
+      const identifier = await Kettle.getIdentifier();
+      const consents = await Kettle.getGrantedConsents();
+
+      if (isMounted) {
+        setIsKettleStarted(status);
+        setKettleIdentifier(identifier);
+        setKettleConsents(consents);
+        setIsBeaconsOnboarded(Object.keys(consents).length > 0);
```
gorandalum, line 43: Since these are all set at the same time, could they be a single state variable?

src/utils/use-beacons.ts
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+
+    if (isBeaconsSupported) checkKettleInfo();
+
+    return () => {
+      isMounted = false;
+    };
+  }, [isKettleStarted, kettleIdentifier, kettleConsents, isBeaconsOnboarded]);
```
gorandalum, line 52: `isBeaconsSupported` need to be in dependency list

gorandalum (approved):
In general it looks good. Some comments that can be looked at.

src/beacons/use-beacons.ts
```diff
@@ -0,0 +1,191 @@
+import { useIsBeaconsEnabled } from './use-is-beacons-enabled';
+import { KETTLE_API_KEY } from '@env';
+import { useCallback, useEffect, useState } from 'react';
+import { NativeModules, Platform } from 'react-native';
+import { Kettle, KettleModules } from 'react-native-kettle-module';
+import { KettleConsents } from 'react-native-kettle-module';
+import { PERMISSIONS, Permission, RESULTS, check, checkMultiple, request } from 'react-native-permissions';
+
+const BEACONS_CONSENTS = [KettleConsents.SURVEYS, KettleConsents.ANALYTICS];
```
marius-at-atb, line 9: Should this be retrieved from remote config?

src/beacons/use-beacons.ts
```diff
@@ -0,0 +1,191 @@
...
+  isKettleStarted: boolean;
+  kettleIdentifier: string | null;
+  kettleConsents: Record<string, boolean> | null;
+  isBeaconsOnboarded: boolean;
+  privacyDashboardUrl?: string;
+  privacyTermsUrl?: string;
+};
+
+export const useBeacons = () => {
+  const [isBeaconsEnabled] = useIsBeaconsEnabled();
```
marius-at-atb, line 29:
```suggestion
  const [isBeaconsEnabled, isBeaconsEnabledDebugOverrideReady] = useIsBeaconsEnabled();
```

src/beacons/use-beacons.ts
```diff
@@ -0,0 +1,191 @@
...
+  const [kettleInfo, setKettleInfo] = useState<KettleInfo>();
+
+  const isBeaconsSupported = isBeaconsEnabled && !!KETTLE_API_KEY && !!kettleInfo?.kettleConsents;
```
marius-at-atb, line 32:
```suggestion
  const isBeaconsSupported = isBeaconsEnabled && isBeaconsEnabledDebugOverrideReady && !!KETTLE_API_KEY && !!kettleInfo?.kettleConsents;
```

src/beacons/use-beacons.ts
jorelosorio: Good question, since this is the only thing we need from them, not sure if in the future we should change that. But in the case is in the remote config we can not just change to new consents because haven't been approved.

marius-at-atb: Good point, it would probably require more logic which is perhaps unnecessary now

src/beacons/use-beacons.ts
```diff
@@ -0,0 +1,209 @@
...
+        privacyTermsUrl,
+      });
+    }
+  }, [setKettleInfo, isBeaconsSupported, isKettleSDKInitialized]);
+
+  // This function must be called before consent is granted
+  const initializeBeaconsSDK = useCallback(async () => {
+    // This will set kettle if KETTLE_API_KEY is set on the native side
+    if (!isKettleSDKInitialized && isBeaconsSupported) {
+      await NativeModules.KettleSDKExtension.initializeKettleSDK();
```
marius-at-atb, line 64: I think this should also run
`const permissions = await allowedPermissionForKettle();`
and then init only if `permissions.length > 0`

jorelosorio: It is a double check, because is supposed that if the user went that far is because at least Bluetooth was accepted.

PR #3975, opened 2023-10-25
fix: Parking violations: Load user position only once

KristianSelnas:
Problem: The user position returned from `useGeolocationState` is a "subscription" that will return an updated position regularly. This causes a reload of the parking violation feature's `init` data. This is unnecessary and init data should be loaded only once.

Solution: Expose `getCurrentPosition` from `GeolocationContext` as a means to get the position only once.

Found during testing of https://github.com/AtB-AS/kundevendt/issues/7908


gorandalum (commented):
I know we talked about doing it like this, exposing the `Geolocation.getCurrentPosition` from the `GeoLocationContext`. However, after thinking about it a little, I think it might be better to just save the current location (from the subscription) in a ref. That way we utilise the location we already actually know through the existing subscription, and the real benefit is that getting the current location is no longer an async operation, which can be a little cumbersome in some situations.

It would be something like this-ish in the `GeoLocationContext`:
```ts
const locationRef = useRef();
useEffect(() => {
  locationRef.current = state.location;
}, [state.location]);
// And returning the locationRef in the provider
```

What do you think?

KristianSelnas:
I could add this solution locally in my feature, or do you think it's useful to add it to `GeolocationContext`?

KristianSelnas:
How would this work when position is undefined initially?

src/parking-violations-reporting/use-parking-violations.tsx
```diff
@@ -1,24 +1,73 @@
...
-  } = useInitQuery(position ?? {latitude: 0, longitude: 0});
+    getCurrentPosition,
+    status: locationPermissionStatus,
+    requestPermission,
+  } = useGeolocationState();
+
+  const requestLocationPermission = useCallback(requestPermission, []);
+
+  useEffect(() => {
+    if (locationPermissionStatus !== 'granted') {
```
jorelosorio, line 32: In this case, the `locationPermissionStatus` could be `.denied` not prompting anything to the user because it was already rejected. I think we could use a warning message indicating that location is not being shared. I assume that this prompt will just pop up without any communication, for the case when status is `notDetermined`. 

NOTE: There is a PR that will prompt an alert indicating go to the settings.

gorandalum:
> I could add this solution locally in my feature, or do you think it's useful to add it to `GeolocationContext`?

> How would this work when position is undefined initially?

I think it would be nice to have in `GeoLocationContext`. I won't be surprised if we have more places in the app where we just want the position at the moment without listening to a subscription, and I think some of those places will come to light when adding exhaustive-deps ref.

Note that either the `locationRef` or a function returning the value (e.g. `getCurrentLocation: useCallback(() => locationRef.current, [])`) can be exposed through the provider. Both variations are fine I think.

KristianSelnas:
I don't think the `locationRef` approach will work, since the location is always `null` initially. I'm more fan of querying the position when you need it, as it's done now.

src/parking-violations-reporting/use-parking-violations.tsx
KristianSelnas: I'm not sure if I follow. What this code currently does is request permission again if the `locationPermissionStatus` is any other than `granted`. If that request is not granted the warning message indicating that the location is not shared is shown.  (`PermissionRequiredError` is picked up in `Root_ParkingViolationsSelect.tsx`)

gorandalum:
> I don't think the `locationRef` approach will work, since the location is always `null` initially. I'm more fan of querying the position when you need it, as it's done now.

What happens in the implementation now? I think returning `undefined` if permission is not shared is better than throwing an error.

jorelosorio:
Now we are discussing a bit about location, why are we today requesting locations all the time? it could be better just to ask for location in those cases when is needed only? Right now with the introduction of the beacons, the option for background location has been activated and this produces a blue status bar in iOS which is caused by requesting location all the time in this Geolocation context, there is a solution right now to only request location when the app becomes active. but if we use location smart just once when needed would be nice, but not sure if that breaks something else. @gorandalum @KristianSelnas we can discuss it a bit off this PR if you want :)

src/parking-violations-reporting/use-parking-violations.tsx
```diff
@@ -12,6 +12,11 @@ export class PermissionRequiredError extends Error {
     super('Permission required');
   }
 }
+export class NoLocationError extends Error {
+  constructor() {
+    super('Location missing');
```
jorelosorio, line 17: Should this be translated?

KristianSelnas: It's not displayed to the user. It's used to differentiate the various error situations. The message is useful e.g. when debugging.

gorandalum (approved):

PR #3979, opened 2023-10-25
refactor: Bugsnag instead of PostHog on loading timeout

gorandalum:
### Background
This is so we can see more relevant metadata for the user by being
able to see the Bugsnag breadcrumbs.

### Acceptance criteria
- [ ] Verify that loading timeouts are registered in Bugsnag, and that the error includes the parameters used in the evaluation og the loading state, and relevant breadcrumbs.


src/loading-screen/use-notify-bugsnag-on-timeout-status.ts
```diff
@@ -0,0 +1,14 @@
...
+import {LoadingState} from '@atb/loading-screen/types';
+import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
+
+export const useNotifyBugsnagOnTimeoutStatus = (
+  status: LoadingState['status'],
+  paramsRef: LoadingState['paramsRef'],
+) => {
+  useEffect(() => {
+    if (status === 'timeout') {
+      notifyBugsnag('Loading boundary timeout', {...paramsRef.current});
```
marius-at-atb, line 11: Perhaps the `'Loading boundary timeout'` string should be accepted as a parameter?

`useNotifyBugsnagOnTimeoutStatus` indicates that the function is general, but `'Loading boundary timeout'` is specific

marius-at-atb (approved):
Lgtm!

src/loading-screen/use-notify-bugsnag-on-timeout-status.ts
gorandalum: Good suggestion. I'll skip it for now since it as of now isn't a general hook, as it is tightly linked to loading both in typing, and that it isn't exposed out of the module through the index file.

tormoseng:
Tested OK!

PR #3980, opened 2023-10-25
feat: add car station clusters

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/4069

- Adds clustering to car stations, where the count is the sum of all leaf nodes. 
- Also updates some zoom levels for bike and car stations. 

https://github.com/AtB-AS/mittatb-app/assets/1774972/f5a9caea-a36c-458c-b0e4-35629d0167ac


KristianSelnas (approved):
LGTM! Smart solution using clusters. Should we do the same for bike stations, while we are at it?

rosvik:
> Should we do the same for bike stations, while we are at it?

Yes! Looking into it now.

PR #3985, opened 2023-10-26
feat: improve scooter cluster zoom level on click

rosvik:
- Updates the scooter cluster zoom logic to zoom at least "2 steps", or to the point where the cluster expands, whichever is furthest.
- Changes [clusterRadius](https://github.com/rnmapbox/maps/blob/main/docs/ShapeSource.md#clusterradius) from 50 (the default), to 40. This means there are slightly less clusters than before, and more individual scooters. 
- Changes zoom animation for scooter clusters from 400ms to around 200ms (scaled to how far we zoom)
- Changes zoom animation for station clusters from 400ms to 200ms

https://github.com/AtB-AS/mittatb-app/assets/1774972/052d979a-2bac-4865-8ca6-4526bc738893

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -124,8 +124,9 @@ export const Map = (props: MapProps) => {
           )}
           {props.vehicles && (
             <Vehicles
-              mapCameraRef={mapCameraRef}
               vehicles={props.vehicles.vehicles}
+              mapCameraRef={mapCameraRef}
+              mapViewRef={mapViewRef}
               onClusterClick={(feature) => {
                 onMapClick({
                   source: 'cluster-click',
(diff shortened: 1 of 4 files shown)


marius-at-atb (approved):
Very nice!

KristianSelnas (approved):
👍

rosvik:
## Acceptance criteria

- [ ] Clicking on scooter clusters zooms a "sensible" distance. The cluster is always split after the zoom.
- [ ] We noticed that for some mapbox settings (cluserRadius set to less than 40), scooters would sometimes disappear when they were close together, instead of turning into a cluster. Haven't seen that for these settings, but something to keep an eye out for.
- [ ] The animation is faster than before.

PR #3986, opened 2023-10-26
feat: add bike station clusters

rosvik:
Same as https://github.com/AtB-AS/mittatb-app/pull/3980, but for bike stations. Adds clustering to bike stations, where the count is the sum of all leaf nodes.


gorandalum (approved):
LGMT 👍

KristianSelnas (approved):
👍

PR #3987, opened 2023-10-27
feat: Dismissible announcements

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/10321

Adds an X to the top right corner of each announcement that dismisses the announcement when pressed. Which announcements the user has dismissed is saved in the store.

Simulator Screenshot - iPhone 13 - 2023-10-27 at 08 36 27

### DIFF
--- a/src/announcements/AnnouncementsContext.tsx
+++ b/src/announcements/AnnouncementsContext.tsx
@@ -1,4 +1,10 @@
-import {createContext, useContext, useEffect, useState} from 'react';
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 import firestore from '@react-native-firebase/firestore';
 import {AnnouncementRaw, AnnouncementType} from './types';
 import {mapToAnnouncements} from './converters';
@@ -42,16 +48,19 @@ const AnnouncementsContextProvider: React.FC = ({children}) => {
     return () => unsubscribe();
   }, []);
 
-  const dismissAnnouncement = (dismissedAnnouncement: AnnouncementType) => {
-    addDismissedAnnouncementInStore(dismissedAnnouncement.id);
-    setAnnouncements((announcements) =>
-      announcements.filter((a) => a.id !== dismissedAnnouncement.id),
-    );
-  };
+  const dismissAnnouncement = useCallback(
+    (dismissedAnnouncement: AnnouncementType) => {
+      addDismissedAnnouncementInStore(dismissedAnnouncement.id);
+      setAnnouncements((announcements) =>
+        announcements.filter((a) => a.id !== dismissedAnnouncement.id),
+      );
+    },
+    [addDismissedAnnouncementInStore],
+  );
 
-  const resetDismissedAnnouncements = () => {
+  const resetDismissedAnnouncements = useCallback(() => {
     setDismissedAnnouncementInStore([]);
-  };
+  }, [setDismissedAnnouncementInStore]);
 
   return (
     <AnnouncementsContext.Provider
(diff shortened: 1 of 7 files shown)


src/announcements/AnnouncementsContext.tsx
```diff
@@ -29,10 +42,23 @@ const AnnouncementsContextProvider: React.FC = ({children}) => {
...
+  const dismissAnnouncement = (dismissedAnnouncement: AnnouncementType) => {
+    addDismissedAnnouncementInStore(dismissedAnnouncement.id);
+    setAnnouncements((announcements) =>
+      announcements.filter((a) => a.id !== dismissedAnnouncement.id),
+    );
+  };
+
+  const resetDismissedAnnouncements = () => {
+    setDismissedAnnouncementInStore([]);
+  };
```
gorandalum, line 54: Wrap these with `useCallback`, just in case. When we return them from a Context, we kinda "loose control" of how they are used, and the function hash should be stable so using them in other hooks don't cause rerenders.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -34,22 +45,42 @@ export const Announcement = ({announcement}: Props) => {
...
+        style={style.close}
+        role="button"
+        hitSlop={12}
+        accessibilityHint={t(
+          DashboardTexts.announcemens.announcement.closeA11yHint,
+        )}
+        onPress={() => onDismiss(announcement)}
+      >
+        <ThemeIcon size={closeIconSize} svg={Close} />
+      </PressableOpacity>
```
gorandalum, line 58: Could we align the implementation of the dismiss button in this component and in `MessageBox`? Now one is done with absolute positioning while the other is done without. I don't have a preference of which implementation to use, just that I think it should be the same so the screen reader experience is equal between them.

gorandalum (changes requested):
Nice work! Some minor comments.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
KristianSelnas: I'll have a look

KristianSelnas: The `MessageBox` has two columns, where one contains the text and the other the close icon. This approach is very safe in terms of text overlapping the icon, but it also means that longer messages will appear to have an unnaturally large right margin. 

The absolute positioning approach makes long texts look more natural, but there is more risk of text overlapping the icon. Absolute positioning works as long as we know that the height of the icon is approximately the same height as the heading line height, and when we know that the heading is required. The `marginRight` set on the header prevents this from overflowing the icon.

There is no "best" solution here. Both have tradeoffs, as described. I'm not sure if there's even any point in using the same approach in both places. What's your opinion, @gorandalum ?

gorandalum: Good summary. Since the heading is required on announcement it makes sense that it can have absolute position, so I'm ok with different implementations. Just need to be sure that screen reader focus order is correct both on iOS and Android.

KristianSelnas: I've done a few tests now and the difference is not that big. Since the text is not justified the line wrapping will often lead to large right margins anyway, so the benefit of the absolute approach is not that great. For the sake of safety and consistency I'll use the same approach as `MessageBox`.

KristianSelnas:
Updated now to use the same strategy for close button placement and accessibility as `MessageBox`.

gorandalum (approved):

PR #3988, opened 2023-10-27
fix: show favorites in search results

adriawh:
### 🆕 Changes: 
- Not filtering out previous locations from results
- Deleted filter function

### Preview - now shows favorites in results
https://github.com/AtB-AS/mittatb-app/assets/70323886/fa035aae-c39c-4a78-a418-d83f4c3c384e


marius-at-atb:
This will show both favourites and previous search results in the current search results. Is this what's desired?

If the only change should be to show favourites as well, then consider sending `history` instead of `previousLocations` to be used in `filterCurrentLocation`

adriawh:
> This will show both favourites and previous search results in the current search results. Is this what's desired?
> 
> If the only change should be to show favourites as well, then consider sending `history` instead of `previousLocations` to be used in `filterCurrentLocation`

@marius-at-atb
When searching I expect to get the result that corresponds the best to my search, regardless of me having searched for it previously. As the search now hides previous locations and favorites while searching, I believe it should always show all results. 

What are your thoughts on this? @oyvindmikkelsenatb

marius-at-atb:
I agree that it makes sense to remove the filter, just making sure it is a conscious decision to remove it completely 👍

gorandalum (approved):

marius-at-atb (approved):

PR #3989, opened 2023-10-31
refactor: Better encapsulation of mobile token

gorandalum:
### Background
The mobile token logic was (and probably still is) quite brittle, and since it is used so many places throughout the app it has large consequences when it is not correct.

### Solution
This PR tries to simplify the usage around the app by encapsulating the logic better in the mobile-token module. This includes evaluation of fallbacks and what barcode to show, evaluating type and inspectable status of remote tokens during mapping of them, and simplified returned fields which is more focused on what the components really need.

Unfortunately changing this gives quite a large PR, and a large need for testing. Hopefully next time around doing changes to the mobile token won't have these large repercussions as the logic is more encapsulated.

### Acceptance criteria
- [ ] Test purchase ticket flow
- [ ] Test active tickets and details/QR code screen
- [ ] Test select travel token screen
- [ ] Test mobile token onbarding screen screen

All of these should be tested in different statuses, like inspectable, other token inspectable, loading, error.


src/mobile-token/MobileTokenContext.tsx
```diff
@@ -193,8 +206,7 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
       const cancelTimeoutHandler = timeoutHandler(() => {
         // When timeout has occured, we notify errors in Bugsnag
         // and set state that indicates timeout.
-        setTimeoutFlag(true);
-        setIsLoading(false);
+        setStatus('timeouted');
```
hanne-at-atb, line 209: I think it's spelled "timed-out", like "something has timed out"?
Ref. https://www.oxfordlearnersdictionaries.com/definition/american_english/time-out

hanne-at-atb (approved):
Great clean-up 🧹 🧼 
I just had a few minor things that I noticed 🌵

PR #3995, opened 2023-11-02
refactor: Make use-pollable-resource exhaustive deps fix

gorandalum:
- Adding prevState to dependency array of reload funcion is safe as
  it is a reference.
- Adding setisLoading to dependency array of reload function is
  safe as it is referentially stable.
- Adding reload to dependency array of the reload-on-focus
  useEffect is safe because the original
  load-on-reload-function-change useEffect now sets firstLoadIsDone
  to false.
- AbortController changed to ref as we don't want to add it in the
  dependency list for the useInterval hook.

And as a final note we should probably look into using useReactQuery
for this logic, as it supports most of what usePollableResource is
trying to achieve.

### Acceptance criteria
- [ ] Trip pattern updates should work as before: Should update every 20 seconds, not update when not focused, and always update when getting focus.

### DIFF
--- a/src/utils/use-pollable-resource.ts
+++ b/src/utils/use-pollable-resource.ts
@@ -47,7 +47,7 @@ export const usePollableResource = <T, E extends Error = Error>(
   const [isLoading, setIsLoading] = useIsLoading(false);
   const [error, setError] = useState<E | undefined>(undefined);
   const [state, setState] = useState<T>(initialValue);
-  const [abortController, setAbortController] = useState<AbortController>();
+  const abortControllerRef = useRef<AbortController>();
   const pollTime = pollingTimeInSeconds * 1000;
 
   const prevState = usePrevious(state);
@@ -75,12 +75,12 @@ export const usePollableResource = <T, E extends Error = Error>(
           }
         });
     },
-    [callback],
+    [callback, prevState, setIsLoading],
   );
 
   useEffect(() => {
     const abortController = new AbortController();
-    setAbortController(abortController);
+    abortControllerRef.current = abortController;
     reload('WITH_LOADING', abortController);
     return () => abortController.abort();
   }, [reload]);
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):

src/utils/use-previous.ts
```diff
@@ -1,5 +1,9 @@
 import {useEffect, useRef} from 'react';
 
+/**
+ * Returns a current reference to the previous value. The returned value is safe
+ * to use in dependency arrays as refs won't trigger reruns.
+ */
```
marius-at-atb, line 6: Good to have comments like this imo :+1:

mikaelbr: Agreed. Easy to skip, but jsdoc/tsdoc like comments (as included on `usePollableResource`) is very handy to understand context for parameters and intention of function as well. So could even add parameters here and description of return value 😄

mikaelbr (approved):
Haven't seen any technically wrong things, but kind of hard to say for this hook I think. It's fairly complex. Suppor the idea of either using something like react query or swr type of semantics at some point which also relies more heavily on caching and mutations.

src/utils/use-pollable-resource.ts
```diff
@@ -75,26 +75,27 @@ export const usePollableResource = <T, E extends Error = Error>(
...
         });
     },
-    [callback],
+    [callback, prevState, setIsLoading],
   );
 
   useEffect(() => {
     const abortController = new AbortController();
-    setAbortController(abortController);
+    firstLoadIsDone.current = false;
```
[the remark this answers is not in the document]
gorandalum: I have tested a bit locally, and it seems fine. 

The ref is only used in the `useEffect` which reloads on focus, and I think the reason is so reloads doesn't happen until the initial query is done.

I think it is more correct with the change, as earlier if a screen was focused and callback function changed at the same time, both the initial query and focus reload would be run. With the change now the focus reload won't happen on callback change.

PR #3997, opened 2023-11-02
refactor: Some low hanging exhaustive deps fixes

gorandalum:
### Acceptance criteria
- [ ] Logging of PostHog events should work as before
- [ ] Changing between dark/light mode on OS level and app settings should both work as before


marius-at-atb (approved):
Lgtm!

PR #3999, opened 2023-11-02
refactor: Deps fix in useDoOnceWhen

gorandalum:
The input function (fn) was not in the dependency list of the
useEffect running the function. JUst naively adding it could not
be done, as that would make the effect rerun on every render as
input functions are not stable references.

The solution would be either:
- Add the function to the dependency list, and users of the hook
  would need to ensure that they only send in stable references,
  for example by wrapping function with useCallback.
- Use the input function as a ref, which would not rerun the effect
  even though the function changes.

The latter option was selected as that is easier on the client, and
is the same as it is working before.


marius-at-atb (approved):
Very nice testing!

PR #4001, opened 2023-11-02
build: rename NFK dev and staging apps

toretefre:
Adding (dev) for debug and (staging) for stagingapp in visible name on device.


toretefre:
Might require some rebuilding of profiles.

mortennordseth (approved):

PR #4002, opened 2023-11-02
refactor: rename useDebounce.ts to match the naming convention

reidzeibel:
Inside `utils` folder we have a file `useDebounce.ts` which is a hook but is named differently to other hooks in the same folder.
This PR should change `useDebounce.ts` into `use-debounce.ts` so it matches the other hooks.


rosvik (approved):
LGTM 👍 🎉

marius-at-atb (approved):
Nice!

tormoseng:
App as before! OK.

PR #4006, opened 2023-11-03
refactor: Fix medium-to-low hanging exhaustive-deps errors

gorandalum:


src/stacks-hierarchy/Root_PurchaseTariffZonesSearchByMapScreen.tsx
```diff
@@ -37,17 +37,17 @@ export const Root_PurchaseTariffZonesSearchByMapScreen = ({
...
 
   useEffect(() => {
-    setSelectedZones({
-      ...selectedZones,
+    setSelectedZones((prev) => ({
+      ...prev,
       to: toTariffZone,
-    });
+    }));
   }, [toTariffZone]);
```
jorelosorio, line 51: Isn't it necessary to add setSelectedZones to the dependency list?

gorandalum: > The set functions returned by useState also have stable identity, so you will often see them omitted from the dependencies too. If the linter lets you omit a dependency without errors, it is safe to do.

https://react.dev/learn/synchronizing-with-effects#why-was-the-ref-omitted-from-the-dependency-array

gorandalum:
> There are several objects in dependencies arrays here (fromPlace, toPlace, tariffZones) that I think will cause updates on every re-render. Is that something that should be fixed?

I tested locally now, and they are stable references. Either memoed or gotten through react navigation params. However, I found some issue with `UserProfilesWithCount` being recreated too often, so I made some improvements now. The `updateOffer` request is now sent in the same manner as before, not more often because of these changes.

rosvik (approved):
LGTM then!

One thing I thought about: I find it a bit scary that we have reusable components that require some params to be stable in order to be performant. (Example from this PR: `RecentFareContract` takes `recentFareContracts` that's used in a dependency array) Is there a way to "mark" these props, so that next time someone uses the component, they know the prop needs to be stable?

> Either memoed or gotten through react navigation params.

Is this an universal thing that navigation params are always stable? 👀

gorandalum:
From Johannes:
> One thing I thought about: I find it a bit scary that we have reusable components that require some params to be stable in order to be performant. (Example from this PR: `RecentFareContract` takes `recentFareContracts` that's used in a dependency array) Is there a way to "mark" these props, so that next time someone uses the component, they know the prop needs to be stable?


Good question! I agree that it can be quite scary, especially when network requests can be triggered way to often if objects are recreated every render. In the example you provide it is not really that scary, since it is just an `useMemo`, and I even think in this situation (and maybe most situations?) the `useMemo` could probably just be removed. Some related quotes from the [doc](https://react.dev/reference/react/useMemo):
> You should only rely on useMemo as a performance optimization. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

>In general, unless you’re creating or looping over thousands of objects, it’s probably not expensive. 

Regarding those more scary situations with network requests, one solution is to avoid triggering network requests in `useEffect` hooks when it is possible. A related quote from the [doc](https://react.dev/learn/you-might-not-need-an-effect):

> You don’t need Effects to handle user events. For example, let’s say you want to send an /api/buy POST request and show a notification when the user buys a product. In the Buy button click event handler, you know exactly what happened. By the time an Effect runs, you don’t know what the user did (for example, which button was clicked). This is why you’ll usually handle user events in the corresponding event handlers.

This would for example mean we change the PurchaseOverviewScreen from code like this (totally pseudocode):
```tsx
const [tariffZone, setTariffZone] = useState<TariffZone>();

useEffect(() => { 
  updateOffer(tariffZone); // Updating offer based on state change
}, [tariffZone])

<Button onPress=(() => setTariffzone(selectedTariffZone)) />
```
to this:
```tsx
const [tariffZone, setTariffZone] = useState<TariffZone>();

<Button onPress=(() => {
  updateOffer(selectedTariffZone) // Updating offer directly in user event handler
  setTariffZone(selectedTariffZone);
}) />
```

With the latter example the state is purely used for what the app should render, not for performing network requests, which is safer. This would make it necessary to often do the network request in several places, for instance every event handler that updates tariff zone, time, product,  or travellers. But it is safer than doing this:
```tsx
useEffect(() => {
  updateOffer(tariffZone, time, product, travellers);
}, [tariffZone, time, product, travellers])
```

Another question from Johannes:
> Is this an universal thing that navigation params are always stable? 👀

React Navigation params are not updated for every render, they are set when navigating (with `navigation.navigate`) or if updating them with `navigation.setParams`. So in general it is pretty safe to use `useEffect` on React Navigation params to trigger network requests, and also necessary as we can't do it in the user event that triggered the navigation.

rosvik:
@gorandalum Thanks for the thorough answer 💯

The part I'm still a bit unsure about is this:

> In general, unless you’re creating or looping over thousands of objects, it’s probably not expensive.

Sure, looping over a short list of objects isn't expensive in itself, and if I'm unsure, it's not that hard to test. But if I'm rendering a list of complex components, it's harder to know. I can connect Flipper and dig through the render tree to see, but it's a bit of a hassle, especially if I need to do it every time I change a bit of related state logic.

So I either need to become more confident that my state isn't causing excessive re-renders, or I need to be confident that my component isn't too slow to render. Is there a good way to measure how long a component takes to render in React?

gorandalum:
> @gorandalum Thanks for the thorough answer 💯
> 
> The part I'm still a bit unsure about is this:
> 
> > In general, unless you’re creating or looping over thousands of objects, it’s probably not expensive.
> 
> Sure, looping over a short list of objects isn't expensive in itself, and if I'm unsure, it's not that hard to test. But if I'm rendering a list of complex components, it's harder to know. I can connect Flipper and dig through the render tree to see, but it's a bit of a hassle, especially if I need to do it every time I change a bit of related state logic.
> 
> So I either need to become more confident that my state isn't causing excessive re-renders, or I need to be confident that my component isn't too slow to render. Is there a good way to measure how long a component takes to render in React?

Regarding measuring, the doc has some tips on that in the paragraph [How to tell if a calculation is expensive?](https://react.dev/reference/react/useMemo#how-to-tell-if-a-calculation-is-expensive).

It is hard to find to give a definitive answer regarding when to memo and when to not, maybe discuss with someone when in doubt.

PR #4007, opened 2023-11-03
chore: remove unused state for tickeing info

rosvik:
Removes some remaining state for some old ticketing information, that's no longer in use.


marius-at-atb (approved):
Lgtm!

PR #4009, opened 2023-11-06
feat: add location permission onboarding screen

marius-at-atb:
Resolves https://github.com/AtB-AS/kundevendt/issues/8622

Several things going on in this PR:
- Introduces the new `OnboardingScreen` module which is used for push notifs, location and share travel habit onboarding screens.
- Adds the new `Root_LocationWhenInUsePermissionScreen` (using OnboardingScreen, see screenshot below).
- Adds the `locationWhenInUsePermissionOnboarded` flag in AppState, with attached complete and restart methods. This is used to show the location permission screen at the right time.
- Requests location permission in the new screen
- Refactored the `getCurrentPosition` method to also handle the case when no permission granted. An issue with the useRef solution [discussed here](https://github.com/AtB-AS/mittatb-app/pull/3975) is when permission is just granted, location will still take some unknown time to be ready, and is still undefined on the following line. Also location permission handling doesn't really belong in other places such as useParkingViolations. Much better to contain it in GeolocationContext. The new solution essentially combines the best of @KristianSelnas's and @gorandalum's solutions.
- Adds the `useInitialCoordinates` hook

Notes:
- Removing the old way to request permission when onboarding completed, is separated into [its own PR](https://github.com/AtB-AS/mittatb-app/pull/4017).
- Regarding naming, I originally went with "locationWhenInUsePermission" to separate it from background location permission. However I'm considering whether to switch to just "locationPermission" throughout, let me know if you have any thoughts about this.

The new location permission screen:


src/permission-screen/PermissionScreen.tsx
rosvik, line 1: When I talked with Øyvind, I mentioned the option to use this screen pattern for different onboarding screens as well (If different OMS orgs want to show different info). Should we call it a `OnboardingScreen` instead of `PermissionScreen` in that case? Since there's really nothing that makes this specific to permissions.

marius-at-atb: Yup, renaming to `OnboardingScreen`!

KristianSelnas:
Solid piece of work! Taking time to do refactors like this is crucial for future maintenance. Good stuff. I especially like the `OnboardingScreen` component. A pet peeve of mine in this code base is the extensive use of local styling, rather than using UI components as building blocks. This is one small step in the right direction.

Your refactor in `getCurrentLocation` resembles my original approach, while @gorandalum argued for using the locationRef approach and making it synchronous. I think your arguments for using `Geolocation.getCurrentPosition` and making it async are sound. However, I suppose Gøran might think otherwise?

marius-at-atb:
@KristianSelnas Thanks, I agree!
I just pushed another commit with a `useInitialCoordinates` hook as well that uses the new `getCurrentLocation` method, and extracts the logic from `TariffZonesSelectorMap`.
The usage is now as simple as `const initialCoordinates = useInitialCoordinates();` for any component.

src/utils/use-initial-coordinates.ts
gorandalum, line 1: Some jsdoc maybe, now that it is in src/utils.

src/GeolocationContext.tsx
```diff
@@ -255,17 +257,60 @@ export const GeolocationContextProvider: React.FC = ({children}) => {
...
     checkPermission();
-  }, [appStatus]);
+  }, [appStatus, state.status]);
 
   useEffect(() => {
     locationRef.current = state.location;
   }, [state.location]);
 
-  const getCurrentPosition = useCallback(() => locationRef.current ?? null, []);
+  const getCurrentLocation = useCallback(
```
gorandalum, line 266: I don't really like [using errors for flow control](https://web.archive.org/web/20140430044213/http://c2.com/cgi-bin/wiki?DontUseExceptionsForFlowControl). In this code there is nothing really wrong if the user hasn't shared location, and I think it is better to have this represented in an object with a status and location.

src/onboarding-screen/OnboardingScreen.tsx
gorandalum, line 1: Nice!

gorandalum (changes requested):

src/utils/use-initial-coordinates.ts
marius-at-atb: Yes!

src/GeolocationContext.tsx
marius-at-atb: Refactored to use state instead of errors :+1:

gorandalum (approved):
Great!

PR #4010, opened 2023-11-06
fix: Beacons was not properly reading permissions

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/8615


src/beacons/use-beacons.ts
```diff
@@ -111,10 +111,11 @@ export const useBeacons = () => {
     }
 
     if (granted) {
-      await storage.set(storeKey.beaconsConsent, "true");
       // Initialize beacons SDK after consent is granted
-      await initializeBeaconsSDK();
-      Kettle.grant(BEACONS_CONSENTS);
+      await initializeBeaconsSDK().then(async () => {
+        Kettle.grant(BEACONS_CONSENTS);
+        await storage.set(storeKey.beaconsConsent, "true");
```
jorelosorio, line 117: Make sure is granted only when the initialization went OK and granted also went OK, before it just updated this value first and then the initialization failed, letting the app in a wrong state.

src/beacons/use-beacons.ts
```diff
@@ -166,14 +167,14 @@ const requestAndroidPermissions = async () => {
   );
 
   if (locationStatus !== RESULTS.GRANTED) {
-    return false;
+    return true;
```
jorelosorio, line 170: According to the Figma, if this is granted or not, then stop and continue with the onboarding, and is true because the Bluetooth permission was granted.

gorandalum (approved):

PR #4013, opened 2023-11-07
test: e2e-test improvements

tormoseng:
The E2E-tests started to fail when there was added several announcements (the add favorites button "disappeared" on frontpage). Here, the frontpage tests first closes all dissmissible global messages and announcements before doing the tests.

In this PR:
- test-ids on global messages and announcements
- remove messages test case on frontpage

Note! Also added an index.js in its own folder that removes all log messages for local testing with metro. Can be copied to root during tests.

### DIFF
--- a/e2e/test/specs/travelsearch.e2e.ts
+++ b/e2e/test/specs/travelsearch.e2e.ts
@@ -31,6 +31,8 @@ describe('Travel search', () => {
      */
 
     try {
+      timeIsEqual('28 min', 'Times is: 29 minutes');
+
       await ElementHelper.waitForElement('id', 'searchFromButton');
       await FrontPagePage.searchFrom.click();
       await SearchPage.setSearchLocation(departure);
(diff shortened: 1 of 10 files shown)


e2e/test/specs/travelsearch.e2e.ts
```diff
@@ -81,9 +83,27 @@ describe('Travel search', () => {
...
       await NavigationHelper.back();
+
+      /**
+       * Check if two times are equal
+       * Note! Simple check for this search wher travel time is in minutes
+       * @param time1 String that contains some digits to compare
+       * @param time2 String that contains some digits to compare
+       * @param allowedDiff Allow a time difference
+       */
+      function timeIsEqual(
```
marius-at-atb, line 97: Could consider renaming to e.g. `timeIsCorrect` or `timeIsMatching` since it is not a strict equals check. In theory it could be a mistake that matches..

marius-at-atb (approved):
Nice! Lgtm :+1:

jorelosorio (approved):

e2e/test/specs/travelsearch.e2e.ts
tormoseng: I agree! Fixed.

tormoseng:
Related fix for notification onboarding: https://github.com/AtB-AS/mittatb-app/pull/4022

tormoseng:
Related fix for empty state: https://github.com/AtB-AS/mittatb-app/pull/4032

tormoseng:
Tests OK: https://github.com/AtB-AS/mittatb-app/actions/runs/6861039536

PR #4011, opened 2023-11-06
feat: Register for push notifications

KristianSelnas:
This PR solves two issues:
1. https://github.com/AtB-AS/kundevendt/issues/13902: Adds a hook for registering devices for push notifications. This will also trigger a permission dialog if required.
2. Partially solves https://github.com/AtB-AS/kundevendt/issues/4286: A notifications screen is added in settings, but the only option added from #4286 is the push on/off option. Fine grained push settings will be added in a separate pr.

<details>
<summary>Screenshots</summary>

Simulator Screenshot - iPhone 13 - 2023-11-07 at 11 05 19
Simulator Screenshot - iPhone 13 - 2023-11-07 at 11 11 02
Simulator Screenshot - iPhone 13 - 2023-11-07 at 11 05 40
Simulator Screenshot - iPhone 13 - 2023-11-07 at 11 06 08

</details>

fixes https://github.com/AtB-AS/kundevendt/issues/13902
partially fixes https://github.com/AtB-AS/kundevendt/issues/4286

PR #4019, opened 2023-11-09
chore: Add Platform.Version as http header

KristianSelnas:
Discussed here: https://github.com/AtB-AS/mittatb-app/pull/4011#discussion_r1385068257

Closes https://github.com/AtB-AS/kundevendt/issues/15395

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -4,9 +4,11 @@ import {API_BASE_URL, APP_VERSION, IOS_BUNDLE_IDENTIFIER} from '@env';
 import {getAxiosErrorMetadata, getAxiosErrorType} from './utils';
 import Bugsnag from '@bugsnag/react-native';
 import {
+  AppIdentifierHeaderName,
   AppVersionHeaderName,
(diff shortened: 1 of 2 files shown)


src/api/client.ts
```diff
@@ -86,6 +87,7 @@ function requestHandler(config: AxiosRequestConfig): AxiosRequestConfig {
   }
 
   config.headers[AppVersionHeaderName] = APP_VERSION;
+  config.headers[PlatformVersionHeaderName] = Platform.Version;
```
mjenssen, line 90: This looks fine for adding the platform version, but why isn't platform added in this function (it's also missing from `src/api/headers.ts`). Same with app id.

mjenssen (approved):

src/api/client.ts
gorandalum: I agree. We should move adding of the 'atb-app-identifier' and 'atb-app-platform' headers to here as well, instead of in `requestIdTokenHandler`.

PR #4021, opened 2023-11-10
fix: remove usage of hideTravellerDescription

adriawh:
❌ **Problem:**
> Info texts are not shown for users which had toggled of the "Show info" toggle which we previously had in the app.

✅ **Solution:** Removed the usage of the preference


gorandalum (approved):

marius-at-atb (approved):
:+1:

PR #4024, opened 2023-11-10
refactor: Fix outdated id token error

gorandalum:
### Background
We had seen situations in production where the id token was for an anonymous account, while the user was actually logged in. That means the user purchased periodic tickets (only available on logged in accounts), but the ticket actually ended up on another anonymous account. The reason is probably that the user got id token for the anonymous account used on that phone before the user logged in.

After looking at it it seems like an issue with Firebase Auth React Native lib. I am not sure if it is a bug or not, but I couldn't find any sources documenting it. The issue is something like (pseudocode):
```tsx
const [userId, setUserId] = useState<string>();

useEffect(() => {
  auth().onUserChanged(user => setUserId(user?.uid));
}, [])

useEffect(() => {
  if (userId) {
    /*
    Whoops, auth().currentUser gives outdated user, and not the same user as was
    in onUserChanged callback that happened just before (which updated the userId)
    */
    auth().currentUser.getIdToken().then(idToken => {...});
  }
}, [userId])
```

The fix is in essence to instead store the user object (still pseudocode):
```tsx
const [user, setUser] = useState<string>();

useEffect(() => {
  auth().onUserChanged(newUser => setUser(newUser));
}, [])

useEffect(() => {
  user?.getIdToken().then(idToken => {...});
}, [user])
```

Just a note that the `onUserChanged` triggers multiple times with the same user object, so it was necessary to add an `isEqual` check so we don't set a new user object to often, which would then refetch the id token too often.

In addition it was changed so that we now store the `user` and the `idToken` in the `AuthContext` state. This is slightly better as we inside of the auth module store the "original source" of the data, but expose the specific fields out through the provider.


src/auth/use-fetch-id-token-with-custom-claims.ts
gorandalum, line 1: This file is mainly the same as the old `use-check-if-account-creation-finished`, it is just that GitHub didn't pick up that this is a moved file.

marius-at-atb (approved):
Looks good as far as I can tell without being an expert on the auth logic..
Very helpful PR description :+1:

tormoseng:
Tested OK! Will just have to see if it happens in prod on versions > 1.43.

PR #4025, opened 2023-11-10
refactor: Make translate func a stable reference

gorandalum:
The translate function (t) from lobot was not a stable reference,
and recreated on every render. This ment it could not be added to
dependency lists without causing the hooks to be run too often.

In this PR we wrap the translate function with 'useCallback', and
add the 't' function to the dependency lists.

I also created an [issue at the lobo-t repo](https://github.com/leile/lobo-t/issues/7), and offered creating a
PR there, but no response yet.

### DIFF
--- a/src/beacons/use-beacons.ts
+++ b/src/beacons/use-beacons.ts
@@ -127,7 +127,7 @@ export const useBeacons = () => {
         buttonPositive,
       },
     };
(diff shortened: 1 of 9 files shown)


marius-at-atb (approved):
Brilliant!

PR #4028, opened 2023-11-10
refactor: Some more low hanging exhaustive-deps fixes

gorandalum:


src/components/map/Map.tsx
```diff
@@ -32,8 +32,8 @@ export const Map = (props: MapProps) => {
     () =>
       initialLocation && initialLocation?.resultType !== 'geolocation'
         ? initialLocation.coordinates
-        : currentLocation?.coordinates || FOCUS_ORIGIN,
-    [],
+        : getCurrentPosition()?.coordinates || FOCUS_ORIGIN,
+    [getCurrentPosition, initialLocation],
```
marius-at-atb, line 36: Note: could be that the [useInitialCoordinates](https://github.com/AtB-AS/mittatb-app/blob/68ec54af07b9925613f2194e4c3cc56f11c9fe97/src/utils/use-initial-coordinates.ts) hook can simplify some things here when available

src/fare-contracts/ValidityLine.tsx
```diff
@@ -142,7 +142,7 @@ const useAnimatedVerticalLineOffset = (animate: boolean | undefined = true) => {
         easing: Easing.linear,
       }),
     ).start();
-  }, [animate]);
+  }, [animate, animatedOffset]);
```
marius-at-atb, line 145: Just a quick question - why is `animatedOffset` defined as `useRef(new Animated.Value(0)).current` instead of just inputting `new Animated.Value(0)` inline to `Animated.timing`??

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/DeparturesWidget.tsx
```diff
@@ -49,7 +49,7 @@ export const DeparturesWidget = ({
   const {location} = useGeolocationState();
   const {state, loadInitialDepartures, searchDate} = useFavoriteDepartureData();
 
-  useEffect(() => loadInitialDepartures(), [favoriteDepartures]);
+  useEffect(() => loadInitialDepartures(), [loadInitialDepartures]);
```
marius-at-atb, line 52: Could this break functionality where the data for the widget is updated when `favoriteDepartures` changes?

marius-at-atb (approved):
Lgtm!

src/fare-contracts/ValidityLine.tsx
gorandalum: The first argument to the `Animated.timing` is the output value, and is continuously updated during the animation effect. The value is then used in an actual animated component like, in the `transform` function in an `Animated.View`. If just using inline `new Animated.Value(0)`, we would have no reference to it and not able to use it on animated components.

https://reactnative.dev/docs/animations

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/DeparturesWidget.tsx
gorandalum: That is handled, as this is the signature for `loadInitialDepartures`:
```tsx
const loadInitialDepartures = useCallback(
  () =>
    dispatch({
      type: 'LOAD_INITIAL_DEPARTURES',
      locationOrStopPlace,
      favoriteDepartures,
    }),
  [locationOrStopPlace?.id, favoriteDepartures],
);
```

marius-at-atb: Great!

PR #4029, opened 2023-11-13
feat: use abt-time for fare contract validity

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/10074

- Added a new TimeContextProvider to supply the application with the server-synchronized time, provided by abt-mobile-client-sdk. It provides a `now` field, which is the current time in milliseconds updated every 2,5 seconds.
- Everything related to ticket inspection uses `now` from TimeContext instead of Date.now(), to prevent using device time to influence ticket validity.
- More of the util functions in `src/ticketing/utils.ts` now takes `now` as a parameter, instead of using Date.now()
- Added the current server time (Now) under mobile token state in the debug menu.


src/time/TimeContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+   * validity of fare contracts.
+   */
+  now: number;
+};
+
+const TimeContext = createContext<TimeContextState | undefined>(undefined);
+
+export const TimeContextProvider: React.FC = ({children}) => {
+  const [now, setNow] = useState(Date.now());
+  useInterval(() => clock.currentTimeMillis().then(setNow), 2500);
```
marius-at-atb, line 18: Should probably send the `triggerImmediately` argument as `true` to `useInterval` to avoid 2,5s where the time comes from `Date.now()`

marius-at-atb: Also, what if the device is offline?
Perhaps the `TimeContextProvider` should separate between something like `serverNow` and `deviceNow`.
Supporting both seem to be a good idea imo, and `serverNow` could then be undefined as default?

marius-at-atb (changes requested):
Nice solution!
Good thinking to add it in debugInfoScreen as well.

However I think the `now` in `TimeContext` should be looked at more closely.
See comments.

I actually like the idea to use a time value from a context not only for validation, but also e.g. for syncing the time used by different components. For example flexible transport search result filtering and messaging is a use case for `TimeContext` where normal device time is sufficient.

src/time/TimeContext.tsx
rosvik: > what if the device is offline?

The library isn't very transparent about this right now, and it seems like it also fails silently by returning device time if it can't fetch from the server. So since we can't trust the clock 100% anyway, I think having the default to Date.now is easier to use in the app, since we never have to add fallback logic like `serverTime || deviceTime`.

rosvik:
@marius-at-atb 

> I actually like the idea to use a time value from a context not only for validation, but also e.g. for syncing the time used by different components. For example flexible transport search result filtering and messaging is a use case for TimeContext where normal device time is sufficient.

I also like the sound of this! It's sometimes hard to keep track of how often real time data is updated, so using a global timer could be a good way of doing it. I'm not sure if 2.5s is always a good balance between staying up-to-date and not doing too many re-renders though. E.g. in departures I think we use 10s. Also need to look at if it's a good idea performance wise to do all time-releated re-renders on the same interval, or if it's better to space them out.

Definitely something to consider, but out of the scope of this PR i think.

gorandalum (approved):
Nice!

If I could wish for something is that we could still have something like `Date.now()`-ish, which could get a more exact time, and not up to 2.5 seconds old time.

Maybe something like this could be added? (Totally pseudocode and of the top of my head atm)
```tsx
let diff = 0;

clock.currentTimeMillis().then(now => {
  diff = Date.now() - now;
})

const DateNow = () => Date.now() - diff; // -? +? surprisingly hard ^-^
```

rosvik:
@gorandalum That makes sense, I think. I can make a new PR with the suggestion 👍 

@marius-at-atb Does this look good to you?

src/time/TimeContext.tsx
marius-at-atb: Oh right! Good to catch that :+1:
Regarding `triggerImmediately`, maybe the default value in the hook should be true? Perhaps it was unintuitive to even consider this when using it?

marius-at-atb (approved):
_Now_ we are talking 😛 

Definitely many things to consider regarding a now time for sync between components, and can be left for another time :+1:

I also agree with @gorandalum's suggestion.

The offline handling should be looked at closely. As @mikaelbr mentioned previously, it is quite important. 

I would also still suggest to change the name somewhat to indicate what kind of now it is. It would not occur to me if I found a `now` in a `TimeContext`, that it may not work when offline.

src/time/TimeContext.tsx
rosvik: Looks like it's close to 50/50 true/false on use of `triggerImmediately` in the app. Although it can be easy to forget,  I like that default is the same as javascripts setInterval now.

rosvik:
@marius-at-atb

I agree that we should be certain that this works when offline. I've tried to implement this in a way that it's _at least_ as good as device time, and never returns null/undefined. The only time I've seen it fall back to device time, is when having no network connection after reinstalling the app. So I this should be a reliable source for time.

I have tried to cover everything in the acceptance criteria if you want to have a look, or expand it: https://github.com/AtB-AS/kundevendt/issues/10074#issuecomment-1809903673

We could rename the context values if it's not intuitive. Is `serverTime` better do you think?

marius-at-atb:
@rosvik I didn't realise that the Entur library functions handle offline that well.

So in that case it is probably not necessary, but maybe still a good idea to call it e.g. `serverNow`, as it communicates more about what it is, and what the motivation is for using it instead of `Date.now()`?

jorelosorio:
> @rosvik I didn't realise that the Entur library functions handle offline that well.
> 
> So in that case it is probably not necessary, but maybe still a good idea to call it e.g. `serverNow`, as it communicates more about what it is, and what the motivation is for using it instead of `Date.now()`?

I remember that for that clock time I used https://github.com/MobileNativeFoundation/Kronos for iOS they had a similar approach on Android perhaps https://github.com/lyft/Kronos-Android, so they have moved that library into a separated module in their SDK, but at the end is more likely this implementation does.

rosvik:
@jorelosorio Good to know! I was looking for libraries solving this before. I hope Enturs implementation is good enough, if not that's something to we should look into 👍

marius-at-atb:
@rosvik Great – yes, good to go!

PR #4031, opened 2023-11-13
fix: Render crash on TravelTokenBox

gorandalum:
React components shouldn't be created like normal javascript functions,
as that bypasses the rules and optimisations of React. The
ErrorMessages component here did not work as intended, as it made
the app crash with 'Rendered fewer hooks than expected' since the
hooks inside the component was rendered conditionally.

The solution now was inlining the error  message boxes into the
TravelTokenBox component.

Fixes:
https://github.com/AtB-AS/kundevendt/issues/10087#issuecomment-1808164753
https://app.bugsnag.com/atb-as/mittatb/errors/655221f4bdc7f80008277524?filters%5Buser.id%5D=d21ab310-04c3-4ba4-afc5-dcc88679951f&event_id=6552221900cc77ff8eac0000

### DIFF
--- a/src/travel-token-box/TravelTokenBox.tsx
+++ b/src/travel-token-box/TravelTokenBox.tsx
@@ -20,7 +20,8 @@ export function TravelTokenBox({
 }) {
   const styles = useStyles();
   const {t} = useTranslation();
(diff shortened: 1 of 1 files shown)


marius-at-atb (approved):
:+1:

PR #4032, opened 2023-11-13
test: handle empty state in e2e

tormoseng:
The e2e tests are not using location, so with the empty state change, a test-id was not found.

Added test-ids to the empty state, and updated the e2e-frontpage test.


marius-at-atb (approved):
Looks good!

PR #4033, opened 2023-11-14
feat: Enable exhaustive deps eslint

gorandalum:
At the same time disable the rule for the places where we still
break the rule, and we will fix them as we go.


marius-at-atb (approved):
There are still quite a lot of them, so let's keep pushing to actually fix this in the coming sprints!

PR #4034, opened 2023-11-14
chore: update react-native-webview to v13.6.2

reidzeibel:
Related to https://github.com/AtB-AS/kundevendt/issues/15480

There are some issues happening when adding payment method in the Android version of the app, this dependency update hopefully fixes that and we should update the version anyway to help with stability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "react-native-screens": "^3.19.0",
     "react-native-svg": "^13.7.0",
     "react-native-tab-view": "^3.3.4",
-    "react-native-webview": "^11.26.0",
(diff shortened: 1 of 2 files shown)


gorandalum (approved):
Note that if this doesn't work it can also be possible to look at using react-native-inappbrowser-reborn, which was introduced in the app for the Vipps login functionality.

marius-at-atb (approved):

reidzeibel:
@gorandalum seems like it fixes the issue, it was reproducible using the staging app from yesterday, and once I update the app, the crash no longer happens (at least when I tried it multiple times, like maybe around 50 times at the minimum).

PR #4040, opened 2023-11-15
fix: Not move to antarctica when opening filters sheet

gorandalum:
The reason for the bug was that the Map tried to show a walking
path, but since the filters sheet uses so much of the screen it
wasn't enough space to show to walk path, even if it zoomed all
the way out (showing Antarctica).

The fix is that when the filter sheet is opened, we needed to
reset the map state that said that the departures sheet and the
walk path should be visible.

One drawback of this solution is that there is no animation when
switching from the departures sheet to the filters sheet, but I
couldn't get that to work without causing havoc.

### DIFF
--- a/src/components/map/components/filter/MapFilter.tsx
+++ b/src/components/map/components/filter/MapFilter.tsx
@@ -1,31 +1,17 @@
 import React from 'react';
 import {Button} from '@atb/components/button';
-import {useBottomSheet} from '@atb/components/bottom-sheet';
-import {MapFilterSheet} from './MapFilterSheet';
-import {MapFilterType} from '../../types';
 import {StyleSheet} from '@atb/theme';
 import {shadows} from '../shadows';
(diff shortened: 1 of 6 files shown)


rosvik (approved):
Nice! Looks good to me 🇦🇶

src/components/map/components/filter/MapFilter.tsx
```diff
@@ -1,40 +1,29 @@
...
 import {Filter} from '@atb/assets/svg/mono-icons/actions';
 import {useAnalytics} from '@atb/analytics';
 
 type MapFilterProps = {
-  onFilterChanged: (filter: MapFilterType) => void;
+  onPress: () => void;
   isLoading: boolean;
 };
-export const MapFilter = ({onFilterChanged, isLoading}: MapFilterProps) => {
+export const MapFilter = ({onPress, isLoading}: MapFilterProps) => {
```
marius-at-atb, line 12: Could be renamed to MapFilterButton

PR #4045, opened 2023-11-16
feat: Use native user location pin

gorandalum:
The non-native user location pin has an onPress handler which is
hard to circumvent, and the hitbox is quite wide. This means when
the user location pin was close to an entity that should be
clickable (like bus stop, bike station, etc.) the click was not
registered by the map.

To fix this we change to the native location pin. The native
location pin is not interactable, which is kinda exactly what we
want here, as we don't want to register clicks on the location pin.

Another minor drawback is that it is not as pretty as the
non-native pin, but this was ok by designers.

Relevant issue about UserLocation pin blocking presses on
underlying components:
https://github.com/rnmapbox/maps/issues/241

Gif before/after:
<div>
<image width=350 src='https://github.com/AtB-AS/mittatb-app/assets/675421/e1c58a8f-531e-4c40-818a-9da840b4e3b1'/>
<image width=350 src='https://github.com/AtB-AS/mittatb-app/assets/675421/6b9b55e5-9f06-4c90-9c73-e95b458890dc'/>
</div>

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -3,7 +3,7 @@ import {useAnalytics} from '@atb/analytics';
 import {FOCUS_ORIGIN} from '@atb/api/geocoder';
 import {StyleSheet} from '@atb/theme';
 import {MapRoute} from '@atb/travel-details-map-screen/components/MapRoute';
-import MapboxGL from '@rnmapbox/maps';
+import MapboxGL, {UserLocationRenderMode} from '@rnmapbox/maps';
 import {MapState} from '@rnmapbox/maps/lib/typescript/components/MapView';
 import {Feature} from 'geojson';
 import React, {useMemo, useRef} from 'react';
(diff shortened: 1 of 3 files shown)


marius-at-atb (commented):
`renderMode={UserLocationRenderMode.Native}` is deprecated according to the docs:
https://github.com/rnmapbox/maps/blob/main/docs/UserLocation.md#rendermode

Better to use `NativeUserLocation`?

gorandalum:
> `renderMode={UserLocationRenderMode.Native}` is deprecated according to the docs: https://github.com/rnmapbox/maps/blob/main/docs/UserLocation.md#rendermode
> 
> Better to use `NativeUserLocation`?

Good catch :) I just didn't bother with it now, as it isn't deprecated in our library version, and our version doesn't support `NativeUserLocation`. It was deprecated [4 days ago](https://github.com/rnmapbox/maps/commit/0078fea553e33f0a869992fe45e5bb4469d98029#diff-66436477f218f421bf944fe38a403b9deb611cf6770c41a0bc693d881a0ea64cR85).

marius-at-atb (approved):

PR #4047, opened 2023-11-17
feat: add getServerNow to TimeContext

rosvik:
ref. https://github.com/AtB-AS/mittatb-app/pull/4029#issuecomment-1810328096

Adds getServerNow() to TimeContext, providing our best estimate for the current server time based on the diff from device time, last time it was updated.

### Example usage

When logging serverNow every time it is updated in TimeContext, and logging the state in an useInterval:

```js
useInterval(
  () => {
    console.log('Date.now()', Date.now(), 'getServerNow()', getServerNow(), 'serverNow', serverNow);
  },
  500,
  [getServerNow],
);
```

Gives this in the console
  
```
 LOG  Date.now() 1700211225129 getServerNow() 1700211225144 serverNow 1700211224576
 LOG  Date.now() 1700211225646 getServerNow() 1700211225661 serverNow 1700211224576
 LOG  Date.now() 1700211226181 getServerNow() 1700211226196 serverNow 1700211224576
 LOG  Date.now() 1700211226696 getServerNow() 1700211226711 serverNow 1700211224576
 LOG  TimeContext: Updating serverNow. Was 1700211224576 with diff -15
 LOG  Date.now() 1700211227679 getServerNow() 1700211227695 serverNow 1700211227130
 LOG  Date.now() 1700211228196 getServerNow() 1700211228212 serverNow 1700211227130
 LOG  Date.now() 1700211228730 getServerNow() 1700211228746 serverNow 1700211227130
 LOG  Date.now() 1700211229246 getServerNow() 1700211229262 serverNow 1700211227130
 LOG  TimeContext: Updating serverNow. Was 1700211227130 with diff -16
 LOG  Date.now() 1700211229763 getServerNow() 1700211229779 serverNow 1700211229702
 LOG  Date.now() 1700211230263 getServerNow() 1700211230279 serverNow 1700211229702
 LOG  Date.now() 1700211230780 getServerNow() 1700211230796 serverNow 1700211229702
 LOG  Date.now() 1700211231282 getServerNow() 1700211231298 serverNow 1700211229702
 LOG  Date.now() 1700211231796 getServerNow() 1700211231812 serverNow 1700211229702
 LOG  TimeContext: Updating serverNow. Was 1700211229702 with diff -16
```

### DIFF
--- a/src/time/TimeContext.tsx
+++ b/src/time/TimeContext.tsx
@@ -1,12 +1,6 @@
 import {useInterval} from '@atb/utils/use-interval';
 import {clock, start} from '@entur-private/abt-time-react-native-lib';
-import React, {
-  createContext,
-  useCallback,
-  useContext,
-  useEffect,
-  useState,
-} from 'react';
+import React, {createContext, useContext, useEffect, useState} from 'react';
 
 type TimeContextState = {
   /**
@@ -15,25 +9,18 @@ type TimeContextState = {
    * validity of fare contracts.
    */
   serverNow: number;
-  /**
-   * Returns server time in milliseconds, based on the difference between
-   * the previous serverNow and device time.
-   */
-  getServerNow: () => number;
 };
 
 const TimeContext = createContext<TimeContextState | undefined>(undefined);
 
+/**
+ * The number of milliseconds the device time is ahead of server time.
+ */
+export let serverDiff = 0;
+
 export const TimeContextProvider: React.FC = ({children}) => {
   const [clockIsRunning, setClockIsRunning] = useState(false);
   const [serverNow, setServerNow] = useState(Date.now());
-
-  // The number of milliseconds the local clock is ahead of the server clock.
-  const [clientServerDiff, setClientServerDiff] = useState<number>();
-
-  const getServerNow = useCallback(() => {
-    return Date.now() - (clientServerDiff || 0);
-  }, [clientServerDiff]);
 
   useEffect(() => {
     start({
@@ -46,7 +33,7 @@ export const TimeContextProvider: React.FC = ({children}) => {
   useInterval(
     () =>
       clock.currentTimeMillis().then((ms) => {
-        setClientServerDiff(Date.now() - ms);
+        serverDiff = Date.now() - ms;
         setServerNow(ms);
       }),
     2500,
@@ -59,7 +46,6 @@ export const TimeContextProvider: React.FC = ({children}) => {
     <TimeContext.Provider
       value={{
         serverNow,
-        getServerNow,
       }}
     >
       {children}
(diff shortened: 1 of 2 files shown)


gorandalum (changes requested):
I think this may cause some subtle bugs. Let's say you want to do a trip search with the server now every time the from or to location changes. The effect would look something like this:

```tsx
useEffect(() => {
  doTripSearch(from, to, getServerNow();
}, [from, to, getServerNow]) // needs getServerNow in dependency list
```

However, this will trigger a new trip search also when there is a new server time diff value, since the `getServerNow` reference changes. 

One solution here is to use a ref for the diff, so it doesn't need to be in the dependency list of the `useCallback` creating the `getServerNow`.

But one solution I just thought of is whether it is possible to put the diff outside of the `TimeContext`? Then it would be like this:
```tsx
// The number of milliseconds the local clock is ahead of the server clock.
let serverDiff = 0;

const serverNow = Date.now() - serverDiff;

const TimeContext = createContext<TimeContextState | undefined>(undefined);
	@@ -17,6 +28,13 @@ export const TimeContextProvider: React.FC = ({children}) => {
  ...
  useInterval(
    () =>
      clock.currentTimeMillis().then((ms) => {
        serverDiff = Date.now() - ms;
        setServerNow(ms);
      }),
    2500,
    [],
    !clockIsRunning,
    true,
  );
  ...
}
```

This means we would not need to rely on the `useTimeContext` to get the `serverNow` function, which will make the code quite a bit cleaner. Since you don't need to send the `serverNow` function down to utility functions, since it can be accessed by them directly.

rosvik:
> I think this may cause some subtle bugs.

Good points, I agree that's probably not how this should work.

> But one solution I just thought of is whether it is possible to put the diff outside of the TimeContext? 

I feel like this is a quite unconventional way of doing it (at least for the app), but now that I tried to implement it, it seems to work well, and should be easy to use in places like util functions.

Does the code look like what you had in mind now?

gorandalum:
> I feel like this is a quite unconventional way of doing it (at least for the app)

Might be unconventional, but I think it is ok practice. The `TimeContext` is like a global singleton, so shouldn't be problem using a variable like this. It isn't possible in hooks since its state isn't shared between usages.

> Does the code look like what you had in mind now?

Yeah, except I think it is better to export a function instead of the `serverDiff` value, like this: `const serverNow = Date.now() - serverDiff`.

Maybe you also can add usage of it to some of the utility files where we needed to send in `now`?

rosvik:
@gorandalum Ok, updated the export now. Not putting it to use just yet, since the TimeContext `serverNow` value works better for fare contracts.

Regarding performance: Seems like the first time we call `clock.currentTimeMillis` can be a bit slow, sometimes a 100ms while i tested it. Subsequent calls are very fast, on average much less than 1ms.

gorandalum (approved):

PR #4049, opened 2023-11-17
feat: update shared mobility bottom sheets

reidzeibel:
Closes https://github.com/AtB-AS/kundevendt/issues/15241

This PR will make the necessary changes to the bottom sheet as shown in the referenced issue. ([Figma link here](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=20438-8092&mode=design&t=hevY3atG9XQhaFoQ-4))

Reviewers should check the `CarSharingBottomSheet.tsx` first as it has the most changes and might have some issues in the logic. 

Depends on https://github.com/AtB-AS/atb-bff/pull/314 for ID retrieval

No need to hold back, fire at will! 😀


src/components/walking-distance/WalkingDistance.tsx
```diff
@@ -4,33 +4,29 @@ import {ThemeIcon} from '@atb/components/theme-icon';
 import {Walk} from '@atb/assets/svg/mono-icons/transportation';
 import {ThemeText} from '@atb/components/text';
 import {useHumanizeDistance} from '@atb/utils/location';
-import {StyleSheet, useTheme} from '@atb/theme';
+import {useTheme} from '@atb/theme';
 
 type Props = {
   distance: number | undefined;
   style?: StyleProp<ViewStyle>;
+  iconStyle?: StyleProp<ViewStyle>;
```
gorandalum, line 12: Do we really need this?  Maybe ask the designers. And if we really need the difference, I don't think we want reusable components to expose fields for internal styling. It would then be better with something like `gapSize: 'small' | 'large'`. But all in all I think this should be avoidable.

src/components/walking-distance/WalkingDistance.tsx
```diff
@@ -4,33 +4,29 @@ import {ThemeIcon} from '@atb/components/theme-icon';
...
+export const WalkingDistance = ({style, iconStyle, distance}: Props) => {
   const humanizedDistance = useHumanizeDistance(distance);
   const {theme} = useTheme();
 
   if (!humanizedDistance) return null;
 
   return (
-    <View style={[style, sheetStyles.distanceLabel]}>
-      <ThemeIcon svg={Walk} fill={theme.text.colors.secondary} />
+    <View style={[style, {flexDirection: 'row', justifyContent: 'center'}]}>
```
gorandalum, line 21: Should not inline this css, should use the `StyleSheet.createThemeHook`.

src/mobility/components/BicycleSheet.tsx
```diff
@@ -71,7 +75,7 @@ export const BicycleSheet = ({vehicleId: id, close}: Props) => {
         )}
         {!isLoading && !isError && vehicle && (
           <>
-            <ScrollView style={style.container}>
+            <View style={style.container}>
```
gorandalum, line 78: Needs to be `ScrollView`. Test it with larger text sizes to see why.

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -74,11 +78,7 @@ export const BikeStationSheet = ({stationId, distance, close}: Props) => {
         )}
         {!isLoading && !isError && station && (
           <>
-            <WalkingDistance
-              style={style.walkingDistance}
-              distance={distance}
-            />
-            <ScrollView style={style.container}>
+            <View style={style.container}>
```
gorandalum, line 81: Needs `ScrollView`

src/mobility/components/ScooterSheet.tsx
```diff
@@ -82,7 +85,7 @@ export const ScooterSheet = ({
         )}
         {!isLoading && !isError && vehicle && (
           <>
-            <ScrollView style={style.container}>
+            <View style={style.container}>
```
gorandalum, line 88: Needs `ScrollView`

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -77,11 +81,7 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
         )}
         {!isLoading && !isError && station && (
           <>
-            <ScrollView style={style.container}>
-              <WalkingDistance
-                style={style.walkingDistance}
-                distance={distance}
-              />
+            <View style={style.container}>
```
gorandalum, line 84: Needs `ScrollView`

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -93,55 +93,79 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+                      <ThemeText type="body__secondary" color="secondary">
+                        {t(CarSharingTexts.stations.carsAvailableLabel)}
+                      </ThemeText>
+                    </View>
+                    <View style={style.carDetailsContainer}>
+                      {station.vehicleTypesAvailable
+                        ?.slice(station.capacity > 2 ? -1 : -2)
+                        .map((vehicle, i) => (
+                          <View
+                            key={i}
```
gorandalum, line 137: Shouldn't use index as key in React elements.
https://react.dev/learn/rendering-lists#why-does-react-need-keys

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -93,55 +93,79 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+                          )}
+                        </ThemeText>
+                      </View>
+                      <ThemeText type="body__secondary" color="secondary">
+                        {t(CarSharingTexts.stations.carsAvailableLabel)}
+                      </ThemeText>
+                    </View>
+                    <View style={style.carDetailsContainer}>
+                      {station.vehicleTypesAvailable
+                        ?.slice(station.capacity > 2 ? -1 : -2)
```
gorandalum, line 134: I don't actually get this part. Maybe move it out into a utility function, where you can add jsdoc regarding what is actually happening 😅

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -93,55 +93,79 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+                      {station.capacity > 2 && (
+                        <View
+                          style={{
+                            height: 40,
+                            width: 40,
+                            borderRadius: 8,
+                            backgroundColor: '#d8d8d8',
+                            alignItems: 'center',
+                            justifyContent: 'center',
+                          }}
```
gorandalum, line 157: Shouldn't inline this.

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -93,55 +93,79 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+                        >
+                          <ThemeText style={{justifyContent: 'center'}}>
+                            +{station.capacity - 1}
```
gorandalum, line 160: Mismatch in logic between Figma and app. In Figma the plus number is only the remaining available, but the plus number in the app is the remaining capacity. Not sure which is correct, but it should match at least.
<div>


</div>

gorandalum (changes requested):
Good work! Some comments here and there, but all in all it looks good, and no major changes necessary :)

src/components/walking-distance/WalkingDistance.tsx
reidzeibel: Reverted the change and made a new component for mobility instead

src/components/walking-distance/WalkingDistance.tsx
reidzeibel: Reverted the change and made a new component for mobility instead

src/mobility/components/BicycleSheet.tsx
reidzeibel: reverted to ScrollView

src/mobility/components/BikeStationBottomSheet.tsx
reidzeibel: reverted to ScrollView

src/mobility/components/ScooterSheet.tsx
reidzeibel: reverted to ScrollView

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: reverted to ScrollView

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: updated using vehicle ID

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: Figma updated with proper documentation, the plus should show the total cars, even if there is no available cars.

https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=20471-12079&mode=design&t=QlA6jv5flbjP5Iyt-4

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: Replaced it with a component.

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: refactored and added documentation

src/mobility/components/CarImage.tsx
```diff
@@ -1,35 +1,63 @@
...
 };
-export const CarImage = ({uri}: Props) => {
+
+export const CarImage = ({uri, plus}: Props) => {
+  const style = useSheetStyle();
+
   return uri ? (
     <Image
-      style={{height: 75, width: 100, resizeMode: 'contain'}}
+      style={[style.container as ImageStyle, style.image as ImageStyle]}
```
KristianSelnas, line 18: I haven't tried, but I suppose this could be solved by creating a separate image style with a type to `createThemeHook`? I.e. something like this: `const useImageStyles = StyleSheet.createThemeHook<ImageStyle>(...)`

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -57,6 +58,28 @@ export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
...
+   * 
+   *  Case can be accessed in Figma link below 
+   *  https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?node-id=20471-12079
+   * 
+   *  Case on version 1.44 : 
+   *  - If the station capacity is 2 or less : show car image(s)
+   *  - If the station capacity is 3 or more : show 1 car image and the plus text
+   */
+
+  function carPreview(
```
KristianSelnas, line 77: A somewhat confusing name for this function. The comment explains it, but rather trivial code as this should be made self explanatory. Could it be a solution to extract the entire car preview stuff into a separate component? `<CarPreview typesAvailable={vehicleTypesAvailable} />`

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -33,8 +32,10 @@ type Props = {
 };
 
 export const CarSharingStationSheet = ({stationId, distance, close}: Props) => {
```
marius-at-atb, line 34: Not a part of your changes, but to keep the name in sync with the file name, this component should be called `CarSharingStationBottomSheet`

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -37,14 +36,18 @@ export const OperatorBenefit = ({
...
               {!isUserEligible && benefit.imageWhenNotActive && (
-                <SvgXml xml={benefit.imageWhenNotActive} />
+                <Image
+                  style={style.benefitImage as ImageStyle}
+                  source={{uri: benefit.imageWhenNotActive}}
+                />
               )}
-            </View>
             <View style={style.benefitContent}>
               <ThemeText type="body__primary--bold">{heading}</ThemeText>
```
marius-at-atb, line 52:
```suggestion
              {heading && (
                <ThemeText type="body__primary--bold">{heading}</ThemeText>
              )}
```

This change will fix the issue where ThemeText takes some room even when blank, causing unwanted whitespace:

| Before | After |
|--------|-------|
|  |  |

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -29,7 +34,6 @@ type Props = {
 
 export const BikeStationSheet = ({stationId, distance, close}: Props) => {
```
marius-at-atb, line 35: Same as the other one - not a part of your changes, but would be nice to update the component name to `BikeStationBottomSheet` to match the file name

src/mobility/components/CarImage.tsx
```diff
@@ -1,35 +1,63 @@
...
+const useSheetStyle = StyleSheet.createThemeHook(() => {
+  return {
+    container: {
+      height: 40,
+      width: 40,
+      borderRadius: 8,
+      backgroundColor: '#d8d8d8',
+    },
+    image: {
+      resizeMode: 'cover',
```
marius-at-atb, line 56: In the Figma sketches, 'contain' is used. However as far as I can see, this only applies to Otto images which are transparent, while Hyre images are not transparent and look better with 'cover'. Potentially worth checking whether the image is transparent to determine which one to use. Thoughts on this @oyvindmikkelsenatb @KristianSelnas? (see car image in the bottom sheet)


| Cover | Contain |
|--------|-------|
|  |  |
|  |  |

KristianSelnas: I agree that cover looks better for non-transparent, rectangular images.  I'm not sure how much work detecting opacity and handling that is, but if it's non-trivial I vote for 'cover'

src/mobility/components/BicycleSheet.tsx
```diff
@@ -86,35 +90,48 @@ export const BicycleSheet = ({vehicleId: id, close}: Props) => {
...
-                    />
-                  ) : (
-                    <VehicleStat
-                      svg={Bicycle}
-                      primaryStat=""
-                      secondaryStat={t(BicycleTexts.humanPoweredBike)}
+                <GenericSectionItem>
+                  <View style={style.content}>
+                    <MobilityStats
+                      top={
```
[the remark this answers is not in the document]
reidzeibel: What do you think if we made it into an array instead? should be more flexible in case we need to add more stats in the future.

src/mobility/components/CarImage.tsx
reidzeibel: I tried but this doesn't work

```
const useImageStyle = StyleSheet.createThemeHook<ImageStyle>(() => {
  return {
    image: {
      resizeMode: 'cover',
    } as ImageStyle,
    defaultImage: {
      justifyContent: 'center',
      alignItems: 'center',
    } as ImageStyle,
  };
});
```

Here's the error
```
Argument of type '() => { image: ImageStyle; defaultImage: ImageStyle; }' is not assignable to parameter of type 'ThemedStyles<NamedStyles<ImageStyle>>'.
  Type '{ image: ImageStyle; defaultImage: ImageStyle; }' has no properties in common with type 'NamedStyles<ImageStyle>'.ts(2345)
```

KristianSelnas: Ok, I see. It's not a big deal. 

I know you haven't worked with Typescript before and it's important to note that adding  `as ...` is not a type cast in a traditional sense, but rather a way of forcing Typescript into accepting whatever as your desired type. You circumvent the type system in a way and as such `as ...` should be used as sparingly as possible.

reidzeibel: Thanks for the explanation, I'll try to look for other solution so we can limit the usage of `as ...`

src/mobility/components/BicycleSheet.tsx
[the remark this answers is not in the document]
KristianSelnas: That would make it "future proof" but possibly harder to read as it is. It could make you believe that sending an array with three elements would magically work. My vote is to leave it as it is or rename the props to something more semantic. No big deal either way.

src/mobility/components/CarImage.tsx
gorandalum: No checking of specific operators allowed in the app code at least 🚫😅 If we want anything like that, it needs to be in the modelling in Firestore. I think just selecting a `resizeMode` is better.

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -12,7 +12,7 @@ export const MobilityTexts = {
       case FormFactor.Scooter:
         return _('Sparkesykkel', 'Electric scooter', 'Elektrisk sparkesykkel');
       case FormFactor.Bicycle:
-        return _('Sykkel', 'Bicycle', 'Sykkel');
+        return _('Bysykkel', 'Bicycle', 'Bysykkel');
```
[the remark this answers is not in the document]
reidzeibel: True, should I use `Sykkel` instead? since the Figma uses `Bysykkel` on the example 
image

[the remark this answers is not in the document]
KristianSelnas: Yes, I think we should use 'Sykkel' as it's more generic

src/mobility/components/BicycleSheet.tsx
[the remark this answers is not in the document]
reidzeibel: True, I agree that it will make it harder to read, and I'll just rename the variable instead for now.

src/translations/screens/subscreens/MobilityTexts.ts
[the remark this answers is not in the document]
reidzeibel: Updated

src/mobility/components/BicycleSheet.tsx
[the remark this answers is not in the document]
reidzeibel: Updated

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: Updated, moved the function to a CarPreviews component on file CarPreviews.tsx

src/mobility/components/CarImage.tsx
```diff
@@ -1,35 +1,63 @@
...
+  const style = useSheetStyle();
+
   return uri ? (
     <Image
-      style={{height: 75, width: 100, resizeMode: 'contain'}}
+      style={[style.container as ImageStyle, style.image as ImageStyle]}
       source={{uri}}
     />
+  ) : plus ? (
+    <PlusContainer plus={plus} />
```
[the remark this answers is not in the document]
reidzeibel: updated, check CarPreviews.tsx

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: Updated

src/mobility/components/BikeStationBottomSheet.tsx
reidzeibel: updated

src/mobility/components/CarImage.tsx
reidzeibel: I think having a `cover` is acceptable for now, at least if the user wants to check a more detailed image, they can just open the operator App instead of looking at the preview. It's just there as an indicator. At least by using `cover`, the one shown in Otto operator is better than using `contain` on Hyre operator.

src/mobility/components/OperatorBenefit.tsx
reidzeibel: updated

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -37,14 +36,18 @@ export const OperatorBenefit = ({
       <Section>
         <GenericSectionItem>
           <View style={style.benefitContainer}>
-            <View style={style.benefitImage}>
               {isUserEligible && benefit.imageWhenActive && (
-                <SvgXml xml={benefit.imageWhenActive} />
+                <Image
```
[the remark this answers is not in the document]
reidzeibel: As discussed on Slack thread: Martin suggested that we shouldn't put the image (even as string) to firestore, as it is not the right way to store images, instead we should use URL on the firestore and put the assets somewhere else.

After Martin commented,  @gorandalum suggested that we should hold a talk with the designer to find the best solution, because right now, having operator png image on the indicator for active promotion is not scalable, might discuss on how to find the best way to implement this part.

src/mobility/components/CarSharingStationBottomSheet.tsx
```diff
@@ -220,36 +215,30 @@ const useSheetStyle = StyleSheet.createThemeHook((theme) => {
...
     },
   };
 });
 
-const isAnyAvailable = (
+const totalAvailableCars = (
   vehicleTypesAvailable: CarAvailabilityFragment[] | undefined,
-) => {
-  const count =
+): number => {
```
KristianSelnas, line 238: Just a tip since you are pretty new to Javascript, @reidzeibel. One liner functions like this can have implicit return. If you omit the brackets {} the function will return the value of your statement. This is equivalent to what you have:
```
const totalAvailableCars = (
  vehicleTypesAvailable: CarAvailabilityFragment[] | undefined,
) => vehicleTypesAvailable
      ?.map((v) => v.count)
      .reduce((sum, count) => sum + count, 0) ?? 0
```
You can also omit the explicit return type (`:number`). A good IDE (with the proper plugins) will deduce this and show it for you. 

image

src/mobility/components/MobilityDistance.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {Walk} from '@atb/assets/svg/mono-icons/transportation';
+import {ThemeText} from '@atb/components/text';
+import {useHumanizeDistance} from '@atb/utils/location';
+import {StyleSheet, useTheme} from '@atb/theme';
+
+type Props = {
+  distance: number | undefined;
+  style?: StyleProp<ViewStyle>;
+};
+export const MobilityDistance = ({style, distance}: Props) => {
```
KristianSelnas, line 13: Why do we need this as specific component for mobility? We should try to minimize duplication. The only differences I can see is the margins and paddings. Could this be achieved by adding som options to the standard `<WalkingDistance />` component instead?

src/mobility/components/MobilityIconText.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {View} from 'react-native';
+import {ThemeIcon} from '@atb/components/theme-icon';
+import {ThemeText} from '@atb/components/text';
+import {useTheme} from '@atb/theme';
+
+type Props = {
+  svg?(props: SvgProps): JSX.Element;
+  text: string;
+};
+export const MobilityIconText = ({svg, text}: Props) => {
```
KristianSelnas, line 12: If we want this as a separate component I believe it could be a local component inside `MobilityStat`, since that is the only place it's used. I also find the name a bit confusing. Is something like `MobilityStatWithIcon` more describing?

src/mobility/components/MobilityStats.tsx
```diff
@@ -0,0 +1,32 @@
+import {View} from 'react-native';
+import React from 'react';
+import {StyleSheet} from '@atb/theme';
+
+type MobilityStatsProps = {
+  first: JSX.Element;
```
KristianSelnas, line 6: Rather than passing a JSX.Element here you could pass an object and rather use the <MobilityStat> component here.
```
type MobilityStat = { 
  svg?(props: SvgProps): JSX.Element;
  primaryStat: string | number;
  secondaryStat?: string | number;
  secondaryStatStyle?: TextStyle;
};

type MobilityStatsProps = {
    first: MobilityStat, <- or use props type exported from MobiltyStat
    second: MobilityStat,
```

src/mobility/components/BicycleSheet.tsx
```diff
@@ -86,35 +90,48 @@ export const BicycleSheet = ({vehicleId: id, close}: Props) => {
...
-                right={
-                  <PricingPlan
-                    operator={operatorName}
-                    plan={vehicle.pricingPlan}
-                  />
-                }
-              />
+                    {brandLogoUrl ? (
+                      <OperatorLogo
+                        operatorName={operatorName}
```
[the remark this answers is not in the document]
reidzeibel: There are 2 places where the `<OperatorLogo>` is used based on Figma : 

image

for example like this 
image

Based on your input, I will update the `OperatorLogo` with fallbacks and adding a constant maxHeight/maxWidth for logo, and add a parameter whether it is a small logo or large logo (boolean), also added some documentation on the OperatorLogo.

src/mobility/components/CarImage.tsx
```diff
@@ -1,15 +1,19 @@
...
+import {Image, ImageStyle, View} from 'react-native';
 import React from 'react';
 import {ThemeIcon} from '@atb/components/theme-icon';
-import {Car} from '@atb/assets/svg/mono-icons/transportation';
+import {Car} from '@atb/assets/svg/mono-icons/transportation-entur';
+import {StyleSheet, useTheme} from '@atb/theme';
 
 type Props = {
-  uri: string | undefined;
+  uri?: string | undefined;
```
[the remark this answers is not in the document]
reidzeibel: oops, seems like I forgot to remove it after testing, thanks!

src/mobility/components/CarSharingStationBottomSheet.tsx
reidzeibel: Thanks for the insight!

src/mobility/components/MobilityDistance.tsx
reidzeibel: can I do something like this instead?

```
type Props = {
  distance: number | undefined;
  style?: StyleProp<ViewStyle>;
  isMobility: boolean;
};
export const WalkingDistance = ({style, distance, isMobility}: Props) => {
  const sheetStyles = useSheetStyle();
  const humanizedDistance = useHumanizeDistance(distance);
  const {theme} = useTheme();

  const labelStyle = isMobility ? sheetStyles.mobilityLabel : sheetStyles.distanceLabel;
  const iconStyle = isMobility ? sheetStyles.icon : {}

  if (!humanizedDistance) return null;

  return (
    <View style={[style, labelStyle]}>
      <ThemeIcon svg={Walk} fill={theme.text.colors.secondary} style={iconStyle}/>
      <ThemeText type="body__secondary" color="secondary">
        {humanizedDistance}
      </ThemeText>
    </View>
  );
};
```

src/mobility/components/MobilityStats.tsx
reidzeibel: I don't think this would be a good idea because sometimes we want to pass something else such as `<PricingPlan>`, for example in `ScooterSheet.tsx` shown below

```
<MobilityStats
  first={
    <MobilityStat
      svg={Battery}
      primaryStat={vehicle.currentFuelPercent + '%'}
      secondaryStat={formatRange(vehicle.currentRangeMeters, language)}
    />
  }
  second={
    <PricingPlan
      operator={operatorName}
      plan={vehicle.pricingPlan}
      eligibleBenefit={
        operatorBenefit && isUserEligibleForBenefit
          ? operatorBenefit.id
          : undefined
      }
    />
  }
/>
```

src/mobility/components/BicycleSheet.tsx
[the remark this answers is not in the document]
reidzeibel: Updated

src/mobility/components/MobilityIconText.tsx
reidzeibel: Refactored, thanks!

reidzeibel:
@KristianSelnas @gorandalum I just added the benefit image, tried to implement it like the one in `<ThemeIcon>`. Need you to review again, thanks!

KristianSelnas:
I'm not sure if `<ThemeIcon>` is the right tool here. Couldn't you just use the svg directly? Maybe @gorandalum has some input here.

reidzeibel:
> I'm not sure if `<ThemeIcon>` is the right tool here. Couldn't you just use the svg directly? Maybe @gorandalum has some input here.

I mean I use `<ThemeIcon>` and `<NotificationIndicator>` as reference to create a new component for the benefit images, but I didn't use them for the images because it is indeed not the right tool. The SVG is added from the design system and to add the check mark I'm adding a new local component based on `NotificationIndicator`.

KristianSelnas:
@reidzeibel I suggest you install ESLint and Prettier plugins to your IDE (VSCode?) to detect lint errors and automatically format code according to prettier config.

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -56,15 +59,70 @@ export const OperatorBenefit = ({
...
 };
 
+const BenefitImageAsset = ({formFactor} : {formFactor: FormFactor}) => {
+  switch (formFactor) {
+    case FormFactor.Car:
+      return <BundlingCarSharing />;
+    case FormFactor.Bicycle:
+      return <BundlingCityBike />;
+    default:
+      return <View />;
```
KristianSelnas, line 69: You could return `null` here instead.

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -56,15 +59,70 @@ export const OperatorBenefit = ({
...
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   benefitContainer: {
     flexDirection: 'row',
+    alignItems: 'center',
+  },
+  benefitContent: {
+    flex: 4,
+    marginStart: theme.spacings.medium,
   },
   benefitImage: {
```
KristianSelnas, line 115: Not in use

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -37,16 +45,11 @@ export const OperatorBenefit = ({
...
-            <View style={style.benefitImage}>
-              {isUserEligible && benefit.imageWhenActive && (
-                <SvgXml xml={benefit.imageWhenActive} />
-              )}
-              {!isUserEligible && benefit.imageWhenNotActive && (
-                <SvgXml xml={benefit.imageWhenNotActive} />
-              )}
-            </View>
+            <BenefitImage eligible={isUserEligible} formFactor={formFactor} />
             <View style={style.benefitContent}>
```
KristianSelnas, line 49: The `marginStart` on `benefitContainer` will create an unwanted margin to the left of the text if `<BenefitImage>` is null.

KristianSelnas (commented):
Lots of good work on this PR. My comments are mostly nitpicking to get you into the coding style and best practices.

src/mobility/components/OperatorNameAndLogo.tsx
```diff
@@ -0,0 +1,87 @@
...
+  logoSize?: "small" | "large";
+  fallback?: JSX.Element
+};
+
+export const OperatorNameAndLogo = ({
+  operatorName,
+  logoUrl,
+  logoSize = "small",
+  fallback
+}: OperatorNameAndLogoProps) => {
```
gorandalum, line 40: I don't think this should be a single component, since it has such a wide variety of how it can end up rendered (can show nothing, just name, small image and name, large image and no name).

gorandalum (changes requested):

src/components/walking-distance/WalkingDistance.tsx
```diff
@@ -9,17 +9,22 @@ import {StyleSheet, useTheme} from '@atb/theme';
 type Props = {
   distance: number | undefined;
   style?: StyleProp<ViewStyle>;
+  isMobility: boolean;
```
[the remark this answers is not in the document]
KristianSelnas: @gorandalum  https://github.com/AtB-AS/mittatb-app/pull/4049#discussion_r1405754039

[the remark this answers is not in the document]
gorandalum: Yeah, I agree, shouldn't be necessary.

src/components/walking-distance/WalkingDistance.tsx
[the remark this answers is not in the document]
reidzeibel: Updated to remove isMobility prop

src/mobility/components/OperatorBenefit.tsx
```diff
@@ -56,15 +59,70 @@ export const OperatorBenefit = ({
...
     marginRight: theme.spacings.medium,
-    justifyContent: 'center',
   },
-  benefitContent: {
-    flex: 1,
+  indicator: {
+    position: 'absolute',
+    right: 0,
+    top: 0,
+    borderRadius: theme.border.radius.circle,
```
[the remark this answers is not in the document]
reidzeibel: You are right, I only need the `right` properties because otherwise it will show like this 
image

If I added the `right` properties it will show correctly as shown in Figma :
image

I will remove the unnecessary properties

src/mobility/components/OperatorBenefit.tsx
reidzeibel: moved the margin to the image so if it's null there will be no margins.

src/mobility/components/OperatorBenefit.tsx
reidzeibel: added null

src/components/walking-distance/WalkingDistance.tsx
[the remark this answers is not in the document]
reidzeibel: attribute removed instead, thanks for pointing it out

src/mobility/components/MobilityStat.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const StatWithIcon = ({svg, text}: StatWithIconProps) => {
+  const {theme} = useTheme();
+  return (
+    <View style={{flexDirection: 'row'}}>
+      {svg && (
+        <ThemeIcon
+          style={
+            (text ?? '').length > 0
```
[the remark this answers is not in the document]
reidzeibel: I removed the `text ?? ''` but kept the style inside the icon, since if the icon is missing, the text will look a bit out of place due to the margin.

src/mobility/components/OperatorBenefit.tsx
reidzeibel: it is now used to add margin to the image

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -162,21 +183,28 @@ const useSheetStyle = StyleSheet.createThemeHook((theme) => {
...
     errorMessage: {
       marginHorizontal: theme.spacings.medium,
     },
     footer: {
       marginBottom: Math.max(bottom, theme.spacings.medium),
       marginHorizontal: theme.spacings.medium,
     },
-    walkingDistance: {
-      marginBottom: theme.spacings.medium,
+    walkingDistanceIcon: {
```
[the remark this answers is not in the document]
reidzeibel: updated

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -74,45 +78,62 @@ export const BikeStationSheet = ({stationId, distance, close}: Props) => {
...
                   benefit={operatorBenefit}
                   isUserEligible={isUserEligibleForBenefit}
+                  formFactor={FormFactor.Bicycle}
+                  style={style.operatorBenefit}
                 />
               )}
               <Section>
                 <GenericSectionItem>
-                  <OperatorLogo
+                  <OperatorNameAndLogo
```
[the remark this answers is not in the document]
reidzeibel: based on the figma here https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=20467-42962&mode=design&t=shZPKo8fvyhZROQD-4, the top part with operator name doesn't have fallback, I will instead split the component into 2 : 
- title part (logo and name) 
- body part (logo and fallback)

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -162,21 +183,28 @@ const useSheetStyle = StyleSheet.createThemeHook((theme) => {
...
+      marginBottom: theme.spacings.medium,
     },
     stationName: {
       flex: 1,
       alignItems: 'center',
     },
+    stationText: {
+      display: 'flex',
+      flexDirection: 'row',
+      marginTop: theme.spacings.xSmall,
```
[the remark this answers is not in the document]
reidzeibel: removed xsmall value from relevant files in this PR

src/mobility/components/BikeStationBottomSheet.tsx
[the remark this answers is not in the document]
reidzeibel: moved the large logo & fallback to BrandingImage.tsx

src/mobility/components/CarPreviews.tsx
```diff
@@ -0,0 +1,113 @@
...
+ *
+ *  Case on version 1.44 :
+ *  - If the station capacity is 2 or less : show car image(s)
+ *  - If the station capacity is 3 or more : show 1 car image and the plus text
+ *
+ *  @param stationCapacity: Station capacity
+ *  @param vehicleTypesAvailable: Types of available cars, array of CarAvailabilityFragment
+ *  @returns array of vehicle types to show
+ */
+function createPreviewArray(
```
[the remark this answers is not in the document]
reidzeibel: Moved to correct place

src/mobility/components/OperatorNameAndLogo.tsx
reidzeibel: Split into BrandingImage.tsx

gorandalum (approved):
Good work 😊

PR #4050, opened 2023-11-17
fix: notification onboarding flow and register for notifications

adriawh:
The notification onboarding will now show:
- After buying a ticket
- When logging in and having an active ticket
- Opening the app after updating and having an active ticket

Screen now also uses the `register` from `usePushNotifications()`

[Figma](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=18256-2778&mode=dev)

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
@@ -70,27 +70,27 @@ export const Root_TabNavigatorStack = ({navigation}: Props) => {
   const {status: notificationStatus} = usePushNotifications();
 
   useEffect(() => {
+    const shouldShowLocationOnboarding =
+      !locationWhenInUsePermissionOnboarded &&
+      locationWhenInUsePermissionStatus === 'denied';
+
     if (!onboarded) {
       InteractionManager.runAfterInteractions(() =>
         navigation.navigate('Root_OnboardingStack'),
       );
     } else {
-      if (
-        !locationWhenInUsePermissionOnboarded &&
-        locationWhenInUsePermissionStatus === 'denied'
-      ) {
+      if (shouldShowLocationOnboarding) {
         InteractionManager.runAfterInteractions(() =>
           navigation.navigate('Root_LocationWhenInUsePermissionScreen'),
         );
       }
     }
-
     if (
       !notificationPermissionOnboarded &&
       pushNotificationsEnabled &&
       validFareContracts.length > 0 &&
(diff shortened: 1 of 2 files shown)


gorandalum:
Could you add something about the background for this PR? It doesn't have reference to an issue, Figma, Slack, or something.

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -68,13 +84,27 @@ export const Root_TabNavigatorStack = ({navigation}: Props) => {
...
+      !notificationPermissionOnboarded &&
+      pushNotificationsEnabled &&
+      validFareContracts.length > 0 &&
+      notificationStatus !== 'granted' &&
+      locationWhenInUsePermissionOnboarded
+    ) {
+      InteractionManager.runAfterInteractions(() =>
+        navigation.navigate('Root_NotificationPermissionScreen'),
+      );
+    }
```
gorandalum, line 98: Would something like this be a little clearer?
```tsx
const shouldOnboardNotifications =
  !notificationPermissionOnboarded &&
  pushNotificationsEnabled &&
  validFareContracts.length > 0 &&
  notificationStatus !== 'granted' &&
  locationWhenInUsePermissionOnboarded;

useEffect(() => {
  if (shouldOnboardNotifications) {
    InteractionManager.runAfterInteractions(() =>
      navigation.navigate('Root_NotificationPermissionScreen'),
    );
  }  
}, [shouldOnboardNotifications]);
```

gorandalum (approved):
LGTM!

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
adriawh: Yes that would be clearer, will make the change

adriawh: With the latest commit I belive it should stay like it is. @marius-at-atb can take a last look maybe?

marius-at-atb (approved):
Looks good!

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
marius-at-atb: Thinking about it, I guess it could be a little clearer when splitting up like @gorandalum suggests. However just fine as is as well :+1:

PR #4051, opened 2023-11-17
chore: Delete useFinishOnboarding

marius-at-atb:
Since the location permission is now requested in its own screen, it should no longer be requested directly after onboarding is finished.

Part of https://github.com/AtB-AS/kundevendt/issues/8622


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/Dashboard_RootScreen.tsx
```diff
@@ -128,19 +128,16 @@ export const Dashboard_RootScreen: React.FC<RootProps> = ({
...
-        const status = await requestLocationPermission();
-        if (status === 'granted') {
-          setCurrentLocationAsFrom();
-        }
       }
-    },
-    [currentLocation, setCurrentLocationAsFrom, requestLocationPermission],
-  );
+    }
+  }, [currentLocation, setCurrentLocationAsFrom, requestLocationPermission]);
```
marius-at-atb, line 140: Note: the only change here is `() => {` instead of `function setCurrentLocationOrRequest() {` since the name was repeated twice

gorandalum (approved):

PR #4054, opened 2023-11-17
chore: Bump to version 1.44.1

toretefre:
Would you like this to be in the existing `release/1.44` or to have a separate `release/1.44.1` branch?


gorandalum:
It can just be in the 1.44 branch

gorandalum (approved):

PR #4056, opened 2023-11-17
feat: support favorite migration also on departures response

marius-at-atb:
Migrate favorites using data received from the `/bff/v2/departures/departures` call

Depends on https://github.com/AtB-AS/atb-bff/pull/313

Resolves https://github.com/AtB-AS/kundevendt/issues/2732
(mentioned in this [comment](https://github.com/AtB-AS/kundevendt/issues/2732#issuecomment-1792286534))

### DIFF
--- a/src/place-screen/hooks/use-departures-data.ts
+++ b/src/place-screen/hooks/use-departures-data.ts
@@ -246,7 +246,8 @@ export function useDeparturesData(
   tickRateInSeconds: number = 10,
 ) {
   const [state, dispatch] = useReducerWithSideEffects(reducer, initialState);
-  const {favoriteDepartures} = useFavorites();
+  const {favoriteDepartures, potentiallyMigrateFavoriteDepartures} =
+    useFavorites();
   const [queryStartTime, setQueryStartTime] = useState<string | undefined>();
   const [timeRange, setTimeRange] = useState<number | undefined>();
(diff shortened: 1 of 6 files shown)


src/place-screen/hooks/use-departures-data.ts
```diff
@@ -337,6 +338,11 @@ export function useDeparturesData(
     !!state.tick,
   );
 
+  useEffect(() => {
+    const estimatedCalls = state.data;
+    estimatedCalls && potentiallyMigrateFavoriteDepartures(estimatedCalls);
+  }, [state.data, potentiallyMigrateFavoriteDepartures]);
```
gorandalum, line 344: Is `potentiallyMigrateFavoriteDepartures` a stable reference?

gorandalum (approved):
Looks good!

src/place-screen/hooks/use-departures-data.ts
marius-at-atb: Good question!
Yes, both code walkthrough and testing indicates that it is stable.

marius-at-atb: Aha 😅
I just looked at the current implementation and whether there was a way to trigger repeated potentiallyMigrateFavoriteDepartures changes from using the app. Will look into safeguarding it for the future as well :+1:

PR #4058, opened 2023-11-17
fix: add install_id to api calls from iOS widget

adriawh:
By adding install_id to the sent requests, we can see how many individual users have the widget active. Was solved this way to avoid adding potshot to the native widget code, and to reduce the amount of logging in posthog.

To find the amount users, check how many individual install_id's have used the `bff/v2/departure-favorites` with a limitPerLine of 50.


gorandalum (approved):
Good to avoid potshot for now 👍

PR #4065, opened 2023-11-23
feat: handle click actions on push notifications

rosvik:
Added a new hook "useOnPushNotificationOpened" that is called when the TabNavigator is rendered. This adds two listeners that reacts to app open events, one from background state and one from closed state. 

If `data.type` is "FARE_CONTRACT_EXPIRY", navigate to ActiveFareProductsTabScreen.

closes https://github.com/AtB-AS/kundevendt/issues/7229

### DIFF
--- a/src/notifications/types.ts
+++ b/src/notifications/types.ts
@@ -17,3 +17,12 @@ export type NotificationConfig = {
   modes: NotificationConfigMode[];
   groups: NotificationConfigGroup[];
 };
+
+export type NotificationPayload = {
+  type: NotificationPayloadType.fareContractExpiry;
+  fareContractId: string;
+};
+
+export enum NotificationPayloadType {
+  fareContractExpiry = 'FARE_CONTRACT_EXPIRY',
+}
--- a/src/notifications/types.ts
+++ b/src/notifications/types.ts
@@ -1,3 +1,5 @@
+import {z} from 'zod';
+
 export type NotificationConfigType = 'mode' | 'group';
 
 export interface NotificationConfigValue {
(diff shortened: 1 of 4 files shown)


src/notifications/use-on-push-notification-opened.ts
```diff
@@ -0,0 +1,55 @@
...
+  const {navigate} = useNavigation<RootNavigationProps>();
+
+  const onMessage = useCallback(
+    (message: FirebaseMessagingTypes.RemoteMessage) => {
+      const isKnownType = Object.values(NotificationPayloadType).includes(
+        message.data?.type as NotificationPayloadType,
+      );
+      if (!isKnownType) return;
+
+      const messageData = message.data as NotificationPayload;
```
KristianSelnas, line 21: How about defining `NotificationPayload` as a Zod object and use Zod to safe parse the data, rather than assuming it's as expected?

rosvik: Good point! Updated now.

KristianSelnas (approved):
LGTM 👍

src/notifications/types.ts
```diff
@@ -17,3 +19,14 @@ export type NotificationConfig = {
   modes: NotificationConfigMode[];
   groups: NotificationConfigGroup[];
 };
+
+export enum PushNotificationPayloadType {
+  fareContractExpiry = 'FARE_CONTRACT_EXPIRY',
+}
```
gorandalum, line 25: @KristianSelnas @rosvik 
I think it might be more scalable if the push notification payloads was phrased based on where they should be navigated in the app, and not on the cause of the navigation? Then we can add new push notifications which may reuse previously registered payloads without feeling "hackish".

PR #4067, opened 2023-11-23
feat: One time popover component

KristianSelnas:
A component for displaying tooltips/popovers once for a given element in the app.  Closes https://github.com/AtB-AS/kundevendt/issues/3854. See issue for screenshots.

Once a user has seen (dismissed) a popover the `oneTimeKey` (id) of that popover is saved to storage, and not displayed again.

The underlying popover library `react-native-popover-view` has three different [view modes](https://www.npmjs.com/package/react-native-popover-view#mode). Only `Popover.MODE.RN_MODAL` is reliably show in the correct location and correct stacking order every time. The other modes have proven challenging to get to work properly. RN_MODAL uses React Native Modal under the hood. The major disadvantage by this approach is that there can only be one modal visible at any given time. The implementation is thus a bit fiddly because popovers need to be added to a queue and shown one at the time.

This pr does not include any use of the popover. Example usage:
```
  const buttonRef = useRef(null);
  const {addPopOver} = usePopOver();

  useEffect(() => {
    if (buttonRef) {
      addPopOver({oneTimeKey: 'example-popover', target: buttonRef});
    }
  }, [buttonRef, addPopOver]);

----

  <Button ref={buttonRef}  .../>
```

New popovers must be added to the `PopOverKey` type and a corresponding text added to `translations/.../OneTimePopOver.ts`.


KristianSelnas:
I've forgotten to fix the accessibility stuff before sending this to review. Just ignore that bit for now.

rosvik (commented):
Looks good to me 👍 Let me know when you have added handling of screen readers.

rosvik (approved):

KristianSelnas:
@gorandalum, @Sebstorvik, and I agreed to postpone the accessibility question for now. If we find it necessary, screen reader support can be added in a separate PR later. The options we have discussed are:

1.  No screen reader support (hide popover from screen readers). Consider the popover as a non critical enhancement that provides som additional support to visual users, but does not prevent screen reader users from accessing functionality.
2. Investigate whether it is feasible to make screen readers focus on the popover and read out it's contents.
3.  Visually hide the popover, and add a screen reader announcement with a customised text, read when the page is loaded.
4. Add an accessibility text/announcement on the element that triggers the popover with a text similar to what the popover contains.

src/popover/PopOverContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const {isSeen, setPopOverSeen} = useOneTimePopover();
+  const {t} = useTranslation();
+  const isScreenReaderEnabled = useIsScreenReaderEnabled();
+  const current = popOvers[0];
+
+  const addPopOver = useCallback(
+    (popOver: PopOverType) => {
+      if (!isSeen(popOver.oneTimeKey)) {
+        // Setting a timeout before adding the popover to allow potential page transitions
+        // or similar animations to complete before the popover is displayed.
```
marius-at-atb, line 36: Should the timeout be removed or optional?
Ideally `addPopOver` will be called when everything is ready for it to be shown?

src/popover/PopOverContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+      }
+    },
+    [isSeen],
+  );
+
+  const onClose = useCallback(
+    async (oneTimeKey: PopOverKey) => {
+      await setPopOverSeen(oneTimeKey);
+      // Remove the first popover from the queue after the close animation has finished.
+      setTimeout(
```
marius-at-atb, line 49: Does it work to use the `onCloseComplete` callback from `RNPopover` instead of a timeout here?

src/popover/PopOverContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+  addPopOver: () => {},
+});
+
+export const PopOverContextProvider: React.FC = ({children}) => {
+  // Queue of popovers to display
+  const [popOvers, setPopOvers] = useState<PopOverType[]>([]);
+  const {isSeen, setPopOverSeen} = useOneTimePopover();
+  const {t} = useTranslation();
+  const isScreenReaderEnabled = useIsScreenReaderEnabled();
+  const current = popOvers[0];
```
marius-at-atb, line 30: Personally I prefer naming to be more specific, such as `currentPopOver`

src/popover/use-one-time-popover.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Fetches from storage the list of one time popovers already seen by the user
+   */
+  const getSeenPopOvers = () =>
+    storage
+      .get(StorageModelKeysEnum.OneTimePopOver)
+      .then((stored) => JSON.parse(stored ?? '[]') as string[]);
+
+  const isSeen = useCallback(
```
marius-at-atb, line 20: Here we se an example why I prefer specific naming. `isSeen` may be easy to understand here, but when used in other components, perhaps not. Also it sounds like a boolean instead of a function.

I would suggest returning `{getPopOverIsSeen, setPopOverIsSeen}` from this hook

gorandalum (approved):
:+1:

marius-at-atb (approved):
Very nice!

src/popover/PopOverContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const addPopOver = useCallback(
+    (popOver: PopOverType) => {
+      if (!isSeen(popOver.oneTimeKey)) {
+        // Setting a timeout before adding the popover to allow potential page transitions
+        // or similar animations to complete before the popover is displayed.
+        setTimeout(() => {
+          setPopOvers((val) => [...val, popOver]);
+        }, 500);
```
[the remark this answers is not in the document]
KristianSelnas: No, I haven't tried that, and I haven't even heard of it. Thanks for the tip! It sounds like a lot more correct solution. I'll test and see if it works as desired.

src/popover/PopOverContext.tsx
KristianSelnas: No, sadly it doen't

src/popover/use-one-time-popover.tsx
gorandalum: Chiming in here, with some pseudocode/pseudosituation 😅 

Generally I don't personally mind `const {isLoading} = useHelicopterState()`, since the helicopter part is implicit by the name of the used hook. I also think in a component like `HelicopterIcon` just using `const {isLoading} = useHelicopterState()` is enough, since that the `isLoading` variable is helicopter-related is understood from the surrounding context.

What we don't want is a large non-helicopter specific component that just uses the name `isLoading` throughout the code, and the context that it's the loading state for the helicopter is lost. In that situation a name like `isHelicopterLoading` is necessary. We can do that with `const {isLoading: isHelicopterLoading} = useHelicopterState()`, but then it is more up to the coder on a case by case basis to use a specific enough name. The option is of course (as suggested by Marius) to enforce it in the variable names so we always have `const {isHelicopterLoading} = useHelicopterState()`, which gives less room for not-explicit-enough named variables.

It might be nice to have a statement on this in the coding guideline, since we do this differently throughout the code. I don't really have a strong opinion on this personally, so we can just do what the team feels is better.

marius-at-atb: Good thinking, would be nice to have some coding guidelines on this. Even though `const {isLoading: isHelicopterLoading} = useHelicopterState()` is possible, I think in most cases it is cleaner to enforce it in the return statement.
I do however see the benefits of a very short and simple variable name when its scope is limited internally to a function/component

src/popover/PopOverContext.tsx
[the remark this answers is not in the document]
KristianSelnas: Works like a charm!

src/popover/PopOverContext.tsx
KristianSelnas: Solved with `InteractionManager.runAfterInteractions()`

src/popover/use-one-time-popover.tsx
KristianSelnas: A lot of this comes down to personal preference. Personally I'm leaning towards explicitly naming properties when/if necessary, as in `const {isLoading: isHelicopterLoading} = useHelicopterState()`. In this case, where the hook is only used locally in the popover feature, and there is very little room for collisions with other hooks or variables with similar name, I prefer the short version.

One argument for renaming it is that the returns from this hook are inconsistently named. They should probably be either `{setPopOverSeen, isPopOverSeen}` or `{setSeen, isSeen}`. In that case, I prefer the first option.

marius-at-atb: `{setPopOverSeen, isPopOverSeen}` sounds good :+1:

marius-at-atb: As you say, it ultimately comes down to just personal preference, so having a coding guideline for it would help with consistency as @gorandalum suggested

PR #4070, opened 2023-11-24
feat: update url for delete profile

hanne-at-atb:
After creating a [new profile service on the BE](https://github.com/AtB-AS/team-platform/issues/446), we now have [new Url's ](https://github.com/AtB-AS/team-platform/issues/446#issuecomment-1823970984)for requests to this service. We therefore should update the url for the DELETE request in the app. 
We will add GET and PATCH as part of work in https://github.com/AtB-AS/kundevendt/issues/4275, but opening a separate PR for changing the existing one (delete) in the app.


rosvik (approved):

gorandalum (approved):

tormoseng:
Tested OK

PR #4072, opened 2023-11-24
feat: Show which scooter/bicycle is selected in map

KristianSelnas:
Highlight the selected scooter, bike station etc. in the map.

Closes https://github.com/AtB-AS/kundevendt/issues/10821

<details>
<summary>Screenshots:</summary>

Simulator Screenshot - iPhone 15 - 2023-11-24 at 15 11 05

Simulator Screenshot - iPhone 15 - 2023-11-24 at 15 09 30

Simulator Screenshot - iPhone 15 - 2023-11-24 at 15 12 46

</details>

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -34,7 +34,7 @@ export const Map = (props: MapProps) => {
     [currentCoordinatesRef, initialLocation],
   );
 
-  const {mapLines, selectedCoordinates, onMapClick} =
+  const {mapLines, selectedCoordinates, selectedFeature, onMapClick} =
     useMapSelectionChangeEffect(
       props,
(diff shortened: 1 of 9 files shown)


gorandalum (commented):
I added one comment on the issue, since it isn't really connected to the implementation, but rather the designed solution.

KristianSelnas:
Closing this PR after deciding we should try to find a more general solution that enables us to highlight any selected item in the map, not just micro mobility stuff. 

Initial research has found that it's not trivial to achieve this with the data coming from Mapbox Studio in RNMapbox. Further investigation is required to determine whether it is possible at all, and whether the cost/benefit of making a general solution is acceptable.

PR #4038, opened 2023-11-15
feat: add notification register and token to debug menu

rosvik:
ref. https://mittatb.slack.com/archives/C0642RNB86L/p1700032726550399

IMG_1514FCF2797B-1

PR #4074, opened 2023-11-27
feat: register notifications on startup and language change

rosvik:

Renamed the `register` function to `requestPermissions`, and pulled the register endpoint -part into it's own function. This is useful when we want to register the app language without prompting the user, which now happends on app start and on language change.

### DIFF
--- a/src/notifications/use-push-notifications.tsx
+++ b/src/notifications/use-push-notifications.tsx
@@ -19,6 +19,7 @@ export const usePushNotifications = () => {
   const {language} = useLocaleContext();
   const [status, setStatus] = useState<NotificationsStatus>('loading');
   const {mutation: registerMutation} = useRegister();
+  const mutateRegister = registerMutation.mutate;
   const {query: configQuery, mutation: configMutation} = useConfig();
 
   const checkPermissions = useCallback(() => {
@@ -55,10 +56,9 @@ export const usePushNotifications = () => {
   const register = useCallback(async () => {
     const token = await messaging().getToken();
     if (!token) return;
-    registerMutation.mutate({token, language: language});
+    mutateRegister({token, language: language});
     return token;
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [language, registerMutation.mutate]);
+  }, [language, mutateRegister]);
 
   const requestPermissions = useCallback(async () => {
     setStatus('updating');
--- a/src/notifications/use-push-notifications.tsx
+++ b/src/notifications/use-push-notifications.tsx
@@ -53,23 +53,31 @@ export const usePushNotifications = () => {
   }, []);
 
   const register = useCallback(async () => {
+    const token = await messaging().getToken();
+    if (!token) return;
+    registerMutation.mutate({token, language: language});
+    return token;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [language, registerMutation.mutate]);
+
+  const requestPermissions = useCallback(async () => {
     setStatus('updating');
     const permissionStatus = await requestUserPermission();
-    const token = await messaging().getToken();
+    const token = await register();
     if (!token) {
       setStatus('error');
       return;
     }
-    registerMutation.mutate({token, language});
     setStatus(permissionStatus);
     return token;
-  }, [language, registerMutation]);
+  }, [register]);
 
   return {
     status,
     config: configQuery.data,
     updateConfig: configMutation.mutate,
     checkPermissions,
+    requestPermissions,
     register,
   };
 };
--- a/src/notifications/use-push-notifications.tsx
+++ b/src/notifications/use-push-notifications.tsx
@@ -1,11 +1,12 @@
-import {useCallback, useState} from 'react';
-import {useRegister} from './use-register';
-import {useConfig} from './use-config';
-import {PermissionsAndroid, Platform} from 'react-native';
+import {useLocaleContext} from '@atb/LocaleProvider';
+import Bugsnag from '@bugsnag/react-native';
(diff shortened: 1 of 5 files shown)


src/notifications/use-push-notifications.tsx
```diff
@@ -53,23 +53,31 @@ export const usePushNotifications = () => {
   }, []);
 
   const register = useCallback(async () => {
+    const token = await messaging().getToken();
+    if (!token) return;
+    registerMutation.mutate({token, language: language});
+    return token;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [language, registerMutation.mutate]);
```
rosvik, line 61: exhaustive-deps wanted me to add `registerMutation` instead of `registerMutation.mutate`, but that would call on every render.

gorandalum: Yeah, I feel like exhaustive deps could be more forgiving and allow `registerMutation.mutate` as it does with other non-function fields.

But the answer is to destructure `registerMutation` and use only `mutate` (maybe with a better name) in the `useEffect`.

rosvik: Ah, of course 💡

src/notifications/use-push-notifications.tsx
```diff
@@ -53,23 +54,30 @@ export const usePushNotifications = () => {
   }, []);
 
   const register = useCallback(async () => {
+    const token = await messaging().getToken();
```
KristianSelnas, line 57: This could potentially return an error if `requestPermission` has not been called yet. This _should_ not happen if the onboarding works correctly. But as you know, everything that should not happen happens.

KristianSelnas: I only know what the doc say: "On iOS you'll need to register for remote notifications before calling this method, you can do this by calling registerDeviceForRemoteMessages or requestPermission as part of your app startup. If you have not registered and you call this method you will receive an 'unregistered' error code."

src/notifications/use-push-notifications.tsx
```diff
@@ -53,23 +55,34 @@ export const usePushNotifications = () => {
...
-  }, [language, registerMutation]);
+  }, [register]);
 
   return {
     status,
     config: configQuery.data,
     updateConfig: configMutation.mutate,
     checkPermissions,
+    requestPermissions,
     register,
```
gorandalum, line 86: `requestPermissions` and `register` returns the token, but is the token necessary to expose outside of the notifications module?

src/notifications/use-push-notifications.tsx
```diff
@@ -53,23 +55,34 @@ export const usePushNotifications = () => {
...
+
+  const requestPermissions = useCallback(async () => {
     setStatus('updating');
     const permissionStatus = await requestUserPermission();
-    const token = await messaging().getToken();
+    const token = await register();
     if (!token) {
       setStatus('error');
       return;
     }
```
gorandalum, line 75: Setting status `error` if no token should only be done when invoking `requestPermissions`, and not when doing just `register`?

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -68,9 +68,18 @@ export const Root_TabNavigatorStack = ({navigation}: Props) => {
     fareContracts,
     serverNow,
   );
-  const {status: notificationStatus} = usePushNotifications();
+
+  const {register: registerForNotifications, status: notificationStatus} =
+    usePushNotifications();
```
gorandalum, line 73: Should `usePushNotications` be a Context instead of a hook? Hooks doesn't have shared state between invocations, so every usage of it has its own status etc. By first glance, I am not exactly sure if `notificationStatus` can be anything other than `loading` in this file 🤔

gorandalum (commented):
Some general questions, not directly related to your changes.

src/notifications/use-push-notifications.tsx
rosvik: That was added to make it visible in the debug menu, which has been used for testing: https://github.com/AtB-AS/mittatb-app/pull/4038

src/notifications/use-push-notifications.tsx
rosvik: I've made the register function as fire-and-forget, since when I use it directly, it's not an issue if it fails once in a while, and it doesn't necessarily mean notifications aren't working. When we use `requestPermissions`, we should handle errors.

The functionality is maybe a bit specific for how it's used, but not sure how to improve it.

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
rosvik: Maybe 🤔 This doesn't seem correct at least.

src/notifications/use-push-notifications.tsx
rosvik: Huh, that's not how it worked in practice for me. We should keep an eye on bugsnag, then, since it might lead to a lot of events if the docs turn out to be right.

rosvik:
@gorandalum / @KristianSelnas 

Looking more at it, I think it makes sense to turn the usePushNotification hook into a context. That is somewhat unrelated to these changes (the `register` and `requestPermissions` API should work the same), so I think I'll make a separate PR for it. With that in mind, does this change look OK as is?

gorandalum (approved):

KristianSelnas (approved):

PR #4076, opened 2023-11-27
feat: add loading state on button

hanne-at-atb:
Would like to have a loading state for submit button in https://github.com/AtB-AS/kundevendt/issues/4275. 
Therefore adding a loading prop in Button component for triggering a loading indicator. 
- Showing ActivityIndicator in right side of button
- Replacing LoadingSpinner with ActivityIndicator for consistent use of spinners (in agreement with @oyvindmikkelsenatb)
- Adding button with loading state to Design system page

https://github.com/AtB-AS/mittatb-app/assets/126664682/2e03bc68-a8d4-4550-91ff-b46f8c32f334


src/components/button/Button.tsx
```diff
@@ -190,6 +193,11 @@ export const Button = React.forwardRef<any, ButtonProps>(
               <ThemeIcon fill={textColor} {...rightIcon} />
             </View>
           )}
+          {!rightIcon && loading && (
+            <View style={rightStyling}>
+              <ActivityIndicator size="small" color={styleText.color} />
+            </View>
+          )}
```
gorandalum, line 200: Does it make sense that a button with an icon on the right side can't have a loading state? 🤔

gorandalum (approved):
Code looks good 👍 Just added a comment regarding right icon/loading state, as I didn't find that part intuitive.

gorandalum:
@oyvindmikkelsenatb And one more thing. The reason we don't have this from before is that the guideline has been to replace buttons with a loading spinner while loading, which can be seen several places in the app. I think this way, with a loading spinner on the button, is a better solution, but just doing it one place makes the app less similar in form between screens. At least if changing it, we should decide that this is the established practice in the app, and make an issue to change it the other places as well.

src/components/button/Button.tsx
hanne-at-atb: The ButtonIconProps used for rightButton includes loading. So currently a button which has an icon can show a spinner instead of the icon when loading. So this will allow a button which doesn't have an icon to show the loader as well. 
That's sort of what I hackishly tried to get around here: https://github.com/AtB-AS/mittatb-app/pull/3924/files#diff-878b44656b464aeb7303c00c30a3f811dee1b5609a0f2dd6d7449540ad09b804R228-R235
On the other hand we could maybe remove loading from ButtonIconProps, and either show ThemeIcon or ActivityIndicator a few lines above instead: https://github.com/AtB-AS/mittatb-app/pull/4076/files/c99a60d6cba4c642d1dcfb57837dd188638a43be#diff-dbb0a2706cf9edd17c0ff507bbe8bc37e950d6e7cdb77116e7ca9d5d0c29aa72R193

gorandalum: Yeah, I think it actually can be removed from `ButtonIconProps` and `ThemeIconProps`, and that we just use the `loading` parameter for the `Button` instead :+1:

gorandalum (approved):

src/components/button/Button.tsx
hanne-at-atb: Made changes accordingly now 👩🏼‍🚀

marius-at-atb:
@hanne-at-atb That's a possibility, especially if you sometimes want it to be clickable while loading. However I'm guessing that is not desired, in which case it is a good idea to handle it inside `Button` instead.

PR #4079, opened 2023-11-28
feat: Add new iOS devices for Kogenta Testers

jorelosorio:
Adding new devices.


tormoseng:
Will be tested by Kogenta

tormoseng:
OK

PR #4081, opened 2023-11-29
refactor: update component naming for token toggle info

reidzeibel:
refactor component naming to match the guideline https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md#naming


marius-at-atb (approved):
Lgtm :+1:

PR #4083, opened 2023-11-29
refactor: refactor usePushNotifications hook to a context

rosvik:
Changes the usePushNotifications hook to a context, in order to have shared state. 

Also adds a "Notifications" section to the debug menu, that reflects this state.

### DIFF
--- a/src/notifications/use-push-notifications.tsx
+++ b/src/notifications/use-push-notifications.tsx
@@ -3,8 +3,10 @@ import Bugsnag from '@bugsnag/react-native';
 import messaging, {
   FirebaseMessagingTypes,
 } from '@react-native-firebase/messaging';
-import {useCallback, useState} from 'react';
+import {createContext, useCallback, useContext, useState} from 'react';
 import {PermissionsAndroid, Platform} from 'react-native';
+import {NotificationConfigUpdate} from './api';
+import {NotificationConfig} from './types';
 import {useConfig} from './use-config';
 import {useRegister} from './use-register';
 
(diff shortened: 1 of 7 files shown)


src/notifications/use-push-notifications.tsx
```diff
@@ -74,19 +92,35 @@ export const usePushNotifications = () => {
...
+        register,
+        fcmToken,
+      }}
+    >
+      {children}
+    </NotificationContext.Provider>
+  );
 };
 
+export function useNotificationContextState() {
```
KristianSelnas, line 114: I can see that the naming of these context hooks is a bit inconsistent throughout the app. Some have only `Context` as postfix, some have only `State`, and some have no postfix. Personally I think I prefer the latter, i.e. just `useNotifications`.

KristianSelnas (approved):
LGTM 👍

src/notifications/use-push-notifications.tsx
rosvik: Agreed 👍

marius-at-atb (approved):
Lgtm.
Regarding naming:
- When is "PushNotifications" used vs just "Notifications"?
- I agree with you and @KristianSelnas that `useNotificationContextState` was a bit long. However I also think it is nice to have something in the name that separates the ContextState hooks from any other hook. Including `State` in the end is therefore a good solution imo. This is probably another point that would be nice to have coding guidelines for @gorandalum 😄

rosvik:
> When is "PushNotifications" used vs just "Notifications"?

I have tried to use "push notifications" when it refers to just phone notifications, and just "notifications" when it also includes email notification. Noticed now that this was not the case in the debug menu.

rosvik:
> I also think it is nice to have something in the name that separates the ContextState hooks

This is probably something we should agree on a convention for, yes. For now, I think this aligns with some of the most used contexts in the app (e.g. useTranslation, useFavorites) and it's easy enough to change later, so I'll move this over to testing.

PR #4084, opened 2023-11-29
feat: User preferences in trip search

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/3107

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@adrianso/react-native-device-brightness": "^1.2.7",
-    "@atb-as/config-specs": "^3.9.0",
+    "@atb-as/config-specs": "^3.10.1",
     "@atb-as/generate-assets": "^9.8.0",
     "@atb-as/theme": "^7.1.1",
     "@bugsnag/react-native": "^7.21.0",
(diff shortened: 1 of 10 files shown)


gorandalum (changes requested):
Good and efficient work! 🚀 A couple comments:

- Should probably update the `walkSpeed` (and maybe `walkReluctance`?) for the non-transit trips query as well.
- I think there is something fishy with the handling of the `tripSearchPreferences` from the Preferences context. I think the easiest and least risky option is to just remove it from the context all together, as it is just developer-stuff under the debug-menu. We can instead customise these filters if we want to try different values in the trip search.

KristianSelnas:
Yeah, I agree that taking the values from Preferences context into account is a bit dodgy. The reason I added it was because I seem to remember having seen screenshots from Profile where the user can adjust his defaults (not just debug users). I cannot really see the use case for this in Profile though, as long as you are able to save filter settings.

KristianSelnas:
> Just remove trip search preferences from the debug screen and the `PreferencesContext` all together. How it ends up now where the filter overrides the debug setting doesn't really make sense, but I would rather just remove it than using time on fixing it.

I believe removing this from debug and `PreferencesContext` should be done in a separate PR. It's not really related.

KristianSelnas:
As discussed with @gorandalum on Slack, `tripSearchPreferences` is now removed

PR #4086, opened 2023-11-30
feat: Show popover when flexible transport suggestion is dismissed

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/3855

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
@@ -37,7 +37,7 @@ import {
 import Bugsnag from '@bugsnag/react-native';
 import {TFunc} from '@leile/lobo-t';
 import {useIsFocused, useNavigation} from '@react-navigation/native';
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {ActivityIndicator, Platform, RefreshControl, View} from 'react-native';
 import {DashboardScreenProps} from '../navigation-types';
 import {SearchForLocations} from '../types';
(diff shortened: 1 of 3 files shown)


rosvik (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
```diff
@@ -345,24 +348,29 @@ export const Dashboard_TripSearchScreen: React.FC<RootProps> = ({
...
-                  rightIcon={{
-                    svg: Filter,
-                    notification: filtersState.anyFiltersApplied
-                      ? {color: 'valid', backgroundColor: 'background_accent_0'}
-                      : undefined,
-                  }}
-                  viewContainerStyle={style.filterButton}
-                  ref={filtersState.onCloseFocusRef}
-                />
+                <View ref={filterButtonWrapper}>
```
gorandalum, line 351: Just thinking out loud, not saying it should be changed in this PR:
Seeing that we now need a wrapper around the button to add another ref makes me think the bottom sheet close ref might be better to implement another way. Now the `useBottomSheet` hook creates the ref for the user, but it might be better that the user sends in the ref. That way we could just create a ref for a specific component, and that same ref could be used both as bottom sheet close ref and for popover target ref.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
```diff
@@ -397,6 +405,10 @@ export const Dashboard_TripSearchScreen: React.FC<RootProps> = ({
                     onDismiss={() => {
                       filtersState.enabled &&
                         filtersState.disableFlexibleTransport();
+                      addPopOver({
+                        oneTimeKey: 'trip-search-flexible-transport-dismissed',
+                        target: filterButtonWrapper,
```
gorandalum, line 410: Personal preference slightly towards `targetRef: filterButtonWrapperRef`.

gorandalum (approved):
Great!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
KristianSelnas: That sounds like a good idea, though slightly outside the scope of this pr.

PR #4088, opened 2023-11-30
feat: Add travel habits message in assistant tab

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/8616

This PR also connect the UI with the solution implemented on the debug menu.


scripts/android/run-android.sh
```diff
@@ -2,4 +2,11 @@
 
 envprop() { grep "^${1}=" ./.env | cut -d'=' -f2 | head -n 1; }
 
-react-native run-android --appId "$(envprop 'ANDROID_APPLICATION_ID')"
\ No newline at end of file
+appId=$(envprop 'ANDROID_APPLICATION_ID')
+
+# Check if KETTLE_API_KEY is set in the .env file
+if [ "$(envprop 'KETTLE_API_KEY')" ]; then
+  react-native run-android --appId "$appId" --mode=beaconsDebug
```
gorandalum, line 9: What does this part do? It sets `beaconsDebug` also in production for other users if that api key is available?

gorandalum (approved):

scripts/android/run-android.sh
jorelosorio: Hi, this is when you run in your machine `yarn android` because now we have two variations app and beacons, by default it lunches app but since ATB has beacons enabled it fails because beacon implementation is not loaded, so to avoid this if API is set then switch the app to beacons, but only when developing.

src/beacons/use-maybe-show-share-travel-habits-screen.tsx
```diff
@@ -25,25 +18,23 @@ export const useMaybeShowShareTravelHabitsScreen = (
...
     onboarded &&
-    isBeaconsEnabledDebugOverrideReady &&
-    isBeaconsEnabled &&
+    isBeaconsSupported &&
     hasSeenShareTravelHabitsScreen !== null &&
     !hasSeenShareTravelHabitsScreen;
 
   const maybeShowShareTravelHabitsScreen = useCallback(async () => {
-    if (await shouldShowShareTravelHabitsScreen()) {
+    if (!kettleInfo?.isBeaconsOnboarded) {
```
marius-at-atb, line 34: This will now show even if `hasLocationWhenInUsePermission` is false.
If I remember correctly, you can't get locationAlways permission without first getting locationWhenInUse permission on iOS?

marius-at-atb: Btw should it also check that `hasSeenShareTravelHabitsScreen` is `false` here?

src/utils/permissions.ts
```diff
@@ -1,101 +0,0 @@
...
-}
-
-async function placeholderReturnAfter1Sec(val: boolean): Promise<boolean> {
-  return new Promise((resolve) => {
-    setTimeout(() => {
-      resolve(val);
-    }, 1000);
-  });
-}
-
```
marius-at-atb, line 25: Ideally we should have implemented these functions instead of removing them, but ok 😅

marius-at-atb (approved):
Looks good, a couple of comments

PR #4090, opened 2023-12-01
bug: Benefit isLoading is not working as expected when enabled is false

KristianSelnas:
Found during testing of https://github.com/AtB-AS/kundevendt/issues/15779


marius-at-atb (approved):

gorandalum (commented):
Does this change have an effect on this part? Is `isLoading` still `true` if not enabled?
```ts
const {isUserEligibleForBenefit, isLoading, isError} =
  useIsEligibleForBenefit(benefit);

if (isLoading || isError) return null;
```

KristianSelnas:
That is kinda what this pr is trying to fix. If enabled is never set to true `isLoading` would be true forever.

PR #4092, opened 2023-12-04
feat: add analytics for stop place travelFrom and travelTo button

reidzeibel:
This will add analytics when user presses the `Travel From` and `Travel To` button on the stop place view, with stop place ID as property.

context : https://github.com/AtB-AS/kundevendt/issues/15241#issuecomment-1831750663


marius-at-atb (approved):
Lgtm :+1:

PR #4096, opened 2023-12-04
refactor: Use reducer in MobileTokenContext

gorandalum:
To get better control of the internal state a reducer is used
instead of multiple states.


hanne-at-atb (approved):
Nice refactor! 💯

PR #4099, opened 2023-12-05
feat: add granular notification settings

rosvik:
- Fetches notificationConfig from firestore / @atb-as/config-specs.
- Displays a list of notificationConfig groups on the Notification screen in the app, based on firestore data.
- notificationConfig modes (push and email) are however more hard coded, since they have a bunch of functionality linked to them. Open for suggestions to improve this.

closes https://github.com/AtB-AS/kundevendt/issues/15351


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
```diff
@@ -72,19 +113,15 @@ export const Profile_NotificationsScreen = () => {
                 ProfileTexts.sections.settings.linkSectionItems.notifications
                   .pushToggle.subText,
               )}
-              value={
-                permissionStatus === 'granted' &&
-                isConfigEnabled(config?.modes, 'push')
-              }
+              value={isConfigEnabled(config?.modes, 'push')}
```
KristianSelnas, line 116: Why is `permissionStatus === 'granted'`` removed? Should this have been `value={pushEnabled}`?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
```diff
@@ -117,6 +153,29 @@ export const Profile_NotificationsScreen = () => {
...
+            text={t(
+              ProfileTexts.sections.settings.linkSectionItems.notifications
+                .groupsHeading,
+            )}
+          />
+          <Section>
+            {notificationConfig?.groups.map((group) => (
+              <ToggleSectionItem
+                key={group.id}
+                text={group.toggleTitle[0].value ?? ''}
```
KristianSelnas, line 166: These language props should use `getTextForLanguage`, e.g. `getTextForLanguage(group.toggleTitle, language)`. 

I suppose `toggleTitle` is always defined in practice, but in theory it could be `undefined`, and `getTextForLanguage` is typed to potentially return undefined. The `?? ''` is therefore necessary. If the `toggleTitle` for any stange reason actually is undefined I think it makes little sense to show this toggle. I think this toggle should be shown only if `getTextForLanguage(group.toggleTitle, language)` actually returns  a value. If you decide to do that I think the code will be more readable if this toggle section item is extracted to a separate component.

rosvik: Oups, this is some remains from a dirty first draft 🙈 Will fix! Can turn it into a component as well. We should support an empty toggleDescription, but probably not toggleTitle, so I can hide it in that case.

rosvik: Added a `.filter` and proper getTextForLanguage now. Making a separate component got quite messy, especially since the `<Section>` doesn't like it when you give it non-section-item-components.

KristianSelnas (approved):

PR #4103, opened 2023-12-05
fix: Consent dashboard message

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/8616
Overview:


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -5,71 +5,138 @@ import {
...
+import {AnnouncementSheet} from './AnnouncementSheet';
+import {useBottomSheet} from '@atb/components/bottom-sheet';
+import {animateNextChange} from '@atb/utils/animation';
+import {useAnalytics} from '@atb/analytics';
+import {useAnnouncementsState} from '@atb/announcements';
 
 type Props = {
   announcement: AnnouncementType;
-  onDismiss: (announcement: AnnouncementType) => void;
+  withBottomPadding: boolean;
```
KristianSelnas, line 28: As far as I know we have stopped adding properties like this. If you need something like this at all, it should be a `style` prop that's applied to the outer element of the component.

reidzeibel:
Would this also close this issue: https://github.com/AtB-AS/kundevendt/issues/15746 ? since it is also related to multiple announcements?

KristianSelnas:
@reidzeibel No it doesn't. There has been support for multiple announcements all the time. They are shown beneath each other, as shown in the screen shot. 

I'm generally a bit critical to the direction announcements are taken with this PR. I don't think adding links to the announcement summary on the dashboard page is the best idea. One of the reasons is that it makes stacking multiple announcements horizontally look weird (very different heights). Another reason is that some of the announcements are clickable on the entire surface, and opens the content in bottom sheets, while others are not clickable and just have a link attached to the bottom. This will most likely feel confusing and inconsistent to the user.

The solution as it's presented here feels a bit rushed to me, especially from a design/UX perspective. I encourage @oyvindmikkelsenatb, @jorelosorio, and @ckbilstad to look at this again to see if you can come up with a solution that addresses the issues I've outlined above.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
gorandalum: https://github.com/AtB-AS/kundevendt/issues/15884

gorandalum: Note also that it is possible to use the `gap` style on the parent view if we want the same space between each listed announcement.

jorelosorio: I am going to explore this gap and fix it.

jorelosorio:
> @reidzeibel No it doesn't. There has been support for multiple announcements all the time. They are shown beneath each other, as shown in the screen shot. However, this was just a temporary solution. It is not sustainable in the long run, as it will clutter the dashboard significantly when announcements are used more extensively.
> 
> I'm generally a bit critical to the direction announcements are taken with this PR. I don't think adding links to the announcement summary on the dashboard page is the best idea. One of the reasons is that it makes stacking multiple announcements horizontally (as suggested in [AtB-AS/kundevendt#15746](https://github.com/AtB-AS/kundevendt/issues/15746)) look weird, since the have very different heights. Another reason is that some of the announcements are clickable on the entire surface, and opens the content in bottom sheets, while others are not clickable and just have a link attached to the bottom. This will most likely feel confusing and inconsistent to the user.
> 
> The solution as it's presented here feels a bit rushed to me, especially from a design/UX perspective. I encourage @oyvindmikkelsenatb, @jorelosorio, and @ckbilstad to look at this again to see if you can come up with a solution that addresses the issues I've outlined above.


I can sum up a bit by saying that for the context of bikes, this button sheet option seems fine, but opening a link behaves differently, and supports the same bottom sheet for this might add two or three steps more with the same information for privacy, while it could be just directly by using the link, we could also add a button to read more in the announcements, or avoid the use of bottom sheet and sends to information in another screen or link, or move all announcements to their section, so is up to how we want to design it.


I can sum up a bit by saying that for the context of bikes, this button sheet option seems fine, but for opening a link behaves differently, and make it fit the same bottom sheet might add two or three steps with the same information for privacy on bottom sheet and links, while it could be just directly by using the link, we could also add a button to read more in the announcements, or avoid the use of bottom sheet and sends to see more information in another screen or link, or move all announcements to their own section, so is up to how we want to design it. @oyvindmikkelsenatb

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
jorelosorio: Fixed

oyvindmikkelsenatb:
Since the plan is to make multiple announcements scrollable horizontally, then I agree that the component should have the same height to make things tidy visually. To achieve this, maybe we have to set a max amount of text, so that bigher amounts of text does not make the component bigger (view screeenshot). We could also use the same solution as in the repeat purchase component to size the component

If we do not make it possible to link directly from the components then we're making more clicks for the user as Jorge is mentioning, so just have that in mind. That being said, the bottom sheet solution might work just fine for what we're trying to achieve with the information. The main goal with this information is not to have the user click the link, but rather to make the user feel safe about sharing info. Also if we decide to only shown a set amount of text in the component, then there would have to be a place for the user to read the full text either way, meaning we would need to use the bottom sheet either way.

An other thing I want to mention is that the "byskkel compnent" does not clearly indicate that clicking the component opens a bottom sheet. Adding a button field at the bottom of the components as Jorge suggest solves this. We could of cource concider a more space saving solution only with icon or something (view screenshot).

@jorelosorio @KristianSelnas 
Skjermbilde 2023-12-06 kl  13 32 02

KristianSelnas:
I'm not particularly biased towards any of the solutions you suggest, @oyvindmikkelsenatb. All I'm saying is that I don't think the solution as it sits right now is a good one, and that it needs some more thoughtwork.

jorelosorio:
> Since the plan is to make multiple announcements scrollable horizontally, then I agree that the component should have the same height to make things tidy visually. To achieve this, maybe we have to set a max amount of text, so that bigher amounts of text does not make the component bigger (view screeenshot). We could also use the same solution as in the repeat purchase component to size the component
> 
> If we do not make it possible to link directly from the components then we're making more clicks for the user as Jorge is mentioning, so just have that in mind. That being said, the bottom sheet solution might work just fine for what we're trying to achieve with the information. The main goal with this information is not to have the user click the link, but rather to make the user feel safe about sharing info. Also if we decide to only shown a set amount of text in the component, then there would have to be a place for the user to read the full text either way, meaning we would need to use the bottom sheet either way.
> 
> An other thing I want to mention is that the "byskkel compnent" does not clearly indicate that clicking the component opens a bottom sheet. Adding a button field at the bottom of the components as Jorge suggest solves this. We could of cource concider a more space saving solution only with icon or something (view screenshot).
> 
> @jorelosorio @KristianSelnas Skjermbilde 2023-12-06 kl 13 32 02

I like option A

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -5,71 +5,134 @@ import {
...
+export const Announcement = ({announcement, style}: Props) => {
+  const styles = useStyle();
   const {t} = useTranslation();
   const {language} = useTranslation();
   const {theme} = useTheme();
+  const analytics = useAnalytics();
+  const {dismissAnnouncement} = useAnnouncementsState();
+  const {open: openBottomSheet, close: closeBottomSheet} = useBottomSheet();
+
+  const handleDismiss = (announcement: AnnouncementType) => {
```
marius-at-atb, line 40: Now that `handleDismiss` is inside Announcement, `announcement` actually doesn't have to be a parameter and can instead be used directly.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -5,71 +5,134 @@ import {
...
+            accessible={true}
+            accessibilityRole={announcement.actionButton ? 'button' : undefined}
+            accessibilityHint={
+              announcement.actionButton !== undefined
+                ? t(DashboardTexts.announcemens.button.accessibility)
+                : undefined
+            }
+          >
+            {announcement.summaryImage && (
+              <View style={styles.imageContainer}>
```
marius-at-atb, line 63: I noticed that in the `imageContainer` style, there is a negative padding value set. I don't think that is valid and will have no effect, so it should be removed I think (see line 151)

marius-at-atb (approved):
Lgtm :+1:

src/translations/screens/Dashboard.ts
```diff
@@ -20,6 +20,19 @@ const DashboardTexts = {
     announcement: {
       closeA11yHint: _('Lukk melding', 'Close announcement', 'Lukk melding'),
     },
+    buttonAction: {
+      external: _(
+        `Aktiver for å lese mer på ekstern side`,
+        `Activate to read more (external content)`,
+        `Aktiver for å lese meir på ekstern side`,
+      ),
+      deeplink: _('Aktiver for å opne lenke', 'Activate to open link', 'Aktiver for å opne lenke'),
```
marius-at-atb, line 29:
```suggestion
      deeplink: _('Aktiver for å åpne lenke', 'Activate to open link', 'Aktiver for å opne lenke'),
```

PR #4107, opened 2023-12-06
bug: Parameter transferSlack is not included in trip search when value is 0

KristianSelnas:
When the value for `transferSlack` in trip search filter is set to 0 (which is a totally valid value) the parameter was not included in the request to BFF.


rosvik (approved):
Good catch 👍 Just use the `fix: ` prefix instead of `bug: `

gorandalum (approved):
:+1:

PR #4108, opened 2023-12-06
fix: only register if push notifications are enabled

rosvik:
We don't want to call register if notifications are disabled in remote config. 


KristianSelnas (approved):

tormoseng:
Tested OK

PR #4113, opened 2023-12-07
feat: FullScreenView and new header on purchase overview

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/15898

This pr contains the new header, appart from what's below the horizontal line:
Overview (Periodic ticket)

The `nowrap`/`flexShrink: 1` change in `<TransportModes>` is done in order to make the text wrap as shown in the screen shot below. In my opinion this does not affect other places this component is used negatively, see below.

<details>
<summary>Text wrap example</summary>

Simulator Screenshot - iPhone 15 - 2023-12-07 at 15 24 19

</details>

<details>
<summary>Transport modes before/after</summary>

Simulator Screenshot - iPhone 15 - 2023-12-07 at 15 26 01
Simulator Screenshot - iPhone 15 - 2023-12-07 at 15 25 35

</details>


KristianSelnas:
I have tested it with screen reader and it works ok. It might read out a bit too much, but I suspect seasoned screen reader users skip past this very quickly anyways.

KristianSelnas:
I know that `DepartureDetailsScreenComponent` uses just `TransportationIconBox` and `ThemeText` now. But this is (yet another) example of components that look the same, walk the same and quack the same, and should hence be the same.

KristianSelnas:
The alternative could be to create something like `<TransportModesWithText>` that can be used by purchase overview and departure details, and which the current `<TransportModes>` can use internally.

gorandalum:
Might be `TransportModesWithText` works better if we want that kind of reusability, since the text doesn't really have anything to do with the transport modes here. But doesn't it crash with screen reader on Android without the `accessible={true}` flag on the parallax content?

KristianSelnas:
I can refactor this into with a `TransportModesWithText` component, and see how that works.

I wasn't aware of the issue with screen reader and parallax content on Android, but I'll make sure to test that as well

gorandalum:
> I wasn't aware of the issue with screen reader and parallax content on Android, but I'll make sure to test that as well

Not an issue with parallax content, just a general thing on Android that things that should get focus needs to be accessible.

marius-at-atb:
Is it just me, or is it a bit confusing that TransportModesWithText uses TransportModes, rather than the other way around?

KristianSelnas:
> Is it just me, or is it a bit confusing that TransportModes uses TransportModesWithText, rather than the other way around?

I kinda agree, it's just that `<TransportModes>` does a bit more than `<TransportModesWithText>` at the moment. I can refactor this a bit more to make it more clear.

KristianSelnas:
@gorandalum I decided to leave the idea of refactoring similar uses of transport mode icons and texts into a single component, due to screen reader issues. Refactoring this without breaking any of the existing screen reader labels or focus proved to be a little more work than I'm prepared to do in the scope of this PR. However, I strongly suggest this is handled at a later stage.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -127,17 +128,24 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
...
-        setFocusOnLoad={!params.onFocusElement}
-      />
-
+        },
+        setFocusOnLoad: !params.onFocusElement,
+        globalMessageContext: GlobalMessageContextEnum.appTicketing,
+      }}
+      parallaxContent={(focusRef?: React.MutableRefObject<null>) => (
+        <FareProductHeader
+          ref={focusRef}
```
gorandalum, line 143: Maybe not set `foucsRef` if there is an `onFocusElement`? 🤔 You can test when going back from the zone selector.

gorandalum (approved):

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
KristianSelnas: It works as expected, although I have to admit I haven't fully gotten the hang of this accessibility stuff. Whether that `focusRef` should actually be there, I honestly don't know. 

When testing this I've found that navigating with voice over in general has some strange issues and inconsistencies. I believe we (@HanneLH ?) created a task for going through accessibility and focus in general, but I can't find this at the moment. I recommend that a review of this should be prioritised.

HanneLH: @KristianSelnas  We have two different task for screen reader.
https://github.com/AtB-AS/kundevendt/issues/14120 - For scooter parking.
https://github.com/AtB-AS/kundevendt/issues/10921 -  Collective task for screen reader issue

gorandalum: I see there are now three competing focuses:
- The original screen header
- The parallax content
- The on focus element set when navigating.

We should clean this up a little. Just reach out if we should talk it through.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -127,17 +128,23 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
...
-        leftButton={{
+    <FullScreenView
+      headerProps={{
+        title: getTextForLanguage(params.fareProductTypeConfig.name, language),
+        leftButton: {
           type: 'cancel',
           onPress: closeModal,
-        }}
-        globalMessageContext={GlobalMessageContextEnum.appTicketing}
-        setFocusOnLoad={!params.onFocusElement}
```
gorandalum, line 138: Should be set to `false` instead of removed, as the default is `true`.

KristianSelnas: Doh 😵‍💫

gorandalum (approved):

PR #4116, opened 2023-12-08
fix: should not auto-capitalize email input

hanne-at-atb:
Turning off auto-capitalization for email input in Edit Profile Screen


rosvik (approved):

tormoseng:
OK!

PR #4117, opened 2023-12-08
fix: allow empty email address when editing account details

hanne-at-atb:
Allow empty email address when editing account details
Closes: https://github.com/AtB-AS/kundevendt/issues/15944


rosvik (approved):
👍

PR #4119, opened 2023-12-08
fix: replace import to fix pull-to-refresh issue on Android

reidzeibel:
**The bug seems to be only reproducible in Android**

Go to the ticket screen with any active tickets, and try to do pull-to-refresh, it will be very difficult to execute it properly.

Here's an example screen recording that shows the bug:

<details>

<summary>View video</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/2d8af1a6-666a-4bdd-b1f7-835ae640bcde

</details>

Apparently after trying to disable components and looking at StackOverflow, the issue was wrong import, if we just change the import of `ScrollView` as shown in this PR, everything works fine as shown in video below.

<details>

<summary>After updating import</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/a4690c0b-f2c9-4a82-9673-fba868d6b26e

</details>


gorandalum (changes requested):
Regarding using [react-native-gesture-handler](https://docs.swmansion.com/react-native-gesture-handler/docs/), it is to have better performance, especially on older phones. From their doc:
> The motivation for building this library was to address the performance limitations of React Native's Gesture Responder System and to provide more control over the built-in native components that can handle gestures.

Christian [had a comment in 2020](https://mittatb.slack.com/archives/C018UPET2Q2/p1598276867078700) about why we added it in the app:
> Handler om perf., gesture handler bruker native gesture response i stedet for JS. Merkes ikke på nyere devices, men vil være merkbart på eldre. Spesielt i forhold til scroll views o. l, men også for knappene.

When searching for other people having the same problem I find some issues, like this one:
https://github.com/software-mansion/react-native-gesture-handler/issues/1067

Also it seems that in 2022 it was added a possibility to use a `RefreshControl` from react-native-gesture-handler to mitigate this problem:
https://github.com/software-mansion/react-native-gesture-handler/pull/2137

Maybe we could try that one first?

And when we are touching this, maybe set up the eslint-rule `import-warning` to specify which import(s) are correct.

reidzeibel:
@gorandalum apparently switching everything to `react-native-gesture-handler` also works perfectly so I have changed the fix to that. As for the import warning for eslint rule I can make another PR along with the necessary changes. Thanks for the suggestion and for the information about `react-native-gesture-handler`

gorandalum (approved):

tormoseng:
👍

PR #4123, opened 2023-12-11
chore: update android dependencies

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/15989

- Updated some libraries that can be updated without any issues (soLoader, NDK, desugar, splashscreen).
- Updated Android bugsnag integration based on https://docs.bugsnag.com/platforms/android/#installation
- Updated Android bugsnag initialization (as shown in the issue)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -4,8 +4,7 @@ apply plugin: "com.facebook.react"
 
 apply plugin: "kotlin-android"
 
-apply from: "../../node_modules/@bugsnag/react-native/bugsnag-react-native.gradle"
-
+apply plugin: "com.bugsnag.android.gradle"
 /**
  * Set this to true to Run Proguard on Release builds to minify the Java bytecode.
  */
@@ -151,11 +150,11 @@ android {
 
 dependencies {
 
-    implementation 'com.facebook.soloader:soloader:0.9.0+'
+    implementation 'com.facebook.soloader:soloader:0.10.5'
     implementation "org.jetbrains.kotlin:kotlin-stdlib:${KOTLIN_VERSION}"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${KOTLIN_VERSION}"
 
-    coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:1.1.5"
+    coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:2.0.4"
 
     implementation fileTree(dir: "libs", include: ["*.jar"])
 
@@ -163,7 +162,7 @@ dependencies {
     implementation("com.facebook.react:react-android")
 
     implementation 'androidx.multidex:multidex:2.0.1'
-    implementation "androidx.core:core-splashscreen:1.0.0"
+    implementation "androidx.core:core-splashscreen:1.0.1"
 
     implementation 'io.intercom.android:intercom-sdk-base:9.+'
     implementation 'io.intercom.android:intercom-sdk:9.+'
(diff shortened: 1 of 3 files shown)


gorandalum (commented):
Are we sure about the bugsnag stuff? Not mentioned in the RN guide:
https://docs.bugsnag.com/platforms/react-native/react-native/

reidzeibel:
@gorandalum that's what I'm also not sure about, why do we need the integration on the Android side if we already integrate bugsnag on React Native? Does it even used?

gorandalum:
Following the react-native installation guide for Android adds some stuff into some files in the android folder I think.

I think updating bugsnag is good, but try following the React Native guide instead and see how it ends up. Might be the same...

reidzeibel:
@gorandalum I checked the manual setup guide https://docs.bugsnag.com/platforms/react-native/react-native/manual-setup/ and looks like the setup is similar to the one I updated on the Android folders. 

image

android/app/build.gradle
```diff
@@ -151,19 +150,19 @@ android {
 
 dependencies {
 
-    implementation 'com.facebook.soloader:soloader:0.9.0+'
```
jorelosorio, line 154: Not sure if this version is required for the current react-version or belongs to the newest, however it says 0.9.0+ might be that we are using the latest 🤔

android/app/build.gradle
```diff
@@ -151,19 +150,19 @@ android {
...
-    coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:1.1.5"
+    coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:2.0.4"
 
     implementation fileTree(dir: "libs", include: ["*.jar"])
 
     // The version of react-native is set by the React Native Gradle Plugin
     implementation("com.facebook.react:react-android")
 
     implementation 'androidx.multidex:multidex:2.0.1'
-    implementation "androidx.core:core-splashscreen:1.0.0"
```
jorelosorio, line 166: There is a react-plugging in the package.json that uses it, or does this belongs to the React SDK. If both are in the same, I think then is OK to upgrade.

reidzeibel: Can you point which SDK needs it?

android/app/build.gradle
reidzeibel: having + on the version actually tries to use the latest version available, but using + on the version is considered a bad thing because it adds some uncertainties to the build. I think this blog gives a good explanation on it : https://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

android/app/build.gradle
jorelosorio: https://github.com/AtB-AS/mittatb-app/blob/ef0ecc6c32b97435062e1ed9060715f02a76c9d8/package.json#L95

Maybe is a dependency of that one, not quite sure, good article.

reidzeibel: nevermind found it, and it specifies 1.0.0 so I'll revert the change

reidzeibel: reverted

reidzeibel:
We should probably update Intercom integration as well, since the package we are using now are deprecated : https://www.npmjs.com/package/react-native-intercom

And Intercom has provided an official package for react-native, although we would need to update the minimum iOS version to 13. https://github.com/intercom/intercom-react-native

reidzeibel:
This is the warning that comes when running our app on Android : 

```
[Bugsnag] "bugsnag-react-native.gradle" should not applied manually. 
Install the Bugsnag Android Gradle Plugin and remove this from your Gradle file.
```

gorandalum (approved):
Looks fine by me! Can create a staging-build from the branch to test that it works as expected before mering.

PR #4124, opened 2023-12-12
feat: add register email button in notifications

hanne-at-atb:
Added button to add e-mail to profile for e-mail notifications.

Closes: https://github.com/AtB-AS/kundevendt/issues/7225

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
@@ -110,10 +110,22 @@ export const Profile_NotificationsScreen = ({
                             .notifications.emailToggle.noEmailPlaceholder,
                     )
               }
+              disabled={data?.email === ''}
               value={isConfigEnabled(config?.modes, 'mail')}
               onValueChange={(enabled) => handleModeToggle('mail', enabled)}
             />
           </Section>
+          {data?.email === '' && (
+            <Button
+              onPress={() => navigation.navigate('Profile_EditProfileScreen')}
+              text={t(
+                ProfileTexts.sections.settings.linkSectionItems.notifications
+                  .button,
+              )}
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
```diff
@@ -104,10 +110,22 @@ export const Profile_NotificationsScreen = () => {
                             .notifications.emailToggle.noEmailPlaceholder,
                     )
               }
+              disabled={data?.email === ''}
               value={isConfigEnabled(config?.modes, 'mail')}
               onValueChange={(enabled) => handleModeToggle('mail', enabled)}
             />
           </Section>
+          {data?.email === '' && (
```
gorandalum, line 118: Is the thought here that if `data` is `undefined` then this button shouldn't be shown? In that case I think it is a little to "hidden", and kinda hard to be aware of on later refactors.

A tip when using `useQuery`, check for the status `success` (or use `isSuccess`), so you get proper typing on the returned data. 

Also I think we want to show the button if email is falsey, to not to be too brittle on later changes.

gorandalum (approved):
Looks good! One comment on something I think could be slightly improved.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_NotificationsScreen.tsx
hanne-at-atb: Made some changes now. 
Email should always be set on the response, but if there is no email registered for the profile it will be empty string. `email: ''`
And that is the scenario (profile has no email) where we would like to show the button. 
With this in mind, was the change according to what you requested @gorandalum ? :)

gorandalum: I think it is more clear :)

rosvik (approved):

PR #4126, opened 2023-12-13
feat: pre-fill e-mail from profile when requesting a receipt

hanne-at-atb:
Closes: https://github.com/AtB-AS/kundevendt/issues/15408

### DIFF
--- a/src/stacks-hierarchy/Root_ReceiptScreen.tsx
+++ b/src/stacks-hierarchy/Root_ReceiptScreen.tsx
@@ -10,11 +10,12 @@ import {
   useTranslation,
 } from '@atb/translations';
 import {isValidEmail} from '@atb/utils/validation';
-import React, {useState} from 'react';
-import {View} from 'react-native';
+import React, {useEffect, useState} from 'react';
+import {ActivityIndicator, View} from 'react-native';
 import {RootStackScreenProps} from '../stacks-hierarchy/navigation-types';
 import {Section, TextInputSectionItem} from '@atb/components/sections';
 import {useAnalytics} from '@atb/analytics';
+import {useProfileQuery} from '@atb/queries';
 
 type Props = RootStackScreenProps<'Root_ReceiptScreen'>;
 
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_ReceiptScreen.tsx
```diff
@@ -28,13 +29,20 @@ type MessageState =
...
   const [email, setEmail] = useState('');
   const [reference, setReference] = useState<string | undefined>(undefined);
   const [state, setState] = useState<MessageState>(undefined);
   const {t} = useTranslation();
   const a11yContext = useAccessibilityContext();
   const analytics = useAnalytics();
 
+  useEffect(() => {
+    if (isSuccess) {
+      setEmail(data.email);
```
marius-at-atb, line 42: Should it be `data?.email`?

marius-at-atb (approved):
Lgtm!
Could consider receiving data like in `Profile_EditProfileScreen`: `data: customerProfile`

src/stacks-hierarchy/Root_ReceiptScreen.tsx
hanne-at-atb: Since we are checking isSuccess we know that we have received data from profile, and email is a mandatory field we don't need the ? 😄

src/stacks-hierarchy/Root_ReceiptScreen.tsx
```diff
@@ -28,13 +29,20 @@ type MessageState =
 export function Root_ReceiptScreen({route}: Props) {
   const {orderId, orderVersion} = route.params;
   const styles = useStyles();
+  const {data: customerProfile, status: profileStatus} = useProfileQuery();
```
gorandalum, line 32: Regarding the changes how we handle profile info:
- Will profile data changes still be reflected in Firestore?
- Can the app just read data from Firestore instead, or should we do http requests to read from the profile service?

Maybe @mjenssen have a thought about this?

mjenssen: For the time being, Firestore will be updated (due to backward compatibility). However, we do want to eventually move away from Firestore since it has been giving us way too many problems, especially wrt writing and updating data in a robust manner. Additionally, we have wanted to have the ability to provide functionality such as "if the data is in Firestore, return it, otherwise fetch that data from another places", which isn't possible in a scalable way.

The recommendation is to use the profile service going forward.

PR #4127, opened 2023-12-14
feat: Ticket information screen

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/15899

This pr contains the ticket information page with just the description section. Mobility benefits and generic tips and information are handled in separate PRs.

<details>
<summary>Screen shots</summary>

image (2)
image (3)

</details>

Related pr: https://github.com/AtB-AS/firestore-configuration/pull/397

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
@@ -14,6 +14,7 @@ import {StyleSheet} from '@atb/theme';
 import {TransportationIconBoxList} from '@atb/components/icon-box';
 import {Button} from '@atb/components/button';
 import {Info} from '@atb/assets/svg/mono-icons/status';
+import {useIsTicketInformationEnabled} from '@atb/stacks-hierarchy/Root_PurchaseOverviewScreen/use-is-ticket-information-enabled';
 
 type Props = {
   fareProductTypeConfig: FareProductTypeConfig;
@@ -33,6 +34,7 @@ export const FareProductHeader = forwardRef<View, Props>(
   ) => {
     const {t, language} = useTranslation();
     const styles = useStyle();
+    const [isTicketInformationEnabled] = useIsTicketInformationEnabled();
 
     return (
       <View style={style}>
@@ -49,26 +51,28 @@ export const FareProductHeader = forwardRef<View, Props>(
             {getTextForLanguage(fareProductTypeConfig.name, language) ?? ''}
           </ThemeText>
         </View>
-        <View style={styles.headerSubSection}>
-          <ThemeText
-            type="body__secondary"
-            color="background_accent_0"
-            style={styles.ticketDescription}
-            numberOfLines={1}
-          >
-            {getTextForLanguage(
-              preassignedFareProduct.description ?? [],
-              language,
-            )?.replaceAll('\n', ' ')}
-          </ThemeText>
-          <Button
-            type="pill"
-            leftIcon={{svg: Info}}
-            interactiveColor="interactive_1"
-            text={t(PurchaseOverviewTexts.ticketInformation.button)}
-            onPress={onTicketInfoButtonPress}
-          />
-        </View>
+        {isTicketInformationEnabled && (
+          <View style={styles.headerSubSection}>
+            <ThemeText
+              type="body__secondary"
+              color="background_accent_0"
+              style={styles.ticketDescription}
+              numberOfLines={1}
+            >
+              {getTextForLanguage(
+                preassignedFareProduct.description ?? [],
+                language,
+              )?.replaceAll('\n', ' ')}
+            </ThemeText>
+            <Button
+              type="pill"
+              leftIcon={{svg: Info}}
+              interactiveColor="interactive_1"
+              text={t(PurchaseOverviewTexts.ticketInformation.button)}
+              onPress={onTicketInfoButtonPress}
+            />
+          </View>
+        )}
       </View>
     );
   },
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
@@ -1,33 +1,74 @@
-import {getTextForLanguage, useTranslation} from '@atb/translations';
+import {
+  getTextForLanguage,
+  PurchaseOverviewTexts,
+  useTranslation,
+} from '@atb/translations';
 import {StyleProp, View, ViewStyle} from 'react-native';
-import {FareProductTypeConfig} from '@atb/configuration';
+import {
+  FareProductTypeConfig,
+  PreassignedFareProduct,
+} from '@atb/configuration';
 import React, {forwardRef} from 'react';
 import {ThemeText} from '@atb/components/text';
 import {StyleSheet} from '@atb/theme';
 import {TransportationIconBoxList} from '@atb/components/icon-box';
+import {Button} from '@atb/components/button';
+import {Info} from '@atb/assets/svg/mono-icons/status';
 
 type Props = {
   fareProductTypeConfig: FareProductTypeConfig;
+  preassignedFareProduct: PreassignedFareProduct;
+  onTicketInfoButtonPress: () => void;
   style?: StyleProp<ViewStyle>;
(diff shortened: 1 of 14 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
```diff
@@ -1,33 +1,78 @@
...
+            {getTextForLanguage(fareProductTypeConfig.name, language) ?? ''}
+          </ThemeText>
+        </View>
+        {isTicketInformationEnabled && (
+          <View style={styles.headerSubSection}>
+            <ThemeText
+              type="body__secondary"
+              color="background_accent_0"
+              style={styles.ticketDescription}
+              numberOfLines={1}
```
gorandalum, line 60: Should it support markdown here as well? In case it starts with bold or something.

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
```diff
@@ -0,0 +1,77 @@
...
+        title: t(
+          PurchaseOverviewTexts.ticketInformation.informationDetails.title,
+        ),
+        leftButton: {type: 'close'},
+      }}
+    >
+      <ScrollView style={styles.container}>
+        {preassignedFareProduct.productDescription && (
+          <>
+            <SectionHeading>
```
gorandalum, line 35: I think `ContentHeading` is the component to use for these headings going forwards. [It was created](https://github.com/AtB-AS/mittatb-app/pull/4111) so that these headings have the same look and feel around the app.

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
```diff
@@ -0,0 +1,77 @@
...
+import {SectionHeading} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/SectionHeading';
+import {TransportationIconBoxList} from '@atb/components/icon-box';
+
+type Props = RootStackScreenProps<'Root_TicketInformationScreen'>;
+
+export const Root_TicketInformationScreen = (props: Props) => {
+  const {t, language} = useTranslation();
+  const styles = useStyle();
+  const fareProductTypeConfig = props.route.params.fareProductTypeConfig;
+  const preassignedFareProduct = props.route.params.preassignedFareProduct;
```
gorandalum, line 21: Maybe just pass the id of the `preassignedFareProduct` here, and retrieve the `preassignedFareProduct` and `fareProductTypeConfig` objects with it. That is more inline with the [React Navigation doc](https://reactnavigation.org/docs/params/#what-should-be-in-params):
> Params are like options for a screen. They should only contain information to configure what's displayed in the screen. Avoid passing the full data which will be displayed on the screen itself (e.g. pass a user id instead of user object)

gorandalum (approved):
Good work! Some non-blocking comments added :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
```diff
@@ -1,33 +1,78 @@
...
+            >
+              {getTextForLanguage(
+                preassignedFareProduct.productDescription ?? [],
+                language,
+              )?.replaceAll('\n', ' ')}
```
marius-at-atb, line 65: Should it also look for `<br>`, `<br/>` and `<br />`?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
KristianSelnas: I get your point, but I'm not sure what's the best option. I've added `.replaceAll('\n', ' ')` to get the text to flow as one line. I think rendering markup here could have unknown consequences, e.g. if the markdown contains a url that's broken by the line length limit etc. If anything, I think the best option would actually be to strip the markdown markup, but I don't know if that is trivial.

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
```diff
@@ -0,0 +1,77 @@
...
+          </>
+        )}
+      </ScrollView>
+    </FullScreenView>
+  );
+};
+
+const useStyle = StyleSheet.createThemeHook((theme) => ({
+  container: {
+    backgroundColor: theme.static.background.background_accent_0.background,
```
marius-at-atb, line 69: I think that the background color should instead be set with `contentColor="background_accent_0"` on `FullScreenView`, to avoid this when scrolling:

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
KristianSelnas: I agree that passing id's is cleaner, but I thought passing objects was fine as long as it's serialised in a predictable way?

marius-at-atb (approved):
Lgtm! :+1:

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
KristianSelnas: I get the point after reading the docs more thoroughly. Thank you!

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
KristianSelnas: Ideally, I think all markup should be should be stripped: https://github.com/AtB-AS/mittatb-app/pull/4127#discussion_r1427975624

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
KristianSelnas: It doesn't seem like Marked har any extensions for stripping markup: https://marked.js.org/using_advanced#extensions

My two-minutes research haven't found any other small, well maintained 3rd party libraries for this either. Writing one from scratch with regexes and stuff is certainly not impossible, but perhaps more work and risk than we want for this particular use case.

A safe option could be to add a separate property to preassigned fare product, e.g. `productDescriptionShort`. We are already adding a custom property for this. What do you think, @gorandalum and @marius-at-atb ?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
KristianSelnas: Resolving this and moving discussion to https://github.com/AtB-AS/mittatb-app/pull/4127#discussion_r1428035977

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FareProductHeader.tsx
marius-at-atb: I kinda agree that even if there is markdown in the text, it is not desirable to show that in this header text.
However it would be preferable imo to solve it with code. Could e.g. [strip-markdown](https://github.com/remarkjs/strip-markdown) be an option?

KristianSelnas: As far as I can tell, strip-markdown is a plugin for Remark. We are using Marked not Remark for markdown, so I don't think it's a good idea to include Remark just to use that one plugin.

KristianSelnas: I've created a simple `strip-markdown` function now using Marked.

src/stacks-hierarchy/navigation-types.ts
```diff
@@ -59,6 +59,11 @@ type FareContractDetailsRouteParams = {
   orderId: string;
 };
 
+type TicketInformationScreenParams = {
+  fareProductTypeConfigType: string;
+  preassignedFareProductId: string | undefined;
```
gorandalum, line 64: I think the id should be required. Doesn't really make sense to navigate to a ticket info screen if one doesn't have an id for the ticket product to show.

KristianSelnas: Yes it does. It has the generic ticket info section, and potentially mikro mobility benefits.

PR #4130, opened 2023-12-14
feat: enable obfuscation and resource shrinking for Android

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/3744

This PR will enable obfuscation and resource shrinking in Android builds for store variant. 

In order to test the build, you can go to the GitHub Actions, then click on the "Build staging android project", select "run workflow", select branch `ridwan/obfuscation-android`, set force new build to true, and then run the workflow. 

The build will be for "Store" variant even though its "Build staging android project" (just for testing the build). Will only build for AtB, and will publish the result to AppCenter.

image

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -8,7 +8,7 @@ apply plugin: "com.bugsnag.android.gradle"
 /**
  * Set this to true to Run Proguard on Release builds to minify the Java bytecode.
  */
-def enableProguardInReleaseBuilds = false
+def enableProguardInReleaseBuilds = true
 
 /**
  * The preferred build flavor of JavaScriptCore.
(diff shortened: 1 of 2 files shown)


reidzeibel:
Obfuscation result on MainApplication.java 
image

Non-obfuscated result 

image

reidzeibel:
The result of the latest changes is pretty much the same, classes are obfuscated properly and we just need to test that everything works as intended.

reidzeibel:
@gorandalum do you have any further input on this task? or should we just try to test it when it gets merged?

gorandalum:
I'm okay with this. If we are unsure if this is OK by Entur we can provide them the apk and they can take a look.

gorandalum (approved):

PR #4135, opened 2023-12-15
fix: Error while beacons is activated

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/8621

This issue fix a non-covered case when requesting kettle info but without having initialized the SDK, so it might close the app is the state is not right.

### DIFF
--- a/ios/BridgeModules/KettleSDKExtension.m
+++ b/ios/BridgeModules/KettleSDKExtension.m
@@ -5,19 +5,22 @@
 #import <KettleKit/KettleKit.h>
 #import <KettleKit/KettleKit-Swift.h>
 static void InitializeKettle(NSDictionary *launchOptions) {
+  NSString *apiKey = [NSString stringWithFormat:@"%@", KETTLE_API_KEY];
+  if ([apiKey length] > 0) {
     KTLConfig* config = [KTLConfig KTLDefaultConfig];
-    NSString *apiKey = [NSString stringWithFormat:@"%@", KETTLE_API_KEY];
-    #if DEBUG
-        config.developmentApiKey = apiKey;
-    #else
-        config.productionApiKey = apiKey;
-    #endif
-
+    
+#if DEBUG
+    config.developmentApiKey = apiKey;
+#else
+    config.productionApiKey = apiKey;
+#endif
+    
     config.developmentLogLevel = KTLLogLevelDebug;
     config.productionLogLevel = KTLLogLevelNone;
-
+    
     // Initialize Kettle
     [KTLKettle prepare:config launchOptions:launchOptions];
+  }
 }
 #endif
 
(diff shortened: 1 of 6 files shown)


ios/BridgeModules/KettleSDKExtension.m
```diff
@@ -5,19 +5,22 @@
 #import <KettleKit/KettleKit.h>
 #import <KettleKit/KettleKit-Swift.h>
 static void InitializeKettle(NSDictionary *launchOptions) {
+  NSString *apiKey = [NSString stringWithFormat:@"%@", KETTLE_API_KEY];
```
jorelosorio, line 8: Extra security to make sure that the KETTLE_API_KEY is present and have a Key on it, not allow empty.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PrivacyScreen.tsx
```diff
@@ -98,7 +99,9 @@ export const Profile_PrivacyScreen = () => {
               }}
               testID="privacyButton"
               onPress={async () => {
-                await Linking.openURL(kettleInfo.privacyDashboardUrl!);
+                const privacyDashboardUrl = await getPrivacyDashboardUrl();
+                privacyDashboardUrl &&
+                  (await Linking.openURL(privacyDashboardUrl));
```
gorandalum, line 104: Not really sure that this is the best handling for the user. Earlier we didn't show the button if there was no `privacyDashboardUrl`. Now we show the button, but clicking it has no effect if no `privacyDashboardUrl`. 

Is it possible to make the handling more safe than before, but still having the url's available on the kettle info object?

jorelosorio: I had a talk with Kogenta and seems like this method always returns a URL so that is why is there but if for some reason it fails no action as you said, before this URL was inside KettleInfo, but we moved it out because the URL must be generated every time is requested because it expires in 15m so in order to show the button based on if there is a URL or not is fine but might be that when the user really presses the button the link expires, so that is why is now upon requests.

marius-at-atb (approved):
Lgtm :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -49,16 +49,7 @@ export const Announcement = ({announcement, style}: Props) => {
...
-            style={styles.content}
-            accessible={true}
-            accessibilityRole={announcement.actionButton ? 'button' : undefined}
-            accessibilityHint={
-              announcement.actionButton !== undefined
-                ? t(DashboardTexts.announcemens.button.accessibility)
-                : undefined
-            }
-          >
+          <View style={styles.content} accessible={true}>
```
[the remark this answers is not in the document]
jorelosorio: It was part of the review process that accessibility was wrong

PR #4136, opened 2023-12-18
feat: provide enabled flag to notification register endpoint

rosvik:
Provides the enabled flag to notifications /register endpoint, and initializes the permissionStatus field on app launch by calling `checkPermissions` instead of `register` directly. `checkPermissions` now calls register after the permissions are determined.

closes https://github.com/AtB-AS/kundevendt/issues/15905


marius-at-atb (approved):
Nice!

KristianSelnas (approved):

PR #4139, opened 2023-12-19
feat: delete collected data on account deletion

adriawh:


adriawh:
@jorelosorio is taking over this issue and PR

jorelosorio:
I made the test and now seems to work after removing the user, this message appears.

Screenshot_1704368452

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DeleteProfileScreen.tsx
```diff
@@ -34,8 +35,25 @@ export const Profile_DeleteProfileScreen = ({
...
 
-  const doDeleteProfile = async () => {
-    await Alert.alert(
+  const {deleteCollectedData} = useBeaconsState();
+
+  const handleDeleteProfile = async () => {
+    try {
+      const deleteOK = await deleteProfile();
+      if (deleteOK) {
+        await deleteCollectedData();
```
jorelosorio, line 44: The internal functions handle if beacons is enabled or not for the organization and if is already initialized.

jorelosorio (approved):

src/beacons/BeaconsContext.tsx
```diff
@@ -132,7 +132,12 @@ const BeaconsContextProvider: React.FC = ({children}) => {
   const deleteCollectedData = useCallback(async () => {
     if (!isBeaconsSupported) return;
     if (isInitializedRef.current) {
-      await Kettle.deleteCollectedData();
+      try {
+        await Kettle.deleteCollectedData();
+        return Promise.resolve();
+      } catch (error) {
+        return Promise.reject(error);
+      }
```
gorandalum, line 140: Isn't this just the same as `return Kettle.deleteCollectedData()`?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DeleteProfileScreen.tsx
```diff
@@ -33,9 +28,24 @@ export const Profile_DeleteProfileScreen = ({
...
+        await signOut();
+      } else {
+        setDeleteError(true);
+      }
+    } catch (error) {
+      setDeleteError(true);
```
gorandalum, line 43: I don't think we can set delete error when delete collected data fails. Especially if we delete profile first. When we have deleted profile I think we are past the point of no return, and need to sign out the user.

gorandalum (changes requested):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DeleteProfileScreen.tsx
jorelosorio: I made a fix to this as in the same way it was done in `deleteProfile()`

src/beacons/BeaconsContext.tsx
```diff
@@ -132,12 +135,15 @@ const BeaconsContextProvider: React.FC = ({children}) => {
...
-      }
+      const deleteOK = await Kettle.deleteCollectedData()
+        .then(() => {
+          return true;
+        })
+        .catch((error) => {
+          Bugsnag.notify(error);
+          return false;
+        });
+      return deleteOK;
```
marius-at-atb, line 146: `deleteOK` is being returned, but is never used where `deleteCollectedData` is used as far as I can see?
Maybe it can be simplified to something like
```
await Kettle.deleteCollectedData().catch((error) => {
  Bugsnag.notify(error);
});
```
(and then also update deleteCollectedData type definition and init value)

jorelosorio: Is not being used yes, we can simplify it a bit

src/beacons/BeaconsContext.tsx
```diff
@@ -135,15 +133,9 @@ const BeaconsContextProvider: React.FC = ({children}) => {
...
-      const deleteOK = await Kettle.deleteCollectedData()
-        .then(() => {
-          return true;
-        })
-        .catch((error) => {
-          Bugsnag.notify(error);
-          return false;
-        });
-      return deleteOK;
+      await Kettle.deleteCollectedData().catch((error) => {
```
marius-at-atb, line 136:
```suggestion
     Kettle.deleteCollectedData().catch((error) => {
```
Sorry I meant to exclude await from the previous suggestion, so it can be removed I think

marius-at-atb (approved):
Looks good 😄

src/beacons/BeaconsContext.tsx
```diff
@@ -153,10 +154,12 @@ const BeaconsContextProvider: React.FC = ({children}) => {
...
   }, [isBeaconsSupported]);
 
-  const deleteCollectedData = useCallback(async () => {
+  const deleteCollectedData = useCallback(() => {
     if (!isBeaconsSupported) return;
     if (isInitializedRef.current) {
-      await Kettle.deleteCollectedData();
+      Kettle.deleteCollectedData().catch((error) => {
+        Bugsnag.notify(error);
+      });
```
gorandalum, line 162: If you have revoked permissions, will the initialised ref be false and data not deleted? Or is data deleted when you revoke?

jorelosorio: I think data is not removed when revoke collect data, only `delete collected data` removes it, as far I know.

marius-at-atb: Good catch, I think it would actually require to init the sdk to delete the data in that case

marius-at-atb: So maybe it is needed to init, delete, then stop again when `isInitializedRef.current` is false?

jorelosorio: But call revoke do not turn false: isInitializedRef.current, once is initialized the SDK, it remains initialized. Which is different from running.

jorelosorio: So in the case is false is because is either not onboarded or disabled in the remote config, for those cases in the debug menu beacons is not show and there is PR to do not show the privacy beacons related stuff if not onboarded.

marius-at-atb: @jorelosorio Yes I confused stop with un-init, which is not a method and so won't happen until the user restarts the app.
So the beacons privacy will be hidden for the false case after another upcoming PR? I guess then the problem is avoided :+1:

gorandalum: I'm not quite following what problem is avoided 😅  We still need to be able to delete when uninitialised? 🤔 

Since it seems like initialisation only happen when user are sharing:
```ts
if (permissions.length > 0) {
  await NativeModules.KettleSDKExtension.initializeKettleSDK();
  isInitializedRef.current = true;
}
```

marius-at-atb: I was thinking if the button is hidden, but realising now that the Delete My Profile button should be there regardless of that flag. In which case, problem not avoided

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DeleteProfileScreen.tsx
```diff
@@ -33,9 +28,20 @@ export const Profile_DeleteProfileScreen = ({
     filterActiveOrCanBeUsedFareContracts(fareContracts, serverNow).length > 0;
 
   const [deleteError, setDeleteError] = useState<boolean>(false);
+  const {deleteCollectedData} = useBeaconsState();
+
+  const handleDeleteProfile = async () => {
+    const isProfileDeleted = await deleteProfile();
+    if (isProfileDeleted) {
+      await deleteCollectedData();
+      await signOut();
```
marius-at-atb, line 37: Is the `navigation.navigate('Profile_RootScreen');` not needed anymore?

marius-at-atb (approved):
Nice work!

src/beacons/BeaconsContext.tsx
```diff
@@ -97,36 +86,54 @@ enum storeKey {
...
+  );
+
+  const stopBeacons = useCallback(async () => {
+    const permissions = await allowedPermissionsForBeacons();
+    if (
+      permissions.length > 0 &&
+      beaconsInfo?.isStarted &&
+      isInitializedRef.current
+    ) {
+      Kettle.stop(permissions);
```
gorandalum, line 124: How important is it that this list of permissions when stopping matches the list it was started with?

From the looks of it, it seems that we fetch permissions from the OS layer before stopping, and it might happen that someone has changed OS permissions between starting and stopping the Kettle module.

gorandalum (approved):

src/beacons/BeaconsContext.tsx
jorelosorio: I have changed this to reflect the changes I made in the stop, this stop now should stop all as it was fixed for the auto-stopping.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DeleteProfileScreen.tsx
jorelosorio: No, it was removed due to an issue, signOut nows handles the redirection.

PR #4140, opened 2023-12-19
feat: toggle emailfield with remote config

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/16096

This PR hides the email input field from the profile edit page in the app. FRAM needs this to avoid mixing app accounts with phone authentication and webshop accounts with email authentication. 

We want to avoid the following scenario:
1. User signs in with phone number in the app
2. User adds email adress to account in app
3. Email is already in use in Entur systems, giving error message when user tries to create webshop account with email.

A disable flag was chosen for remote config because no config would then be required for AtB and NFK, and to ensure that the remote config value can be removed as soon as app and webshop accounts are merged at FRAM.


reidzeibel (approved):
Looks good to me!

src/translations/screens/subscreens/EditProfileScreen.ts
```diff
@@ -45,6 +45,11 @@ export const EditProfileTexts = {
         'The e-mail address is already in use',
         'E-postadressa er allereie i bruk',
       ),
+      disabledWithRemoteConfig: _(
+        'Appen støtter foreløpig ikke lagring av e-postadresse.',
+        'The app currently does not support saving your email address',
+        'Appen stettar førebels ikkje lagring av e-postadresse.',
```
marius-at-atb, line 51: stettar -> støttar

toretefre: You sure? https://ordbokene.no/nn/73672

Anyways, I created a PR: https://github.com/AtB-AS/mittatb-app/pull/4152

toretefre: Merged 🤝

PR #4141, opened 2023-12-19
fix: add onCameraChange callback to fix android issue

reidzeibel:
Closes https://github.com/AtB-AS/kundevendt/issues/16093

Working on this fix with @KristianSelnas . 

Issue is caused by RNMapBox issue here : https://github.com/rnmapbox/maps/issues/2902, so basically, in Android, if you show UserLocation, then `onMapIdle` callback no longer triggers properly. In iOS this is not a problem. A workaround offered by one of the RNMapBox developer is to use a debounced `onCameraChanged` event to trigger the change, as mentioned here : https://github.com/rnmapbox/maps/issues/2902#issuecomment-1633003164.

So we are implementing that solution, but currently we have some issues where we can't find a way to implement the enforced eslint exhaustive deps on useEffect. 

Therefore hopefully someone have an idea on how to do it properly. 

This PR already fixes the issue, but the code needs to be improved while keeping the solution working.

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -5,7 +5,7 @@ import {MapRoute} from '@atb/travel-details-map-screen/components/MapRoute';
 import MapboxGL, {UserLocationRenderMode} from '@rnmapbox/maps';
 import {MapState} from '@rnmapbox/maps/lib/typescript/components/MapView';
 import {Feature} from 'geojson';
-import React, {useMemo, useRef} from 'react';
+import React, {useEffect, useMemo, useRef, useState} from 'react';
 import {View} from 'react-native';
 import {MapCameraConfig, MapViewConfig} from './MapConfig';
 import {SelectionPin} from './components/SelectionPin';
@@ -17,6 +17,7 @@ import {useControlPositionsStyle} from './hooks/use-control-styles';
 import {useMapSelectionChangeEffect} from './hooks/use-map-selection-change-effect';
 import {MapProps, MapRegion} from './types';
 import {isFeaturePoint} from './utils';
+import {useDebounce} from '@atb/utils/use-debounce';
 
 export const Map = (props: MapProps) => {
   const {initialLocation} = props;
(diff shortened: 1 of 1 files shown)


marius-at-atb:
Related issue: https://github.com/AtB-AS/mittatb-app/pull/4045

marius-at-atb:
I'm guessing the answer is yes, but does the same Android issue occur when using `<NativeUserLocation />` instead of `UserLocation` with `renderMode={UserLocationRenderMode.Native}`?

src/components/map/Map.tsx
```diff
@@ -73,6 +76,24 @@ export const Map = (props: MapProps) => {
...
+  useEffect(() => {
+    if (debouncedState) {
+      loadMobility({
+        visibleBounds: [
+          debouncedState.properties.bounds.ne,
+          debouncedState.properties.bounds.sw,
+        ],
+        zoomLevel: debouncedState.properties.zoom,
+        center: debouncedState.properties.center,
+      });
```
marius-at-atb, line 88:
```suggestion
      onMapIdle(debouncedState)
```

src/components/map/Map.tsx
```diff
@@ -73,6 +76,24 @@ export const Map = (props: MapProps) => {
...
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
marius-at-atb, line 90: With the suggestion above, maybe it will work when wrapping `onMapIdle` and `loadMobility` with `useCallback`?

reidzeibel: Unfortunately, no, it still triggers multiple debounce due to location change if we put the dependency to the useEffect.

reidzeibel:
@marius-at-atb we are using `renderMode={UserLocationRenderMode.Native}`, can't seem to find `NativeUserLocation` on the RNMapbox, can you point out where it is?

marius-at-atb:
https://github.com/rnmapbox/maps/blob/main/docs/NativeUserLocation.md
Might require rnmapbox to be upgraded

reidzeibel:
@gorandalum @marius-at-atb please hold off reviewing this, I think i found something when trying to update the RNMapBox dependency version, will let you know after lunch the latest. 

@KristianSelnas

reidzeibel:
Closing this PR because there is a better solution: Updating the RNMapBox dependency, works on both Android and iOS normally.

toretefre:
I guess my review is not needed anymore as well. Hope dependency update works 🤞

PR #4143, opened 2023-12-19
fix: Show header in FullScreenView without parallax content

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/16103


hanne-at-atb (approved):
Arrriba! 🎉 🌮

marius-at-atb (approved):

PR #4144, opened 2023-12-20
fix: remove realtime position in details on cancelled departures

adriawh:
Now removes the realtime icon and the text with position if the route is cancelled. Also shows the warning seen below now.

### DIFF
--- a/src/place-screen/components/EstimatedCallList.tsx
+++ b/src/place-screen/components/EstimatedCallList.tsx
@@ -61,7 +61,7 @@ export const EstimatedCallList = ({
         },
         existingFavorite,
       ),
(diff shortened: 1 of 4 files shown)


src/place-screen/components/EstimatedCallList.tsx
```diff
@@ -61,7 +61,7 @@ export const EstimatedCallList = ({
         },
         existingFavorite,
       ),
-      // eslint-disable-next-line react-hooks/exhaustive-deps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
hanne-at-atb, line 64: 👀

hanne-at-atb (approved):
Good job! 💯

PR #4145, opened 2023-12-20
chore: update app name for staging and debug builds

reidzeibel:
On android app, there is no easy way to distinguish between app builds. All build types have same app name and app icon as the production build, which makes it difficult to distinguish between build types if we installed multiple build types on the same phone.

This PR will try to fix that by adding some build type suffix in parentheses, for the debug, staging, and prodstaging builds. The app names will be as follows: 

- Dev: `APP_NAME=AtB` --> `APP_NAME=AtB (Dev)`
- Staging : `APP_NAME=AtB` --> `APP_NAME=AtB (Staging)`
- ~~ProdStaging: `APP_NAME=AtB` --> `APP_NAME=AtB (ProdStaging)`~~ Reverted

This will especially help the testers when they work with multiple build types.


marius-at-atb:
Is it sufficient to rename dev and staging for debugging? If so maybe we don't have to revert the change later

reidzeibel:
@marius-at-atb ok I asked both Tor and Hanne and they don't know anything about ProdStaging so I'll revert that one. I'll ask Gøran about it when he's back to work.

reidzeibel:
@marius-at-atb reverted the change on ProdStaging

marius-at-atb (approved):

reidzeibel:
image

No more confusion finally!

HanneLH:
Tested ok ✅

PR #4147, opened 2023-12-20
fix: Update Beacons starting useEffect

jorelosorio:
While testing, the beacons were not properly started, so after refactoring this class a small piece was missing to keep it on after close the app.

Closes: https://github.com/AtB-AS/kundevendt/issues/8621

### DIFF
--- a/src/beacons/BeaconsContext.tsx
+++ b/src/beacons/BeaconsContext.tsx
@@ -136,14 +136,13 @@ const BeaconsContextProvider: React.FC = ({children}) => {
     }
   }, [isBeaconsSupported]);
 
-  const isOnboardedButNotStarted =
-    !kettleInfo?.isKettleStarted && kettleInfo?.isBeaconsOnboarded;
-
   useEffect(() => {
     (async function () {
       if (!isBeaconsSupported) return;
+      const consentGranted =
+        (await storage.get(storeKey.beaconsConsent)) ?? 'false';
 
-      if (isOnboardedButNotStarted) {
+      if (consentGranted === 'true') {
         const permissions = await initializeKettleSDK();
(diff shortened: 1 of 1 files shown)


src/beacons/BeaconsContext.tsx
```diff
@@ -136,14 +136,13 @@ const BeaconsContextProvider: React.FC = ({children}) => {
...
   }, [isBeaconsSupported]);
 
-  const isOnboardedButNotStarted =
-    !kettleInfo?.isKettleStarted && kettleInfo?.isBeaconsOnboarded;
-
   useEffect(() => {
     (async function () {
       if (!isBeaconsSupported) return;
+      const consentGranted =
+        (await storage.get(storeKey.beaconsConsent)) ?? 'false';
```
marius-at-atb, line 143: I think best practice is to use `parseBoolean`

src/beacons/BeaconsContext.tsx
```diff
@@ -175,12 +174,14 @@ const getKettleInfo = async (): Promise<KettleInfo> => {
   const status = await Kettle.isStarted();
   const identifier = await Kettle.getIdentifier();
   const consents = await Kettle.getGrantedConsents();
+  const consentGranted =
+    (await storage.get(storeKey.beaconsConsent)) ?? 'false';
```
marius-at-atb, line 178: Same as above regarding `parseBoolean`

marius-at-atb (approved):
Looks good!

gorandalum (commented):
By just looking at this I don't really grasp what "picks up" that the consent has been granted, since I don't really see what is triggering the retrieval from async storage. But that I don't see it doesn't mean it doesn't work, and it will be tested.

PR #4148, opened 2023-12-20
feat: Make announcement scroll horizontally when more than one 

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/15746

Announcements are stretched vertically to be come as tall as the tallest visible announcement.

In addition to the horizontal scroll, this pr also ensures that legacy announcement  are opened in bottom sheet.

The announcement close icon is moved inside the header. This leaves more room for the content, which is more important now, since the announcements will be narrower.

<details>
<summary>Screen shots</summary>

Simulator Screenshot - iPhone 15 - 2023-12-20 at 12 55 26

Simulator Screenshot - iPhone 15 - 2023-12-20 at 12 55 38

</details>

### DIFF
--- a/src/components/sections/items/GenericSectionItem.tsx
+++ b/src/components/sections/items/GenericSectionItem.tsx
@@ -1,5 +1,5 @@
 import React, {Children, PropsWithChildren} from 'react';
-import {AccessibilityProps, View} from 'react-native';
+import {AccessibilityProps, StyleProp, View, ViewStyle} from 'react-native';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
 import {useSectionStyle} from '../use-section-style';
@@ -7,6 +7,7 @@ import {useSectionStyle} from '../use-section-style';
 type Props = PropsWithChildren<
   SectionItemProps<{
     accessibility?: AccessibilityProps;
+    style?: StyleProp<ViewStyle>;
   }>
 >;
 export function GenericSectionItem({children, accessibility, ...props}: Props) {
@@ -14,7 +15,7 @@ export function GenericSectionItem({children, accessibility, ...props}: Props) {
   const style = useSectionStyle();
 
   return (
-    <View style={topContainer} {...accessibility}>
+    <View style={[topContainer, props.style]} {...accessibility}>
       {Children.map(children, (child) => (
         <View style={style.spaceBetween}>{child}</View>
       ))}
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
@@ -45,6 +45,11 @@ export const Announcement = ({announcement, style}: Props) => {
     });
   };
 
+  const summaryTitle = getTextForLanguage(
+    announcement.summaryTitle ?? announcement.fullTitle,
+    language,
+  );
+
   return (
     <Section style={style} key={announcement.id} testID="announcement">
       <GenericSectionItem>
@@ -69,12 +74,7 @@ export const Announcement = ({announcement, style}: Props) => {
               </View>
             )}
             <View style={styles.textContainer}>
-              <ThemeText type="body__primary--bold">
-                {getTextForLanguage(
-                  announcement.summaryTitle ?? announcement.fullTitle,
-                  language,
-                )}
-              </ThemeText>
+              <ThemeText type="body__primary--bold">{summaryTitle}</ThemeText>
               <ThemeText style={styles.summary}>
                 {getTextForLanguage(announcement.summary, language)}
               </ThemeText>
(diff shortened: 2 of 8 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
```diff
@@ -45,91 +45,95 @@ export const Announcement = ({announcement, style}: Props) => {
...
-              announcement.actionButton?.actionType === 'external'
-                ? 'external-link'
-                : 'arrow-right'
-            }
-            accessibility={{
-              accessibilityHint: t(
-                DashboardTexts.announcemens.buttonAction[
-                  announcement.actionButton.actionType
-                ],
+      {announcement.actionButton?.actionType && (
```
gorandalum, line 88: With the fallback to `bottom_sheet` in the converter, the `actionType` doesn't need to be optional in the typing now?

src/components/sections/items/GenericSectionItem.tsx
```diff
@@ -1,20 +1,21 @@
...
-import {AccessibilityProps, View} from 'react-native';
+import {AccessibilityProps, StyleProp, View, ViewStyle} from 'react-native';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
 import {useSectionStyle} from '../use-section-style';
 
 type Props = PropsWithChildren<
   SectionItemProps<{
     accessibility?: AccessibilityProps;
+    style?: StyleProp<ViewStyle>;
```
gorandalum, line 10: I feel like it is a little can of worms to expose the internal `style` of a section item. Could consider using a more specific field, like `shouldGrow: boolean` or something here.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
```diff
@@ -36,23 +36,34 @@ export const Announcements = ({style: containerStyle}: Props) => {
...
-      marginBottom: theme.spacings.medium,
-    },
-  };
-});
+const useStyle = StyleSheet.createThemeHook((theme) => ({
+  headerWrapper: {
+    marginHorizontal: theme.spacings.medium,
+  },
+  scrollView: {
+    marginHorizontal: theme.spacings.medium,
```
gorandalum, line 63: There is something iffy about using margins on `ScrollView`. And when testing this I see there is an issue with a missing right margin
Screenshot 2023-12-21 at 11 36 30

My tip is to try using `gap` and `paddingHorizontal` on the `ScrollView`, and remove the margin on the announcement.

gorandalum (approved):
Nice!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
KristianSelnas: Thanks. I've noticed myself. I'll try your solution

src/components/sections/items/GenericSectionItem.tsx
KristianSelnas: Yes, I was considering something like that myself, but wasn't sure what I felt was the best option. I believe a prop like `shouldGrow` is kinda similar to `withPadding`, in that you use a prop to dictate appearance. In that sense i like the `style` approach better. But with increased freedom also comes increased responsibility. It needs to be used with caution.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcement.tsx
KristianSelnas: Yes, I suppose that's true 👍

KristianSelnas: Fixed in https://github.com/AtB-AS/mittatb-app/pull/4148/commits/f24e8f6902468c0ca282d03c8d8573277df608eb

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
KristianSelnas: Fixed in https://github.com/AtB-AS/mittatb-app/pull/4148/commits/94bd79b35b70792058587fb648a613edbd4fe01b

src/components/sections/items/GenericSectionItem.tsx
KristianSelnas: @gorandalum What do you think?

gorandalum: Not meaning to delay the merge on such minor details 😅 

My preference still goes towards `shouldGrow`. The reason for moving away from `withPadding` and such was that we saw that they were too limiting, as `Section` can't foresee every usage from the ouside (thus over time also adding `withFullPadding`, `withTopPadding`, `withBottomPadding`). In this case it is internally in the `Section`, and then we generally want to be as limiting as possible.

KristianSelnas: If we come back to this prop in a year I think it will be a bit hard to grasp what `shouldGrow` is actually doing. As you say, to not stall this pr any longer I'll just merge it with the `style` prop. Personally I don't think that will case any major issues. Most devs in our are reasonably responsible, and the code review process should probably pick up the rest.

PR #4149, opened 2023-12-20
feat: add screen reader support for disabled ToggleSectionItems

rosvik:
ref. QA feedback https://github.com/AtB-AS/kundevendt/issues/15351#issuecomment-1857972212

- Adds a hint to ToggleSectionItem when toggle is disabled. (`accessibilityState` with disabled=true wasn't picked up on iOS)
- Prevents toggles from being toggleable with screan reader when they are disabled.

closes https://github.com/AtB-AS/kundevendt/issues/15351

### DIFF
--- a/src/components/sections/items/ToggleSectionItem.tsx
+++ b/src/components/sections/items/ToggleSectionItem.tsx
@@ -10,6 +10,7 @@ import {InteractiveColor, TextNames} from '@atb/theme/colors';
 import {LabelType} from '@atb/configuration';
 import {LabelInfo} from '@atb/components/label-info';
 import {SvgProps} from 'react-native-svg';
+import {SectionTexts, useTranslation} from '@atb/translations';
 
 type Props = SectionItemProps<{
   text: string;
@@ -40,6 +41,7 @@ export function ToggleSectionItem({
   const {topContainer} = useSectionItem(props);
   const sectionStyle = useSectionStyle();
   const styles = useStyles();
+  const {t} = useTranslation();
 
   /*
    Need to maintain a internal copy of the checked state for the accessibility
@@ -52,6 +54,7 @@ export function ToggleSectionItem({
   const [checked, setChecked] = useState(value);
   useEffect(() => setChecked(value), [value]);
   const onChange = (v: boolean) => {
+    if (disabled) return;
     setChecked(v);
     setTimeout(() => onValueChange(v), 0);
   };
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):
Lgtm!

src/components/sections/items/ToggleSectionItem.tsx
```diff
@@ -61,13 +64,18 @@ export function ToggleSectionItem({
       style={topContainer}
       accessible={true}
       accessibilityRole="switch"
-      accessibilityState={{checked: checked}}
+      accessibilityState={{checked: checked, disabled: disabled}}
       accessibilityActions={[{name: 'activate'}]}
       onAccessibilityAction={() => onChange(!checked)}
+      accessibilityHint={
+        disabled ? t(SectionTexts.toggleInput.disabled) : undefined
+      }
```
gorandalum, line 72: Should not be necessary to add this disabled hint. The accessibility state `disabled` should add the phrase "dimmed"/"nedtonet" to signal the status (if it doesn't, something is changed in recent RN versions).

We want to lean on the iOS/Android handling of disabled states rather than roll our own. Also we asked in a user test whether the phrase "nedtonet" was understood:
https://mittatb.slack.com/archives/CSCD9JWGP/p1622190495025200

src/components/sections/items/ToggleSectionItem.tsx
```diff
@@ -61,13 +64,18 @@ export function ToggleSectionItem({
       style={topContainer}
       accessible={true}
       accessibilityRole="switch"
-      accessibilityState={{checked: checked}}
+      accessibilityState={{checked: checked, disabled: disabled}}
       accessibilityActions={[{name: 'activate'}]}
       onAccessibilityAction={() => onChange(!checked)}
```
gorandalum, line 69: Would maybe handle the `disabled` flag here instead:
`onAccessibilityAction={disabled ? undefined : () => onChange(!checked)}`

gorandalum (commented):
Some minor comments, not necessary to create a new PR to fix. The comment regarding disabled/"nedtonet" is mostly to be aware of it later.

src/components/sections/items/ToggleSectionItem.tsx
rosvik: Aha, I didn't realize voiceover said "dimmed", so I thought the field wasn't read out at all 🙈

PR #4150, opened 2023-12-20
feat: new bottomSheet header

adriawh:
# New bottomSheet header
The `bottomSheet` now has its own custom header. This header always has a close button, and the button is now an icon instead of text. 

- The title for the `bottomSheet` is set at the `bottomSheetContainer`
- An optional onClose function has been added 
- ScreenReader has been added for the close button and title

<details>
<summary>Images</summary>

</details>

### Edge case
The emojiSheet for the favorites originally had a clear emoji button. This functionality has been pulled out of the bottom sheet and placed on the side of the emoji instead.

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

https://github.com/AtB-AS/mittatb-app/assets/70323886/177b90e5-8f43-4a38-bbb6-919bd034ac67

</details>

### Additional changes:
Removed the "lukk" button in the error message for bikes, scooters etc. after a talk with @oyvindmikkelsenatb. It was necessary to do changes to this code anyways. 

Below is a before picture. The "Lukk" link is now removed


src/components/bottom-sheet/BottomSheetHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+  // Calculate hitSlop values
+  const desiredHitboxSize = 48;
+  const hitSlopSize = (desiredHitboxSize - buttonSize.height) / 2;
+
+  const hitSlop = {
+    top: hitSlopSize,
+    bottom: hitSlopSize,
+    left: hitSlopSize,
+    right: hitSlopSize,
+  };
```
gorandalum, line 52: Shouldn't overthink this hitslop stuff, it is not worth the effort. Just remove it all and use `hitSlop={insets.all(10)}` on the `TouchableOpacity`.

src/components/bottom-sheet/BottomSheetContainer.tsx
```diff
@@ -1,24 +1,33 @@
...
 import {useWindowDimensions, View} from 'react-native';
+import {BottomSheetHeader} from '@atb/components/bottom-sheet';
 
 export type BottomSheetContainerProps = {
   children: ReactNode;
   maxHeightValue?: number;
   fullHeight?: boolean;
   testID?: string;
+  bottomSheetTitle?: string;
+  closeBottomSheet?: () => void;
```
gorandalum, line 11: Just call them `title` and `close`, it is self-explanatory that the params are related to bottom sheet since it is set on the `BottomSheetContainer`.

src/components/bottom-sheet/BottomSheetHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+  title,
+  closeBottomSheet,
+}: BottomSheetHeaderWithoutNavigationProps) => {
+  const styles = useStyle();
+  const {t} = useTranslation();
+
+  const {theme} = useTheme();
+  const {close: closeBottomSheetDefault, onOpenFocusRef} = useBottomSheet();
+
+  const handleClose = closeBottomSheet ?? closeBottomSheetDefault;
```
gorandalum, line 31: I think we can just use the `close` function from the `useBottomSheet` here, not maing it possible to override from the outside? I think the close button should have no side-effects, it should be the same as closing by pressing the backdrop?

gorandalum (changes requested):
Nice work! Some comments to simplify some stuff.

src/components/bottom-sheet/BottomSheetHeader.tsx
adriawh: Yes I agree, however the `useUpdateBottomSheetWhenSelectedEntityChanges` has a function that needs to run when any of the bottom sheets in the map is to be closed. I could try to change it to a onClose instead, so there can be added code to be ran when the close button is clicked.

gorandalum (approved):

PR #4155, opened 2023-12-21
fix: change default periodic ticket duration

adriawh:
Setting default ticket to 30 day ticket for atb.

The issue mentions to talk with NFK and Fram, however their default is already the 30day ticket and as the solution is done in this specific way of identifying the ticket I didn't see it necessary to contact them.


gorandalum (changes requested):
Not allowed to reference any product ids in the app code. It should be agnostic to the internal details of our products.

We either need to do the same as Fram, change the product order so the 30 days is first.

Or we need to add the possibility to set which product is default in Firestore, as we have on tariff zones.

PR #4157, opened 2023-12-21
fix: use local status in notification toggle values

rosvik:
- If the user haven't given permission to receive notifications, the toggle value is always false.
- If the user doesn't have an email registered on their profile, the toggle value is always false.

might fix some of the issues in https://mittatb.slack.com/archives/C0642RNB86L/p1703160462775469


hanne-at-atb (approved):
Looks good :)

tormoseng:
Tested OK on iOS.

On Android (v.10) I'm not able to test it due to this observation. Repeated here
```
On my Android, notifications are allowed by default when opening a freshly installed app (no other AtB-apps on the device). 
- Remove all AtB-apps on the device
- Install latest master
- Walk through normal onboarding
- Check permissions on the AtB (staging) app in settings
```

HanneLH:
Tested ok on Android (V.13) ✅

PR #4160, opened 2023-12-22
fix: Show announcements vertically on screen reader

KristianSelnas:
Found during testing of https://github.com/AtB-AS/kundevendt/issues/15746

Screen reader on iOS reads horisontal content horizontally, which means all the titles will be read as one "row", then all the descriptions as the next row, etc.

The easies solution to fix this is to align announcements vertically for screen readers since they have to scroll through them anyways.

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
@@ -9,6 +9,7 @@ import {useNow} from '@atb/utils/use-now';
 import {StyleSheet} from '@atb/theme';
 import {useHasSeenShareTravelHabitsScreen} from '@atb/beacons/use-has-seen-share-travel-habits-screen';
 import {useBeaconsState} from '@atb/beacons/BeaconsContext';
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
```diff
@@ -62,6 +67,7 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
   scrollView: {
     paddingHorizontal: theme.spacings.medium,
     columnGap: theme.spacings.medium,
+    rowGap: theme.spacings.medium,
```
gorandalum, line 70: Possible to just use `gap`, it should work both ways.

gorandalum (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/components/Announcements.tsx
KristianSelnas: I saw it when I created the pr, and kinda closed my eyes and hoping you woundn't notice 🙈 😆

PR #4163, opened 2023-12-27
feat: Use <FullScreenView> with parallax scroll header on Profile screens

KristianSelnas:
Closes https://github.com/AtB-AS/kundevendt/issues/16180

[Relevant Figma sketches](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=18886-7579&mode=design&t=W7XjcszLkTyO0vNb-0)

This PR does includes the use of `<FullScreenView>` on `Profile_RootScreen`, but that does make this screen look according to the Figma sketches. The FullScreenView currently doesn't support the case where the base header is initially hidden and only the parallax content is shown (which is how I interpret the sketches to be). I think this case should be handled in a separate PR, possibly after a discussion with the designers about how this should actually work, i.e. should there be some kind of parallax effect in this case? If so, what should happen?


gorandalum (changes requested):
Looks generally good, but remember that everything focusable needs `accessible` on Android to not crash.

KristianSelnas:
I wasn't aware of that. What does that mean in practice? That the `ScreenHeader` component needs `accessible` on it's wrapper view element (the one with `ref`)?

gorandalum:
Yeah. If not accessible, then a react tag can't be found to give focus to.
https://github.com/facebook/react-native/issues/22807

KristianSelnas:
Good catch there @gorandalum . I've fixed this in https://github.com/AtB-AS/mittatb-app/pull/4163/commits/c8d713d7a4f012161c7e374be14b8ea0fabb87ec

Verified both the crash and the fix on a physical Android

gorandalum (approved):

PR #4165, opened 2023-12-27
fix: Remove exhaustive deps errors in mobility

KristianSelnas:
Fixes exhaustive deps errors in mobility. 

Partly closes https://github.com/AtB-AS/kundevendt/issues/15526

### DIFF
--- a/src/mobility/components/OperatorActionButton.tsx
+++ b/src/mobility/components/OperatorActionButton.tsx
@@ -1,7 +1,7 @@
 import {useAnalytics} from '@atb/analytics';
 import {getTextForLanguage, useTranslation} from '@atb/translations';
 import {useAppMissingAlert} from '@atb/mobility/use-app-missing-alert';
-import React, {useCallback} from 'react';
+import React from 'react';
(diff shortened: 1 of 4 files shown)


src/mobility/components/OperatorActionButton.tsx
```diff
@@ -61,8 +61,16 @@ export const OperatorActionButton = ({
...
+  }, [
+    analytics,
+    operatorName,
+    benefit,
+    isUserEligibleForBenefit,
+    rentalAppUri,
+    valueCode,
+    showAppMissingAlert,
+    appStoreUri,
+  ]);
```
gorandalum, line 73: Is this `handleCallToAction` just an event handler? In that case is doesn't need `useCallback` at all.

gorandalum (approved):
Looks good! Can probably do a quick check with network logger to see that no unwanted side effects regarding the reloading of vehicles.

KristianSelnas:
Hmm.... Yes, there are indeed consequences of adding the `reload` to deps.  Not sure what the best solution is though.

PR #4167, opened 2023-12-27
feat: Check if token should be renewed

gorandalum:
Every 6 hours, check if token needs to be renewed.

Checks by using the native token validity end. Won't catch changes
to the validity done online. To catch that a validation (on app
start) is necessary.

The original plan was to validate with a given interval, but as first 
measure we try this instead, which is less intruding and doesn't
need network requests.


reidzeibel (approved):
Looks good to me, also the comments added are nice to give more context to other developers!

PR #4168, opened 2023-12-27
chore: remove Entur SDK

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/16187

This PR will : 
- Move some types/enums/interfaces to `sdk.ts` as temporary place before further refactoring
- Remove entur sdk dependency from `package.json`
- Remove some unused types from `sdk.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "@bugsnag/source-maps": "^2.3.1",
     "@entur-private/abt-mobile-barcode-javascript-lib": "1.3.4",
     "@entur-private/abt-mobile-client-sdk": "1.3.4",
-    "@entur/sdk": "^5.3.0",
     "@leile/lobo-t": "^1.0.5",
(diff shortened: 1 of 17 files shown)


gorandalum (approved):
Looks good!

To be more finished with this cleanup we should replace the types in `sdk.ts` with types from `journey_planner_v3_types.ts` or with GQL Fragments copied from bff. If this isn't done in this issue, then maybe create a new one for that part of the cleanup?

reidzeibel:
@gorandalum I'll try replacing them with the types from `journey_planner_v3_types.ts` that is already available. Will notify after finished

reidzeibel:
@gorandalum I made some more changes on sdk.ts, removed some more types & replaced some with `journey_planner_v3.ts` types I think for other types we should do on separate PR because it will need more files to be changed.

gorandalum (approved):

PR #4174, opened 2023-12-29
fix: Bottom sheet focused by screen reader

gorandalum:
Changed implementation of use-focus-on-load to use an ref callback
instead, which will be called every time the component is mounted
and unmounted.

Also the InteractionsManager.runAfterInteractions were not always
enough, so now also a setTimeout of 100ms was added before
focusing the element.


reidzeibel (approved):

reidzeibel:
@gorandalum I did a trace and the crashes started on this PR https://github.com/AtB-AS/kundevendt/issues/16389

PR #4178, opened 2024-01-02
fix: Atb bestill button text placement with larger text

adriawh:
### Text placement on AtB bestill button

> "... has reported a UI error on the button in the travel search when AtB Bestill comes up. We should fix this so that the text is closer to the symbol and make the buttons more similar. The point is the space between the icon and the text on the phone."

| Android Before                                      | Android After                                       |
|-----------------------------------------------------|-----------------------------------------------------|
|  |  |
| iOS Before                                          | iOS After                                           |
|  |  |

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -145,7 +145,7 @@ export const Button = React.forwardRef<any, ButtonProps>(
       flex: isInline ? undefined : 1,
       alignItems: 'center',
       marginHorizontal: textMarginHorizontal,
-      flexShrink: 1,
+      flexShrink: isInline ? 1 : undefined,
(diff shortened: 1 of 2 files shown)


reidzeibel (approved):

gorandalum (changes requested):
Implementation looks good, but there are no examples of left-oriented text in the Figma designs for pill button. That should be there if we have it in the app.

https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components?type=design&node-id=3912%3A12606&mode=design&t=lJ5dpxA0vVRnXG6d-1

gorandalum (approved):

PR #4179, opened 2024-01-02
fix: use correct color for icons in ToggleSectionItem

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/16243

Should be pretty straightforward, I pulled out the `getFill` function from `ThemeIcon.tsx` and put it into `src/theme/colors.ts`, and then added the fill to the `LeftImage` svg inside `ToggleSectionItem`.

Any theme-affected images/icons will switch colors based on the theme, while all static illustrations/images will keep their color.

This will also fix the indentation issue mentioned in slack https://mittatb.slack.com/archives/C0116FMPX4Y/p1704187943688439

### Acceptance Criteria

- [x] Correct icons (light colors) on Travel search filter on dark mode.
- [x] Correct icons (light colors) on Form factor filter ("map" bottom navigation) on dark mode.
- [x] Correct image on "on behalf of" bottom sheet traveller selection. (no color change should happen).

### DIFF
--- a/src/components/sections/items/ToggleSectionItem.tsx
+++ b/src/components/sections/items/ToggleSectionItem.tsx
@@ -1,12 +1,12 @@
 import React, {useEffect, useState} from 'react';
 import {AccessibilityProps, View} from 'react-native';
-import {StyleSheet, Theme} from '@atb/theme';
+import {StyleSheet, Theme, useTheme} from '@atb/theme';
 import {ThemeText} from '@atb/components/text';
 import {useSectionItem} from '../use-section-item';
 import {SectionItemProps} from '../types';
 import {useSectionStyle} from '../use-section-style';
 import {Toggle} from '@atb/components/toggle';
-import {InteractiveColor, TextNames} from '@atb/theme/colors';
+import {InteractiveColor, TextNames, getFill} from '@atb/theme/colors';
 import {LabelType} from '@atb/configuration';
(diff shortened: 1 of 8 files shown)


gorandalum (changes requested):
A couple comments:
- We have tried to encapsulate as much as possible of the theming logic inside specific components like `ThemeIcon`, `ThemeText` and such. It breaks that encapsulation when we do theme-specific logic inside a specific section item.
- This always applies fill on all left images, but I'm not completely sure if we always want to do it. It is only mono-icons which are meant to be applied fill to.

Could we look into a solution where we instead send in a `ThemeIcon` as the left image in the situations where we want it themed?

reidzeibel:
> A couple comments:
> 
>     * We have tried to encapsulate as much as possible of the theming logic inside specific components like `ThemeIcon`, `ThemeText` and such. It breaks that encapsulation when we do theme-specific logic inside a specific section item.
> 
>     * This always applies fill on all left images, but I'm not completely sure if we always want to do it. It is only mono-icons which are meant to be applied fill to.
> 
> 
> Could we look into a solution where we instead send in a `ThemeIcon` as the left image in the situations where we want it themed?

Sure, let me look into it! @gorandalum

gorandalum (approved):
Great! Remember to add the places where `ToggleSectionItem` is used to the acceptance criteria.

PR #4180, opened 2024-01-02
fix: Not use expired recurring payment card

gorandalum:
Expired payment card was used for purchase even after expiration,
which ended up in error for the user, and the error message "The
customer has no active recurring payment with given id" from
Entur/Nets.


reidzeibel (approved):

rosvik (approved):
Makes sense for the PurchaseConfirmationScreen, but do we show information about the card being expired somewhere else in the app? I wouldn't expect a payment card to just "disappear" from the app once it expires.

gorandalum:
> Makes sense for the PurchaseConfirmationScreen, but do we show information about the card being expired somewhere else in the app? I wouldn't expect a payment card to just "disappear" from the app once it expires.

Yeah, they just disappear now. The endpoint in the ticket service for retrieving stored payment cards doesn't return expired cards, so they don't show up in the "Payment options" under "My profile" either.

Although it would be better with a more elegant handling, I won't start going down that path in this bugfix-issue.

HanneLH:
Verifisert i staging nå som januar er over ✅ 🎉

PR #4183, opened 2024-01-02
fix: notification onboarding logic

adriawh:
### Notification onboarding logic fix
This PR updates the logic for displaying the notification onboarding. With these changes:
- The onboarding will only be shown to users who have a ticket with notifications set to default enabled.
- If the user has already enabled notifications through their settings, the onboarding will not be displayed.

### DIFF
--- a/src/notifications/utils.ts
+++ b/src/notifications/utils.ts
@@ -1,9 +1,38 @@
-import {NotificationConfigValue} from './types';
+import {NotificationConfig, NotificationConfigValue} from './types';
 import {isDefined} from '@atb/utils/presence';
+import {
+  findReferenceDataById,
+  PreassignedFareProduct,
+} from '@atb/configuration';
+import {FareContract} from '@atb/ticketing';
 
 export function isConfigEnabled<T extends NotificationConfigValue>(
   config: T[] | undefined,
   key: T['id'],
 ): boolean {
   return config?.filter(isDefined).find((v) => v.id === key)?.enabled ?? false;
+}
+
+export function hasValidFareContractWithActivatedNotification(
+  validFareContracts: FareContract[],
+  preassignedFareProducts: PreassignedFareProduct[],
+  config?: NotificationConfig,
+): boolean {
+  if (!config) return false;
+  return validFareContracts.some((fareContract) => {
+    const fareProductRef = fareContract.travelRights[0]?.fareProductRef;
+
+    if (!fareProductRef) {
+      return false;
+    }
+
+    const preassignedFareProduct = findReferenceDataById(
+      preassignedFareProducts,
+      fareProductRef,
+    );
+
+    return config?.groups.some(
+      (group) => group.id === preassignedFareProduct?.type && group.enabled,
+    );
+  });
 }
--- a/src/notifications/utils.ts
+++ b/src/notifications/utils.ts
@@ -14,12 +14,12 @@ export function isConfigEnabled<T extends NotificationConfigValue>(
 }
 
 export function hasFareContractWithActivatedNotification(
-  validFareContracts: FareContract[],
+  fareContracts: FareContract[],
   preassignedFareProducts: PreassignedFareProduct[],
   config?: NotificationConfig,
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -99,12 +104,21 @@ export const Root_TabNavigatorStack = ({navigation}: Props) => {
         );
       }
     }
+
+    const notificationsNotGranted =
+      pushNotificationPermissionStatus == 'undetermined' ||
+      pushNotificationPermissionStatus == 'denied';
```
rosvik, line 110: Do you know how this will work on Android? From what I understand, `'denied'` can mean both "we haven't asked the user yet" and "the user have actively denied permissions". Which might lead to us showing the onboarding more times that strictly necessary? But if that's true, that might be ok?

adriawh: I did read up about it quickly and tested it, but will double check it

adriawh: the undetermined was not necessary and the [denied handles it for both iOS and android ](https://github.com/zoontek/react-native-permissions#understanding-permission-flow)

src/notifications/utils.ts
```diff
@@ -1,9 +1,38 @@
...
   return config?.filter(isDefined).find((v) => v.id === key)?.enabled ?? false;
 }
+
+export function hasFareContractWithActivatedNotification(
+  validFareContracts: FareContract[],
+  preassignedFareProducts: PreassignedFareProduct[],
+  config?: NotificationConfig,
+): boolean {
+  if (!config) return false;
+  return validFareContracts.some((fareContract) => {
```
rosvik, line 22:
```suggestion
  fareContracts: FareContract[],
  preassignedFareProducts: PreassignedFareProduct[],
  config?: NotificationConfig,
): boolean {
  if (!config) return false;
  return fareContracts.some((fareContract) => {
```

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
rosvik: I don't think removing 'undetermined' makes sense here, since that is the initial state on iOS after a new install. Maybe a good idea to have a flow chart like that for the app, but I don't think this one is correct for notifications, since we use Firebase Messaging instead of react-native-permissions. 

If I understand the Android behavior correctly, I think this should look like:

```
Platform.OS === 'ios' 
	? pushNotificationPermissionStatus === 'undetermined'
	: pushNotificationPermissionStatus === 'denied'
```

adriawh: This clarifies, I for some reason believed it was handled by the same permissions library. Will look into it, and look at the right place now

adriawh:
You were correct with the permissions @rosvik , have tested it and seems to be working as it should. Tested buying single ticket and not getting onboarded, then logging in to user with ticket and getting onboarded. And tested with turning on notifications in settings and then getting ticket that can get notifications, and not getting onboarded.

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -99,12 +104,21 @@ export const Root_TabNavigatorStack = ({navigation}: Props) => {
         );
       }
     }
+
+    const notificationsNotGranted =
```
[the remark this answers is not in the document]
rosvik: Now that we have this const again, I think we should update the name 👆

rosvik (approved):
Good work 🎉 
Just one remaining nit 😄

PR #4184, opened 2024-01-02
refactor: Add exhaustive deps rule to useInterval

gorandalum:
You get the power of the exhaustive deps eslint rule on your
custom hooks by adding them in the 'additionalHooks' argument
in the rule options.

To make this work the first argument of the custom hook needs to
be the function, while the second argument needs to be the deps.

Added this to the 'useInterval' hook and fixed the errors that
showed up.


rosvik (approved):
Nice!

PR #4186, opened 2024-01-03
refactor: Fix gap on payment options screen

gorandalum:
Remove margins from message boxes as there is a gap set on the
container level.

Before/After (constructed example with all message boxes set to visible):
<div>

</div>


HanneLH (approved):
Looks good for me 🌸

marius-at-atb (approved):

PR #4187, opened 2024-01-03
fix: bottom sheet close button color

adriawh:
@ckbilstad discovered that we used the wrong interactive color on the[ button on the bottom-sheet.](https://www.figma.com/file/2QTjAdekdIPuLFovQhVrY3/Components?type=design&node-id=4573-10756&mode=dev)

| Before | After |
| ------ | ----- |
|  |  |


rosvik (approved):

hanne-at-atb (approved):
👍🏼 🙂

PR #4190, opened 2024-01-03
fix: refetch notification config on auth user change

rosvik:
This fixes a bug where stale notification config data would stay after the auth user had changed. By providing userId as part of the useQuery key, this is automagically refetched when the userId changes.

### Acceptance criteria

- [ ] Reproduce the bug / verify the fix:
	1. While logged in, go to My profile -> Notifications
	2. Enable notifications from some tickets other that the defaults
	3. Log out
	4. Open the Notifications page again,
		- Before this fix, the settings would show as the same as for the logged in user, even though this is a new anonymous user with default configs. Once you toggle something, the default configs will be loaded.
		- After this fix, the default configs should be shown.

### DIFF
--- a/src/notifications/use-config.tsx
+++ b/src/notifications/use-config.tsx
@@ -1,16 +1,19 @@
 import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query';
 import {getConfig, NotificationConfigUpdate, updateConfig} from './api';
+import {useAuthState} from '@atb/auth';
 
 export const useConfig = () => {
   const queryClient = useQueryClient();
+  const {userId} = useAuthState();
 
+  const queryKey = ['notification/config', userId];
   const query = useQuery({
-    queryKey: ['notification/config'],
+    queryKey,
     queryFn: getConfig,
   });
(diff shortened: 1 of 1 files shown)


hanne-at-atb (approved):
👏🏼

gorandalum (commented):
Aha. The user id/id token is actually a dependency of all queries where `authWithIdToken` is set to `true`, but of course it won't be caught by the exhaustive deps rule since the id token is injected in the axios client.

I have a feeling this will be a pitfall going forward, and that we should think of a solution to not introduce bugs at a later time. One catch-all solution is to invalidate all queries on user change, maybe that is the safest way to go?

rosvik:
@gorandalum Good point, it probably should be solved in another way. But I couldn't find an obvious way to solve it right now.

If you think this looks fine for now, I'll merge this to the release branch, and make an issue for invalidating queries on auth change.

gorandalum (approved):

HanneLH:
Verified ok on Android ✅

HanneLH:
Verified ok on iOS ✅

PR #4198, opened 2024-01-04
feat: onboarding refactor animations

marius-at-atb:
This PR adds custom animations for onboarding transitions, and customizes some back/close buttons.


src/anonymous-purchase-consequences-screen/AnonymousPurchaseConsequencesScreenComponent.tsx
```diff
@@ -56,18 +59,25 @@ export const AnonymousPurchaseConsequencesScreenComponent = ({
...
+                <ThemeIcon
+                  svg={consequencesIcons[i]}
+                  colorType={themeColor}
+                  size="large"
+                />
+              }
+            />
+          ),
+        )
+      }
```
rosvik, line 80: I appreciate the drive to get rid of hard coded content, but I feel like this maybe is a bit overkill? Its a bit harder to reason about, plus it breaks if we add more than 3 messages, since there's only 3 items in `consequencesIcons`.

What do you think?

rosvik (commented):
Nice! One comment for the consequence screen.

Also, not introduced in this PR, but I noticed that the LoginOptionsScreen is missing a Back/Cancel button when logging in from my profile and ticketing. Does it makes sense to fix that in this PR as well?

src/anonymous-purchase-consequences-screen/AnonymousPurchaseConsequencesScreenComponent.tsx
marius-at-atb: Yea might be overkill in this case. I guess we don't expect it to be a long list anyway, so I'll revert it.

src/stacks-hierarchy/Root_LoginPhoneInputScreen.tsx
```diff
@@ -29,7 +31,12 @@ export const Root_LoginPhoneInputScreen = ({
   return (
     <View style={styles.container}>
       <FullScreenHeader
-        leftButton={{type: 'close'}}
+        leftButton={{
+          type:
+            transitionPreset === TransitionPresets.ModalSlideFromBottomIOS
+              ? 'close'
+              : 'back',
+        }}
```
gorandalum, line 39: I think this is a little stretch, that we change what is rendered to the user based on the transition preset used. I would rather be more precise on it, and have its own argument in the params for the back/close button logic.

in the params for the `Root_PurchaseConfirmationButton` we have the `headerLeftButton` as a prop for this.

gorandalum (changes requested):
Nice to have this! Just one comment about a non-intuitive usage of the transition preset.

PR #4200, opened 2024-01-04
fix: Update Beacons context to handle better status updates

jorelosorio:
This PR handles a case where the `KettleInfo` state changed due to a revoke, onboard but the `useEffect` was not aware of it, so it did not trigger start again the beacons, so `kettleInfo` had to be added in the deps list in order to know if it is running or not and depending on changes, and be able to start it again.

### DIFF
--- a/src/beacons/BeaconsContext.tsx
+++ b/src/beacons/BeaconsContext.tsx
@@ -75,14 +75,13 @@ const BeaconsContextProvider: React.FC = ({children}) => {
   const updateKettleInfo = () => getKettleInfo().then(setKettleInfo);
 
   const initializeKettleSDK = useCallback(async () => {
-    const permissions = await allowedPermissionForKettle();
     if (!isInitializedRef.current) {
+      const permissions = await allowedPermissionForKettle();
       if (permissions.length > 0) {
         await NativeModules.KettleSDKExtension.initializeKettleSDK();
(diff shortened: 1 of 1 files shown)


tormoseng:
🟢  Tested to turn on beacons through the privacy page: Beacons status was then `Running`

🔴  Can't seem to get announcements to work after seeing the onboarding and allowing all permissions. The status is `Running`. Handle in separate issue https://github.com/AtB-AS/kundevendt/issues/16286

🔴  Pt. 1 in this [issue](https://github.com/AtB-AS/kundevendt/issues/16140) is still not handled. Handle in separate issue https://github.com/AtB-AS/kundevendt/issues/16285

marius-at-atb (approved):
Lgtm!

gorandalum (commented):
The PR comment mentions something about revoking, but I don't really understand that part by looking at the code. The user has revoked a consent, and that should trigger something?

jorelosorio:
> The PR comment mentions something about revoking, but I don't really understand that part by looking at the code. The user has revoked a consent, and that should trigger something?

I mentioned revoking as an action that needs to trigger an update but when the user tried onboard again, nothing have changed after revoking, therefore the start never happened. I did not change something on revoking, but they way how it reacts after onboarding again.

tormoseng:
👍

PR #4201, opened 2024-01-04
fix: remove duplicate enum declaration

reidzeibel:
Fixes issue from this comment : https://github.com/AtB-AS/kundevendt/issues/16187#issuecomment-1875087306

Remove duplicate declaration of enum `TransportSubmode` causing issue with Rail replacement bus message. 

Due to duplicate enum, the rail replacement bus message component took the wrong import and can't find the correct enum, after removing the duplicate entry, the rail replacement bus appear as intended.


gorandalum (approved):
Good catch.

Could it also be possible to just use the value in the journey planner v3 types file? That is the one that most closely represents the types we can get from our bff endpoints.

src/sdk.ts
```diff
@@ -108,45 +108,6 @@ export declare type Feature = {
...
-  REGIONAL_RAIL = "regionalRail",
-  SCHOOL_BUS = "schoolBus",
-  SHUTTLE_BUS = "shuttleBus",
-  SIGHTSEEING_BUS = "sightseeingBus",
-  SIGHTSEEING_SERVICE = "sightseeingService",
-  TELECABIN = "telecabin",
-  TOURIST_RAILWAY = "touristRailway"
-}
-
 export declare enum TransportSubmode {
```
rosvik, line 111: Huh, you made me look up the documentation for [`declare`](https://github.com/Microsoft/TypeScript/blob/v1.8.10/doc/spec.md#11-ambient-declarations) here, but I guess this is a remnant of a .d.ts file, that probably should be removed. Same with `Feature` on [line 82](https://github.com/AtB-AS/mittatb-app/blob/ridwan/fix-rail-replacement-bus-text-missing/src/sdk.ts#L82). 

```suggestion
export enum TransportSubmode {
```

reidzeibel: Updated both enums, thank you!

reidzeibel:
@gorandalum @rosvik I removed more enums and replaced them with the journey-planner-v3 enums, thanks for pointing it out.

PR #4202, opened 2024-01-04
chore: adding button to storybook

adriawh:
###  📖 Adding the button component to storybook

- Added the button component with block, pill and inline  
- Changed theme selection to radio buttons on existing stories for faster switching 🏎️ 

### ❗ Note: 
Another possibility would also to have the switch between button modes(pill, inline, block) to be a selection in the addons instead of separate pages, however I belive it is more structured this way and also the addons page already had a lot of options.

### Showcase

<details>
<summary> 📷  Images </summary>

Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 01 57
Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 01 48
Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 02 04
Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 02 07
</details>

<details>
<summary> 📹 Video </summary>

https://github.com/AtB-AS/mittatb-app/assets/70323886/3f8e0961-4d27-4b5f-85a0-074ad6060d78

</details>

### DIFF
--- a/.storybook/storybook.requires.js
+++ b/.storybook/storybook.requires.js
@@ -47,6 +47,7 @@ try {
 
 const getStories = () => {
   return {
+    "./src/storybook/stories/Button.stories.js": require("../src/storybook/stories/Button.stories.js"),
     "./src/storybook/stories/MessageInfoBox.stories.js": require("../src/storybook/stories/MessageInfoBox.stories.js"),
     "./src/storybook/stories/MessageInfoText.stories.js": require("../src/storybook/stories/MessageInfoText.stories.js"),
     "./src/storybook/stories/ThemeText.stories.js": require("../src/storybook/stories/ThemeText.stories.js"),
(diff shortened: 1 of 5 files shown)


rosvik (approved):
Nice!

gorandalum (commented):
The implementation looks good, but I want more! 😈 

We said at some point earlier that when we add things to Storybook we also fix up the component, and align it with Figma. I will be more clear on this in Storybook-issues I create going forward.

But what I would like on the Button-component:
- Align Figma and Storybook. The page for the button component in Figma has many examples which are not in Storybook. They should probably not be in Storybook, but moved to another place in Figma.
- Fix illogicalities in the implementation/design. For example changing the secondary interactive_1 button to active makes it go from white to black, which doesn't make sense on almost any background.
- Remove the possible "hacks" to change the internal design of the Button component through using `textStyle` and `textContainerStyle`.

adriawh:
Got it, will do some research on the component in the code and in Figma and then organize a meeting with designers. Just need to figure out how my schedule for next week will be first.

PR #4203, opened 2024-01-04
feat: Live position in map not reliant on vehicle position

gorandalum:
We checked on whether the estimated call had realtime data to
determine whether to show live bus position or not, but it turns
out a vehicle may have live position even if it doesn't have
realtime data.

Changed the logic to be based on departure time instead. If there
is less than 10 minutes until the service journey starts, or it has
started, then we will start to fetch vehicle live position.

### DIFF
--- a/src/travel-details-screens/utils.ts
+++ b/src/travel-details-screens/utils.ts
@@ -1,6 +1,6 @@
 import {
   dateWithReplacedTime,
-  iso8601DurationToSeconds,
+  iso8601DurationToSeconds, minutesBetween,
   secondsBetween,
 } from '@atb/utils/date';
 import {Leg, TripPattern} from '@atb/api/types/trips';
@@ -17,6 +17,7 @@ import {
 } from '@atb/api/types/generated/journey_planner_v3_types';
 import {dictionary, TranslateFunction} from '@atb/translations';
 import {APP_ORG} from '@env';
+import {EstimatedCallWithMetadata} from "@atb/travel-details-screens/use-departure-data";
 
 const DEFAULT_THRESHOLD_AIMED_EXPECTED_IN_MINUTES = 1;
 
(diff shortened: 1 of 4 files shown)


rosvik (approved):
💯

toretefre (approved):

src/travel-details-screens/utils.ts
```diff
@@ -427,3 +428,16 @@ export function getIsTooLateToBookTripPattern(
 ): boolean {
   return tripPattern?.legs?.some((leg) => getIsTooLateToBookLeg(leg, now));
 }
+
+export const getShouldShowLiveVehicle = (
```
marius-at-atb, line 432: Should it return `false` if
- cancelled?
- `aimedArrivalTime`/`actualArrivalTime` passed > 1 hour ago?

gorandalum: > Should it return `false` if
> 
> * cancelled?
> * `aimedArrivalTime`/`actualArrivalTime` passed > 1 hour ago?

Regarding cancelled:
Good suggestion and that makes sense, but after looking into it now I found it slightly harder than one would think. The `activeItem` in this file has an `isTripCancelled` property, but it can't really be used as it only reflects cancellation for the specific stop place which are specified as from-stop-place in the departure details. Even though that stop may be cancelled, the rest of the departure may go as usual. So all in all, since opening departure details for a cancelled departure is probably not happening often, I'll not do anything with this now.

Regarding logic on arrival time:
There is logic handling this now in the app, it just isn't touched here. The main logic is in this expression:
```ts
  const isJourneyFinished =
    vehiclePosition?.vehicleStatus === VehicleStatusEnumeration.Completed ||
    estimatedCallsWithMetadata.every((e) => e.actualArrivalTime);  
```

marius-at-atb: Right. Should be fine as is for the current use case then :+1:

PR #4205, opened 2024-01-05
fix: results with AtB bestill as first leg not showing with larger text

adriawh:
### Problem
On some screens, both android and iOS, the trip results which had _AtB bestill_ as its first leg did not show but were clickable. This was caused by the width of the content being larger than the parent. The following code therefore made it so that the **opacity was always 0**:
```
 const fadeInValueRef = useRef(new Animated.Value(0));

  // Dynamically collapse legs to fit horizontally
  useEffect(() => {
    if (legIconsParentWidth && legIconsContentWidth) {
      if (legIconsContentWidth >= legIconsParentWidth) {
        setNumberOfExpandedLegs((val) => Math.max(val - 1, 1));
      } else {
        Animated.timing(fadeInValueRef.current, {
          toValue: 1,
          duration: 250,
          useNativeDriver: true,
        }).start();
      }
    }
  }, [legIconsParentWidth, legIconsContentWidth]);
```

### Before

<details>
<summary>The results were invisible but clickable</summary>

| Android(Pixel) | iOS(iPhone 13 mini) |
|---------|-----|
|  |  |

</details>

### After

<details>
<summary>The results are visible</summary>

https://github.com/AtB-AS/mittatb-app/assets/70323886/f1d659bc-91cf-4233-b6a6-1e02909c527c

| Android(Pixel) | iOS(iPhone 13 mini) |
|---------|-----|
|  |  |

</details>

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
@@ -176,14 +176,12 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
 
   // Dynamically collapse legs to fit horizontally
   useEffect(() => {
-    if (
-      legIconsParentWidth &&
-      legIconsContentWidth &&
-      legIconsContentWidth >= legIconsParentWidth
-    ) {
-      updateNumberOfExpandedLegsOrFadeIn();
-    } else {
-      fadeIn();
+    if (legIconsParentWidth && legIconsContentWidth) {
+      if (legIconsContentWidth >= legIconsParentWidth) {
+        updateNumberOfExpandedLegsOrFadeIn();
+      } else {
+        fadeIn();
+      }
     }
   }, [
     fadeIn,
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
@@ -32,7 +32,7 @@ import {
   getTranslatedModeName,
 } from '@atb/utils/transportation-names';
 
-import React, {useCallback, useEffect, useRef, useState} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import {
   AccessibilityProps,
   Animated,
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -160,19 +160,25 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
+      legIconsParentWidth &&
+      legIconsContentWidth &&
+      legIconsContentWidth >= legIconsParentWidth
+    ) {
+      setNumberOfExpandedLegs((val) => Math.max(val - 1, 1));
     }
   }, [legIconsParentWidth, legIconsContentWidth]);
 
+  useEffect(() => {
+    if (numberOfExpandedLegs === 1) {
```
marius-at-atb, line 173: Should be `numberOfExpandedLegs >== 1` I think?

marius-at-atb, line 173: This won't work to show resultItems with final `numberOfExpandedLegs > 1`

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -158,20 +158,30 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
+
   // Dynamically collapse legs to fit horizontally
   useEffect(() => {
     if (legIconsParentWidth && legIconsContentWidth) {
       if (legIconsContentWidth >= legIconsParentWidth) {
-        setNumberOfExpandedLegs((val) => Math.max(val - 1, 1));
+        setNumberOfExpandedLegs((val) => {
+          const newVal = Math.max(val - 1, 1);
+          if (newVal === val) {
+            fadeIn();
```
gorandalum, line 176: Not sure about this 🤔 
https://stackoverflow.com/questions/71780110/is-it-okey-to-use-side-effects-in-the-usestate-hook-callback

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -158,20 +158,37 @@ const ResultItem: React.FC<ResultItemProps & AccessibilityProps> = ({
...
+    }).start();
+  }, []);
+
+  const updateNumberOfExpandedLegsOrFadeIn = useCallback(() => {
+    if (numberOfExpandedLegs > 1) {
+      setNumberOfExpandedLegs((val) => val - 1);
+    } else {
+      fadeIn();
+    }
+  }, [fadeIn, numberOfExpandedLegs]);
```
marius-at-atb, line 175: Since `numberOfExpandingLegs` is a dependency here, and `updateNumberOfExpandedLegsOrFadeIn` is a dependency for the useEffect, I think it will trigger as many updates as when using `numberOfExpandedLegs` as a dependency directly on the useEffect.

Perhaps `updateNumberOfExpandedLegsOrFadeIn` can be removed, and the useEffect can have `numberOfExpandedLegs` as a dependency and be simplified to
```ts
if (legIconsParentWidth && legIconsContentWidth) {
  if (
    legIconsContentWidth >= legIconsParentWidth &&
    numberOfExpandedLegs > 1
  ) {
    setNumberOfExpandedLegs((val) => val - 1);
  } else {
    fadeIn();
  }
}
```

adriawh: Yes, correct and I found myself with the same issue as before when testing some more. Have changed my approach

PR #4208, opened 2024-01-08
fix: rename "active" tab title

adriawh:
| English | Bokmål | Nynorsk |
|---------|--------|---------|
| Simulator Screenshot - iPhone 15 Pro - 2024-01-08 at 14 36 52 | Simulator Screenshot - iPhone 15 Pro - 2024-01-08 at 14 36 33 | Simulator Screenshot - iPhone 15 Pro - 2024-01-08 at 14 36 57 |

### DIFF
--- a/src/translations/screens/Ticketing.ts
+++ b/src/translations/screens/Ticketing.ts
@@ -48,11 +48,6 @@ const TicketingTexts = {
   },
   activeFareProductsAndReservationsTab: {
     label: _('Mine billetter', 'My tickets', 'Mine billettar'),
-    a11yLabel: _(
-      'Mine aktive billetter',
-      'My active tickets',
-      'Mine aktive billettar',
-    ),
     noActiveTicketsTitle: _(
       'Ingen aktive billetter',
       'No active tickets',
--- a/src/translations/screens/Ticketing.ts
+++ b/src/translations/screens/Ticketing.ts
@@ -47,8 +47,12 @@ const TicketingTexts = {
     },
   },
   activeFareProductsAndReservationsTab: {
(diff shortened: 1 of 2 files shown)


src/translations/screens/Ticketing.ts
```diff
@@ -47,8 +47,12 @@ const TicketingTexts = {
...
   },
   activeFareProductsAndReservationsTab: {
-    label: _('Aktive', 'Active', 'Aktive'),
-    a11yLabel: _('Aktive billetter', 'Active tickets', 'Aktive billettar'),
+    label: _('Mine billetter', 'My tickets', 'Mine billettar'),
+    a11yLabel: _(
+      'Mine aktive billetter',
+      'My active tickets',
+      'Mine aktive billettar',
+    ),
```
gorandalum, line 55: Accessibility text should be the same as shown text, and there is no meta-info (like image or anything) that signals extra info regarding that the tab is a "My active tickets" tab. And this probably means we don't need to set an `accessibilityLabel` on the component at all.

marius-at-atb (approved):

gorandalum (changes requested):

src/translations/screens/Ticketing.ts
adriawh: Will remove it on the "my tickets" tab. There is also a similar case on the "buy" tab, where it in the a11y labels says "buy tickets". Here I belive it should be kept as just the word "buy" is not very describing in itself. Agree or not? @gorandalum

adriawh: was decided by @hannahATB to keep it as "buy"

gorandalum (approved):

PR #4212, opened 2024-01-09
feat: Show live button in trip screen not reliant on realtime

gorandalum:
The live button was missing in the trip details screen:
https://github.com/AtB-AS/kundevendt/issues/15618#issuecomment-1878666180

Fixed by using same utility function as in:
https://github.com/AtB-AS/mittatb-app/pull/4203


marius-at-atb (approved):
Looks good!

PR #4215, opened 2024-01-10
feat: Stop the beacons if remote config becomes false

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/15762


ios/Podfile.lock
```diff
@@ -1907,13 +1907,13 @@ SPEC CHECKSUMS:
...
-  TokenStateReactNativeLib: 0f6875d9d039385a07456588701f638dc5463a79
+  TokenStateReactNativeLib: ee91eddc1159d7efab18960b8afd90fdfbb6f77c
   Turf: 13d1a92d969ca0311bbc26e8356cca178ce95da2
   Yoga: 3efc43e0d48686ce2e8c60f99d4e6bd349aff981
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
 PODFILE CHECKSUM: 65771909d1de0011499a623efd21303801625f9f
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.12.1
```
marius-at-atb, line 1919: The changes in this file are probably not intended to be included?

marius-at-atb (approved):
Lgtm!

ios/Podfile.lock
jorelosorio: We should keep 1.12.1 because that is what we have defined on the gemfiles, I'm going to create a task to upgrade to the newest version.

PR #4216, opened 2024-01-10
refactor: Add StackParams type

gorandalum:
This type is meant to be used on every stack params specification. It both:
- Checks that each key in the stack param list ends with "Screen" or "Stack"
- Adds the CustomScreenParams to the mapped value type., which makes it
  possible to set transition when navigating.


src/stacks-hierarchy/RootStack.tsx
```diff
@@ -392,19 +392,19 @@ export const RootStack = () => {
                 component={Root_AddPaymentMethodScreen}
               />
               <Stack.Screen
-                name="Root_ParkingViolationsSelect"
+                name="Root_ParkingViolationsSelectScreen"
                 component={Root_ParkingViolationsSelect}
```
marius-at-atb, line 396: Update the component names as well?

rosvik (approved):
Clean ✨

src/stacks-hierarchy/Root_TicketAssistantStack/TicketAssistant_ZonePickerScreen.tsx
```diff
@@ -43,8 +43,8 @@ export const TicketAssistant_ZonePickerScreen = ({
 
   const isApplicableOnSingleZoneOnly = false;
 
-  const fromTariffZone = route.params?.fromTariffZone ?? defaultTariffZone;
-  const toTariffZone = route.params?.toTariffZone ?? defaultTariffZone;
+  const fromTariffZone = defaultTariffZone;
+  const toTariffZone = defaultTariffZone;
```
jorelosorio, line 47: So this does not depend on any more of incoming parameters in the zone?

jorelosorio: OK, got it! Seems like not related but good to have then I think.

jorelosorio (approved):
Good approach

PR #4219, opened 2024-01-11
fix: Check for permissions before start the beacons

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/16142

### DIFF
--- a/src/beacons/BeaconsContext.tsx
+++ b/src/beacons/BeaconsContext.tsx
@@ -180,7 +180,7 @@ const BeaconsContextProvider: React.FC = ({children}) => {
       }
 
       const permissions = await allowedPermissionsForBeacons();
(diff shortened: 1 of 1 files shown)


src/beacons/BeaconsContext.tsx
```diff
@@ -180,7 +180,7 @@ const BeaconsContextProvider: React.FC = ({children}) => {
       }
 
       const permissions = await allowedPermissionsForBeacons();
-      if (consentGranted && permissions && !beaconsInfo?.isStarted) {
+      if (consentGranted && permissions.length > 0 && !beaconsInfo?.isStarted) {
```
jorelosorio, line 183: Permissions is an array and returned as empty if no permissions were granted, the condition only checks if there is an array. Then it sends empty to the SDK when starts and fails internally.

reidzeibel (approved):

PR #4221, opened 2024-01-11
fix: remove redundant close function for bottom sheet in map

adriawh:
When closing any bottom sheet, except for the filter one, on the map on android the app crashed. The reason for the crash was that when pressing the close button the close function was triggered twice leading to an issue with a sending accessibiltyEvent.

I found that the removed close call was not needed for closing the bottom sheet when clicking the map or when clicking the close button.

### DIFF
--- a/src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
+++ b/src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
@@ -168,8 +168,6 @@ export const useUpdateBottomSheetWhenSelectedEntityChanges = (
             />
           );
         }, false);
-      } else {
-        closeBottomSheet();
       }
     })();
     // eslint-disable-next-line react-hooks/exhaustive-deps


hanne-at-atb (approved):

reidzeibel (approved):

marius-at-atb:
Hmm, the error seems to persist in my Android emulator after this change as well.
When removing `runAfterInteractions` and `timeout` from `giveFocus`, the error does not occur.

I suspect it can be an issue related to [this change](https://github.com/AtB-AS/mittatb-app/pull/4174/files#diff-ad9397a771054f883a4397af4d53ce131d6ad31ba85fb9cdbfd075091b50d045R37-R43)

A little hypothesis:
Before we would send in a ref whose current value would be updated also when async logic occurred in giveFocus.
Now we send in a value that doesn't change, and which might be accepted by  `findNodeHandle`/`setAccessibilityFocus` validity checks because of proper format. However if the element no longer exists, this could give an error when executing.

The change that all focus calls now use timeouts, also increases the likelihood of elements being removed before `findNodeHandle`/`setAccessibilityFocus`

PR #4223, opened 2024-01-11
feat: fix session counter for share travel habits screen

reidzeibel:
When using `appStatus !== 'active'` the counter will increase when something appears on top of our app. For example, a permission dialog, or maybe a phone call, which blocks our app and causes the `appStatus` to change to 'background', triggering the `useEffect` because it is a dependency. 
  
In this particular case, we only want the counter to be increased between app sessions, and not by something incidental such as permission request. Therefore using `useFocusEffect` and `useCallback` seems to be the better choice, since it only updates the counter when the actual screen is shown and focused, instead of calling it everytime the app went to `background` state.

This should fix the issue that the `travelHabitSessionCount` increasing more than intended, a minor issue found by @tormoseng while investigating https://github.com/AtB-AS/kundevendt/issues/16288.

### DIFF
--- a/src/beacons/use-should-show-share-travel-habits-screen.tsx
+++ b/src/beacons/use-should-show-share-travel-habits-screen.tsx
@@ -1,12 +1,12 @@
-import {useCallback, useEffect, useRef, useState} from 'react';
+import {useCallback, useRef, useState} from 'react';
 import {storage} from '@atb/storage';
 
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
-import {useAppStateStatus} from '@atb/utils/use-app-state-status';
 
 import {useAppState} from '@atb/AppContext';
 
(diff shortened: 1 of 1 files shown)


jorelosorio (approved):
LGTM

gorandalum (commented):
Haven't tested it, but I got a suspicion this will fire a focus event every time a root screen is closed and the tabnav is focused again.

reidzeibel:
@gorandalum apparently that happens, so I will try to find further fix for this, maybe I have to use appState and combine it with useFocusEffect. I'll think of something else over the weekend.

reidzeibel:
Can't found simple enough solution that doesn't require too much effort, closing this for now, created an issue https://github.com/AtB-AS/kundevendt/issues/16472

PR #4224, opened 2024-01-12
fix: use service journey transport mode in departure

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/16257

BFF part: https://github.com/AtB-AS/atb-bff/pull/322


marius-at-atb (approved):
Lgtm

gorandalum (commented):
💯👍 

Note that the types-files should be updated in the app as well when they are updated in the BFF.

PR #4225, opened 2024-01-12
fix: giveFocus to unmounted node

adriawh:
Issue link: https://github.com/AtB-AS/kundevendt/issues/16389

The problem was the node trying to call `AccessibilityInfo.setAccessibilityFocus(reactTag)` when the node was unmounted. 

As @marius-at-atb said in [a comment](https://github.com/AtB-AS/mittatb-app/pull/4221#issuecomment-1887463960):

> When giveFocus receives a ref, then its current value will later be modified, and this change will be reflected inside giveFocus.
> 
> When giveFocus receives a ref.current directly, that value will later be reassigned, and the change won't be reflected inside giveFocus.
> 
> Therefore when using async logic inside giveFocus and accepting a ref.current directly, the value can be outdated by the time it is used, causing a crash.

This fix addresses the bottom-sheet crashing when pressing the close button in the map, and @reidzeibel's case of the app crashing after logging in with vipps on a fresh install.

### DIFF
--- a/src/components/bottom-sheet/BottomSheetContext.tsx
+++ b/src/components/bottom-sheet/BottomSheetContext.tsx
@@ -68,7 +68,7 @@ export const BottomSheetProvider: React.FC = ({children}) => {
   const close = () => {
     setContentFunction(() => () => null);
     setIsOpen(false);
-    giveFocus(onCloseFocusRef.current);
+    giveFocus(onCloseFocusRef);
   };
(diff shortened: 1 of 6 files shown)


gorandalum (commented):
Missing change in the `focusCallbackRef`?

marius-at-atb (approved):
Nice!

PR #4226, opened 2024-01-12
fix: Stations and vehicles fetched too often

gorandalum:
There was two error that together caused this:
- The onMapIdle listener fires way too often, which make it seems
like the viewport has changed when it really hasn't.
- Bug in the code comparing the previously loaded area with the new
area, which caused refetching data when it was not necessary.

Since onMapIdle fired too many times (with the same map region
visible) and there was a bug when comparing two map regions, we
ended up fetching data more than we should.

This is now fixed by:
- The onMapIdle listener updates a state storing the map region, but
only if it has really changed. So now, even though the onMapIdle
listener fires more than it should, it shouldn't update the state.
- Fixed the logic comparing two areas, where we now transform
both previous area and new area to features of the same format
before comparing them.

After this there should be no multiple-spinners after each other
on viewport change, and also no spinner if just clicking the map.


marius-at-atb (approved):
Very nice!

PR #4227, opened 2024-01-12
feat: add new permission prompts for notifications and beacons

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/16425

This PR attempts to separate "consents" and "permissions" more clearly by adding message boxes that warns about missing permissions when consent is given. This applies to push notifications, email and beacons. (ref. https://github.com/AtB-AS/kundevendt/issues/15656#issuecomment-1886818199)

In the BeaconsContext, this is reflected by always setting local storage `beaconsConsent = true`, even if the user doesn't give permissions. 

Also changes some variable names and adds some comments :) 

### Screenshots

<div>

</div>


rosvik:
@jorelosorio

PR #4229, opened 2024-01-15
fix: travel token consider change screen onboarding flag

marius-at-atb:
The travel token needs some time before it is ready after a login.
The new onboarding flow showed the `Root_ConsiderTravelTokenChangeScreen` screen immediately, but this caused an error state to be visible. Now the flag for whether that screen should show or not is false until the travel token is ready.
One change this causes, is that it will no longer show as the first screen after login (but atm it has highest priority, so as soon as it is ready, it will be shown.) In most cases it will now be shown right after the location onboarding screen.

Closes https://github.com/AtB-AS/kundevendt/issues/8620#issuecomment-1891961937


gorandalum (approved):
:+1:

PR #4231, opened 2024-01-16
feat: don't show onboarding if beacon consent is granted even once

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/16288

This PR will save the state of the beacons consent granted, if it has ever been toggled once, then we should not show the beacons onboarding. This is to handle if the user toggles the consent directly on the privacy policy screen before they saw the beacons onboarding. 

Also resets the key on the restart beacons onboarding item on the debug screen.

### DIFF
--- a/src/beacons/BeaconsContext.tsx
+++ b/src/beacons/BeaconsContext.tsx
@@ -178,7 +178,7 @@ const BeaconsContextProvider: React.FC = ({children}) => {
     }
 
     return permissionsGranted;
-  }, [isBeaconsSupported, rationaleMessages, initializeKettleSDK]);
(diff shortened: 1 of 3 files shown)


rosvik (commented):
I briefly looked at this bug myself now, and I think there might be a simpler fix for this. I haven't tested my solution (it takes some time to get my test device ready here 😅), but pushed my changes to a91b6afb3cdaab01f6bfcbedf0d30d7ff7204472. 

Do you think that could work instead?

reidzeibel:
@rosvik looks like yours are shorter and easier, I'll test it!

reidzeibel:
@rosvik yours work and is simpler, so let's go with yours!

rosvik:
@reidzeibel Nice! Local testing of my PR looks good here as well 🙌

PR #4230, opened 2024-01-15
fix: set screenreader focus on title for OnboardingScreenComponent

marius-at-atb:
When using screenreader, set the focus on the title for OnboardingScreenComponent.

Prevents issues like https://github.com/AtB-AS/kundevendt/issues/8620#issuecomment-1878544889 where the screenreader can get stuck in the title.

Resolves https://github.com/AtB-AS/kundevendt/issues/8619

PR #4234, opened 2024-01-16
fix: missing accessible for android

marius-at-atb:
Fix for android, see [comment here](https://github.com/AtB-AS/mittatb-app/pull/4230/files/04c722a41280339fe713a4b85f69403ba2178c31#r1453319773)

### DIFF
--- a/src/onboarding-screen/OnboardingScreenComponent.tsx
+++ b/src/onboarding-screen/OnboardingScreenComponent.tsx
@@ -50,7 +50,7 @@ export const OnboardingScreenComponent = ({
       footerDescription={footerDescription}
     >
       <View style={styles.header}>{illustration}</View>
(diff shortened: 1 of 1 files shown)


gorandalum (approved):

gorandalum:
And one more thing, have you tested it on Android? Might need `collapsable={false}` as well.

marius-at-atb:
@gorandalum In the android simulator it crashes without `accessible`, but not when it is added.
So `collapsable={false}` doesn't seem to be necessary, but should we add it anyway?

gorandalum:
Yeah, I think it only crashed in staging without it. Might create an `AccessibleView` or something for this, where ref is mandatory and it sets accessible and collapsable correctly.

tormoseng:
Tested OK!

PR #4236, opened 2024-01-16
fix: fix qa issues on permissions flow

rosvik:
ref. https://github.com/AtB-AS/kundevendt/issues/16425#issuecomment-1893576632

Fixes a few issues from QA:

- Dismissing the permissions prompt during onboarding no longer sets consent to true.
- The prompt to open phone settings on the privacy page is now based on if we have _any_ beacons permission, not just bluetooth
- The prompt to open phone settings on the privacy page is now removed when permissions are allowed

### DIFF
--- a/src/beacons/BeaconsContext.tsx
+++ b/src/beacons/BeaconsContext.tsx
@@ -64,9 +64,11 @@ type BeaconsContextState = {
    * permissions are granted, the Kettle SDK will be started, and beaconsInfo
    * will be initialized. `shareTravelHabitsOnboarded` will also be set to true.
    *
+   * @param alreadyConsented When true, consent will be set to true regardless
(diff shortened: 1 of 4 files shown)


marius-at-atb (approved):
Nice!

gorandalum:
> * The prompt to open phone settings on the privacy page is now based on if we have _any_ beacons permission, not just bluetooth

Does the beacons sdk run if we don't have bluetooth enabled, but have given the other permissions? I was under the impression that bluetooth was kinda mandatory, and that's why we don't ask for the other permissions if they say no to bluetooth.

rosvik:
@gorandalum 

Looks to me like we only need one of the permissions in order to start the SDK. Do you know if that's something that should be changed, @jorelosorio?

PR #4238, opened 2024-01-17
feat: remove emojis from tips and information

hanne-at-atb:
Removing the usage of emojis in tips and information

Closes: https://github.com/AtB-AS/kundevendt/issues/16299


marius-at-atb (approved):
Looks good!

PR #4240, opened 2024-01-17
fix: make expanded part of expandable section item accessible

hanne-at-atb:
The screen reader can now access and read out the content of the expanded section in ExpandableSectionItem, both on android and iOS 👾 

Fixed as part of: https://github.com/AtB-AS/kundevendt/issues/16299
Also closing the reported bug: https://github.com/AtB-AS/kundevendt/issues/10098

### DIFF
--- a/src/components/sections/items/ExpandableSectionItem.tsx
+++ b/src/components/sections/items/ExpandableSectionItem.tsx
@@ -45,7 +45,7 @@ export function ExpandableSectionItem({
   testID,
   ...props
 }: Props) {
-  const {contentContainer, topContainer} = useSectionItem(props);
+  const {contentContainer, topContainer, spacing} = useSectionItem(props);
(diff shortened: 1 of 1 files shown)


rosvik (approved):

gorandalum (commented):
Tested both on iOS and Android? My experience is that nested accessible doesn't play well on iOS.

marius-at-atb:
Doesn't seem to work on my iPhone 15

hanne-at-atb:
> Tested both on iOS and Android? My experience is that nested accessible doesn't play well on iOS.

Tested on android 😢 Will look into it!

gorandalum (approved):
One tip: Sit together with tester and show them the solution locally to decide if it's good enough (if you haven't already). The feedback loop is much faster then than it going back and forth in test.

PR #4241, opened 2024-01-17
feat: Add BorderedInfoBox component

gorandalum:
Adds a new component to the src and storybook. This is the bordered
info box introduced in the on behalf of product initiative. Pretty
basic component, where the main cavaet is that the border is based
on the text color with 0.1 opacity. This was since we didn't have
an appropriate color to use in the design-system, and we didn't
want to introduced a new one just for this use case.

<details>
<summary>Screenshots from Storybook:</summary>

Screenshot 2024-01-17 at 12 07 50
Screenshot 2024-01-17 at 12 07 57
Screenshot 2024-01-17 at 12 08 06
Screenshot 2024-01-17 at 12 08 16
Screenshot 2024-01-17 at 12 08 24

</details>

### DIFF
--- a/src/utils/add-opacity.ts
+++ b/src/utils/add-opacity.ts
@@ -1,0 +1,15 @@
+/**
+ * Adds opacity to the input hex color. Should be used with colors from the
+ * design system, as that is the hex format this function supports. It doesn't
+ * work with 3 letter hex codes, or hex codes already containing the
+ * opacity/alpha channel.
+ *
+ * Example usage:
+ * addOpacity(theme.static.background.background_0.background, 0.2)
+ */
+export const addOpacity = (hex: string, opacity: number) => {
+    const r = parseInt(hex.slice(1, 3), 16);
+    const g = parseInt(hex.slice(3, 5), 16);
+    const b = parseInt(hex.slice(5, 7), 16);
+    return `rgba(${r}, ${g}, ${b}, ${opacity})`;
+};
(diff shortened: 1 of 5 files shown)


gorandalum:
Adding everyone on PR since it is a new component, and nice for every one to keep track of those.

src/utils/add-opacity.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Adds opacity to the input hex color. Should be used with colors from the
+ * design system, as that is the hex format this function supports. It doesn't
+ * work with 3 letter hex codes, or hex codes already containing the
+ * opacity/alpha channel.
+ *
+ * Example usage:
+ * addOpacity(theme.static.background.background_0.background, 0.2)
+ */
+export const addOpacity = (hex: string, opacity: number) => {
```
marius-at-atb, line 10: Should we add a safety check and maybe also a warning if the hex string is not of the required format?

marius-at-atb (approved):
Lgtm!

src/utils/add-opacity.ts
gorandalum: Good suggestion! I added a safety check and fallback to return the hex value unopacitiated (that's surely a word). Didn't add Bugsnag warning since that would really spam Bugsnag if this happened.

reidzeibel: We definitely need to add `unopacitiated` into the vocabulary

marius-at-atb: Nice! I was thinking of just a `console.warn`, but the safety check should be enough :+1:

reidzeibel (approved):

PR #4246, opened 2024-01-17
feat: Use BorderedInfoBox on ticket components

gorandalum:
Replace the previously used InfoChip with the new
BorderedInfoBox component. The main reason is that the
blue InfoChip looks to similar to a button, which should not
be a problem with the MessageInfoBox.

How it was before:

Places it has been changed:
<div>

</div>


marius-at-atb (approved):
Lgtm

PR #4247, opened 2024-01-17
refactor: Delete InfoTag and InfoChip

gorandalum:
After replacing most InfoChip usages with BorderedInfoBox the only usage left was the 'new' tag on the ticket assistant. That could be removed as it isn't new anymore, and voilà we could delete both InfoTag and InfoChip. Replaced InfoChip with LabelInfo in the design system.


marius-at-atb (approved):
Voilà indeed 😛

PR #4249, opened 2024-01-18
Implement fetching customer account id using phone number

reidzeibel:
closes  https://github.com/AtB-AS/kundevendt/issues/16506
also closes https://github.com/AtB-AS/kundevendt/issues/15550

- Implement API for fetching customer account id from phone number.
- Moved some login texts that are no longer relevant to phone input texts.
- Also implemented sending ticket to others since we just need to pass the fetched customer ID to the vipps/card payment and it only takes around 15 minutes to do that and test it.
- Ticket sending should work, but the UI is not updated, so both account (sender and receiver) will have both tickets, how to handle the tickets is being discussed in this slack thread https://mittatb.slack.com/archives/C02DL21RK7C/p1705654896301609
- Ticket history filtering is done in other PR(s)

<details>
<summary>Android video</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/593ed40a-1542-4140-8493-7f5b81651f93

</details>

<details>
<summary>iOS video</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/616151fe-c427-49e4-a16b-225789342852

</details>


src/components/phone-input/PhoneInput.tsx
```diff
@@ -62,29 +65,38 @@ export const PhoneInput = ({
       return;
     }
 
-    const errorCode = await onSubmitPromise(phoneValidation.phoneNumber);
-    if (!errorCode) {
+    const result = await onSubmitPromise(phoneValidation.phoneNumber);
+
+    setIsSubmitting(false);
+
+    if (!result) {
```
jorelosorio, line 72: We might simplify the conditions by just using:

```
if (isError(result)) {
        setError(result);
      } else {
        setError(undefined);
        onSubmitAction(phoneValidation.phoneNumber, result);
      }
```

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
```diff
@@ -23,6 +24,21 @@ export const Root_ChooseTicketReceiverScreen: React.FC<Props> = ({
...
+  const handleValidatePhoneNumber = async (phoneNumber: string) => {
+    const result = await getCustomerAccountId(phoneNumber);
+    if (result) {
+      switch (typeof result) {
+        case 'string': // customer account id successfully returned
+          return result;
+        case 'object': // phone number not associated to any account or invalid number
+          return 'no_associated_account';
+        case 'undefined': // unknown error
+          return 'unknown_error';
```
jorelosorio, line 36: For this case, I think the result is not undefined never because of the condition.

jorelosorio: What about:

return typeof result === 'string' ? result : 'no_associated_account';

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
```diff
@@ -58,12 +74,12 @@ export const Root_ChooseTicketReceiverScreen: React.FC<Props> = ({
           <PhoneInput
             submitButtonText={t(PurchaseOverviewTexts.summary.button)}
             submitButtonTestId="toPaymentButton"
-            // TODO: this is just a placeholder, update when API is ready
-            onSubmitPromise={() => new Promise((r) => setTimeout(r, 1000))}
-            onSubmitAction={(number) => {
+            onSubmitPromise={(number) => handleValidatePhoneNumber(number)}
```
jorelosorio, line 77: As far I understand, this is a validation of the phone, then if ok call the action. In that case I would suggest better names for it. ValidatePhoneAction, OnCompletedPhoneValidationAction or something like that... or

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
reidzeibel: the `getCustomerAccountId` function may return 3 types: string (if request success), object (if request fails and there's a message from backend), and undefined (other errors, server issue, for instance), so I want these types to be supported, what do you think?

src/components/phone-input/PhoneInput.tsx
reidzeibel: thanks!

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
jorelosorio: I mean if the 3 states are needed then is OK but because there is a condition first that check result is not undefined but inside ask for undefined type which was excluded at the top then I just see two cases if response a string then is good if not then an error that includes undefined and object.

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
reidzeibel: Updated, thanks!

src/components/phone-input/PhoneInput.tsx
reidzeibel: Updated!

src/api/profile.ts
```diff
@@ -38,3 +38,37 @@ export async function deleteProfile(opts?: AxiosRequestConfig) {
...
+ * @param phoneNumber phone number with prefix
+ *
+ * @returns customer account ID as string if request is success
+ *
+ * Will return one the following errors if request unsuccessful:
+ * -  Error 404 with object {"error": "cannot find user from phone number: \"{phoneNumber}\""}
+ * -  Error 400 with object {"error": "Invalid phone number as per country code"}
+ * -  other than those 2 errors mentioned earlier, returns undefined.
+ */
+export const getCustomerAccountId = async (phoneNumber: string) => {
```
gorandalum, line 54: Could you type the return type from this function? A little hard to see just by reading the code.

src/components/phone-input/PhoneInput.tsx
```diff
@@ -1,24 +1,27 @@
...
+import {GetAccountByNumberErrorCode} from '@atb/on-behalf-of/types';
+
+type PhoneInputErrorCode = PhoneSignInErrorCode | GetAccountByNumberErrorCode;
 
 type Props = {
   style?: StyleProp<ViewStyle>;
   submitButtonText: string;
   submitButtonTestId: string;
-  onSubmitPromise: (
+  phoneNumberValidation: (
```
gorandalum, line 20: Kinda hard to understand that it is a function based on the parameter name when it is just a noun. It is easier to understand when using "on" as prefix as that is a pretty standard convention. Or something like `validatePhoneNumber` is also clearer that is a function, when it starts with a verb.

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
```diff
@@ -23,6 +24,14 @@ export const Root_ChooseTicketReceiverScreen: React.FC<Props> = ({
 
   const focusRef = useFocusOnLoad();
 
+  const handleValidatePhoneNumber = async (phoneNumber: string) => {
+    const result = await getCustomerAccountId(phoneNumber);
```
gorandalum, line 28: New api requests in the app should use `useQuery`

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
```diff
@@ -23,6 +24,14 @@ export const Root_ChooseTicketReceiverScreen: React.FC<Props> = ({
...
+    if (result) {
+      return typeof result === 'string' ? result : 'no_associated_account';
+    }
+    return 'unknown_error';
+  };
```
gorandalum, line 33: Handle this in the api-method? So not every place in the code where we use the api method needs to handle it.

src/api/profile.ts
```diff
@@ -38,3 +38,37 @@ export async function deleteProfile(opts?: AxiosRequestConfig) {
...
+      },
+    )
+    .then((response) => response.data.customerAccountId)
+    .catch((error) => {
+      const metadata = getAxiosErrorMetadata(error);
+      const responseStatus = metadata.responseStatus;
+      if (responseStatus === 404 || responseStatus === 400) {
+        return metadata.responseData && JSON.parse(metadata.responseData);
+      }
+      return undefined;
```
gorandalum, line 72: Can we instead return undefined when no phone number was found? That is more intuitive to me than returning an error when phone number not found, and returning undefined if unknown error occured.

gorandalum (changes requested):
Good work! Some comments here and there 😊

src/components/phone-input/PhoneInput.tsx
reidzeibel: renamed into ´validatePhoneNumber´

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
```diff
@@ -18,11 +19,19 @@ export const Root_ChooseTicketReceiverScreen: React.FC<Props> = ({
...
+  const {mutateAsync: getAccountIdByPhone} = useGetAccountIdByPhoneMutation();
+  const handleValidatePhoneNumber = async (phoneNumber: string) => {
+    try {
+      const result = await getAccountIdByPhone(phoneNumber);
+      return result;
+    } catch (error) {
+      return error as string;
+    }
+  };
+
```
marius-at-atb, line 34: I think this can be simplified to
```suggestion
const handleValidatePhoneNumber = async (phoneNumber: string) =>
    await getAccountIdByPhone(phoneNumber);
```

marius-at-atb (approved):
Nice!

src/stacks-hierarchy/Root_ChooseTicketReceiverScreen/Root_ChooseTicketReceiverScreen.tsx
marius-at-atb: Or maybe you can even use `getAccountIdByPhone` directly, omitting `handleValidatePhoneNumber` altogether?

reidzeibel: Yeah apparently that works! Thanks for the suggestion!

rosvik (approved):
This looks good to me 💯

reidzeibel:
@gorandalum could you do a recheck on this one?

src/api/profile.ts
```diff
@@ -38,3 +39,37 @@ export async function deleteProfile(opts?: AxiosRequestConfig) {
...
+          error.response?.status === 404 || error.response?.status === 400,
+      },
+    )
+    .then((response) => response.data.customerAccountId as string)
+    .catch((error) => {
+      const metadata = getAxiosErrorMetadata(error);
+      const responseStatus = metadata.responseStatus;
+      if (responseStatus === 404 || responseStatus === 400) {
+        return 'no_associated_account';
+      }
```
gorandalum, line 72: I would still prefer returning `undefined` here, as account not found is a normal situation, and not considered an error.

I think this touches on this old school principle:
https://web.archive.org/web/20140430044213/http://c2.com/cgi-bin/wiki?DontUseExceptionsForFlowControl

src/api/profile.ts
```diff
@@ -38,3 +39,37 @@ export async function deleteProfile(opts?: AxiosRequestConfig) {
...
+ * @param phoneNumber phone number with prefix
+ * @returns {string | GetAccountByPhoneErrorCode}
+ *  - customer_account_id as string if successful,
+ *  - "no_associated_account" if account ID not found,
+ *  - "unknown_error" if there's other error
+ *
+ */
+export const getCustomerAccountId = async (
+  phoneNumber: string,
+): Promise<string | GetAccountByPhoneErrorCode> => {
```
gorandalum, line 55: I think returning string or error codes that are just strings is a little unspecific typing. If we want to return both error codes and string then I would either:
- Return `string` or `Error` (not errors for "normal" situations, like account not found).
- Type it up more specifically, maybe a return type like this: 
`{ number: string } | { errorCode: GetAccountByPhoneErrorCode }`

src/components/phone-input/PhoneInput.tsx
```diff
@@ -123,6 +132,12 @@ export const PhoneInput = ({
   );
 };
 
+function isError(error: string): error is PhoneInputErrorCode {
+  return ['invalid_phone', 'no_associated_account', 'unknown_error'].includes(
```
gorandalum, line 136: This should refer to constants. Then we would avoid having outdated values, like 'invalid_phone' is here (I think).

gorandalum (approved):
The solution mainly looks good, so it is approved. But there are some comments on best practice javascript/typescript usage, which I think is nice to be aligned with.

src/api/profile.ts
reidzeibel: Updated!

src/api/profile.ts
reidzeibel: Updated! No longer relevant since it now returns `string` or `undefined`

src/components/phone-input/PhoneInput.tsx
reidzeibel: Updated! Not needed anymore

PR #4250, opened 2024-01-18
Only use giveFocus timeouts for android

marius-at-atb:
The timeouts caused the iOS screenreader to be set incorrectly, and end up in a locked state.

Only use timeouts for Android.

Closes https://github.com/AtB-AS/kundevendt/issues/8619#issuecomment-1898076894

### DIFF
--- a/src/utils/use-focus-on-load.ts
+++ b/src/utils/use-focus-on-load.ts
@@ -3,6 +3,7 @@ import {
   AccessibilityInfo,
   findNodeHandle,
   InteractionManager,
+  Platform,
 } from 'react-native';
 import {useNavigationSafe} from '@atb/utils/use-navigation-safe';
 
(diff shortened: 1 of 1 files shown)


src/utils/use-focus-on-load.ts
```diff
@@ -38,15 +39,21 @@ export function useFocusOnLoad(setFocusOnLoad: boolean = true): Ref<any> {
...
   if (focusRef && focusRef.current) {
     InteractionManager.runAfterInteractions(() => {
-      setTimeout(() => {
-        const reactTag = findNodeHandle(focusRef?.current);
-        reactTag && AccessibilityInfo.setAccessibilityFocus(reactTag);
-      }, 100);
+      if (Platform.OS === 'ios') {
+        focusOnRef(focusRef);
+      } else {
+        setTimeout(() => focusOnRef(focusRef), 100);
```
jorelosorio, line 55: I know is not ready yet but would be nice to invalidate the timer in case of a user go back or something else, before the timer is fired otherwise will create an issue. It looks like this is impossible during the 100ms delay, but might be safe to call `clearInterval(intervalId)`

marius-at-atb: I like this suggestion!

marius-at-atb:
Closed in favor of https://github.com/AtB-AS/mittatb-app/pull/4254

PR #4251, opened 2024-01-18
Refactor use-ticket-info and remove Root_CarnetDetailsScreen

reidzeibel:
fixes https://github.com/AtB-AS/mittatb-app/pull/4228#pullrequestreview-1823146388

This PR will : 
- Remove navigation part from use-ticket-info hook
- Integrate carnet ticket detail screen to Root_FareContractDetailsScreen
- Remove Root_CarnetDetailsScreen

Acceptance criteria: 
- [ ] Carnet ticket details shows correctly
- [ ] Regular ticket details shows correctly
- [ ] Can navigate to ticket detail screen for both regular and carnet tickets


src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
```diff
@@ -1,29 +1,43 @@
...
 import React from 'react';
 import {ScrollView, View} from 'react-native';
 import {RootStackScreenProps} from '../stacks-hierarchy/navigation-types';
 import {useTimeContextState} from '@atb/time';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
+import {useAnalytics} from '@atb/analytics';
 
 type Props = RootStackScreenProps<'Root_FareContractDetailsScreen'>;
 
+export type TicketType = 'normal' | 'carnet';
```
jorelosorio, line 15: I can suggest to align the ticket type to something more aligned, `normal` perhaps is a type that might not be on our wording.

This is what I have heard:

| 'PreActivatedSingleTicket'
    | 'PreActivatedPeriodTicket'
    | 'CarnetTicket'
    | 'NightTicket'
    | 'SingleBoatTicket'
    | 'PeriodBoatTicket'
    | 'UnknownTicket';
    
    There is something called `NormalTravelRight` 🤔

src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
```diff
@@ -1,29 +1,43 @@
...
   useApplePassPresentationSuppression();
 
+  const navigateToTicketInfoScreen = () => {
+    if (ticketInfoParams) {
+      analytics.logEvent(
+        'Ticketing',
+        'Ticket information button clicked',
+        ticketInfoParams,
+      );
+      navigation.navigate('Root_TicketInformationScreen', ticketInfoParams);
```
jorelosorio, line 37: It might not be part of this PR, but we extract `ticketInfoParams` from `useTicketInfo(route.params.orderId)` and send it to `Root_TicketInformationScreen` which inside finds the `preassignedFareProduct` that we already have as a result from `useTicketInfo(route.params.orderId)`. It would be better to just send the `TicketInfo` to the screens to avoid search the same object multiple times. I saw this patter on many other places, so might not require an action here.

marius-at-atb: I remember it was a discussed in [this PR](https://github.com/AtB-AS/mittatb-app/pull/4127#discussion_r1427773994).

Basically the guideline for using navigation parameters is to pass ids, not entire objects. So I guess that is why this pattern is found in many places.

src/stacks-hierarchy/navigation-types.ts
```diff
@@ -45,16 +45,13 @@ type PaymentParams = {
...
-  orderId: string;
-};
-
 type ReceiptScreenRouteParams = {
   orderId: string;
   orderVersion: string;
 };
 
 type FareContractDetailsRouteParams = {
+  ticketType: 'normal' | 'carnet';
```
marius-at-atb, line 54: Should use the same `TicketType` variable as defined in `src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx`

marius-at-atb (approved):
Looks good!
I'm not sure if we actually need the `ticketType` parameter.
In `getFareContractInfoDetails` this info is found like 
```ts
const carnetTravelRights =
    fareContract.travelRights.filter(isCarnetTravelRight);
  const isACarnetFareContract = carnetTravelRights.length > 0;
```
where `isCarnetTravelRight` is a util function.

How about defining a new util like:
```ts
export function isCarnet(fareContract: FareContract) {
  return fareContract.travelRights.some(isCarnetTravelRight);
}
```
And then use it both in `Root_FareContractDetailsScreen.tsx` and `getFareContractInfoDetails`

src/stacks-hierarchy/navigation-types.ts
reidzeibel: Now that we can use the function to check if it is a carnet ticket, we should not need to pass on parameters like this, just check the `fareContract` and it just works.

src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
reidzeibel: Apparently we can use `fareContract` to find out if it is a carnet or not, so this type is no longer needed.

marius-at-atb (approved):
Lgtm!

src/stacks-hierarchy/Root_FareContractDetailsScreen.tsx
gorandalum: That is correct, we send ids not objects. Because when using ids we get up-to-date data from the Firestore subscription, but if we start sending around objects then we risk that the data becomes outdated.

tormoseng:
👍

Issue #807, opened 2020-12-31
NdkLinkError in Runtime.java:366


mikaelbr:
## Error in MittAtb

**NdkLinkError** in **Runtime.java:366**
Native library could not be linked. Bugsnag will not report NDK errors. See https://docs.bugsnag.com/platforms/android/ndk-link-errors

[View on Bugsnag](https://app.bugsnag.com/atb-as/mittatb/errors/5feddf66b1d9fa0017629187?event_id=5feddf6600659649bf6e0000&i=gh&m=ci)

## Stacktrace

    Runtime.java:366 - java.lang.Runtime.loadLibrary

[View full stacktrace](https://app.bugsnag.com/atb-as/mittatb/errors/5feddf66b1d9fa0017629187?event_id=5feddf6600659649bf6e0000&i=gh&m=ci)

*Created automatically via Bugsnag*

PR #4253, opened 2024-01-19
chore: update button component

adriawh:


src/components/button/Button.tsx
```diff
@@ -38,8 +44,8 @@ const DefaultModeStyles: {[key in ButtonMode]: ButtonSettings} = {
 };
 
 type ButtonTypeAwareProps =
-  | {text?: string; type: 'inline'}
   | {text: string; type?: 'block'}
+  | {text?: string; type: 'inline'}
   | {text: string; type: 'pill'};
```
gorandalum, line 49: In Figma now the types are `large | medium | small`, and not `block | inline | pill`.

src/components/button/Button.tsx
```diff
@@ -48,20 +54,21 @@ type ButtonIconProps = {
...
-  viewContainerStyle?: StyleProp<ViewStyle>;
-  textContainerStyle?: StyleProp<ViewStyle>;
-  textStyle?: StyleProp<TextStyle>;
   leftIcon?: ButtonIconProps;
   rightIcon?: ButtonIconProps;
   active?: boolean;
   compact?: boolean;
   loading?: boolean;
   style?: StyleProp<ViewStyle>;
+  hasShadow?: boolean;
```
gorandalum, line 69: Not any shadows on the component page in Figma. Ask christian if we still need it. Then it should be in Figma as well.

Or another option is wrapping the button with a view with shadow.

gorandalum (changes requested):
Good work! But replace the types so they are aligned with Figma.

src/components/button/Button.tsx
adriawh: @ckbilstad decided to not add in figma, so wrapped in a view

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -280,23 +279,9 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
   paddedContainer: {
     paddingHorizontal: theme.spacings.medium,
   },
-  purchaseButton: {
```
adriawh, line 283: found this unused style when looking into shadows

src/components/map/components/ShadowWrapper.tsx
adriawh, line 1: Noticed that this sends this warning:

>(ADVICE) View #807 of type RCTView has a shadow set but cannot calculate shadow efficiently. Consider setting >a background color to fix this, or apply the shadow to a more specific component.

However, by setting a backgroundcolor it will make it look different while being pressed because of the opacity of the button. Cannot apply the shadow to a more sepcific component, as that would be the button itself, which should not have a shadow attribute.

gorandalum: How does it look without the shadows? Is it noticable?

adriawh: Yes, it is quite noticeable and I don't belive removing it would be a good option. 

| Skjermbilde 2024-02-01 kl  13 50 02 | Skjermbilde 2024-02-01 kl  13 50 11 |
|:---:|:---:|

It is especially noticeable if the buttons are placed over a solid bright background.

| Skjermbilde 2024-02-01 kl  13 51 46 | Skjermbilde 2024-02-01 kl  13 51 53 |
|:---:|:---:|

src/components/button/Button.tsx
adriawh: Decided to go with this solution as it was the easiest way to solve it in regards to the button structure

adriawh:
Re-Added the hasShadows, Should be ready for review again now @gorandalum

gorandalum (approved):
Good work! :+1:

adriawh:
Just merged this big one into master, it touches every regular button in the app but they should not appear any different. But here's a heads up incase you notice anything while testing other issues @tormoseng @HanneLH

PR #4256, opened 2024-01-19
fix: invalidate all use queries

hanne-at-atb:
* invalidating all queries on change of userId
* move checkPermissions to notificationContext and only execute when authStatus is authenticated

Closes: https://github.com/AtB-AS/kundevendt/issues/16268


gorandalum (approved):
Nice! :+1:

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -77,10 +79,16 @@ export const RootStack = () => {
...
   const navRef = useNavigationContainerRef<RootStackParamList>();
   useFlipper(navRef);
+  const {userId} = useAuthState();
+  const queryClient = useQueryClient();
 
   useBeaconsState();
   useTestIds();
 
+  useEffect(() => {
+    queryClient.invalidateQueries();
```
jorelosorio, line 89: I have a small question here, should we somehow know that `authStatus === 'authenticated' is not authenticated or logged out before invalidate. It might be that the app enters to this useEffects always when the app restart?

jorelosorio (approved):
Small comment but code looks good, nice work.

rosvik (approved):

src/stacks-hierarchy/RootStack.tsx
hanne-at-atb: Removed it here and now rather doing the invalidation on log out and login both for Vipps and with one time code :) 
@gorandalum or @marius-at-atb would possibly have some input on this solution vs. the previous one with useEffect in RootStack?

gorandalum (commented):
I liked the simple `useEffect` better. Was there any negative side-effects from using it?

jorelosorio:
> I liked the simple `useEffect` better. Was there any negative side-effects from using it?

Hi, Yes. @hanne-at-atb and I were looking at this useEffect, and it works as soon as the user do everything in the same session or without refresh the app but, as soon as the user closes and open the app the useEffect is trigged and invalidation everything. So to avoid that, then the solution was moved it to a place when is needed and avoid being fired many times.

gorandalum:
> > I liked the simple `useEffect` better. Was there any negative side-effects from using it?
> 
> @gorandalum, Yes. @hanne-at-atb and I were looking at this useEffect, and it works as soon as the user do everything in the same session or without refresh the app but, as soon as the user closes and open the app the useEffect is trigged and invalidation everything. So to avoid that, then the solution was moved it to a place when is needed and avoid being fired many times.

As discussed in the office, react-query is only in-memory, so no problem just invalidating on app startup.

> React Query, per default, only stores data in-memory and nowhere else. If you reload your browser page, the cache is gone.

PR #4257, opened 2024-01-19
feat: Not navigate after login

gorandalum:
Two main changes in this PR:
- We complete onboarding before the last step of the authentication
process, like sending in the confirmation code. This is so we don't
end up in a race condition between Firebase Auth switching user and
setting the onboarded flag to true.
- We do not navigate anywhere after login, as changing user is
basically a reload of the app, and we end up back at the starting
screen no matter what.

### DIFF
--- a/src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
+++ b/src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
@@ -1,7 +1,10 @@
-import React, {useEffect, useState} from 'react';
+import React, {useState} from 'react';
 import {LoginTexts, useTranslation} from '@atb/translations';
-import {useAuthState} from '@atb/auth';
-import {ConfirmationErrorCode, PhoneSignInErrorCode} from '@atb/auth';
+import {
+  ConfirmationErrorCode,
+  PhoneSignInErrorCode,
+  useAuthState,
+} from '@atb/auth';
 import {useFocusOnLoad} from '@atb/utils/use-focus-on-load';
 import {
   ActivityIndicator,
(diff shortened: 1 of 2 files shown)


rosvik (approved):
LGTM 👍

src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
```diff
@@ -20,35 +23,34 @@ import {StyleSheet, useTheme} from '@atb/theme';
...
     ConfirmationErrorCode | PhoneSignInErrorCode
   >();
   const [isLoading, setIsLoading] = useState(false);
   const focusRef = useFocusOnLoad();
-  const completeOnboardingAndEnterApp = useCompleteOnboardingAndEnterApp();
+  const {completeOnboarding} = useAppState();
 
   const onLogin = async () => {
     setIsLoading(true);
+    completeOnboarding();
```
jorelosorio, line 48: So if the user was onboarded, but code fails or wrong / expired / etc, does it end as completed but not fully completed because of the error?

marius-at-atb: @jorelosorio Yes. If the error code fails, and the app is restarted, it will continue as if you chose continue without logging in. This is not great obviously, but not a catastrophe either I guess 😅 
Ideally we would set the `onboarded` flag when we know there was no error and before entering the app. (What we want to avoid is entering the app before the onboarded flag has been set.)

marius-at-atb (approved):
Looks good!

We may want to come back to the issue of where `completeOnboarding` is called to avoid a current edge case that will likely confuse users if they encounter it.

marius-at-atb:
Btw we may also want to rename `onboarded` and `completeOnboarding` since there are still permission screens as part of the onboarding coming after it.
E.g. `loginOnboardingCompleted`, `initialOnboardingCompleted`, `unconditionalOnboardingCompleted` or something similar could be more precise for the way it is currently used.

src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
gorandalum: > So if the user was onboarded, but code fails or wrong / expired / etc, does it end as completed but not fully completed because of the error?

Note that the function `completeOnboarding` only sets the onboarded flag to `true` in local storage, it doesn't navigate or something. So if the the user types wrong code, he stays in the onboarding with an error msg saying "wrong code", and can try again or request a new one.

PR #4260, opened 2024-01-22
fix: Two prompts for bluetooth

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/16576

### DIFF
--- a/ios/Shared/BeaconsPermissions.swift
+++ b/ios/Shared/BeaconsPermissions.swift
@@ -66,7 +66,9 @@ class BluetoothPermission: NSObject, PermissionRequestable, CBCentralManagerDele
 
   func request(callback onComplete: @escaping PermissionCallback) {
     self.onComplete = onComplete
-    bluetoothManager = CBCentralManager(delegate: self, queue: nil)
+    // A fix for double prompt for Bluetooth when CBCentralManagerOptionShowPowerAlertKey is not explicitly set to false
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):
LGTM

src/beacons/permissions.ts
```diff
@@ -115,7 +115,17 @@ export const checkPermissionStatuses = async () => {
...
+    const majorVersionIOS = parseInt(Platform.Version, 10);
+    // For iOS 13 and above, Bluetooth permission is not required
+    // Then it should send bluetooth as on in the kettle module
+    // NOTE: This is a fix for iOS 12 and above were the SDK crashes if bluetooth is not on
+    if (majorVersionIOS < 13) {
+      kettleModulesArray.push(KettleModules.BLUETOOTH);
+    } else if (permissionStatuses.bluetooth) {
+      kettleModulesArray.push(KettleModules.BLUETOOTH);
+    }
+  } else if (permissionStatuses.bluetooth) {
```
marius-at-atb, line 128: To make this logic more clear I can suggest something like
```suggestion ts
  const majorPlatformVersion = parseInt(String(Platform.Version), 10);
  const bluetoothPermissionNotNeeded =
    Platform.OS === 'ios' && majorPlatformVersion < 13;

  if (permissionStatuses.bluetooth || bluetoothPermissionNotNeeded) {
```

marius-at-atb (approved):

PR #4265, opened 2024-01-23
feat: hide new pill for privacy choice if beacons are disabled

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/16587

Not able to test locally, so maybe the reviewer with a functional setup could verify this works as intended? 🙏

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
@@ -301,24 +301,14 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
             onPress={() => navigation.navigate('Profile_LanguageScreen')}
             testID="languageButton"
           />
-          {isBeaconsEnabled ? (
-            <LinkSectionItem
-              text={t(
-                ProfileTexts.sections.settings.linkSectionItems.privacy.label,
-              )}
-              label="new"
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -299,14 +301,24 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
...
           />
-          <LinkSectionItem
-            text={t(
-              ProfileTexts.sections.settings.linkSectionItems.privacy.label,
-            )}
-            label="new"
-            onPress={() => navigation.navigate('Profile_PrivacyScreen')}
-            testID="privacyButton"
-          />
+          {isBeaconsEnabled ? (
```
marius-at-atb, line 304: Haven't tested, but is it possible to just use `label={isBeaconsEnabled ? "new" : undefined}` instead of duplicating it?

toretefre: Absolutely, good idea 🙌

jorelosorio (approved):
Good, it is also possible to `isBeaconsSupported` that also checks that there is no variable set on .env. Both works.

toretefre:
Changing to `isBeaconsSupported` since it seems to be preferrable

marius-at-atb (approved):

PR #4268, opened 2024-01-24
Revert "fix: Add a new target to handle beacons"

jorelosorio:
This revert changes made on iOS where two targets were created in order to avoid wrong configurations copies or permissions for those apps that not need extra features.
Compiling this run OK, but the build breaks on runtime, so reverting will allow compiling a build to allow QA team to test.


reidzeibel (approved):

jorelosorio:
Tested and new builds works as expected.

PR #4270, opened 2024-01-24
fix: change token screen back button

marius-at-atb:
Since the screen transitions in popping up, it should have a close button instead of back button.
Also deletes a now unused profile screen.

Resolves https://github.com/AtB-AS/kundevendt/issues/8619#issuecomment-1903824259


reidzeibel (approved):
LGTM!

PR #4271, opened 2024-01-24
feat: add benefit tiles to fare contract details

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/16297

https://github.com/AtB-AS/mittatb-app/assets/1774972/f42010bb-1c05-41e3-94c3-a2f6e5c24adb


src/mobility/api/api.ts
rosvik, line 1: I'm renaming things here so that they're not confused with `FareProductBenefitType` in use-operator-benefits-for-fare-product.tsx.

src/fare-contracts/details/DetailsContent.tsx
```diff
@@ -170,6 +179,18 @@ export const DetailsContent: React.FC<Props> = ({
...
+            <ThemeText type="body__secondary" color="secondary">
+              {t(FareContractTexts.label.indludedBenefits)}
+            </ThemeText>
+            <BenefitTiles
+              benefits={benefits}
+              onNavigateToMap={onNavigateToMap}
+              interactiveColor="interactive_2"
+            />
+          </GenericSectionItem>
+        )}
```
gorandalum, line 193: To follow the pattern from the other benefit stuff, could this be a section item? We have `MobilityBenefitsInfoSectionItem`, maybe this could be `MobilityBenefitsActionsSectionItem`.

src/components/tile/TileWithButton.tsx
```diff
@@ -55,25 +56,27 @@ export function TileWithButton({
...
+          >
+            {buttonText}
+          </ThemeText>
+          <ThemeIcon
+            size={mode === 'spacious' ? 'normal' : 'xSmall'}
+            svg={buttonSvg}
+            fill={color.outline.text}
+          />
+        </View>
+      )}
```
gorandalum, line 79: Can the button be disabled instead of removing it? I don't think it makes sense to have a `TileWithButton` without the button. If they want an info box without button they can show a `BorderedInfoBox` instead.

Also by skimming the code, maybe we doesn't need support for undefined `onPress` at this time.

src/mobility/components/BenefitTile.tsx
```diff
@@ -0,0 +1,149 @@
...
+  MapFilterType,
+  MobilityMapFilterType,
+  useUserMapFilters,
+} from '@atb/components/map';
+import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
+import {getNewFilterState} from '../utils';
+import {useOperators} from '../use-operators';
+
+type BenefitCardProps = {
+  interactiveColor: InteractiveColor;
```
gorandalum, line 21: Is this parameter used?

src/mobility/components/BenefitTile.tsx
```diff
@@ -0,0 +1,149 @@
...
+  style?: ViewStyle;
+  benefit: FareProductBenefitType;
+  onNavigateToMap?: (initialFilters: MapFilterType) => void;
```
gorandalum, line 24: Can this really be `undefined`?

src/mobility/components/BenefitTile.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const {getMapFilter, setMapFilter} = useUserMapFilters();
+  const operators = useOperators();
+
+  const onPress = onNavigateToMap
+    ? async () => {
+        const storedFilters = await getMapFilter();
+
+        let mobilityFilters: MobilityMapFilterType = storedFilters.mobility;
+        benefit.formFactors.forEach((formFactor) => {
+          const allOperators = operators.byFormFactor(FormFactor.Scooter);
```
gorandalum, line 48: May just be me missing the point, but should it be `byFormFactor(FormFactor.Scooter)` here?

rosvik: Good catch! You're not missing anything, just some temporary code I forgot to replace. Interesting that it still worked like that, tho 😄

src/mobility/components/BenefitTile.tsx
rosvik: Not more than once. Would you prefer if I didn't expose it?

src/mobility/use-operator-benefits-for-fare-product.tsx
```diff
@@ -12,21 +12,36 @@ export type FormFactorTicketDescriptionPair = {
...
+export type FareProductBenefitType = {
+  operatorId: string;
+} & OperatorBenefitType;
+
 export const useOperatorBenefitsForFareProduct = (
   productId: PreassignedFareProductId | undefined,
 ): {
   status: UseQueryResult['status'];
-  benefits?: OperatorBenefitType[];
+  benefits?: Array<FareProductBenefitType>;
```
reidzeibel, line 23: What is the difference between `Array<FareProductBenefitType>` and `FareProductBenefitType[]`, I tried looking in here https://www.typescriptlang.org/docs/handbook/2/generics.html and it looks like using both `Array<Type>` and `Type[]` works. 

It was originally `OperatorBenefitType[]`, why the new type is not `FareProductBenefitType[]` and instead an `Array<FareProductBenefitType>`?

rosvik: Good point, there shouldn't be a difference! Changing it back, since there's no point in changing it, and I think `[]` makes it more readable in this case 👍 

I usually use `Array<>` to make things more readable if it's a long complex type, which this was at some point during the making of this PR.

reidzeibel: If it was same, then it should be fine, because I really don't know which ones are better or more used in our code. 🤣

reidzeibel (approved):
⏰ L G T M ⏰

src/mobility/components/BenefitTile.tsx
rosvik: Not anymore 👍

src/fare-contracts/details/DetailsContent.tsx
rosvik: Sure, updated it now.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_RootScreen.tsx
```diff
@@ -9,12 +10,18 @@ import {useIsScreenReaderEnabled} from '@atb/utils/use-is-screen-reader-enabled'
...
+};
+
 export const Map_RootScreen = ({
   navigation,
+  route,
 }: MapScreenProps<'Map_RootScreen'>) => {
   const isScreenReaderEnabled = useIsScreenReaderEnabled();
-  const vehicles = useVehicles();
-  const stations = useStations();
+  const vehicles = useVehicles(route.params?.initialFilters?.mobility);
```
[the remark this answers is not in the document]
rosvik: sure :)

jorelosorio (approved):
Looks good, :) I have a question when selecting and navigating to the map seems like it shows what the ticket specify isn't? or should be all categories of shared mobility enabled? Bikes: All, Car: All instead of only Hyre?

rosvik:
@jorelosorio I talked to Christian about this, and we landed on only enabling the filter for the specific opertator. Since if you have e.g. Hyre included, it might be confusing if a bunch of Otto cars show up. It would make it harder to find the one kind you have included when they look the same in the map.

gorandalum (approved):

PR #4273, opened 2024-01-24
fix: Remove beacons usage if not in use

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/16030 (Read internally for more details)


marius-at-atb (approved):
Lgtm

PR #4275, opened 2024-01-24
chore: use correct nav flow logic for extended onboarding

marius-at-atb:
The extended onboarding wasn't properly supported by the new onboarding flow.
One issue with it was that if you restarted the app after completing it, you had to go through it again.
Now the implementation is in line with the new way to deal with the onboarding flow.

- Added `extendedOnboardingOnboarded` flag in AppContext (not a great name, but refactoring `AppContext` will have to wait)
- Using `extendedOnboardingOnboarded` in the new `useShouldShowExtendedOnboarding` in `useOnboardingFlow`

### DIFF
--- a/src/AppContext.tsx
+++ b/src/AppContext.tsx
@@ -11,6 +11,7 @@ import {register as registerChatUser} from './chat/user';
 import {storage} from '@atb/storage';
 
 enum storeKey {
+  extendedOnboardingOnboarded = '@ATB_extended_onboarding_onboarded',
   userCreationOnboarded = '@ATB_onboarded', // variable renamed, but keep old storage key
   mobileTokenOnboarding = '@ATB_mobile_token_onboarded',
   mobileTokenWithoutTravelcardOnboarding = '@ATB_mobile_token_without_travelcard_onboarded',
@@ -20,6 +21,7 @@ enum storeKey {
 }
 type AppState = {
   isLoading: boolean;
+  extendedOnboardingOnboarded: boolean;
   userCreationOnboarded: boolean;
   mobileTokenOnboarded: boolean;
   mobileTokenWithoutTravelcardOnboarded: boolean;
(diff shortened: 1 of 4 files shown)


src/AppContext.tsx
```diff
@@ -11,6 +11,7 @@ import {register as registerChatUser} from './chat/user';
 import {storage} from '@atb/storage';
 
 enum storeKey {
+  extendedOnboardingOnboarded = '@ATB_extended_onboarding_onboarded',
```
gorandalum, line 14: Will this rerun the extended onboarding for those who have already seen it? Or is this just readding a previously removed storage value?

src/AppContext.tsx
```diff
@@ -20,6 +21,7 @@ enum storeKey {
 }
 type AppState = {
   isLoading: boolean;
+  extendedOnboardingOnboarded: boolean;
```
jorelosorio, line 24: Is not a bit intuitive the meaning of this attr.

src/AppContext.tsx
```diff
@@ -257,6 +287,20 @@ export const AppContextProvider: React.FC = ({children}) => {
     restartShareTravelHabitsOnboarding,
   } = useMemo(
     () => ({
+      completeExtendedOnboarding: async () => {
+        await storage.set(
+          storeKey.extendedOnboardingOnboarded,
+          JSON.stringify(true),
```
jorelosorio, line 293: It can be also just as 'true'

src/AppContext.tsx
marius-at-atb: It would have if it wasn't for [the second commit](https://github.com/AtB-AS/mittatb-app/pull/4275/commits/34d05b7f882c72e74b15a4f23a14bac0be582ea3).
It is a brand new storage key afaik, not sure if we have a list of previously used ones?

src/AppContext.tsx
marius-at-atb: Yeah I agree. Problem was that the pattern in `AppContext` is a flag with `Onboarded` at the end, and also there is a flag in remote config `enable_extended_onboarding`. So either break the pattern, introduce a new (and kinda duplicate) remote config variable, or accept `extendedOnboardingOnboarded` 😬

src/AppContext.tsx
marius-at-atb: Yep. I think AppContext could use a refactor to simplify and have less duplicate code for each variable.
For now keeping it like this since it follows the exact pattern of all the other ones

gorandalum (approved):

PR #4277, opened 2024-01-24
test: make the e2e tests run

tormoseng:
The E2E-tests started a while ago to fail with the message `SyntaxError: Cannot use import statement outside a module` (e.g. [link](https://github.com/AtB-AS/mittatb-app/actions/runs/7296810814)). It seems that there is a problem with the tsconfig and what module is used. A bit peculiar since no change happened for the tests at least from one day to another. 

However, after some tries, I've taken the tests out of the main `package.json` and created a separat `package.json` and `tsconfig.json` in the e2e-folder.

Along with this change, I'm not able to make the import look like e.g. `import ElementHelper from '../utils/element.helper';`, but need to have it as `import ElementHelper from '../utils/element.helper.ts';`. Anyway, I didn't think the effort in gaining it was worth the effort.

In addition, these changes are done:
- Added a test-id for the button text in the Button component
- Updated the config files with new paths
- Updated the workflow-yml files with new paths
- Updated the emulator used to the latest
- Updated tests with a new way to handle scroll - as the previous was deprecated


reidzeibel (commented):
@tormoseng Is the `DEFAULT_BRANCH: 'main'` correct? If this refers to to mittatb repo, we use `master` instead of `main`.

tormoseng:
> @tormoseng Is the `DEFAULT_BRANCH: 'main'` correct? If this refers to to mittatb repo, we use `master` instead of `main`.

You are so correct! Was working in parallell with webshop 🤦  Fixed now

reidzeibel (approved):
Looks good to me sir, thank you!

PR #4283, opened 2024-01-26
fix: favorite toggle departures view with screen reader

adriawh:
Making the favorite star in departures accessible with screen reader.

https://github.com/AtB-AS/mittatb-app/assets/70323886/0e9586fd-8cac-45a3-b754-04cfa99f4a1f

A new issue to improve this flow has also been created: https://github.com/AtB-AS/kundevendt/issues/16648


jorelosorio (approved):
Looks good :)

src/place-screen/components/EstimatedCallItem.tsx
```diff
@@ -108,16 +111,20 @@ export const EstimatedCallItem = memo(
...
-                onMarkFavourite={() =>
-                  onPressFavorite(departure, existingFavorite)
-                }
-              />
+            {mode !== 'Favourite' && (
+              <DepartureTime accessible={false} departure={departure} />
             )}
           </View>
+
+          {mode !== 'Map' && (
```
jorelosorio, line 119: If not a Map mode, then is Favourite? and if not a Favourite is, then a Map? not sure if we could change the conditionals to understand which component is associated with what.

src/place-screen/components/EstimatedCallItem.tsx
```diff
@@ -108,16 +111,20 @@ export const EstimatedCallItem = memo(
...
+            <FavouriteDepartureToggle
+              existingFavorite={existingFavorite}
+              toggleFavouriteAccessibilityLabel="Favorite"
```
gorandalum, line 122: Should be translated? And do we need some a11y hints?

src/place-screen/components/EstimatedCallItem.tsx
adriawh: its the fact that the component is used several places, and the star is not to be shown in all of them

PR #4289, opened 2024-01-30
fix: Fix avoid Apple complains about missing keys in infoPlist

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/16030

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -81,9 +81,7 @@ platform :ios do
     unless beacons_enabled
       languages = ['nb', 'en', 'nn']
       keys = [
-        'NSBluetoothAlwaysUsageDescription', 'NSBluetoothPeripheralUsageDescription',
-        'NSLocationAlwaysAndWhenInUseUsageDescription', 'NSLocationAlwaysUsageDescription',
-        'NSMotionUsageDescription'
+        'NSLocationAlwaysUsageDescription'
       ]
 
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -103,10 +101,28 @@ platform :ios do
...
+            key = match[1]
+            value = match[2]
+            lines_hash[key] = value
+          end
+        end
 
-        File.open(file_path, 'w') do |file|
-          file.puts(updated_lines)
+        # Apple requires that the key NSLocationAlwaysAndWhenInUseUsageDescription is present even if the app only uses when in use
+        # because some of the libraries we use require always permission even if we don't use it.
```
rosvik, line 115: > some of the libraries we use require always permission

Not actually connected to your changes, but this made me check the App Store to see what permissions is displayed there, and sure enough, they do say that we could use location in the background for 1.45.

Screenshot 2024-01-30 at 14 00 00

Not sure how that works, but is it something we can change about the related libraries?

rosvik: Same applies to FRAM / NFK

rosvik (approved):

fastlane/Fastfile
jorelosorio: In this PR the background location was removed, I think it might change that message as well, I hope.

PR #4290, opened 2024-01-31
feat: streamline onboarding logic with config and useOnboardingSections

marius-at-atb:
Simplifications:
- `onboarded` flags are handled generally in `AppContext`, allowing lots of duplicated code to be removed
- Repeated code in `Profile_DebugInfoScreen` replaced
- `completeOnboardingSection` is called in `continueFromOnboardingSection`, removing need for calling custom complete functions before continuing

The onboarding flow is configured statically in `staticOnboardingSectionsInPrioritizedOrder`. The combination of this and the general solution for `onboarded` flags in `AppContext`, actually opens up the possibility of fetching the onboarding config remotely and running A/B user tests in the future.

Note that a distinction has been made between onboarding sections and onboarding screens, as a part of the onboarding can be more than just one screen. E.g. `getNextOnboardingScreen` has been renamed to `getNextOnboardingSection` for this reason.

Essentially this simplifies and enforces correct usage of the new onboarding flow.

Resolves https://github.com/AtB-AS/kundevendt/issues/16700
Also relevant issue: https://github.com/AtB-AS/kundevendt/issues/16699


gorandalum:
I haven't really taken a look yet, just a glance, but I think it looks like a nice improvement.

Just note that I think this should be added to the board as a stand-alone entity as the task it is linked to is already approved, and its storypoints was counted for in the last sprint.

marius-at-atb:
@gorandalum Good thinking! Issue created and linked now.

src/utils/use-onboarding-flow.ts
```diff
@@ -133,97 +66,28 @@ export const useOnboardingFlow = (
...
-      ? pushNotificationPermissionStatus === 'undetermined'
-      : pushNotificationPermissionStatus === 'denied';
-
-  return (
-    !notificationPermissionOnboarded &&
-    pushNotificationsEnabled &&
-    pushNotificationPermissionsNotGranted &&
-    validFareContracts.length > 0 &&
-    hasFareContractWithActivatedNotification
-  );
```
marius-at-atb, line 228: Note: these useShouldShow hooks have been moved to `use-onboarding-sections.ts`, and the prefix is instead useCustomShouldShow. Also the onboarded state from `AppContext` (such as e.g. `notificationPermissionOnboarded`) is no longer needed as conditions inside these hooks, since that is now checked in `getNextOnboardingSection` anyway

src/utils/use-onboarding-sections.ts
```diff
@@ -0,0 +1,276 @@
...
+        ...loadedOnboardingSection,
+        customShouldShow: getCustomShouldShow(
+          loadedOnboardingSection.onboardingSectionId,
+        ),
+      })),
+    [getCustomShouldShow, loadedOnboardingSections],
+  );
+};
+
+const useGetCustomShouldShow = (
```
marius-at-atb, line 134: Do you think there are too many things gathered in this file?
If so, I guess it could be a good idea to move `useGetCustomShouldShow` and all the `useCustomShouldShow` hooks it uses to a new `use-onboarding-custom-should-show.ts` file.

src/utils/use-onboarding-navigation.ts
```diff
@@ -1,47 +1,56 @@
...
       InteractionManager.runAfterInteractions(() => {
+        if (!screen?.name) return;
         if (replace) {
-          navigation.dispatch(StackActions.replace(screenName, params));
+          navigation.dispatch(
+            StackActions.replace(screen.name, screen?.params),
+          );
         } else {
-          navigation.navigate(screenName, params);
+          navigation.navigate(screen.name, screen?.params);
```
jorelosorio, line 25: should not be `screen?.name` as `screen.params?`

src/utils/use-onboarding-sections.ts
```diff
@@ -0,0 +1,276 @@
...
+
+  const customShouldShowMobileTokenWithoutTravelcardOnboarding =
+    useCustomShouldShowMobileTokenWithoutTravelcardOnboarding();
+
+  return useCallback(
+    (onboardingSectionId: OnboardingSectionId) => {
+      switch (onboardingSectionId) {
+        case 'extendedOnboarding':
+          return customShouldShowExtendedOnboarding;
+        // case 'userCreation':
```
jorelosorio, line 162: Perhaps remove the commented lines

src/utils/use-onboarding-sections.ts
```diff
@@ -0,0 +1,276 @@
...
+      customShouldShowLocationOnboarding,
+      customShouldShowShareTravelHabitsScreen,
+      customShouldShowNotificationPermissionScreen,
+      customShouldShowMobileTokenOnboarding,
+      customShouldShowMobileTokenWithoutTravelcardOnboarding,
+    ],
+  );
+};
+
+/* --- useCustomShouldShow hooks --- */
```
jorelosorio, line 189: All this hooks could be moved out of this file

src/utils/use-onboarding-sections.ts
```diff
@@ -0,0 +1,276 @@
...
+  return shouldShowTravelTokenOnboarding && !disable_travelcard;
+};
+
+const useCustomShouldShowMobileTokenWithoutTravelcardOnboarding = () => {
+  const shouldShowTravelTokenOnboarding = useShouldShowTravelTokenOnboarding();
+  const {disable_travelcard} = useRemoteConfig();
+  return shouldShowTravelTokenOnboarding && disable_travelcard;
+};
+
+/* --- utils --- */
```
jorelosorio, line 255: Same for utils, move it out if possible.

jorelosorio:
In general looks like a good improvement, there are some small comments here and there, however the namings that use wording as: should, extended, section and there are some others like: useOnboardingSectionIsOnboarded is not clear. comingFromOnboardingSectionId might be just previousSection or previousScreen. It is just an opinion that it looks a bit confuse.

src/utils/use-onboarding-navigation.ts
marius-at-atb: Not in this case because of `if (!screen?.name) return;` on line 19.
Basically params can be defined or undefined, but navigating to a screen with undefined screenName will fail and should never happen.

marius-at-atb: Aha, I read it the other way if (`screen.name` should be `screen?.name`). Updated :+1:

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DebugInfoScreen.tsx
```diff
@@ -182,6 +179,30 @@ export const Profile_DebugInfoScreen = () => {
...
+            text="Restart all onboarding sections"
+            onPress={() => restartAllOnboardingSections()}
+          />
+          <ExpandableSectionItem
+            text="Individual onboarding restarts"
+            showIconText={true}
+            expandContent={onboardingSections.map((onboardingSection) => {
+              const {onboardingSectionId} = onboardingSection;
+              return (
+                <LinkSectionItem
```
gorandalum, line 192: Using `LinkSectionItem` here doesn't really make sense as we can't have section item inside section item. I know it is just in the debug menu, but still think it should be changed so someone doesn't use this as an example of how to use `ExpandableSectionItem` at a later time.

src/utils/use-onboarding-sections.ts
marius-at-atb: Update: they have been moved

src/utils/use-onboarding-sections.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export type LoadedOnboardingSection = StaticOnboardingSection & {
+  isOnboarded: boolean; // loaded and added in AppContext
+};
+
+export type OnboardingSection = LoadedOnboardingSection & {
+  customShouldShow?: boolean; // calculated and added in useOnboardingSections
+};
+
+export const staticOnboardingSectionsInPrioritizedOrder: StaticOnboardingSection[] =
```
gorandalum, line 43: Nice!

src/utils/use-onboarding-get-custom-should-show.ts
```diff
@@ -0,0 +1,138 @@
...
+      customShouldShowLocationOnboarding,
+      customShouldShowShareTravelHabitsScreen,
+      customShouldShowNotificationPermissionScreen,
+      customShouldShowMobileTokenOnboarding,
+      customShouldShowMobileTokenWithoutTravelcardOnboarding,
+    ],
+  );
+};
+
+/* --- useCustomShouldShow hooks --- */
```
gorandalum, line 74: What does the term custom mean here actually? I don't understand at first glance.

src/utils/use-onboarding-navigation.ts
jorelosorio: The changes are not reflected yet 🤔

src/utils/use-onboarding-get-custom-should-show.ts
marius-at-atb: To determine whether an onboarding section should be shown, there is always a check on whether it is already onboarded, which is locally stored with AppContext.
In addition, for each onboarding section, there is also the option to add _custom conditions_ for whether it should be shown. E.g. for location permission onboarding, it should not show if location permission is already granted, and for travel token onboarding, it should only show if the token is active on another device..

src/utils/use-onboarding-navigation.ts
marius-at-atb: Maybe you have a commit filter active?
https://github.com/AtB-AS/mittatb-app/pull/4290/commits/b705ec786780bddda62d4494ac912fcab496b365

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DebugInfoScreen.tsx
marius-at-atb: True! Updated

marius-at-atb:
@jorelosorio
Regarding `extended`, I guess it is now kinda `prepended` since it is first. However it stays in sync with the name used in the existing `enable_extended_onboarding` remote config flag, and remains accurate even if e.g. FRAM wish to change the order.

`comingFromOnboardingSectionId` is not yet the previous onboardingSectionId when received as a parameter in `getNextOnboardingSection`, but it could be called `currentOnboardingSectionId`. Not sure if that will lead to better use/understanding than `comingFromOnboardingSectionId` however 🤔

marius-at-atb:
> `comingFromOnboardingSectionId` is not yet the previous onboardingSectionId when received as a parameter in `getNextOnboardingSection`, but it could be called `currentOnboardingSectionId`. Not sure if that will lead to better use/understanding than `comingFromOnboardingSectionId` however 🤔

Also an interesting distinction between `comingFromOnboardingSectionId` and `currentOnboardingSectionId` is that in order to update the `nextOnboardingSection` state variable, it should not specify the current onboarding section:
```ts
useEffect(() => {
    setNextOnboardingSection(
      getNextOnboardingSection(undefined, assumeUserCreationOnboarded),
    );
  }, [getNextOnboardingSection, assumeUserCreationOnboarded]);
```
In this case it is not coming from a special onboardingSection, so it is left undefined. With this in mind, I think it should still be called `comingFromOnboardingSectionId`

marius-at-atb:
For future reference: it can be considered if `initialScreen` should be added by `useOnboardingSections` instead of directly in `staticOnboardingSectionsInPrioritizedOrder`. If so, see [this commit](https://github.com/AtB-AS/mittatb-app/commit/aa765b2f846d0ec7f3fcc988121a75156b6adb03)

PR #4293, opened 2024-01-31
fix: show correct transportModeIcon for commercial express boat services

toretefre:
Closing https://github.com/AtB-AS/kundevendt/issues/16684


toretefre:
Is there any reason why we can't do this, @marius-at-atb ? Seems like this was introduced with https://github.com/AtB-AS/mittatb-app/pull/3736/files

The problem is that Møre og Romsdal has [commercial express boat](https://www.geirangerfjord.no/fjord-cruise-alesund-geiranger) options in the Entur data, so resetting all flex offers to bus does not work 😅

marius-at-atb:
@toretefre Good question! It currently uses `flex` to get the correct `themeColor` for AtB Bestill from `useThemeColorForTransportMode` inside `TransportationIconBox`, so we need to keep that functionality.
The easiest might be to only use `flex` info for `themeColor`, not the `svg` icon. I'll look further into this tomorrow.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/ResultItem.tsx
```diff
@@ -586,7 +586,7 @@ const TransportationLeg = ({
   return (
     <TransportationIconBox
       style={style}
-      mode={isLegFlexibleTransport(leg) ? 'flex' : leg.mode}
```
jorelosorio, line 589: I think it might be something else, this could break our flex. What it does that internally is to check if flexibleLineType is true or present.

marius-at-atb:
Update: me and @Sebstorvik will ask a few relevant people tomorrow to figure out what color should be shown when a leg is flexible but not with bus. The icon should definitely be correct for the `mode`.
It could be that we end up requiring `isFlexible && mode==='bus'` to show the orange flex color.

marius-at-atb:
It was decided that the icon should always be right from mode, and the color should be flexible for all flexible trips.
So basically we do as you suggest @toretefre, and then also add support for flex color (see new commit).
In the AtB theme the flex color is orange, so it now looks like this:

marius-at-atb:
Btw it would also be nice if someone has time to either continue this PR or create a new one, where `getTransportationColor` in `colors.ts` is deleted and replaced it with `useTransportationColor`?
And then also rename `useThemeColorForTransportMode` to `getThemeColorForTransportMode`.

gorandalum:
Note from my side:
- Good improvement showing correct transportation mode. Doesn't make sense that all flex lines are bus.
- The OMS should together decide if flex transports should have its own color or if they should just show the color for the transport mode.
- Marius told me that the boat transports in question here also shows up in departures, where we don't have support for showing flex lines. I was kinda surprised by this since other departures that needs booking are just normal non-flex departures, but with the `bookingArrangement` field. I think we together with Entur need need to align what differs a flex line from a non-flex line which requires booking.

gorandalum:
Now updated so also the map lines have correct color (ref. image above where icon still had correct color, but map line were grey when it should have been the flex color (orange).

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -252,7 +252,7 @@ const LiveVehicleMarker = ({
   isError,
 }: VehicleIconProps) => {
   const {theme} = useTheme();
-  const fillColor = useTransportationColor(mode, subMode, 'background');
+  const fillColor = useTransportationColor(mode, subMode, false, 'background');
```
marius-at-atb, line 255: Shouldn't it be orange if it is an AtB Bestill vehicle?

marius-at-atb (approved):
Nice!

src/api/types/trips.ts
```diff
@@ -10,6 +10,10 @@ export type TripPattern =
 export type TripMetadata = Required<Types.TripsQuery>['trip']['metadata'];
 export type Leg =
   Required<Types.TripsQuery>['trip']['tripPatterns'][0]['legs'][0];
+export type Line = Required<
+  Required<Types.TripsQuery>['trip']['tripPatterns'][0]['legs'][0]
```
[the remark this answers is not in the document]
gorandalum: Not really sure, I just followed the pattern. Maybe the thought is that `Types.TripsQuery` is the source and all types should be derived explicitly from that one.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
gorandalum: Yes, should have been, some reasons why not:
- It wasn't supported from before
- We don't have support for "See live" on flex travels in the app today
- We don't have enough data in the `TravelDetailsMapScreenComponent` to know if the live vehicle is a flex route or not, and it requires some slightly larger refactoring to work

marius-at-atb: I see. Not worth the extra work for now then.

PR #4294, opened 2024-02-01
feat: add popover for on-behalf–of

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/16033

Pretty straightforward PR, things added:

- ref to the new label.
- useFocusEffect to trigger popover when view is rendered and screen is shown.
- text for popover.

Preview below

image

image

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
@@ -83,11 +83,13 @@ export function TravellerSelection({
 
   useFocusEffect(
     useCallback(() => {
-      addPopOver({
-        oneTimeKey: 'on-behalf-of-new-feature-introduction',
-        target: onBehalfOfIndicatorRef,
-      });
-    }, [addPopOver]),
+      if (isOnBehalfOfEnabled) {
+        addPopOver({
+          oneTimeKey: 'on-behalf-of-new-feature-introduction',
+          target: onBehalfOfIndicatorRef,
+        });
+      }
+    }, [isOnBehalfOfEnabled, addPopOver]),
   );
 
   if (selectionMode === 'none') {
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from 'react';
+import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {AccessibilityProps, StyleProp, View, ViewStyle} from 'react-native';
 
 import {PurchaseOverviewTexts, useTranslation} from '@atb/translations';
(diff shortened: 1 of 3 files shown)


marius-at-atb (approved):
Lgtm!

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -187,7 +203,15 @@ export function TravellerSelection({
             </View>
 
             {/* remove new label when requested */}
-            {isOnBehalfOfEnabled && <LabelInfo label="new" />}
+            {isOnBehalfOfEnabled && (
+              <View
+                ref={onBehalfOfIndicatorRef}
+                renderToHardwareTextureAndroid={true}
```
jorelosorio, line 209: I am just curious about this, we haven't used it before.

gorandalum (approved):

jorelosorio:
@reidzeibel not sure if before a tap action on a tag was done before but looks strange without more context, is not clear that it is possible to tap on it and see more information, perhaps adds a ? icon or something. Once is clicked it should be more explicit about how it works isnt?

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: Sometimes the popover doesn't show in Android, we need to add that line based on this source: https://www.npmjs.com/package/react-native-popover-view#show-from-a-ref-not-working

reidzeibel:
@jorelosorio This popover view is just to notify users that there are something to look into. About your suggestion of adding a "?", do you have any example of how would it look? maybe a rough sketch or something?

jorelosorio:
> @jorelosorio This popover view is just to notify users that there are something to look into. About your suggestion of adding a "?", do you have any example of how would it look? maybe a rough sketch or something?

I guess that @oyvindmikkelsenatb could help us to identify it, because at least is not intuitive that we can tap on that label new at least it does not work for that for other places where is used.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
jorelosorio: Interesting, thanks for the explanation.

gorandalum:
> > @jorelosorio This popover view is just to notify users that there are something to look into. About your suggestion of adding a "?", do you have any example of how would it look? maybe a rough sketch or something?
> 
> I guess that @oyvindmikkelsenatb could help us to identify it, because at least is not intuitive that we can tap on that label new at least it does not work like that for other places where is used.

The label isn't clickable, the whole section item is. So the question is just where on the section item to point the arrow from the popover. You feel it should be pointed somewhere else @jorelosorio?

jorelosorio:
> > > @jorelosorio This popover view is just to notify users that there are something to look into. About your suggestion of adding a "?", do you have any example of how would it look? maybe a rough sketch or something?
> > 
> > 
> > I guess that @oyvindmikkelsenatb could help us to identify it, because at least is not intuitive that we can tap on that label new at least it does not work like that for other places where is used.
> 
> The label isn't clickable, the whole section item is. So the question is just where on the section item to point the arrow from the popover. You feel it should be pointed somewhere else @jorelosorio?

OK, I understand now, I missed then the part of the useEffect that opens it. However, would be nice explains a bit more how it works. But is not related to this task.

PR #4299, opened 2024-02-02
fix: crash when opening list of tickets

reidzeibel:
Fixes crash when opening list of tickets page.

Apparently checking `benefits?.length && validityStatus === 'valid'` results in 0, which then results in `shouldShowBundlingInfo` to be always defined. 

Since `shouldShowBundlingInfo` is always defined, the `MobilityBenefitsInfoSectionItem` will be rendered, but there are no value to be rendered, causing a crash.


marius-at-atb (approved):

gorandalum (commented):
:+1: 🙈

PR #4300, opened 2024-02-02
chore: onboarding context

marius-at-atb:
With `useOnboardingSections` as a hook, the calculations of the `shouldShow` onboarding values would be re-calculated many times for no good reason.

This PR
- renames `AppContext` to `OnboardingContext`, and moves it to be the innermost `ContextProvider` so that all other context and hooks are available for calculations of `shouldShow` values for onboarding sections
- moves the logic from the now deleted `useOnboardingSections` hook into `OnboardingContext`
- generalises the `onboarding-screen` folder to just `onboarding`, and moves hooks related to onboarding from utils into this folder
- replaces `isLoading` from `AppContext` with `useIsLoadingAppState`
- removes need for the annoying `utilizeThisHookForSessionCounting` param for `useShouldShowShareTravelHabitsScreen`
- replaces the `useCustomShouldShow` hooks with a combination of `useShouldShowArgs` and `shouldShowPredicate` methods directly in the config, and renames  `customShouldShow` to just `shouldShow`

Resolves https://github.com/AtB-AS/kundevendt/issues/16744


src/beacons/use-should-show-share-travel-habits-screen.tsx
```diff
@@ -37,19 +41,12 @@ export const useShouldShowShareTravelHabitsScreen = (
...
-        setSessionCount(newCount);
-      } else {
-        setSessionCount(currentCount);
-      }
-    },
-    [utilizeThisHookInstanceForSessionCounting],
-  );
+  const updateCount = useCallback(async (currentCount: number) => {
+    const newCount = currentCount + 1;
+    storage.set(shareTravelHabitsSessionCountKey, JSON.stringify(newCount));
```
jorelosorio, line 46: Is no matter of this PR, but I was wondering why is not possible to just send newCount directly and in the wrappers we could encode, decode the json as in the documentation: https://react-native-async-storage.github.io/async-storage/docs/usage

src/loading-screen/use-is-loading-app-state.ts
```diff
@@ -0,0 +1,6 @@
+import {useOnboardingState} from '@atb/onboarding';
+
+export const useIsLoadingAppState = () => {
+  const {isLoading: onboardingIsLoading} = useOnboardingState();
+  return onboardingIsLoading; // also wait for e.g. popover flags ++?
```
jorelosorio, line 5: If those flags are needed, then is good to have it ready when fully loaded. 👍🏽

src/onboarding/OnboardingContext.tsx
```diff
@@ -144,51 +161,72 @@ export const AppContextProvider: React.FC = ({children}) => {
...
+      <OnboardingDispatch.Provider value={dispatch}>
+        {children}
+      </OnboardingDispatch.Provider>
+    </OnboardingContext.Provider>
   );
 };
 
-export function useAppState() {
-  const context = useContext(AppContext);
+export function useOnboardingState() {
```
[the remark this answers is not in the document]
jorelosorio: I saw that mostly of context have useOnboardingState the word State at the end: `useGeolocationState`, `useBeaconsState`, etc.

src/onboarding/onboarding-config.ts
```diff
@@ -0,0 +1,57 @@
+import {StaticOnboardingSection} from '@atb/onboarding';
+
+export const staticOnboardingSectionsInPrioritizedOrder: StaticOnboardingSection[] =
```
jorelosorio, line 3: I like the idea of define the screens to show in the onboard, need to have a more deep look and see how it works, It's nice to see it is more dynamic and is a good approach, I will suggest in the future add more attributes like next Screen, order and parameters to be expected to receive. and the possibility to configure it from the remote Config, but of course might not be part of this PR.

src/beacons/use-should-show-share-travel-habits-screen.tsx
marius-at-atb: I also think it should be fine to use JSON.stringify inside `storage.set`.
Loading values from storage however I think needs to know specifically whether it should be a string, int, json object etc

src/onboarding/onboarding-config.ts
marius-at-atb: The config format can be adapted to new feature requests for sure :+1:

rosvik (approved):
A bit of work reading through this PR, but looking at the result, I like how this turned out! Really nice that everything is in one organized /onboarding folder, and with the scaffolding in place, it looks like it will be a lot easier to work with going forward 🎉 

Great work! A++

src/onboarding/OnboardingContext.tsx
```diff
@@ -0,0 +1,334 @@
...
+};
+
+export const OnboardingContextProvider: React.FC = ({children}) => {
+  const [state, dispatch] = useReducer<OnboardingReducer>(
+    onboardingReducer,
+    defaultOnboardingState,
+  );
+
+  // loadedOnboardingSections is omitted for contextState as onboardingSections should always be used instead
+  const {loadedOnboardingSections, ...contextState} = state;
```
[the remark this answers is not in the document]
gorandalum: Not my comment, but I'll answer how I interpret it:
In the internal state of OnboardingContext we have the `loadedOnboardingSections`, before they are enriched with the `shouldShow` value and become `onboardingSections`. These internal `loadedOnboardingSections` is not ment to be exposed outside of the context like the rest of the context state.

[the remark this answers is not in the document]
marius-at-atb: Good suggestion :+1:

PR #4302, opened 2024-02-02
fix: wrong color on button in TravelTokenBox

adriawh:
AS mentioned by @tormoseng in [this comment](https://github.com/AtB-AS/kundevendt/issues/15986#issuecomment-1923614202), the color was wrong on the button inside `TravelTokenBox `.

`TravelTokenBox ` has `InteractiveColor` as its color, and the button component therefore needed added support for when an `interactiveColor` is the `backgroundColor`.

After changes:

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -261,7 +261,7 @@ const useButtonStyle = StyleSheet.createThemeHook((theme: Theme) => ({
 }));
 
 const useColor = (color: StaticColor | InteractiveColor): string => {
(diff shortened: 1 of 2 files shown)


jorelosorio (approved):
Good :)

jorelosorio (approved):
Nice, not sure if this make something else on the onboarding @marius-at-atb

adriawh:
I belive the padding was removed, since the button was removed in 1.46. So when merging the release branch into master, the padding was still commented out. So I think it should be fine @jorelosorio

PR #4304, opened 2024-02-05
refactor: rename some notification-related file and functions

reidzeibel:
Just a small refactor to make it more intuitive: 

useConfig ➡️ useNotificationConfig
use-config.tsx ➡️ use-notification-config.tsx
getConfig ➡️ getNotificationConfig
updateConfig ➡️ updateNotificationConfig

### DIFF
--- a/src/notifications/use-push-notifications.tsx
+++ b/src/notifications/use-push-notifications.tsx
@@ -13,7 +13,7 @@ import {
 import {PermissionsAndroid, Platform} from 'react-native';
 import {NotificationConfigUpdate} from './api';
 import {NotificationConfig} from './types';
-import {useConfig} from './use-config';
+import {useNotificationConfig} from './use-notification-config';
 import {useRegister} from './use-register';
 import {getLanguageAndTextEnum} from '@atb/translations/utils';
 import {usePushNotificationsEnabled} from '@atb/notifications/use-push-notifications-enabled';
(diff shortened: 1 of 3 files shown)


src/notifications/use-push-notifications.tsx
```diff
@@ -47,7 +47,8 @@ export const NotificationContextProvider: React.FC = ({children}) => {
   const [fcmToken, setFcmToken] = useState<string>();
   const {mutation: registerMutation} = useRegister();
   const mutateRegister = registerMutation.mutate;
-  const {query: configQuery, mutation: configMutation} = useConfig();
+  const {query: configQuery, mutation: configMutation} =
```
jorelosorio, line 50: Perhaps use getConfigQuery or getNotificationConfig as well updateNotificationConfig here?

jorelosorio (approved):
Nice

src/notifications/use-push-notifications.tsx
reidzeibel: I think leaving it as-is would be preferable, to keep it aligned with tanstack-query standards. It would be weird if we call `updateNotificationConfig.mutate`. It might be better with `getConfigQuery.data` but it doesn't add that much value anyway. So keeping it as-is for now, unless we want to start renaming mutations and queries. :)

jorelosorio: OK, but I was more thinking:

```ts
const {query: getNotificationConfig, mutation: updateNotificationConfig} = useConfig();
```

not: `updateNotificationConfig.mutate` but I see that it is a standard

reidzeibel: Thanks Jorge, Sorry I got a typo there on my comment hahaha.

PR #4308, opened 2024-02-05
Feat: add listener to `sentFareContracts` firestore collection

reidzeibel:
This PR is part of : 
- https://github.com/AtB-AS/kundevendt/issues/15879
- https://github.com/AtB-AS/kundevendt/issues/16304

This will add 'sentFareContracts' collection from firestore to the listener in app. Added the snapshot listener and reducer.
Please let me know if there's anything to fix! I'm new to this part of the code. 😄

### DIFF
--- a/src/fare-contracts/FareContractAndReservationsList.tsx
+++ b/src/fare-contracts/FareContractAndReservationsList.tsx
@@ -35,22 +35,11 @@ export const FareContractAndReservationsList: React.FC<Props> = ({
   const navigation = useNavigation<RootNavigationProp>();
   const analytics = useAnalytics();
 
-  /**
-   * This const will filter out completed reservation to not be shown,
-   * so when a user only have sent tickets to someone else,
-   * the completed reservations are not included in the calculation,
-   * showing a proper empty page.
-   */
-  const filterFinishedReservation = reservations?.filter((reservation) => {
-    reservation.paymentStatus !== 'CAPTURE';
-  });
-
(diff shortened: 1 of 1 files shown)


src/ticketing/TicketingContext.tsx
```diff
@@ -65,6 +70,19 @@ const ticketingReducer: TicketingReducer = (
...
+    case 'UPDATE_SENT_FARE_CONTRACTS': {
+      const currentFareContractOrderIds = action.fareContracts.map(
+        (fc) => fc.orderId,
+      );
+      return {
+        ...prevState,
+        sentFareContracts: action.fareContracts,
+        reservations: prevState.reservations.filter(
+          (r) => !currentFareContractOrderIds.includes(r.orderId),
+        ),
```
gorandalum, line 82: We remove reservations when we retrieve the fare contract. We may need to do that when we receive the sent fare contract as well, but I am not completely sure.

gorandalum (approved):

src/ticketing/TicketingContext.tsx
reidzeibel: Then we will need to do that as well, so the `FareContractAndReservationList` can work as intended, thanks for pointing it out!

reidzeibel: I can confirm that the reservations are filtered out when we receive the sent fare contract. 😄

PR #4310, opened 2024-02-05
fix: Update use-push-notifications

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/16615

So after many findings it ended up doing two small fixes. First the fcmToken it is OK to get it once, so it was moved to a use Effect and became part of the dependencies for when registering, so is not often this changes.


src/notifications/use-push-notifications.tsx
```diff
@@ -53,21 +53,20 @@ export const NotificationContextProvider: React.FC = ({children}) => {
...
         });
-        return token;
+        return fcmToken;
       } catch (e) {
         Bugsnag.notify(`Failed to register for push notifications: ${e}`);
       }
     },
-    [language, mutateRegister],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [language, fcmToken],
```
jorelosorio, line 69: adding `mutateRegister` here was causing fire this register function multiple times, is not needed, so the solution was to add this ESLint exception rule. fcmToken became part of the dependencies.

src/notifications/use-push-notifications.tsx
```diff
@@ -126,6 +125,14 @@ export const NotificationContextProvider: React.FC = ({children}) => {
...
   }, [pushNotificationsEnabled, checkPermissions, authStatus]);
 
+  useEffect(() => {
+    (async function () {
+      const token = await messaging().getToken();
+      if (!token) return;
+      setFcmToken(token);
+    })();
+  }, []);
+
```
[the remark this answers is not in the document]
jorelosorio: Not sure if understood your question, but the token is not attached to an account, is the use of mutation function that we set the token to a specific user.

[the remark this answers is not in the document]
gorandalum: Alright. I wasn't sure if you needed a new token after Firebase auth user changed.

jorelosorio:
The following changes ensure that fcmtoken is requested once when component start, and included on the dependencies and removed a couple of async definitions that are not needed any longer.

image


register is being called only when between switches.

src/notifications/use-push-notifications.tsx
```diff
@@ -127,6 +124,14 @@ export const NotificationContextProvider: React.FC = ({children}) => {
       checkPermissions();
   }, [pushNotificationsEnabled, checkPermissions, authStatus]);
 
+  useEffect(() => {
+    (async function () {
+      const token = await messaging().getToken();
+      if (!token) return;
```
gorandalum, line 130: Do we know in what situations the token can be `undefined`. As it is now we just try once at startup, and if the token is `undefined` then it will never be retried until a new app start.

src/notifications/use-push-notifications.tsx
```diff
@@ -110,13 +107,13 @@ export const NotificationContextProvider: React.FC = ({children}) => {
   const requestPermissions = useCallback(async () => {
     setStatus('updating');
     const permissionStatus = await requestUserPermission();
-    const token = await register(permissionStatus === 'granted');
-    if (!token) {
+    if (!fcmToken) {
       setStatus('error');
```
gorandalum, line 111: When the status is error we show an error message saying "An error ocurred. Please try again later 🤞". If try again here means trying to enable to token toggle again, it will never work as `fcmToken` will always be `undefined`.

gorandalum (changes requested):
This looks better. I think we need some error handling as the error message says "try again later", but trying again will never work as it is implemented now.

src/notifications/use-push-notifications.tsx
jorelosorio: The promise always returns a nil as far I can see, so then I guess is because an explicit exception going on.

src/notifications/use-push-notifications.tsx
jorelosorio: Fixed.

src/notifications/use-push-notifications.tsx
```diff
@@ -110,13 +113,15 @@ export const NotificationContextProvider: React.FC = ({children}) => {
   const requestPermissions = useCallback(async () => {
     setStatus('updating');
     const permissionStatus = await requestUserPermission();
-    const token = await register(permissionStatus === 'granted');
+    // Retry to get the token after the user has granted permission
+    const token = await getFcmToken();
```
gorandalum, line 117: Note that this triggers register, since `fcmToken` is in dependency list of the register function.

gorandalum (approved):

src/notifications/use-push-notifications.tsx
jorelosorio: So, I am putting back some of the stuff to do this retry, however the register callback it is not called himself if token changes, but it might fire other useEffect to be triggered.

PR #4313, opened 2024-02-06
feat: implement handling of tickets sent to others

reidzeibel:
closes
https://github.com/AtB-AS/kundevendt/issues/15879
https://github.com/AtB-AS/kundevendt/issues/16304

Both issues practically uses the same layout.

[Figma link for reference](https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=20920-15395&mode=design&t=6mTCaFpOu6j9Cw2q-4)

This PR will make the necessary adjustments for sent tickets, some of them are:

- Adding routes and params to pass the sent ticket status.
- Add handling for validity status `sent`.
- Add text for sent tickets.
- Mark `sent` tickets as invalid for inspection.

<details>
<summary>Videos</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/cb08a9c0-63b5-44d4-abbf-16b4366f5962

https://github.com/AtB-AS/mittatb-app/assets/1777333/890f2912-e812-4335-aab6-c277bc32ac07

</details>

### DIFF
--- a/src/fare-contracts/FareContractInfo.tsx
+++ b/src/fare-contracts/FareContractInfo.tsx
@@ -146,6 +146,9 @@ export const FareContractInfoDetails = (
     toTariffZone,
   );
 
+  const isValidOrSentFareContract: boolean =
+    isValidFareContract(status) || isSentFareContract(status);
+
   return (
     <View style={styles.container} accessible={true}>
       <View style={styles.fareContractDetails}>
@@ -163,9 +166,11 @@ export const FareContractInfoDetails = (
             />
           )}
         </View>
-        {isValidFareContract(status) && <InspectionSymbol {...props} />}
-        {isSentFareContract(status) && (
-          <InspectionSymbol {...props} sentTicket={true} />
+        {isValidOrSentFareContract && (
+          <InspectionSymbol
+            {...props}
+            sentTicket={isSentFareContract(status)}
+          />
         )}
       </View>
     </View>
--- a/src/fare-contracts/FareContractInfo.tsx
+++ b/src/fare-contracts/FareContractInfo.tsx
@@ -46,7 +46,7 @@ export type FareContractInfoProps = {
   status: ValidityStatus;
   testID?: string;
   preassignedFareProduct?: PreassignedFareProduct;
-  fareContract?: FareContract;
+  sentToCustomerAccountId?: string;
 };
 
 export type FareContractInfoDetailsProps = {
@@ -67,7 +67,7 @@ export const FareContractInfoHeader = ({
   status,
   testID,
   preassignedFareProduct,
-  fareContract,
+  sentToCustomerAccountId,
 }: FareContractInfoProps) => {
   const styles = useStyles();
   const {t, language} = useTranslation();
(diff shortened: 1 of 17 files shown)


src/stacks-hierarchy/navigation-types.ts
```diff
@@ -48,6 +48,7 @@ type ReceiptScreenRouteParams = {
 
 type FareContractDetailsRouteParams = {
   orderId: string;
+  isSentFareContract?: boolean;
```
rosvik, line 51: Not sure exactly how this works, but is there a way to avoid send this param? Could FareContratDetails figure out itself if the fare contract is sent or not?

reidzeibel: We could, using `abtCustomerId` from `useAuthState()`, if we use that instead, we can remove all these parameter passing. 

The problem with `abtCustomerId` is that it can be undefined, and we will need to constantly check for its value on different screens/classes. 

Do you have any inputs/preference/best practice for this?

src/fare-contracts/FareContractInfo.tsx
```diff
@@ -150,6 +164,9 @@ export const FareContractInfoDetails = (
           )}
         </View>
         {isValidFareContract(status) && <InspectionSymbol {...props} />}
+        {isSentFareContract(status) && (
+          <InspectionSymbol {...props} sentTicket={true} />
+        )}
```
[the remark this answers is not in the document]
reidzeibel: Technically, a sent fare contract should not be valid since the status is marked as `sent`. But I'll see what I can do to simplify this part. Thanks!

src/fare-contracts/details/DetailsContent.tsx
```diff
@@ -123,6 +126,7 @@ export const DetailsContent: React.FC<Props> = ({
             status={validityStatus}
             testID="details"
             preassignedFareProduct={preassignedFareProduct}
+            fareContract={isSentFareContract ? fc : undefined}
```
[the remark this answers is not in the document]
reidzeibel: I agree, it can be simplified to that. Will change it so something else.

src/stacks-hierarchy/navigation-types.ts
rosvik: I believe the performance cost of fetching the id from useAuthState is next to zero, so I think that's a better way of doing it. We should do whatever reduces complexity, but in general I think it's nice to have as few route params as possible.

src/fare-contracts/FareContractInfo.tsx
```diff
@@ -79,6 +83,10 @@ export const FareContractInfoHeader = ({
   const warning = useNonInspectableTokenWarning(preassignedFareProduct?.type);
   const showTwoWayIcon = travelRight.direction === TravelRightDirection.Both;
 
+  const {data: phoneNumber} = useGetPhoneByAccountIdQuery(
+    fareContract?.customerAccountId,
+  );
+
```
gorandalum, line 89: I think the logic that if a fare contract is sent as a param to `FareContractInfoHeader`, then we know it is a sent contract and should show the phone number is a little too subtle. And it being subtle may cause bugs later when someone needs another value from the fare contract here, and ends up providing it as a param without knowing the side effect it causes.

gorandalum (approved):
:+1:

src/stacks-hierarchy/navigation-types.ts
reidzeibel: Updated 👍

PR #4315, opened 2024-02-08
fix: remove global messages from app purchase overview

adriawh:
Removing the `app-purchase-overview` context from the global messages, and removing the global message component in the `PurhaseOverviewScreen`. 

~~After merging the messages should be removed from firebase.~~

The purchaseMessage looks similar to the global message, but gives more urgent information and is therefore kept. In addition to informing when the active token is another device, It also says whether or not the ticket must be used on a phone, which Is something a user should be reminded of before buying.  An example of this is with the express boat tickets.

<details>

</details>

### DIFF
--- a/src/global-messages/GlobalMessagesContext.tsx
+++ b/src/global-messages/GlobalMessagesContext.tsx
@@ -17,7 +17,7 @@ import {
   GlobalMessageRaw,
   GlobalMessageType,
 } from '@atb/global-messages/types';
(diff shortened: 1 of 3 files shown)


gorandalum:
I think we definitively should keep the option to show messages here.

Sebstorvik:
Sorry, a bit slow to comment, but definitively keep the option. Setting a maxVersion sounds like a good solution to me👍

PR #4316, opened 2024-02-08
feat: Show booking info on non-flex departures

gorandalum:
Now showing booking info and booking options on all departures that
have relevant booking info. The bottom sheet with extra info is
available just for AtB flex departures.

### DIFF
--- a/src/travel-details-screens/components/BookingInfoBox.tsx
+++ b/src/travel-details-screens/components/BookingInfoBox.tsx
@@ -1,39 +1,40 @@
 import {
-  TripDetailsTexts,
   Language,
   TranslateFunction,
+  TripDetailsTexts,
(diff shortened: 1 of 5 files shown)


rosvik (approved):
Nice!

jorelosorio:
> Now showing booking info and booking options on all departures that have relevant booking info. The bottom sheet with extra info is available just for AtB flex departures.
> 
> Note that in the pictures the yellow notifiction icon and yellow message box is not connected to the booking info. It is a message from the situations api.
> 
> Before:
> 
>  
> After:
> 
>  

Nice to see that booking might not be related to flex only, as it was a while a go.

Good fix.

On the other hand, seems like the spacing between buttons might need a check. This is how it looks in figma now.

image

PR #4317, opened 2024-02-08
fix: hide token warning message if buying ticket for someone else

reidzeibel:
fixes issue from slack : https://mittatb.slack.com/archives/CSFC6PG23/p1707392965566429

When sending to someone else, we should not show any token warning message.
This PR will hide the warning if `on-behalf-of` toggle is activated.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -436,7 +436,7 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
             </View>
           </GenericSectionItem>
         </Section>
-        {inspectableTokenWarningText && (
+        {inspectableTokenWarningText && !phoneNumber && (
           <MessageInfoBox
             type="warning"
             message={inspectableTokenWarningText}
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -436,7 +436,7 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
             </View>
           </GenericSectionItem>
         </Section>
-        {inspectableTokenWarningText && (
+        {inspectableTokenWarningText && !phoneNumber && (
```
reidzeibel, line 439: Having `phoneNumber` means that "this ticket is sent to `phoneNumber`, then we should hide the warning.

jorelosorio, line 439: is phoneNumber at that point valid? I mean existing customer? or is just non-empty number.

reidzeibel: if `on-behalf-of` toggle is enabled, then `phoneNumber` will have a value from API and is passed on from the previous screen.

Otherwise if `on-behalf-of` toggle is disabled, then `phoneNumber` will be undefined.

jorelosorio (approved):

gorandalum (approved):

PR #4318, opened 2024-02-08
fix: add top padding to sectionHeader subtext

hanne-at-atb:
Ref. comment in: https://github.com/AtB-AS/kundevendt/issues/12515#issuecomment-1933575059


jorelosorio (approved):
🐶

rosvik (approved):

PR #4320, opened 2024-02-08
fix: map zooming all the way out on bicycle station in map

adriawh:
The problem was similar to the one in @gorandalum's PR https://github.com/AtB-AS/mittatb-app/pull/4040. I found the issue to be that we were taking font scale into the calculation of the padding. On the note 10, where the issue was, the scale was 0.8 making the padding bigger which means less space for the walking path. This led to it zooming all the way out since the path never fit inside the remaining space.

The scaling is also kind of counter intuitive, as when the font is smaller and things take less space, the padding was getting bigger. And vice versa with larger font. This made the path not centered when the font was scaled.

So this fixes both the very specific bug, and the map for larger fonts.

Below is before and after with larger font:

<details>

</details>


jorelosorio (approved):
Nice, so we made an upgrade on the maps a while ago, after this PR was merged. At that point, the scales didn't match between the phone scales and the map scales, for iOS was OK, that is why this is made only for Android. However, by just reading your findings might be that for Android seems to be correct now, so if is it possible to check that works for iOS then we could remove the Android condition since then would be the same for both.

PR #4323, opened 2024-02-12
feat: Red icon and message if booking time expired

gorandalum:
Also some other changes and simplifications, discussed with design:
- Not change text based on time until deadline. Earlier we showed relative minutes when the booking deadline was closed, bus this wasn't worth the effort.
- Always use warning status level when a trip requires booking. Earlier there was info level if more than one hour until deadline, and warning if less than one hour.
- Went through each message and aligned them with Figma.

### DIFF
--- a/src/travel-details-screens/utils.ts
+++ b/src/travel-details-screens/utils.ts
@@ -19,6 +19,7 @@ import {
 import {dictionary, TranslateFunction} from '@atb/translations';
 import {APP_ORG} from '@env';
 import {BookingArrangementFragment} from '@atb/api/types/generated/fragments/booking-arrangements';
+import {BookingStatus, TripPatternBookingStatus} from './types';
 
 const DEFAULT_THRESHOLD_AIMED_EXPECTED_IN_MINUTES = 1;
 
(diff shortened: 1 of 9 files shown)


src/travel-details-screens/utils.ts
```diff
@@ -321,138 +322,93 @@ export function getSecondsRemainingToBookingDeadline(
...
-}
+  if (secondsToDeadline < 0) {
+    return 'late';
+  }
 
-export const getTripPatternBookingsRequiredCount = (tp: TripPattern): number =>
-  tp?.legs?.filter((leg) => !!leg.bookingArrangements).length;
+  if (flex_booking_number_of_days_available) {
+    const secondsBeforehandItCanBeBooked =
+      flex_booking_number_of_days_available * 24 * secondsInOneHour;
```
rosvik, line 363: Not sure how critical it is, but will this be off one hour twice a year for summer time / winter time?

rosvik (approved):
Looks good! One question about some date math.

src/travel-details-screens/utils.ts
```diff
@@ -321,138 +322,93 @@ export function getSecondsRemainingToBookingDeadline(
...
-  const isTooLate = getIsTooLateToBook(
+  const secondsToDeadline = getSecondsRemainingToBookingDeadline(
     bookingArrangements,
     aimedStartTime,
     now,
   );
 
-  return !isTooEarly && !isTooLate;
-}
+  if (secondsToDeadline < 0) {
```
jorelosorio, line 357: should this be based on the booking max time?, as far I remember we discussed it to be 3h before departure, but this is not a rule written somewhere because this value should come in the response. It might be OK to have it like now but would be better if is a real time.

jorelosorio: `aimedStartTime` seems to be the departure time. Not the time when it is OK to book it.

src/travel-details-screens/utils.ts
gorandalum: Yeah, but probably not critical. And there is the same issue with tickets, where a 24hour ticket is actually 24 hours, and not from 3pm to 3pm if there is a summer/winter time switch.

src/travel-details-screens/utils.ts
gorandalum: When you say booking max time, do you mean the booking deadline? If so that is calculated in the `getSecondsRemainingToBookingDeadline` based on the data returned from Entur. For example for the Vy bus shown in the pictures the deadline is 15 minutes before the start time. And if seconds to booking deadline is negative, then it is too late to book it.

jorelosorio: OK, that sounds good, I just saw `aimedStart` in the equation. seems like latestBookingTime inside booking arrangement.

PR #4324, opened 2024-02-12
feat: refactor carnet fare contract

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/12053

This PR will: 
- merge carnet and regular ticket components into `FareContractView.tsx`. (open to suggestions)
- add various utility functions to simplify the merged component.
- remove unused components.

There's a possibility of merging the `DetailsContent` and `FareContractView`, it's just that I feel it might be counter-intuitive and they aren't exactly the same (one for list item and the other for detailed content). If we create a reusable layout for both screen, it might become more difficult to maintain when new element(s) are added.

Also, there are `utils` for `Ticketing` and `utils` for `FareContract`, maybe we need to refactor these as well? I feel like the whole farecontract thing can be re-organized to be more tidy, but not sure on what the optimal structure is for now.


src/fare-contracts/FareContractInfo.tsx
```diff
@@ -147,7 +146,7 @@ export const FareContractInfoDetails = (
   );
 
   const isValidOrSentFareContract: boolean =
-    isValidFareContract(status) || isSentFareContract(status);
+    isValidFareContract(status) || status === 'sent';
```
jorelosorio, line 149: we may use a `const isStatusSent = status === 'sent'` and reuse it.

gorandalum (approved):
Looks like a nice improvement! :+1:

src/fare-contracts/FareContractInfo.tsx
reidzeibel: There's so much checking of sent fare contract that I think it's better to check it manually instead of making more functions 😆 , though this one I think I can do it.

PR #4325, opened 2024-02-13
fix: qa feedback for benefit tiles

rosvik:
Fix of QA feedback on bundling actions in ticket details

- Adds a11y role "button" to TileWithButton.
- Adds a11y hint to benefit tiles.
- Only shows benefits when the ticket is active.

fixes https://github.com/AtB-AS/kundevendt/issues/16297

### DIFF
--- a/src/translations/screens/subscreens/MobilityTexts.ts
+++ b/src/translations/screens/subscreens/MobilityTexts.ts
@@ -88,6 +88,11 @@ export const MobilityTexts = {
     'Inkludert i billetten',
   ),
   showInMap: _('Se i kart', 'Show in map', 'Sjå i kart'),
+  showInMapA11yLabel: _(
+    'Aktiver for å se i kart',
+    'Activate to show in map',
+    'Aktiver for å sjå i kart',
+  ),
 };
 
 export const ScooterTexts = {
--- a/src/components/tile/TileWithButton.tsx
+++ b/src/components/tile/TileWithButton.tsx
@@ -43,6 +43,7 @@ export function TileWithButton({
       onPress={onPress}
       accessibilityLabel={accessibilityLabel}
       accessibilityHint={accessibilityHint}
+      accessibilityRole="button"
       testID={testID}
     >
       <View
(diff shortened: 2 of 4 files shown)


reidzeibel (approved):

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -88,6 +88,11 @@ export const MobilityTexts = {
     'Inkludert i billetten',
   ),
   showInMap: _('Se i kart', 'Show in map', 'Sjå i kart'),
+  showInMapA11yLabel: _(
+    'Aktiver for å se i kart',
```
marius-at-atb, line 92: The map is disabled when screenreader is activated, so this won't work?

rosvik: Correct! We (me, @oyvindmikkelsenatb, @HanneLH ++) had a discussion about possible solutions to this yesterday, and we landed on this as a solution that's as close to good as we'll get for now 😅 Please let me know if you have better ideas.

marius-at-atb: Hmm, thinking about it now - if you can't see the screen, you shouldn't drive or ride a bike either. So with the current lineup of benefits it shouldn't matter.

rosvik: I'm sure renting a car can be useful, since someone else can drive it. Would be nice if we could redirect directly to Hyre (or any other provider app) instead.

marius-at-atb: Yea. Ideally you shouldn't have to go through the map at all to activate the benefits, so that would be the best solution.
However if you have someone to drive for you, maybe you also have someone to help you use the benefit in the app?
Another potential issue is if the 3rd party app doesn't support screenreader, but nothing we can do about that..

marius-at-atb: Agreed :+1:

PR #4326, opened 2024-02-13
feat: Add booking info to departures

gorandalum:
Show booking info on:
- As info-box on The departure details screen, with booking action
buttons if currently bookable.
- As icon in the upper left corner (same as situations) in the
departures list v2.
- As right icon on the departure chip in departures list v1.

### DIFF
--- a/src/travel-details-screens/DepartureDetailsScreenComponent.tsx
+++ b/src/travel-details-screens/DepartureDetailsScreenComponent.tsx
@@ -33,8 +33,8 @@ import {TripLegDecoration} from './components/TripLegDecoration';
 import {TripRow} from './components/TripRow';
 import {ServiceJourneyDeparture} from './types';
 import {
-  useDepartureData,
   EstimatedCallWithMetadata,
+  useDepartureData,
 } from './use-departure-data';
 import {useIsScreenReaderEnabled} from '@atb/utils/use-is-screen-reader-enabled';
 import {PaginatedDetailsHeader} from '@atb/travel-details-screens/components/PaginatedDetailsHeader';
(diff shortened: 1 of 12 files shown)


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -433,12 +440,24 @@ function EstimatedCallRow({
 
   const {group, isStartOfGroup, isEndOfGroup, isStartOfServiceJourney} =
     call.metadata;
+  const isFirstOfTrip = group === 'trip' && isStartOfGroup;
```
marius-at-atb, line 443: Maybe this variable name could clarify what it is first of?

marius-at-atb (approved):
Lgtm! Really nice that you added tests for `statusComparator`, as it is not that easy to parse its internal logic when reading through 😄

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
gorandalum: Do you have a suggestion here? Would it be better with `isFirstOfTripGroup`? Kinda hard to find better naming without diverging from the group terms ('passed', 'trip', 'after') we have in the code.

hanne-at-atb (approved):
LGTM ⭐

jorelosorio (approved):
Nice work, a see tests for it.

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
marius-at-atb: Yea I think `isFirstOfTripGroup` would be better :+1:
Maybe also `isStartOfTripGroup` to keep the naming in line with `isStartOfGroup`?

PR #4327, opened 2024-02-14
feat: Add feature flag for server time

gorandalum:
- Add a feature flag to be able to turn of server time if
necessary. The feature flag has the default value true, and can be
set to false in RemoteConfig.
- Also fixed a bug where if the server clock didn't start as it
should, then the interval wasn't fired. This ment that if
something went wrong with the time-lib, we would never get an
updated time. Now the interval always runs every 2500 ms, but
fallbacks to Date.now() if server time not enabled or the time-lib
isn't started.

### Acceptance criteria
- [ ] If the feature toggle is false, then ticket views should use local device time
- [ ] If the feature toggle is true, then ticket views should use server time
- [ ] The app should handle that it is started in flight mode. Hopefully server time, but falling back to client time is ok, just that it doesn't end up having a stale time like was possible before.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -79,11 +79,11 @@ export const App = () => {
       <StorybookContextProvider>
         <ErrorBoundary type="full-screen">
           <AppContextProvider>
-            <TimeContextProvider>
-              <PreferencesContextProvider>
-                <LocaleContextProvider>
-                  <AuthContextProvider>
-                    <RemoteConfigContextProvider>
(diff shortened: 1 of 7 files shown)


hanne-at-atb (approved):
LGTM 🌟

marius-at-atb (approved):

rosvik (approved):
LGTM 👍

tormoseng:
👍

PR #4328, opened 2024-02-14
fix: use small padding around icons in travel search results

rosvik:
This adds some more padding around transport icons in travel search ResultItems.

I was unsure about what kind of prop I should use here, so I'm open to feedback 😊 

1. I could use ContainerSizingType (`'compact' | 'standard' | 'spacious'`), but it didn't quite fit since all I needed was two modes, and default would be compact. 
2. A new prop `type='normal' | 'spacious'`. The `mode` name is taken for Transport Mode.
3. I felt like `spacious=boolean` was the simplest solution, but it doesn't really fit what we've done elsewhere

fixes https://github.com/AtB-AS/kundevendt/issues/3834#issuecomment-1943696655

### DIFF
--- a/src/components/icon-box/TransportationIconBox.tsx
+++ b/src/components/icon-box/TransportationIconBox.tsx
@@ -16,7 +16,7 @@ export type TransportationIconBoxProps = {
   subMode?: AnySubMode;
   lineNumber?: string;
   size?: keyof Theme['icon']['size'];
-  spacious?: boolean;
+  type?: 'compact' | 'standard';
   style?: StyleProp<ViewStyle>;
   disabled?: boolean;
   testID?: string;
@@ -27,7 +27,7 @@ export const TransportationIconBox: React.FC<TransportationIconBoxProps> = ({
   subMode,
   lineNumber,
   size = 'normal',
-  spacious = false,
+  type = 'compact',
   style,
   disabled,
   testID,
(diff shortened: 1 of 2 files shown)


gorandalum (commented):
Why not use `xSmall` padding on the other boxes in the travel search results as well? Discussed that with design? I think they are "transportation boxes" as well, and it would be beneficial to have same look and feel throughout the app.

rosvik:
> Why not use `xSmall` padding on the other boxes in the travel search results as well? Discussed that with design? I think they are "transportation boxes" as well, and it would be beneficial to have same look and feel throughout the app.

@gorandalum That's an option, but I don't think that's a good solution in this case:

- I think these have more in common with "LineChip" in departures, than the other transport icons (although it is implemented differently), so we might want to look at those together.
- The icon width grows as the text below gets longer, so we end up with very long icons like this in some cases, which is a bit odd looking:
Screenshot 2024-02-14 at 14 59 38

But @oyvindmikkelsenatb or @ckbilstad, do you have any opinions?

jorelosorio:
I could suggest using an attribute as padding and use xsmall, medium, large... as we have it, it might that spacious alone might not be clear.

gorandalum:
I'll ask @oyvindmikkelsenatb in the office here, conclusions incoming.

gorandalum:
He said we should have two different variations of the padding. Suggested "compact" and "standard", where "compact" is the `xSmall` padding.

rosvik:
@gorandalum Sounds good! So in typescript, I could set `type='compact' | 'standard'` where default is `'compact'`?

gorandalum (approved):

PR #4338, opened 2024-02-20
feat: add support for showing product alias in product selection

reidzeibel:
part of https://github.com/AtB-AS/kundevendt/issues/12059

This PR will implement the use of `productAlias` on the `fareProductTypeConfig`. I reused the `ProductSelectionByProducts` component instead of making a new one since we can just use a flag for it to reduce code duplication.

This will display the `productAlias` declared in the `preassignedFareProduct` instead of `name`. 

For example, with the data below in `preassignedFareProduct`:

```
      "name": {
        "lang": "nob",
        "value": "Klippekort"
      },
      "productAlias": [
        {
          "lang": "nob",
          "value": "10 enkeltbilletter"
        },
        {
          "lang": "eng",
          "value": "10 single tickets"
        }
      ],
```

by using `productAlias` as the `productSelectionMode`, it will display `10 enkeltbilletter` instead of `Klippekort`. (se skjermbilde)

### `productAlias` (left) vs `product`(right)                   
image
image

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelectionByProducts.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelectionByProducts.tsx
@@ -28,6 +28,7 @@ type ProductSelectionByProductsProps = {
   fareProductTypeConfig: FareProductTypeConfig;
   setSelectedProduct: (product: PreassignedFareProduct) => void;
   style?: StyleProp<ViewStyle>;
+  useAlias?: boolean;
 };
 
 export function ProductSelectionByProducts({
@@ -35,6 +36,7 @@ export function ProductSelectionByProducts({
   fareProductTypeConfig,
   setSelectedProduct,
   style,
+  useAlias,
 }: ProductSelectionByProductsProps) {
   const {t, language} = useTranslation();
   const {preassignedFareProducts} = useFirestoreConfiguration();
@@ -50,6 +52,14 @@ export function ProductSelectionByProducts({
     useTextForLanguage(
       fareProductTypeConfig.configuration.productSelectionTitle,
     ) || t(PurchaseOverviewTexts.productSelection.title);
+
+  const aliasText = useAlias
+    ? getTextForLanguage(selectedProduct.productAlias, language)
+    : undefined;
+
+  const productName = aliasText
+    ? aliasText
+    : getReferenceDataName(selectedProduct, language);
 
   const subText = (fp: PreassignedFareProduct) => {
     const descriptionMessage = getTextForLanguage(fp.description, language);
@@ -89,7 +99,7 @@ export function ProductSelectionByProducts({
           <ContentHeading text={title} />
           <Section>
             <HeaderSectionItem
-              text={getReferenceDataName(selectedProduct, language)}
+              text={productName}
               subtitle={subText(selectedProduct)}
             />
           </Section>
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelection.tsx
@@ -28,6 +28,15 @@ export function ProductSelection({
         />
       );
     case 'productAlias':
+      return (
+        <ProductSelectionByProducts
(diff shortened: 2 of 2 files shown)


rosvik (commented):
I don't think we can use the `productAlias` productSelectionMode for carnets, since that changes how existing tickets look, e.g. [FRAMs periodic tickets](https://github.com/AtB-AS/firestore-configuration/blob/b3eac40a8fd7cace95d3550760f967fbf239b2f4/documents/fram-prod-a7850/fareProductTypeConfigs.yaml#L115).

Would a solution be to keep using the `product` productSelectionMode, but always show aliases if there are any? Need to verify that we don't affect other ticket types that have `product` and aliases.

reidzeibel:
@rosvik ouch, didn't expect that, here's a quick summary of tickets

Tickets with `productSelectionMode: product` in Production
      
### AtB

| ticket type | uses alias? | available on app? |
| ----------- |  ----------- |  ----------- |
| night_v2 |  false | true |

### Fram
| ticket type | uses alias? | available on app? |
| ----------- |  ----------- |  ----------- |
period-area |  true | false
fram-october |  false | false
fram-county |  false | true

### NFK

| ticket type | uses alias? | available on app? |
| ----------- |  ----------- |  ----------- |
night_v2 |  false | true

So if we want to display productAlias if it is available for items with `productSelectionMode: product`, then I'll need to check the Fram part if it breaks anything, I have asked Martin for access to fram firestore and will check again after lunch.

reidzeibel:
@rosvik I updated the code to automatically use alias if it is specified on the referenceData. except for those that have more than 1 type of products (for example, night tickets). Should we enable the alias for multiple product as well?

rosvik:
@reidzeibel Not sure I follow: If there is an alias, is there any reason not to show it? Night tickets doesn't have aliases, right?

If an org doesn't want aliases, they shouldn't be in reference data, I think 😄

rosvik:
So, yes: I think we should enable alias for multiple as well 😊

reidzeibel:
@rosvik ready for another review 😊

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelectionByProducts.tsx
```diff
@@ -98,3 +99,28 @@ export function ProductSelectionByProducts({
...
+  fareProduct: PreassignedFareProduct,
+  language: Language,
+): string {
+  const aliasText = fareProduct.productAlias
+    ? getTextForLanguage(fareProduct.productAlias, language)
+    : undefined;
+
+  const productName = aliasText
+    ? aliasText
+    : getReferenceDataName(fareProduct, language);
```
rosvik, line 123: I think we can simplify this - and at that point it's maybe not necessary with the helper function?

```suggestion
  const productName = fareProduct.productAlias
    ? getTextForLanguage(fareProduct.productAlias, language)
    : getReferenceDataName(fareProduct, language);
```

reidzeibel: Riiight, thank you!

gorandalum (approved):
Looks good! :+1:

And as you touched upon in your discussion the `productSelectionMode: 'productAlias'` doesn't really make sense anymore. I think the correct change here would be:

- Rename the `productSelectionMode``productAlias` into maybe just `name` instead.
- Make the new `name` mode use the same `getProductName` function, so that it shows alias if it is present, and fallbacks to name if no alias.

But I'm not sure if it is worth the effort now or not. 

Whatever we choose we should update the doc here: https://github.com/AtB-AS/docs-private/blob/main/fare-product-type-configs.md
For now it is enough to just specify that `productSelectionMode` `product` also support `productAlias`.

reidzeibel:
@gorandalum will create another issue and update the docs as well, thank you for the reminder!

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/ProductSelectionByProducts.tsx
reidzeibel: updated, but since the `getTextForLanguage` can return undefined, I modified it a bit.

rosvik (approved):
Nice 🚀

PR #4339, opened 2024-02-21
feat: add consumption of carnet tickets

rosvik:
https://github.com/AtB-AS/mittatb-app/assets/1774972/a99ccbc3-49a6-462d-9434-065eb68ae85e

closes https://github.com/AtB-AS/kundevendt/issues/12562

### DIFF
--- a/src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
+++ b/src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
@@ -11,6 +11,7 @@ import {ThemeText} from '@atb/components/text';
 import {StyleSheet} from '@atb/theme';
 import {consumeCarnet} from '@atb/ticketing';
 import {FareContractTexts, useTranslation} from '@atb/translations';
+import Bugsnag from '@bugsnag/react-native';
 import React, {useState} from 'react';
 import {ScrollView} from 'react-native-gesture-handler';
 import {useSafeAreaInsets} from 'react-native-safe-area-context';
@@ -33,7 +34,10 @@ export const ConsumeCarnetBottomSheet = ({fareContractId}: Props) => {
       close();
     } catch (e: any) {
       const errorData = getAxiosErrorMetadata(e);
-      console.error('Error when consuming carnet: ', errorData);
+      Bugsnag.notify({
+        name: `${errorData.responseStatus} error when consuming carnet`,
+        message: `Error: ${JSON.stringify(errorData)}`,
+      });
       setError(true);
     }
     setIsLoading(false);
--- a/src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
+++ b/src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
@@ -1,0 +1,61 @@
+import {Confirm} from '@atb/assets/svg/mono-icons/actions';
+import {BottomSheetContainer} from '@atb/components/bottom-sheet';
+import {Button} from '@atb/components/button';
+import {GenericSectionItem, Section} from '@atb/components/sections';
+import {ThemeText} from '@atb/components/text';
+import {StyleSheet} from '@atb/theme';
+import {FareContractTexts, useTranslation} from '@atb/translations';
+import React from 'react';
+import {ScrollView} from 'react-native-gesture-handler';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+type Props = {
+  fareContractId: string;
+};
+
+export const ConsumeCarnetBottomSheet = ({fareContractId}: Props) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+
+  return (
+    <BottomSheetContainer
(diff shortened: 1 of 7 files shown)


src/ticketing/utils.ts
```diff
@@ -18,7 +18,7 @@ import {getLastUsedAccess} from '@atb/fare-contracts/utils';
 export function isCarnetTravelRight(
   travelRight: TravelRight | undefined,
 ): travelRight is CarnetTravelRight {
-  return travelRight?.type === 'CarnetTicket';
+  return !!travelRight && 'maximumNumberOfAccesses' in travelRight;
```
rosvik, line 21: @gorandalum You've talked a lot about not using the `type` field for logic in the app: Do you think this is this an ok replacement?

src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const ConsumeCarnetBottomSheet = ({fareContractId}: Props) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<boolean>(false);
+  const {close} = useBottomSheet();
+
+  const onConsume = async () => {
```
reidzeibel, line 29: IIRC, we want to start using react-query for new API operations, so we should do that also for this one.

rosvik: I think this is the one type of requests that shouldn't use react-query. There is no state to keep up to date here, and no dependencies to keep track of. Just an user interaction that causes an API request, which doesn't need to be more complicated that this.

reidzeibel: Fair enough! 👍

src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
```diff
@@ -0,0 +1,87 @@
...
+    setIsLoading(true);
+    try {
+      await consumeCarnet(fareContractId);
+      close();
+    } catch (e: any) {
+      const errorData = getAxiosErrorMetadata(e);
+      console.error('Error when consuming carnet: ', errorData);
```
rosvik, line 36: Noticing we should probably handle errors better than this. I guess a bugsnag report would make sense here.

reidzeibel (approved):
LGTM!

src/fare-contracts/components/ConsumeCarnetButton.tsx
```diff
@@ -0,0 +1,40 @@
+import {TicketValid} from '@atb/assets/svg/mono-icons/ticketing';
```
marius-at-atb, line 1: Should this file be called `ConsumeCarnetSectionItem.tsx`?

src/fare-contracts/components/ConsumeCarnetButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {useBottomSheet} from '@atb/components/bottom-sheet';
+import {LinkSectionItem, SectionItemProps} from '@atb/components/sections';
+import {ThemeIcon} from '@atb/components/theme-icon';
+import {useTheme} from '@atb/theme';
+import {FareContractTexts, useTranslation} from '@atb/translations';
+import React from 'react';
+import {ConsumeCarnetBottomSheet} from './ConsumeCarnetBottomSheet';
+
+type ConsumeCarnetButtonProps = SectionItemProps<{
```
marius-at-atb, line 10: Should this be called `ConsumeCarnetSectionItemProps`?

src/fare-contracts/components/ConsumeCarnetButton.tsx
rosvik: Yes, good catch! Fixed this and your other comment in 9665b1ce6fbdeba4e4aae8ba056467e3241ca58f

src/fare-contracts/components/ConsumeCarnetButton.tsx
rosvik: 9665b1ce6fbdeba4e4aae8ba056467e3241ca58f

src/ticketing/utils.ts
gorandalum: Yes, that's great, directly from the source :+1:

src/fare-contracts/components/ConsumeCarnetBottomSheet.tsx
gorandalum: We should use react-query. Then we get error/loading statues, retry possibility, and more out of the box, and we get the same typing/behaviour on all network requests.

src/fare-contracts/components/ConsumeCarnetButton.tsx
gorandalum: Also note that I think just using `Props` is fine when it is not exported.

PR #4340, opened 2024-02-21
feat: add warning icon and add warning info below title on departure

reidzeibel:
Closes https://github.com/AtB-AS/kundevendt/issues/16259

This PR will :


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -217,6 +221,14 @@ export const DepartureDetailsScreenComponent = ({
           style={styles.scrollView__content}
           testID="departureDetailsContentView"
         >
+          {subMode === TransportSubmode.RailReplacementBus ? (
```
marius-at-atb, line 224: Can just use `&&` instead of `? : null`

marius-at-atb (approved):
Lgtm!

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
reidzeibel: Nice, thanks!

PR #4341, opened 2024-02-21
fix: Simulator crashes during onboarding for beacons 

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/17005

NOTE: This solution makes sure for some cases where the initialization of the SDK is called but not started because lack of permissions on the simulator, so then make sure that only run Kettle of SDK was initialized.


marius-at-atb (approved):

reidzeibel (approved):
Well done!

PR #4342, opened 2024-02-21
refactor: update ticketing util function to be more reusable

rosvik:
- The main change here is to split `hasActiveOrUsableCarnetTravelRight()` into two smaller util functions that when used together, can achieve more granular functionality.
- Since fare contract state doesn't give the wanted data for showing ConsumeCarnetSectionItemProps, we now check active state based on `usedAccesses` trough `isCanBeConsumedNowFareContract`.
- Fixes some naming issues (https://github.com/AtB-AS/mittatb-app/pull/4339#discussion_r1497688948)

fixes https://github.com/AtB-AS/kundevendt/issues/12562


reidzeibel (approved):
LGTM!

LGTM

PR #4345, opened 2024-02-22
fix: give receipt LinkSectionItem a11y role "link" instead of "button"

rosvik:
@HanneLH Noticed an inconsistency between the "Get receipt sent" button and "Activate single ticket" LinkSectionItem in ticket details. One said had accessibilityRole "button" and the other "link". 

> - Links take the user to a new location, such as a new web page or new section of the current page.
> - Buttons trigger some action, such as showing content on the page that was previously hidden, playing a video, or submitting a form.
>
> https://www.washington.edu/accesstech/websites/links-buttons/

I think "link" makes the most sense in both cases, given that opening a bottom sheet is considered a new page and not "showing content on the page that was previously hidden".

There are probably lots of inconsistencies around this in the app, this is just one that showed up while testing https://github.com/AtB-AS/kundevendt/issues/12562

### DIFF
--- a/src/fare-contracts/details/DetailsContent.tsx
+++ b/src/fare-contracts/details/DetailsContent.tsx
@@ -234,7 +234,6 @@ export const DetailsContent: React.FC<Props> = ({
         <LinkSectionItem
           text={t(FareContractTexts.details.askForReceipt)}
           onPress={onReceiptNavigate}
-          accessibility={{accessibilityRole: 'button'}}
           testID="receiptButton"
         />
         {isCanBeConsumedNowFareContract(fc, now) && (


gorandalum:
My thought:
- Stuff that link to another screen should be a link (I know we have lots of inconsistency here)
- Stuff that opens bottom sheet should be a button (I think we are fairly consistent on this one)

But maybe ask on slack about this, so the team can decide on the standard going forward.

jorelosorio (approved):
In a web context, it makes sense that links are to navigate and buttons to actions, but on mobile context are considered as primary a secondary actions. Where secondary looks like "links". Those actions appear mostly on login screen, where skip are those "Links".

At least I agree that a Link is not button. 😂

gorandalum:
Is this still valid after the discussion on Slack?

rosvik:
@gorandalum I'm working on changing all LinkSectionItems to use button based on the discussion, but I guess it makes more sense to make a new PR for that than to update this one 👍

PR #4346, opened 2024-02-22
feat: add global message for carnet tickets summary

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/12547

This PR will : 

- Introduce global message on Purchase Confirmation screen.
- Adds a new global message context: `app-purchase-confirmation`.
- Add `TextColor` to `MessageInfoBox`, there was no way to give `secondary` color to the text component otherwise, none of the colors in `StaticColor` works.
- Add styling to necessary parts of `Root_PurchaseConfirmationScreen.tsx`, the flex part is necessary to make the text rendered correctly, especially line wrap on long text.
- Since we cannot get the carnet validity through the offer (only shows the validity of single ticket), we use global message to display the validity time.
- utilize the new `isCollectionOfAccesses` field for fareProductTypeConfig, to determine whether a ticket is a collection of travel access or not. If the value is true, do not show the start period (Oppstart nå information).

Some previews: 

### Klippekort
image

### Enkeltbillett
image

### Periodebillett (30-dagersbillett, oppstart senere)
image

### Enkeltbillett Hurtigbåt
image

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -41,11 +41,11 @@ import {CardPaymentMethod, PaymentMethod, SavedPaymentOption} from '../types';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy/navigation-types';
 import {GenericSectionItem, Section} from '@atb/components/sections';
 import {useAnalytics} from '@atb/analytics';
-import {Info} from '@atb/assets/svg/color/icons/status';
 import {StopPlaceFragment} from '@atb/api/types/generated/fragments/stop-places';
-import {GlobalMessageContextEnum} from '@atb/global-messages';
+import {GlobalMessage, GlobalMessageContextEnum} from '@atb/global-messages';
 import {useShowValidTimeInfoEnabled} from '../Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-show-valid-time-info-enabled';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
+import {MessageInfoText} from '@atb/components/message-info-text';
 
 function getPreviousPaymentMethod(
   previousPaymentMethod: SavedPaymentOption | undefined,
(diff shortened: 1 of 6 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -169,6 +169,8 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
       )
     : t(PurchaseConfirmationTexts.travelDate.now);
 
+  const isCarnet = preassignedFareProduct.type === 'carnet';
```
gorandalum, line 172: Not allowed to check on the type string. This is ment as a "foreign key" between the fare product type configs and preassigned fare products, and doesn't have any semantic meaning. If we wan't some logic especially for this product, it needs to be in the configs.

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -378,17 +381,27 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
                     ),
                   )}
-
-                <View style={[styles.smallTopMargin, {flexDirection: 'row'}]}>
-                  <Info
-                    height={theme.icon.size.normal}
-                    width={theme.icon.size.normal}
-                    style={{marginRight: theme.spacings.small}}
+                {isCarnet ? (
+                  <GlobalMessage
```
gorandalum, line 385: I think we should rather add a global message context not just for carnet, and if we want a message for just carnet we need to specify that in the rules.

gorandalum (changes requested):

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
reidzeibel: how do we know a PreassignedFareProduct is carnet or not then? On a FareContract we can use TravelRight, not sure what to use on PreassignedFareProduct. 🤔

gorandalum: The answer is that we shouldn't know if a PreassignedFareProduct is carnet or not. How we sell the project should be based on configuration in Firestore. So if a feature that is designed needs its own configuration we take a step back and evaluate if the feature is actually important enough that its worth the increase in configuration complexity, and have a talk with the designers and product owner about it.

gorandalum (approved):

gorandalum (approved):

PR #4347, opened 2024-02-22
fix: Correct value for token refresh percentage

gorandalum:
The refresh percentage should be 10 instead of 90. With the value
we had the 30 day token was renewed after 3 days, and not after 27
days as intended.

### Acceptance criteria
- [ ] A token should be renewed if the app is opened while less than 10 percent of the total validity period for the token is remaining

Note that changing the validity periods in BackOffice won't help, as it is the validity period on the local token that is evaluated. If we want to test this in a quicker manner than waiting many days we could modify the token-server in staging to create tokens with shorter validity days than 30.

### DIFF
--- a/src/mobile-token/mobileTokenClient.ts
+++ b/src/mobile-token/mobileTokenClient.ts
@@ -39,8 +39,8 @@ export const mobileTokenClient = {
   Check if token should be renewed. Will return true if either:
    - Token is expired
    - It is less than 12 hours until token expires
-   - More than 90 % of the token validity time has expired
+   - Less than 10 % of the token validity time is left
    */
   shouldRenew: (token: ActivatedToken) =>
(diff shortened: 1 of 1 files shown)


jorelosorio (approved):

tormoseng:
@gorandalum I guess you have tested this locally, so we don't need to repeat it?

gorandalum:
Can test that the token isn't renewed after 3 days anymore, like it was before. And after it has been in prod for a month-ish we can do some quick verification that this works as expected by looking at some logs.

tormoseng:
Will check in 3 days

```
jsonPayload.installId="412023ed-3b38-4e9b-8f3c-5def8e024f87"
```

tormoseng:
No traces of renewals ☝️ . Tested OK!

PR #4348, opened 2024-02-22
feat: remove outdated carnet messages

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/16654

### DIFF
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -107,7 +107,7 @@ export const mapToUserProfilesWithCount = (
         'id' in userProfileWithCount,
     );
 
-export const useNonInspectableTokenWarning = (fareProductType?: string) => {
+export const useNonInspectableTokenWarning = () => {
   const {t} = useTranslation();
   const {barcodeStatus, tokens} = useMobileTokenContextState();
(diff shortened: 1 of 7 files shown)


reidzeibel (approved):
Looks great to me!

src/fare-contracts/utils.ts
```diff
@@ -120,22 +120,14 @@ export const useNonInspectableTokenWarning = (fareProductType?: string) => {
       return t(FareContractTexts.warning.errorWithToken);
     case 'other':
       const inspectableToken = tokens.find((t) => t.isInspectable);
-      if (fareProductType === 'carnet') {
```
jorelosorio, line 123: This is the tricky part fareProductType === 'carnet' was removed, but that removes the two different messages, so I think is fine then.

jorelosorio (approved):
As a removing a message looks good :) . GW

src/fare-contracts/utils.ts
rosvik: It is the tricky part, yes 😄

In practice, it means we treat carnets like other tickets now, since they are also inspectable on phone: We simply message about token being set to another device if that is the case.

PR #4349, opened 2024-02-23
fix: fix VAT amount calculation

mjenssen:
Fixes https://github.com/AtB-AS/kundevendt/issues/17036


rosvik (approved):
Great, that makes more sense!

gorandalum (approved):
Good catch! Also possible to write it like this, if that feels simpler:

`const vatAmount = totalPrice / (100 + vatPercent) * vatPercent`

PR #4350, opened 2024-02-23
fix: The order of the active tickets below my tickets

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/16360

Before:

After:


jorelosorio:
> Have you tested this with active carnet tickets?

Yes Sr. 🙃

rosvik:
@jorelosorio Alright, I was asking because carnets stay "valid" even after one consumption have expired, but looks like getFareContractInfo takes that into account. Maybe keep an eye on the carnet, and verify that it goes below the active fare contracts once the single ticket expires. (I introduced a bug because of that earlier, ref. https://github.com/AtB-AS/kundevendt/issues/12562#issuecomment-1956528161)

I did find the code a bit hard to review, since there are nested ternary (`a ? b : c`) operators on top of sorting math. (E.g: I'm having a hard time figuring out if reservations will show up at the top or bottom of the list 😅) Not sure if that's just me, but could you look at making the code a bit more verbose, so it's easier to work with in the future?

jorelosorio:
> @jorelosorio Alright, I was asking because carnets stay "valid" even after one consumption have expired, but looks like getFareContractInfo takes that into account. Maybe keep an eye on the carnet, and verify that it goes below the active fare contracts once the single ticket expires. (I introduced a bug because of that earlier, ref. [AtB-AS/kundevendt#12562 (comment)](https://github.com/AtB-AS/kundevendt/issues/12562#issuecomment-1956528161))
> 
> I did find the code a bit hard to review, since there are nested ternary (`a ? b : c`) operators on top of sorting math. (E.g: I'm having a hard time figuring out if reservations will show up at the top or bottom of the list 😅) Not sure if that's just me, but could you look at making the code a bit more verbose, so it's easier to work with in the future?

As soon the state changes, yes: `getFareContractInfo` should take the new state if it changes. The `a ? b : c` is because of the magic in `[...(fareContracts || []), ...(reservations || [])]` so we need to treat two types and a condition. I tried it verbose, but is 3 conditions to say 1 or 0. I can change it anyway if it is hard to read, I just found it convenient.

rosvik:
Ok, up to you. If it works, I don't want to be nit-picky about it 😅 

But just as an example, I find this easier to reason about in calculateWeight:

```ts
// Show reservations first
if (!('travelRights' in fcOrReservation)) return 0;

// Next show fare contracts that have valid travel rights right now
return getFareContractInfo(now, fcOrReservation, currentUserId)
  .validityStatus === 'valid'
    ? 1
    : 0;
```

jorelosorio:
> Ok, up to you. If it works, I don't want to be nit-picky about it 😅
> 
> But just as an example, I find this easier to reason about in calculateWeight:
> 
> ```ts
> // Show reservations first
> if (!('travelRights' in fcOrReservation)) return 0;
> 
> // Next show fare contracts that have valid travel rights right now
> return getFareContractInfo(now, fcOrReservation, currentUserId)
>   .validityStatus === 'valid'
>     ? 1
>     : 0;
> ```

Yes, many ways to organize the code. I could also propose another, but I found the implemented one convenient, ✌️. It is hard to know what can result less difficult to read from one to another, but of course I am open to change if it became difficult. 🤷 No rules about how many nested ternary to include in one sentence 😅 but it looks OK when is indented.

gorandalum (commented):
I think a unit test should be added for this.

Also just be skimming through the code it seems that active tickets will always be above reservations? I think it would be better the other way around, that reservations were above the rest of the tickets.

jorelosorio:
> I think a unit test should be added for this.
> 
> Also just be skimming through the code it seems that active tickets will always be above reservations? I think it would be better the other way around, that reservations were above the rest of the tickets.

Changed the logic to make reservations go to the top.

jorelosorio:
> > Ok, up to you. If it works, I don't want to be nit-picky about it 😅
> > But just as an example, I find this easier to reason about in calculateWeight:
> > ```ts
> > // Show reservations first
> > if (!('travelRights' in fcOrReservation)) return 0;
> > 
> > // Next show fare contracts that have valid travel rights right now
> > return getFareContractInfo(now, fcOrReservation, currentUserId)
> >   .validityStatus === 'valid'
> >     ? 1
> >     : 0;
> > ```
> 
> Yes, many ways to organize the code. I could also propose another, but I found the implemented one convenient, ✌️. It is hard to know what can result less difficult to read from one to another, but of course I am open to change if it became difficult. 🤷 No rules about how many nested ternary to include in one sentence 😅 but it looks OK when is indented.

I made the condition more verbose and more explanatory to make more sense of what it is about.

gorandalum (commented):
Looks better now :+1: But could you extract this to a function and create a unit test for it?

src/fare-contracts/FareContractAndReservationsList.tsx
```diff
@@ -33,12 +35,28 @@ export const FareContractAndReservationsList: React.FC<Props> = ({
...
+    [now, currentUserId],
+  );
 
   const fareContractsAndReservationsSorted = useMemo(() => {
-    return [...(fareContracts || []), ...(reservations || [])].sort(
-      (a, b) => b.created.toMillis() - a.created.toMillis(),
-    );
-  }, [reservations, fareContracts]);
+    return [...(fareContracts || []), ...(reservations || [])].sort((a, b) => {
+      const validityWeight = calculateWeight(b) - calculateWeight(a);
```
marius-at-atb, line 54: Could `validityWeight` be called `weightDelta` instead? 
Suggesting this because `calculateWeight` considers more than just validity, and since it is a difference between the weights, which means it can also be -1, not just 1 or 0.

marius-at-atb (approved):
Looks good!
Agree with @gorandalum's suggestion that unit testing is worth adding here.

gorandalum (commented):
On first glance I wonder where the FareContract from Firebase (with the Firebase Timestamp) is mapped to the internal FareContract type (with Date)?

jorelosorio:
> On first glance I wonder where the FareContract from Firebase (with the Firebase Timestamp) is mapped to the internal FareContract type (with Date)?

As far I can see in the TicketingContext.tsx there is not such map from one to another, it was the opposite it converted it to fire store timestamp.

gorandalum:
> > On first glance I wonder where the FareContract from Firebase (with the Firebase Timestamp) is mapped to the internal FareContract type (with Date)?
> 
> As far I can see in the TicketingContext.tsx there is not such map from one to another, it was the opposite it converted it to fire store timestamp.

That doesn't really make sense that we in the app previously converted from Date to Firestore Timestamp, and I don't see where the code is removed in this PR. I think we should know why something works, not just "It works". I'll try to check it out locally.

gorandalum:
It crashed at startup for me locally. Don't you get this @jorelosorio?
`TypeError: firstTravelRight.startDateTime.getTime is not a function (it is undefined)`

jorelosorio:
> > > On first glance I wonder where the FareContract from Firebase (with the Firebase Timestamp) is mapped to the internal FareContract type (with Date)?
> > 
> > 
> > As far I can see in the TicketingContext.tsx there is not such map from one to another, it was the opposite it converted it to fire store timestamp.
> 
> That doesn't really make sense that we in the app previously converted from Date to Firestore Timestamp, and I don't see where the code is removed in this PR. I think we should know why something works, not just "It works". I'll try to check it out locally.

Hello there :), I didn't want to go into details why "it works" here, but good we had a talk. So what I meant by the opposite was a Mock-up that was written to do the opposite (From Date to Timestamp) now that code is not needed and that was I wanted to point out.

With the latest changes, extra mapping was done around timestamps.

jorelosorio:
> It crashed at startup for me locally. Don't you get this @jorelosorio? `TypeError: firstTravelRight.startDateTime.getTime is not a function (it is undefined)`
> 
> 

Fixed.

src/ticketing/firestore.ts
```diff
@@ -20,16 +20,76 @@ export function setupFirestoreListeners(
...
+
+      travelRight.startDateTime = trStartDateTime.toDate();
+      travelRight.endDateTime = trEndDateTime.toDate();
+    }
+
+    if (travelRight.usedAccesses) {
+      travelRight.usedAccesses = travelRight.usedAccesses.map(mapUsedAccesses);
+    }
+
+    return travelRight;
```
gorandalum, line 36: This looks pretty good, just make sure that the mapping functions doesn't modify the input parameter. Instead return a new object.

https://dev.to/smeijer/stop-mutating-in-map-reduce-and-foreach-58bf

gorandalum (approved):

PR #4352, opened 2024-02-26
fix: Update the position always on prompt text

tormoseng:
After the rejected Google submission, they asked us to change the location always on prompt. We have to include either `background`, `when the app is closed`, `always in use` or `when the app is not in use`.

The text has been changed by Marked.

See more in the Slack [thread](https://mittatb.slack.com/archives/C02EEG7D8EL/p1708787438664609).


gorandalum (approved):

tormoseng:
- [x] Tested on Android
- [ ] Tested on iOS

PR #4355, opened 2024-02-27
feat: disable consumption of carnets with several travel rights

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/17062

Have also added a global message in staging with id `FareContractDetailsMultipleCarnets`, that shows for carnet tickets when `numberOfTravelRights > 1`.

### DIFF
--- a/src/ticketing/utils.ts
+++ b/src/ticketing/utils.ts
@@ -174,6 +174,12 @@ function isActiveNowOrCanBeUsedFareContract(f: FareContract, now: number) {
 export function isCanBeConsumedNowFareContract(f: FareContract, now: number) {
   if (!isCarnet(f)) return false;
   const travelRights = f.travelRights.filter(isCarnetTravelRight);
+
+  // @TODO: This is a temporary limitation because of issues with consumption of
+  // carnets with more than one travel right. Should be removed once there is a
+  // solution for this in the backend.
+  if (travelRights.length > 1) return false;
+
   return (
     hasUsableCarnetTravelRight(travelRights, now) &&
     !hasActiveCarnetTravelRight(travelRights, now)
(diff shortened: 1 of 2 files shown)


src/ticketing/utils.ts
```diff
@@ -174,6 +174,12 @@ function isActiveNowOrCanBeUsedFareContract(f: FareContract, now: number) {
 export function isCanBeConsumedNowFareContract(f: FareContract, now: number) {
   if (!isCarnet(f)) return false;
   const travelRights = f.travelRights.filter(isCarnetTravelRight);
+
+  // @TODO: This is a temporary limitation because of issues with consumption of
+  // carnets with more than one travel right. Should be removed once there is a
+  // solution for this in the backend.
+  if (travelRights.length > 1) return false;
```
marius-at-atb, line 181: Should this check depend on a remote config flag, so that once the backend issue has been resolved, we can make it work for all users immediately in the app?

rosvik: Since this only affects people that have purchased their carnets through customer service, and for some reason want to start using it on their phone instead of the t:card they use today, I thought it wasn't necessary to rely on a remote config flag here.

@Sebstorvik What do you think? Should we have the option to enable consumption of multiple carnets in 1.48 once it's supported?

marius-at-atb (approved):

src/ticketing/utils.ts
rosvik: @Sebstorvik Good point! We don't really know if it will be as easy as turning it on later, and even if it is, it's very few users that are affected. 

It's not a lot of work to do, but still not worth it I think 👍

Sebstorvik: Then we skip it👍

PR #4356, opened 2024-02-28
test: handle popovers in e2e test

tormoseng:
The performance test script needs to handle the on-behalf-of popover. Added a step to close a popover if it exists.

NB! Added test-id to the popover component.

### DIFF
--- a/e2e/test/utils/app.helper.ts
+++ b/e2e/test/utils/app.helper.ts
@@ -48,10 +48,10 @@ class AppHelper {
    * Remove the popover if it exists
    */
   async removePopover() {
-    const popoverId = `//*[@resource-id="closePopover"]`;
+    const closePopoverId = `//*[@resource-id="closePopover"]`;
     const exists = await ElementHelper.isElementExisting('closePopover', 2);
     if (exists) {
-      await $(popoverId).click();
+      await $(closePopoverId).click();
     }
   }
 
--- a/e2e/test/utils/app.helper.ts
+++ b/e2e/test/utils/app.helper.ts
@@ -45,6 +45,17 @@ class AppHelper {
   }
 
   /**
+   * Remove the popover if it exists
+   */
+  async removePopover() {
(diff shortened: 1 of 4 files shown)


src/components/popover/PopOver.tsx
```diff
@@ -50,9 +50,9 @@ export const PopOver = ({
       }
       backgroundStyle={style.backdrop}
     >
-      <View style={style.heading}>
+      <View style={style.heading} testID="popover">
```
marius-at-atb, line 53: This testID is not used as far as I can see?

e2e/test/utils/app.helper.ts
```diff
@@ -44,6 +44,17 @@ class AppHelper {
     await driver.saveScreenshot(`${screenshotsFolder}/${fileName}.png`);
   }
 
+  /**
+   * Remove the popover if it exists
+   */
+  async removePopover() {
+    const popoverId = `//*[@resource-id="closePopover"]`;
```
marius-at-atb, line 51: Was this meant to be called `closePopoverId`?

marius-at-atb (approved):
Looks good, see a couple of minor comments

src/components/popover/PopOver.tsx
tormoseng: That's true, but I added it to later be able to add tests for popovers, as well :-)

e2e/test/utils/app.helper.ts
tormoseng: True! Fixed.

src/components/popover/PopOver.tsx
marius-at-atb: Aha 😄

tormoseng: A second thought, if I'll add popover tests I rather need a test-id for the specific popover text. So I'll remove this and rather add proper test-id later 👍

PR #4362, opened 2024-03-01
feat: add logic for selectable travellers and update accessibility

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17099

This PR added some checks needed to disable the traveller selection when there's only 1 selectable profile, and updates the logic for accessibility.

https://github.com/AtB-AS/mittatb-app/assets/1777333/9ffa0bea-63cf-442b-8423-a7e10e9cef32

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
@@ -127,17 +127,20 @@ export function TravellerSelection({
   const accessibility: AccessibilityProps = {
     accessible: true,
     accessibilityRole: canSelectUserProfile ? 'button' : 'none',
-    accessibilityLabel:
-      t(
-        selectionMode == 'multiple'
-          ? PurchaseOverviewTexts.travellerSelection.a11yLabelPrefixMultiple
-          : PurchaseOverviewTexts.travellerSelection.a11yLabelPrefixSingle,
-      ) +
-      ' ' +
-      travellersDetailsText +
-      sendingToOthersAccessibility +
-      newLabelAccessibility +
-      screenReaderPause,
+    accessibilityLabel: canSelectUserProfile
+      ? t(
+          selectionMode == 'multiple'
+            ? PurchaseOverviewTexts.travellerSelection.a11yLabelPrefixMultiple
+            : PurchaseOverviewTexts.travellerSelection.a11yLabelPrefixSingle,
+        )
+      : t(
+          PurchaseOverviewTexts.travellerSelection.a11yLabelPrefixNotSelectable,
+        ) +
+        ' ' +
+        travellersDetailsText +
+        sendingToOthersAccessibility +
+        newLabelAccessibility +
+        screenReaderPause,
     accessibilityHint: canSelectUserProfile
       ? t(PurchaseOverviewTexts.travellerSelection.a11yHint)
       : undefined,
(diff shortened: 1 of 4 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -171,6 +178,7 @@ export function TravellerSelection({
       <Section {...accessibility}>
         <GenericClickableSectionItem
           onPress={travellerSelectionOnPress}
+          disabled={!canSelectUserProfile}
```
gorandalum, line 181: I think we shouldn't use a disabled button here. I know its faster to implement this way and less code, but semantically not correct.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -111,17 +115,18 @@ export function TravellerSelection({
...
 
   const sendingToOthersAccessibility = isOnBehalfOfToggle
     ? screenReaderPause + t(PurchaseOverviewTexts.onBehalfOf.sendToOthersText)
     : '';
 
   const accessibility: AccessibilityProps = {
     accessible: true,
-    accessibilityRole: 'button',
+    accessibilityRole: canSelectUserProfile ? 'button' : 'none',
     accessibilityLabel:
```
gorandalum, line 130: Do a quick check with UX regarding if "Selected travellers:" still is the preferred label for this, when the user can't actually select.

gorandalum (approved):
Nice! 

We had a discussion on https://github.com/AtB-AS/mittatb-app/pull/4359 about how to model this logic, if you haven't seen that. Me and Johannes talked with Sebastian and concluded on that having the logic implicit based on number of user profiles was ok. Just remember to document it here, with screenshots:
https://github.com/AtB-AS/docs-private/blob/main/fare-product-type-configs.md

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: We can leave it clickable, but then it's a bit weird when you can click a button but nothing shows up. Another way is probably switch to another component which are not clickable if they can't select it. Which one do you think is better? I'm open to suggestions and I'll make another PR to update it. 😄

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: I'll consult with @ckbilstad in a few minutes

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
gorandalum: Ment more like dragging out the stuff inside `GenericClickableSectionItem` to be reusable, and then something like this:
```
canSelectUserProfile 
  ? <GenericClickableSectionItem ...>...</GenericClickableSectionItem>
  : <GenericSectionItem ...>...</GenericSectionItem>
```

reidzeibel: Yes that is what i meant by "switch to other component when it's not clickable" 👍

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: Updated to "Traveller:" as discussed with @ckbilstad

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: I'm making a separate PR for this one :)

reidzeibel: https://github.com/AtB-AS/mittatb-app/pull/4364

PR #4364, opened 2024-03-01
refactor: rework TravellerSelection to add non-clickable section

reidzeibel:
context: https://github.com/AtB-AS/mittatb-app/pull/4362#discussion_r1508915274

This PR attempts to refactor and remove the disabled part of the `GenericClickableSectionItem`, and instead use a `GenericSectionItem` if it is not clickable. 

The `<Section>` component is needed to be put around on both `GenericClickableSectionItem` and `GenericSectionItem` to ensure that the corner rounding works properly. 

Open to suggestion on how to implement this in a better way 😊 

### Acceptance Criteria

- [ ] Traveller selection, when it is clickable, should behave normally, and visually the corners are rounded
- [ ] Traveller selection, when there's only 1 selectable profile, should not be clickable, and visually the corners are rounded
- [ ] Screenreader behaves as usual for the clickable part
- [ ] Screenreader says "Traveller" instead of "Selected Traveller" when it is not clickable

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
@@ -54,6 +54,8 @@ export function TravellerSelection({
   isOnBehalfOfToggle,
 }: TravellerSelectionProps) {
   const {t, language} = useTranslation();
+  const styles = useStyles();
+
   const {
     open: openBottomSheet,
     close: closeBottomSheet,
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
```diff
@@ -182,60 +193,85 @@ export function TravellerSelection({
             : t(PurchaseOverviewTexts.travellerSelection.titleNotSelectable)
         }
       />
-      <Section {...accessibility}>
```
gorandalum, line 185: The naming doesn't really make sense since we now have a `TravellerSeletionWrapper` inside the `TravellerSelection` component. So should find a better name for it.

How do you think it looks the other way around, without introducing a new component? Like this ish? (pseudocode)

```tsx
const content = (
   <View style={styles.sectionContentContainer}>
          <View style={{flex: 1}}>
            <ThemeText type="body__primary--bold">
              {multipleTravellerCategoriesSelectedFrom
  ...
);
```

and then:
```tsx
<Section {...props}>
  {canSelectUserProfile
    ? <GenericClickableSectionItem
        onPress={travellerSelectionOnPress}
        ref={ref}
    >
      {content}
    </GenericClickableSectionItem>)
    : <GenericSectionItem>{content}</GenericSectionItem>}
</Section>
```

gorandalum (approved):
Looks good! Just a nitpick about the naming :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/TravellerSelection.tsx
reidzeibel: I think that should also work, I'll try 👍

reidzeibel: Made the change, looks better now I think, thank you!

tormoseng:
@reidzeibel Both for TalkBack and VoiceOver: 
When you have multiple traveller options, the screenreader says "Selected traveller button activate to...", but it doesn't read what traveller category that is selected. After changing and confirming the traveller, it reads e.g. "Senior" back at the purchase details - but this is only after you have changed the traveller. 

Note: It also doesn't read the "new" label. It does so only after you have changed the traveller category.

If you only have a single traveller category, everything is fine.

reidzeibel:
@tormoseng I'll check it out, thanks!

reidzeibel:
@tormoseng fix has been merged, if there's a new staging app, you can on that one.

tormoseng:
Re-tested OK!

PR #4365, opened 2024-03-04
chore: Bump to version 1.49

rosvik:


gorandalum (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "atb",
-  "version": "1.48.0",
+  "version": "1.49.0",
```
rosvik, line 3: @tormoseng I bumped the version for e2e tests as well. Should we add that to the checklist so it's done every time?

tormoseng: Would be nice!

PR #4367, opened 2024-03-04
build: remove symlink to `/usr/local/bin/2to3` on buildscript

reidzeibel:
another attempt to fix build issue, similar to https://github.com/AtB-AS/mittatb-app/pull/4366

### DIFF
--- a/.github/workflows/build-store-ios.yml
+++ b/.github/workflows/build-store-ios.yml
@@ -39,7 +39,8 @@ jobs:
       - name: Generate native assets
         run: |
           rm '/usr/local/bin/2to3'
-          brew install imagemagick && yarn generate-native-assets
+          brew install imagemagick
+          yarn generate-native-assets
       - name: Disable widget
         if: ${{ env.ENABLE_WIDGET != 'true' }}
         run: bundle exec configure_extensions remove ios/atb.xcodeproj app departureWidget AtbAppIntent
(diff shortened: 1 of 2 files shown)


.github/workflows/build-store-ios.yml
```diff
@@ -37,7 +37,9 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.MATCH_SSH_PRIVATE_KEY }}
       - name: Generate native assets
-        run: brew install --overwrite imagemagick && yarn generate-native-assets
+        run: |
+          rm '/usr/local/bin/2to3'
+          brew install imagemagick && yarn generate-native-assets
```
rosvik, line 42:
```suggestion
          rm '/usr/local/bin/2to3'
          brew install imagemagick
          yarn generate-native-assets
```

rosvik (approved):
Worth a shot 🤞 Nit: Might as well use newlines for all commands when we do it for the one you added.

rosvik (changes requested):
I'll trigger a staging build, and we'll see it that works before we merge this

.github/workflows/build-store-ios.yml
rosvik, line 1: @reidzeibel I'm realising we're running the build with the other yml file from the one we're changing 😅 Try making (some of) these changes to build-staging-ios.yml instead.

reidzeibel:
Build success on staging : https://github.com/AtB-AS/mittatb-app/actions/runs/8141619752 

ready for review @rosvik

rosvik (approved):
Great job figuring this out!

This doesn't look like a very robust solution (It will probably break the moment imagemagick updates the python version they rely on? 😅), so after this, I think we should use some time to find a better solution.

But for now, let's hope this get the build up and running again 🚀

PR #4369, opened 2024-03-04
build: replace gxargs with xargs

rosvik:
When trying to run register-local-app-version.sh, me and @hanne-at-atb spent some time looking at what "gxargs" is, where to install it, and how it differs from "xargs". While doing so, we found a alternative solution that uses xargs instead ([source](https://superuser.com/questions/467176/replacement-for-xargs-d-in-osx#comment581887_467192)). Since it comes pre-installed on macOS, this means we don't have to install it for CI, or when running register-local-app-version.sh locally!

An alternative solution would be to add docs for running `brew install findutils` while setting up the project, but I think this is nicer. :)

### DIFF
--- a/scripts/ios/override-config-files.sh
+++ b/scripts/ios/override-config-files.sh
@@ -5,12 +5,8 @@ echo "Copy .netrc file for mapbox api token"
 cat .netrc >> ~/.netrc
 chmod 600 ~/.netrc
 
-echo "Installing pre-build dependencies"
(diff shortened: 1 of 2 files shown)


hanne-at-atb (approved):
🎉

gorandalum (approved):

rosvik:
I'll wait with merging this until the build is working. Just so it's not broken in more than one way at once 😄

PR #4371, opened 2024-03-05
chore: increase FareProductGroup heading marginTop

marius-at-atb:
To more clearly indicate that the FareProductGroup heading belongs to the content below it and not above it, the marginTop is increased. As discussed with @ckbilstad, this was the original intention of the design.
UPDATE: The same done to the ticketAssistantTile heading after a good catch by @rosvik.

| Before | After |
|--------|-------|
|  |  |

### DIFF
--- a/src/components/heading/ContentHeading.tsx
+++ b/src/components/heading/ContentHeading.tsx
@@ -20,7 +20,7 @@ export function ContentHeading({
   const styles = useStyles();
 
   return (
-    <View style={[style, styles.container]}>
+    <View style={[styles.container, style]}>
       <ThemeText
         type="body__secondary"
         color={color}
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/Components/FareProducts/FareProductGroup.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/Components/FareProducts/FareProductGroup.tsx
@@ -91,6 +91,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
   heading: {
     margin: theme.spacings.medium,
     marginLeft: theme.spacings.xLarge,
+    marginTop: theme.spacings.large,
   },
   fareProductsContainer: {
     flex: 1,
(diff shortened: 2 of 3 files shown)


rosvik (commented):
Should there be extra padding on the ticket assistant heading as well?

rosvik (approved):

marius-at-atb:
Note: both instances of the heading style updated here, should ideally have the same source. Several ways to achieve this, but not sure if it's worth it.

rosvik:
Yes, we should usually use ContentHeading, but because of the icons, but because of the icons, it doesn't fit here. Could you use it for the ticket assistant heading?

marius-at-atb:
Yes, it can be like this:
```ts
<ContentHeading
  style={styles.heading}
  color="primary"
  text={t(TicketingTexts.ticketAssistantTile.title)}
/>
```

However in `ContentHeading.tsx`, it looks like this:
```ts
<View style={[styles.container, style]}>
...
container: {
},
```

Which overwrites the margin, so then the heading style would have to use `paddingTop` not to be overwritten.
Do you think we can safely switch the order of `styles.container` and `style` in `ContentHeading.tsx`? Normally the style you pass in should be the one that overwrites, because by definition, it is a custom style..

marius-at-atb:
Update: the only case I can find where `ContentHeading` style is passed in, is in `DeparturesWidget`, which only passes in `marginBottom`, so it should be safe.

rosvik:
I think you're right that the props should be the other way around. But the way I've tried to solve this other places is to add rowGap=medium to the container element (for the whole view), instead of applying margin to each element individually. Could that work here as well?

marius-at-atb:
Using `rowGap` can be done like this: https://github.com/AtB-AS/mittatb-app/pull/4374/files

marius-at-atb:
The duplicated
```ts
heading: {
     marginBottom: theme.spacings.medium,
     marginLeft: theme.spacings.xLarge,
  }
```
still isn't solved there, but I guess it isn't worth e.g. its own `useHeadingStyle` hook. What do you think?

rosvik:
Yeah, no need to add complexity just to avoid one bit of duplicate style 😄 

Thanks for investigating and setting up another PR, but it's a bit hard to say what the best solution is without doing it myself, so what do you suggest? To me it looks a bit unnecessary to use ContentHeading when we end up overwriting most of its style anyway.

marius-at-atb:
In this case I think it is easiest to just keep it as is in this PR. `rowGap` can be fragile as well, especially when using multiple components.
To avoid the duplicated code, I would't mind having this:
```ts
const useFareProductHeadingStyle = StyleSheet.createThemeHook((theme) => ({
  margin: theme.spacings.medium,
  marginLeft: theme.spacings.xLarge,
  marginTop: theme.spacings.large,
}));
```

marius-at-atb:
Is [this](https://github.com/AtB-AS/mittatb-app/pull/4376/files) an accepted practice @gorandalum?

gorandalum:
> Is [this](https://github.com/AtB-AS/mittatb-app/pull/4376/files) an accepted practice @gorandalum?

I don't think we want styles shared between components. Styles should mainly be local, and duplicated if necessary. In general, there isn't a goal to be DRY with our styles. If we want to be DRY, extract components, not styles.

gorandalum (approved):
I think this looks good :+1:

marius-at-atb:
Merged just with these changes then.
I actually think that this was a pretty good case for a shared style hook. When one is changed, we want the others to change as well. However with the complexity that comes with `TransportModes`, I think it was only desirable to share a style, not a component.

PR #4379, opened 2024-03-06
fix: make it possible to choose tcard or mobile when no inspectable token

marius-at-atb:
Slack discussion [here](https://mittatb.slack.com/archives/C0116FMPX4Y/p1709718806225559)


marius-at-atb:
Solved by https://github.com/AtB-AS/mittatb-app/pull/4891

PR #4384, opened 2024-03-07
refactor: Make getSignedToken use react-query

gorandalum:
Using react-query and letting that handle the refetch interval instead.


src/mobile-token/hooks/useGetSignedTokenQuery.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const useGetSignedTokenQuery = () => {
+  const {nativeToken} = useMobileTokenContextState();
+  return useQuery({
+    queryKey: ['getSignedToken', nativeToken],
+    queryFn: async () => {
+      if (!nativeToken) return undefined;
+      try {
+        return await mobileTokenClient.encode(nativeToken);
+      } catch (err: any) {
+        notifyBugsnag(err, {
```
marius-at-atb, line 22: Assuming it is on purpose to no longer send `userId` here?

marius-at-atb (approved):
Lgtm!
Would have been nice with a PR description as well.

PR #4386, opened 2024-03-12
refactor: Use react-query for listing tokens

gorandalum:
This is to enable more fine-grained error handling in mobile
token flow when updating the sdk. When we have decoupled the
different requests they can have their own error handling.

Without this change we would need a stored state specifying where
in the flow the error happened, so the correct step was retried.

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -44,7 +44,6 @@ import {
   LIST_REMOTE_TOKENS_QUERY_KEY,
   useListRemoteTokensQuery,
 } from '@atb/mobile-token/hooks/useListRemoteTokensQuery';
-import {LOAD_NATIVE_TOKEN_QUERY_KEY} from '@atb/mobile-token/hooks/useLoadNativeTokenQuery';
 
 const SIX_HOURS_MS = 1000 * 60 * 60 * 6;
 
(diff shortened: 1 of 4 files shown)


src/mobile-token/MobileTokenContext.tsx
```diff
@@ -97,7 +100,11 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
   const mobileTokenEnabled = hasEnabledMobileToken();
 
   const enabled =
-    mobileTokenEnabled && userId && authStatus === 'authenticated';
+    mobileTokenEnabled && !!userId && authStatus === 'authenticated';
```
rosvik, line 103: Is the `!!userId` change intentional here? If so, why is that done? 🤔

rosvik (approved):
Nice! Looks like a clean solution to me.

src/mobile-token/hooks/useListRemoteTokensQuery.tsx
hanne-at-atb, line 1: As we are adding more hooks with queries, have we found a pattern of some kind for folder structure and where to place them? I notice that this is added to the modules' own `hooks` folder. At the same time we have another query hook outside of this folder: https://github.com/AtB-AS/mittatb-app/blob/master/src/mobile-token/use-token-toggle-details.ts

hanne-at-atb (approved):

src/mobile-token/MobileTokenContext.tsx
gorandalum: To make it typed as `boolean`.
Screenshot 2024-03-12 at 15 24 14

src/mobile-token/hooks/useListRemoteTokensQuery.tsx
gorandalum: I'll move `use-token-toggle-details.ts` inside the hooks folder. About the general principle we'll take a look soon. I think maybe these mobile-token-hooks needs to live in the mobile-token module, but won't conclude yet.

hanne-at-atb: Nice, I see they don't follow the same naming convention either so a little discussion soon would be nice 🌻

gorandalum: That's just my bad 😅

PR #4387, opened 2024-03-12
fix: add environment variables for the update devices workflow

mortennordseth:
FRAM and NFK are not able to run the "update devices" workflow due to missing environment variables, ref https://github.com/AtB-AS/mittatb-app/actions/runs/8246762529/job/22553490228. This fixes that.


reidzeibel (approved):
No LGTM because nothing to look into. 🤷‍♂️ 

All good

PR #4389, opened 2024-03-14
fix: Only show login onboarding if not logged in

gorandalum:
This was not ment to happen, and we are not really sure if it can
happen or not without messing with onboarding flags in the debug
menu. But the safeguard is added just in case.

### Acceptance criteria
- [ ] When resetting onboarding flags while logged in, the login onboarding should not show


marius-at-atb (approved):

rosvik (approved):

tormoseng:
👍

PR #4390, opened 2024-03-14
refactor: Toggle token as mutation

gorandalum:
This also has the added benefit that screens doing toggling now
don't need to have their own internal state for loading/error, they
can use the mutation state.

Also removed token id as key in the list tokens query, as it didn't
really make sense to provide the token id when setting query data
through the client provider. Instead added a useEffect in
MobileTokenContext which invalidates the list tokens query when
native token id changes.


hanne-at-atb (approved):
Nice 🎊

PR #4391, opened 2024-03-14
feat: redesign LoginActiveFareContractWarningScreen

hanne-at-atb:
* Redesign screen for informing users of consequences of logging in when they have active ticket

Closes: https://github.com/AtB-AS/kundevendt/issues/16899

### DIFF
--- a/src/fare-contracts/FareContractInfo.tsx
+++ b/src/fare-contracts/FareContractInfo.tsx
@@ -59,6 +59,7 @@ export type FareContractInfoDetailsProps = {
   now?: number;
   validTo?: number;
   fareProductType?: string;
+  showSummarisedFareContract?: boolean;
 };
 
 export const FareContractInfoHeader = ({
@@ -135,6 +136,7 @@ export const FareContractInfoDetails = (
     userProfilesWithCount,
     status,
     preassignedFareProduct,
+    showSummarisedFareContract,
   } = props;
   const {t, language} = useTranslation();
   const styles = useStyles();
@@ -155,14 +157,22 @@ export const FareContractInfoDetails = (
       <View style={styles.fareContractDetails}>
         <View style={styles.details}>
           <FareContractDetail
-            header={t(FareContractTexts.label.travellers)}
+            header={
+              showSummarisedFareContract
+                ? undefined
+                : t(FareContractTexts.label.travellers)
+            }
             content={userProfilesWithCount.map((u) =>
               userProfileCountAndName(u, language),
             )}
           />
           {tariffZoneSummary && (
             <FareContractDetail
-              header={t(FareContractTexts.label.zone)}
+              header={
+                showSummarisedFareContract
+                  ? undefined
+                  : t(FareContractTexts.label.zone)
+              }
               content={[tariffZoneSummary]}
             />
           )}
--- a/src/fare-contracts/FareContractInfo.tsx
+++ b/src/fare-contracts/FareContractInfo.tsx
@@ -136,7 +136,6 @@ export const FareContractInfoDetails = (
     userProfilesWithCount,
     status,
     preassignedFareProduct,
-    showSummarisedFareContract,
   } = props;
   const {t, language} = useTranslation();
   const styles = useStyles();
@@ -157,22 +156,14 @@ export const FareContractInfoDetails = (
       <View style={styles.fareContractDetails}>
         <View style={styles.details}>
           <FareContractDetail
-            header={
-              showSummarisedFareContract
-                ? undefined
-                : t(FareContractTexts.label.travellers)
-            }
+            header={t(FareContractTexts.label.travellers)}
             content={userProfilesWithCount.map((u) =>
               userProfileCountAndName(u, language),
             )}
           />
           {tariffZoneSummary && (
             <FareContractDetail
-              header={
-                showSummarisedFareContract
-                  ? undefined
-                  : t(FareContractTexts.label.zone)
-              }
+              header={t(FareContractTexts.label.zone)}
               content={[tariffZoneSummary]}
             />
           )}
--- a/src/fare-contracts/FareContractOrReservation.tsx
+++ b/src/fare-contracts/FareContractOrReservation.tsx
@@ -5,6 +5,7 @@ import {TicketingTexts, useTranslation} from '@atb/translations';
 import {SimpleFareContract} from '@atb/fare-contracts';
 import {ErrorBoundary} from '@atb/error-boundary';
 import {TicketHistoryMode} from '@atb/ticket-history';
+import {StyleSheet} from '@atb/theme';
 
 export function FareContractOrReservation({
   fcOrReservation,
@@ -21,6 +22,7 @@ export function FareContractOrReservation({
 }) {
   const {t} = useTranslation();
   const hasActiveTravelCard = false;
+  const styles = useStyles();
 
   if ('transactionId' in fcOrReservation) {
     return (
@@ -44,8 +46,15 @@ export function FareContractOrReservation({
           sentToOthers={mode === 'sent'}
           onPressDetails={onPressFareContract}
           testID={'ticket' + index}
+          style={styles.fareContract}
         />
       </ErrorBoundary>
     );
   }
 }
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  fareContract: {
+    paddingBottom: theme.spacings.large,
+  },
+}));
--- a/src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
+++ b/src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
@@ -5,7 +5,6 @@ import React from 'react';
 import {ScrollView, View} from 'react-native';
 import {Button} from '@atb/components/button';
 import {ThemeText} from '@atb/components/text';
-import {ArrowRight} from '@atb/assets/svg/mono-icons/navigation';
 import {useFocusOnLoad} from '@atb/utils/use-focus-on-load';
 import {StaticColorByType} from '@atb/theme/colors';
 import {
@@ -15,7 +14,6 @@ import {
 import {SimpleFareContract} from '@atb/fare-contracts';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy/navigation-types';
-import {PressableOpacity} from '@atb/components/pressable-opacity';
 import {useTimeContextState} from '@atb/time';
 import {TransitionPresets} from '@react-navigation/stack';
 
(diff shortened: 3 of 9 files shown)


src/translations/screens/subscreens/Login.ts
```diff
@@ -49,22 +49,23 @@ const LoginTexts = {
...
+      'Billetten din slettes ved innlogging',
     ),
     body: _(
-      'Billetten kan **ikke** overføres til din profil og vil slettes ved innlogging. Dersom du trenger billetten, vent til den er utløpt **før** du logger inn.',
-      'This ticket can **not** be transferred to your profile, and will be deleted upon login. If you need the ticket, wait until it has expired **before** logging in.',
-      'Billetten kan **ikkje** overførast til din profil og vil slettast ved innlogging. Dersom du treng billetten, vent til den er gått ut **før** du loggar inn.',
+      'Billetten din kan ikke overføres til en profil, fordi den er kjøpt anonymt. Vent med å logge inn til billetten utløper, for å unngå at den blir slettet. Slettede billetter **refunderes ikke.**',
+      'Your ticket cannot be transferred to a profile because it was purchased anonymously. Please wait until it expires before you log in, to avoid deletion. Deleted tickets are **non-refundable.**',
+      'Billetten din kan ikkje overførast til ein profil, sidan den vart kjøpt anonymt. Vent med å logge inn til billetten går ut, for å unngå at den blir sletta. Sletta billettar blir **ikkje refunderte.**',
```
rosvik, line 61:
```suggestion
      'Billetten din kan ikkje overførast til ei profil, sidan han vart kjøpt anonymt. Vent med å logge inn til billetten går ut, for å unngå at han blir sletta. Sletta billettar blir **ikkje refunderte.**',
```

src/translations/screens/subscreens/Login.ts
```diff
@@ -49,22 +49,23 @@ const LoginTexts = {
...
   },
   activeFareContractPrompt: {
+    header: _('Logg inn', 'Log in', 'Logg inn'),
     title: _(
-      'Vent litt, du har en gyldig billett',
-      'Wait a sec – you have a valid ticket',
-      'Vent litt, du har ein gyldig billett',
+      'Billetten din slettes ved innlogging',
+      'Your ticket will be deleted upon logging in',
+      'Billetten din slettes ved innlogging',
```
rosvik, line 56:
```suggestion
      'Billetten din slettast ved innlogging',
```

src/fare-contracts/FareContractOrReservation.tsx
```diff
@@ -44,8 +46,15 @@ export function FareContractOrReservation({
...
+          style={styles.fareContract}
         />
       </ErrorBoundary>
     );
   }
 }
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  fareContract: {
+    paddingBottom: theme.spacings.large,
```
rosvik, line 58: Do we want this padding to apply to all fare contracts? Could we wrap the SimpleFareContract in a view, and apply padding in the warning screen instead?

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
```diff
@@ -126,27 +118,24 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     marginVertical: theme.spacings.medium,
     textAlign: 'center',
   },
-  errorMessage: {
-    marginBottom: theme.spacings.medium,
+  overlay: {
```
rosvik, line 121: Could we make the ticket transparent instead of adding a overlay?

src/translations/screens/subscreens/Login.ts
marius-at-atb: Not sure about this suggestion?
https://ordbokene.no/nn/search?q=profil&scope=ei&perPage=20

I think it is fine as is.

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
hanne-at-atb: I tried adding `transparent` to the `<GenericSectionItem>` in `<FareContractView>`, also tried to set the opacity there but that didn't work / look so good 😢 @gorandalum suggested the overlay approach, but if you can give me some pointers on other approach with transparent I can try that 😄

src/fare-contracts/FareContractOrReservation.tsx
hanne-at-atb: It's due to removing `withBottomPadding` from FareContractView: https://github.com/AtB-AS/mittatb-app/pull/4391/files#diff-bb41825cddfbc0a8a96c04b2818fc2ce9a9a49d9afe0ad6fade19b1ba9c87aabL104
and we don't want this padding to show for all `SimpleFareContracts`

marius-at-atb:
Question regarding UI/UX - should the buttons be locked to the bottom of the screen instead of inside the scrollable view? With large text size it looks like this:

src/translations/screens/subscreens/Login.ts
hanne-at-atb: Agree with keeping ein, but what about han vs. den? @marius-at-atb / @rosvik

rosvik:
> should the buttons be locked to the bottom of the screen instead of inside the scrollable view? 

I think so! `<FullScreenFooter>` can be used for this

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
marius-at-atb: You can add opacity to the `fareContract` style

src/translations/screens/subscreens/Login.ts
rosvik: Huh, I was sure it was "ei profil", but you're right 😅 I still think we should use "han", since "billetten" is hankjønn. https://www.sprakradet.no/svardatabase/sporsmal-og-svar/den-i-nynorsk-for-hanho/

marius-at-atb:
Also an edge case I just encountered - only `hasActiveFareContracts` is checked for whether the screen should show or not. However, in this state it should probably also show:

src/fare-contracts/FareContractInfo.tsx
```diff
@@ -59,6 +59,7 @@ export type FareContractInfoDetailsProps = {
   now?: number;
   validTo?: number;
   fareProductType?: string;
+  showSummarisedFareContract?: boolean;
```
marius-at-atb, line 62: This one is unused

marius-at-atb (approved):
Looks great!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_FareContractsScreen.tsx
```diff
@@ -224,6 +224,9 @@ const useStyles = StyleSheet.createThemeHook((theme: Theme) => ({
   content: {
     padding: theme.spacings.medium,
   },
+  fareContract: {
+    paddingBottom: theme.spacings.medium,
+  },
```
rosvik, line 229: Not in use?

hanne-at-atb:
> Also an edge case I just encountered - only `hasActiveFareContracts` is checked for whether the screen should show or not. However, in this state it should probably also show: 

Nice catch, fixed now 👏🏼

src/fare-contracts/FareContractInfo.tsx
hanne-at-atb: 😓
Fixed now 😄

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
hanne-at-atb: Done 💯

src/fare-contracts/FareContractOrReservation.tsx
hanne-at-atb: Since removing the overlay, I did as you suggested 😄

src/translations/screens/subscreens/Login.ts
hanne-at-atb: Since both are allowed and `den` was suggested by designers through sketches in Figma, we keep it like this ⛵

rosvik (commented):
Bottom padding around fare contracts are inconsistent in the app now:



I don't think this PR should have to change any styling of fare contracts globally (SimpleFareContract or FareContractView), but handle everything in the warning screen.

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
gorandalum: Not totally sure about using opacity instead of overlay, at least need to check it. The reason it works pretty ok for AtB is that `background_accent` is actually a dark color, but how does it work for Fram?

Screenshot 2024-03-18 at 09 11 21

gorandalum:
> Also an edge case I just encountered - only `hasActiveFareContracts` is checked for whether the screen should show or not. However, in this state it should probably also show:

Might not be worth the effort for this corner case, you can get a feel for it @hanne-at-atb and decide what you think. If we also shows the screen on reservations, then we might need to rephrase the texts in the warning screen for the case where there is only a reservation, and also show that reservation instead of showing fare contract (since it is no fare contract to show). We shouldn't do it half-way, so shouldn't check for reservations to decide whether to show the screen, but only present fare contracts in the screen.

src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
hanne-at-atb: Talked to @hannahATB and we figured this works fine 😄 FRAM normal vs dark mode👇🏼

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_FareContractsScreen.tsx
hanne-at-atb: Removed this and removed the style on SimpleFareContract 😄

gorandalum (approved):

rosvik (approved):
Nice 👏

hanne-at-atb:
> > Also an edge case I just encountered - only `hasActiveFareContracts` is checked for whether the screen should show or not. However, in this state it should probably also show:
> 
> Might not be worth the effort for this corner case, you can get a feel for it @hanne-at-atb and decide what you think. If we also shows the screen on reservations, then we might need to rephrase the texts in the warning screen for the case where there is only a reservation, and also show that reservation instead of showing fare contract (since it is no fare contract to show). We shouldn't do it half-way, so shouldn't check for reservations to decide whether to show the screen, but only present fare contracts in the screen.

Good point, made some adjustments now to show the reservation in this view @gorandalum / @marius-at-atb 🎫

gorandalum:
> Good point, made some adjustments now to show the reservation in this view @gorandalum / @marius-at-atb 🎫

Nice, remember to let designers now so that Figma is up to date regarding what can be shown in the app.

PR #4393, opened 2024-03-15
refactor: Load native token as query

gorandalum:
This ended up with more code changes than anticipated. After all
the PRs are done the mobiletoken functionality need to be
throroughly tested.

The main changes here are:
- Using react-query for loading mobile token. Since it has potential
side effect it could (should?) have been a mutation, but using a
query was more inline with how it was previously used.
- Restructured all query keys for mobile token queries under parent
key 'mobile-token', so when retrying/logging out the parent key may
be used to reset all mobile token requests.
- Merged 'mobileTokenStatus', 'barcodeStatus' and
'deviceInspectionStatus' into just a 'mobileTokenStatus'. For
simplicity also a field 'isInspectable' is now returned from the
MobileTokenContext.


rosvik (approved):
Nice! This looks good 👍

PR #4399, opened 2024-03-20
fix: fixed scroll of ActiveTokenOnPhoneRequiredForFareProductScreen

hanne-at-atb:
When you have toggled to t:card and want to purchase for instance boat or youth ticket you have to toggle to phone. In order to do so we prompt the user to do it by showing a screen. If a user has increased the font size on their device they might not be able to currently read the whole info text or select the confirm "bytt til mobil" button. 

_Acceptance criteria_
- [ ] The user can read all info in the screen and proceed with doing the swap to use token on mobile.

_Before_  👉🏼 👉🏼 👉🏼   _After_ 

Note: This shows the entire screen possible to scroll to, meaning scrolled all the way down. So the bottom text is cut and the button is not visible and clickable in our current solution.

### DIFF
--- a/src/stacks-hierarchy/Root_ActiveTokenOnPhoneRequiredForFareProductScreen.tsx
+++ b/src/stacks-hierarchy/Root_ActiveTokenOnPhoneRequiredForFareProductScreen.tsx
@@ -19,7 +19,6 @@ import MobileTokenOnboarding from '@atb/translations/screens/subscreens/MobileTo
 import Ticketing from '@atb/translations/screens/Ticketing';
 import {getDeviceNameWithUnitInfo} from '@atb/select-travel-token-screen/utils';
 import {useToggleTokenMutation} from '@atb/mobile-token';
-import {FullScreenFooter} from '@atb/components/screen-footer';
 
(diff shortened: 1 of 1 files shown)


gorandalum:
I don't really get the before/after pictures. It only seems like a larger screen on the after picture which enables all info to be shown 😅

hanne-at-atb:
> I don't really get the before/after pictures. It only seems like a larger screen on the after picture which enables all info to be shown 😅

Sorry, it was not very well stated. The screens are scrolled all the way down on both. I.e. before the text was cut and it was not possible to scroll all the way down to the button when the fontSize is increased.

rosvik (approved):

hanne-at-atb:
Initially fixed the button to the bottom of the screen with `FullScreenFooter`, but in discussions with @ckbilstad and @gorandalum we found that the user should read the content at the bottom of the page as it is important. Therefore not fixing the button. 
This PR just made the whole screen scrollable so that all of the text can be read and the button can be reached 😄

tormoseng:
Tested OK!

PR #4400, opened 2024-03-20
fix: Presented sent date should show the created date

gorandalum:
### Background
It was using the valid from date, which doesn't work when
purchasing a ticket with a future start date.

### Solution
Using the created date as the sent date.

Before/After (Notice the difference between "Sendt" and "Kjøpt" dates in the before picture)
<div>

</div>

### Acceptance criteria
- [ ] When purchasing a ticket for others with a future start date, the sent date on the sent tickets screen is the purchase date.


reidzeibel (approved):
Good catch!

rosvik (approved):

PR #4405, opened 2024-03-21
feat: filter empty fare product groups on ticket selection screen

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17365

This PR will filter out empty fare product groups introduced by https://github.com/AtB-AS/firestore-configuration/pull/497.

For users without the ability to purchase HjemJobbHjem tickets, they won't have any products in that group, resulting in an empty group and weird display.

After this PR is merged, the empty product group should not be shown.


rosvik (commented):
I think this can be solved in a simpler way by first calculating groupedFareProducts like before, and then using `groupedFareProducts.filter` to remove the ones with zero fareProducts. 

Could you take a look to see if that works in practice?

gorandalum:
> @rosvik yes we can do this : `.filter((group) => group.fareProducts.length > 0)`, but then we will traverse the array 2 times instead of just once? what do you think?

No problem traversing array twice, the ti to performance is next-to-none, and worth it if it gives cleaner code.

reidzeibel:
@rosvik @gorandalum ready 😊

rosvik (approved):
I like it :)

Screenshot 2024-03-22 at 10 22 27

PR #4406, opened 2024-03-21
feat: disable on-behalf-of when anonymous

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17206

Adds auth state check for on-behalf-of features, if not logged in, then don't show on-behalf-of

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Travellers/SingleTravellerSelection.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Travellers/SingleTravellerSelection.tsx
@@ -12,6 +12,7 @@ import {StyleSheet} from '@atb/theme';
 import {HoldingHands} from '@atb/assets/svg/color/images';
 import {useOnBehalfOfEnabled} from '@atb/on-behalf-of';
 import {TravellerSelectionBottomSheetType} from './types';
+import {useAuthState} from '@atb/auth';
 import {getTravellerInfoByFareProductType} from './../../utils';
 
 export function SingleTravellerSelection({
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Travellers/SingleTravellerSelection.tsx
```diff
@@ -33,6 +34,10 @@ export function SingleTravellerSelection({
   const isOnBehalfOfEnabled =
     useOnBehalfOf() && fareProductTypeConfig.configuration.onBehalfOfEnabled;
 
+  const isLoggedIn = useAuthState().authenticationType === 'phone';
```
rosvik, line 37: Nit: We mostly implement things like this with `const {authenticationType} = useAuthState();`, that way we don't need to call useAuthState more than once when you need more than one field. Just a note, no blocker for merging the PR :)

rosvik (approved):
Looks good 👍

marius-at-atb:
Do we inform the users anywhere that on behalf of is possible, but requires login?
Otherwise it might be a completely undiscovered feature for anonymous users.

Perhaps the option should still show, but be disabled / gray, and if you try to click it, you get a message informing about this?

reidzeibel:
@marius-at-atb Not that I know of, since this feature was requested by the OMS. @Sebstorvik might have an idea for that.

Sebstorvik:
Very good point @marius-at-atb. We talked about the issue in refinement and the team pointed out that this might be a bit rushed without actually taking the time to think through the proper solution. I will ask around in the OMS and see if we should maybe postpone it and do it properly or if they still want it disabled.

PR #4410, opened 2024-03-27
feat: add new global message component on purchase confirmation screen

reidzeibel:
context: https://github.com/AtB-AS/kundevendt/issues/17075#issuecomment-2014785178

Straightforward PR, tried to add rule, but @gorandalum mentioned that it should be a different context. Suggestions to context naming are appreciated. :)

### DIFF
--- a/src/global-messages/types.ts
+++ b/src/global-messages/types.ts
@@ -11,7 +11,7 @@ export enum GlobalMessageContextEnum {
   appTicketing = 'app-ticketing',
   appPurchaseOverview = 'app-purchase-overview',
   appPurchaseConfirmation = 'app-purchase-confirmation',
-  appPurchaseInformation = 'app-purchase-information',
+  appPurchaseConfirmationBottom = 'app-purchase-confirmation-bottom',
   appFareContractDetails = 'app-fare-contract-details',
   appDepartureDetails = 'app-departure-details',
   appTripDetails = 'app-trip-details',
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -453,6 +453,14 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
             isMarkdown={true}
           />
         )}
+        <GlobalMessage
+          style={styles.purchaseInformation}
+          globalMessageContext={GlobalMessageContextEnum.appPurchaseInformation}
+          textColor="secondary"
```
gorandalum, line 459: Is `textColor="secondary"` correct? 

From Figma:
Screenshot 2024-04-02 at 11 24 31

src/global-messages/types.ts
```diff
@@ -11,6 +11,7 @@ export enum GlobalMessageContextEnum {
   appTicketing = 'app-ticketing',
   appPurchaseOverview = 'app-purchase-overview',
   appPurchaseConfirmation = 'app-purchase-confirmation',
+  appPurchaseInformation = 'app-purchase-information',
```
gorandalum, line 14: Hard to intuitively understand where 'app-purchase-information' is located in the app. Maybe something like 'app-purchase-confirmation-bottom' would be more descriptive?

reidzeibel: Updated

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
reidzeibel: Updated, was using non-subtle global message, so the color change wasn't noticeable. Thanks for pointing it out!

rosvik (approved):

marius-at-atb (approved):

PR #4411, opened 2024-04-02
feat: update CouterIconBox styling and behavior

rosvik:
- Updates logic for limit in TransportationIconBoxList to match sketches
- Makes sure that CounterIconBox uses the exact same styling as TransportationIcons, with font scale taken into consideration, except that the box is allowed to grow horizontally if the number is wider than it is tall.
	- `textType` is added as a required param since there isn't a "correct" text type for a given size
	- `spacing` as an optional param, to make it work in ResultItems.

<div>

</div>

closes https://github.com/AtB-AS/kundevendt/issues/17076

### DIFF
--- a/src/components/icon-box/CounterIconBox.tsx
+++ b/src/components/icon-box/CounterIconBox.tsx
@@ -4,6 +4,7 @@ import {ThemeText} from '@atb/components/text';
 import React from 'react';
 import {iconSizes} from '@atb-as/theme';
 import {getTransportationColor, TextNames} from '@atb/theme/colors';
+import {useFontScale} from '@atb/utils/use-font-scale';
 
 interface CounterStyling {
   padding: number;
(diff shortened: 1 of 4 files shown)


src/components/icon-box/CounterIconBox.tsx
```diff
@@ -79,10 +47,13 @@ export const CounterIconBox = ({
           'transport_other',
           'secondary',
         )}
-        type={styling().type}
+        type={textType}
         testID="tripLegMore"
         style={{
-          lineHeight: styling().lineHeight,
+          height: lineHeight * fontScale,
+          minWidth: lineHeight * fontScale,
```
gorandalum, line 54: Should `MAX_FONT_SCALE` be taken into account?

gorandalum (approved):

src/components/icon-box/CounterIconBox.tsx
gorandalum: Never mind, I see that it already is handled in the useFontScale hook.

jorelosorio (approved):

PR #4412, opened 2024-04-03
feat: enable hjemjobbhjem ticket purchase in app

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17191

Changes in this PR: 
- Now uses product endpoint with fallback of firestore configuration to fetch the product list, used Tanstack react-query to call the endpoint.
- Re-introduced `offerEndpoint` setting in firestore.
- Added `City` image for HjemJobbHjem tickets.
- Updated dependencies for assets.

### DIFF
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -172,7 +172,6 @@ export const useTariffZoneSummary = (
   );
   const zoneSelectionModeDisabledForProduct =
     fareProductTypeConfig?.configuration.zoneSelectionMode === 'none';
-
   if (zoneSelectionModeDisabledForProduct) return undefined;
 
   return tariffZonesSummary(fromTariffZone, toTariffZone, language, t);
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -173,10 +173,10 @@ export const useTariffZoneSummary = (
   const zoneSelectionModeDisabledForProduct =
     fareProductTypeConfig?.configuration.zoneSelectionMode === 'none';
 
-  const usesZoneOfferEndpointForProduct =
-    fareProductTypeConfig?.configuration.offerEndpoint === 'zones';
+  const usesAuthorityOfferEndpointForProduct =
+    fareProductTypeConfig?.configuration.offerEndpoint === 'authority';
 
-  if (zoneSelectionModeDisabledForProduct && !usesZoneOfferEndpointForProduct) return undefined;
+  if (zoneSelectionModeDisabledForProduct && usesAuthorityOfferEndpointForProduct) return undefined;
 
   return tariffZonesSummary(fromTariffZone, toTariffZone, language, t);
 };
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -173,10 +173,7 @@ export const useTariffZoneSummary = (
   const zoneSelectionModeDisabledForProduct =
     fareProductTypeConfig?.configuration.zoneSelectionMode === 'none';
 
-  const usesAuthorityOfferEndpointForProduct =
-    fareProductTypeConfig?.configuration.offerEndpoint === 'authority';
-
-  if (zoneSelectionModeDisabledForProduct && usesAuthorityOfferEndpointForProduct) return undefined;
+  if (zoneSelectionModeDisabledForProduct) return undefined;
 
   return tariffZonesSummary(fromTariffZone, toTariffZone, language, t);
 };
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -172,7 +172,11 @@ export const useTariffZoneSummary = (
   );
   const zoneSelectionModeDisabledForProduct =
     fareProductTypeConfig?.configuration.zoneSelectionMode === 'none';
-  if (zoneSelectionModeDisabledForProduct) return undefined;
+
+  const usesZoneOfferEndpointForProduct =
+    fareProductTypeConfig?.configuration.offerEndpoint === 'zones';
+
+  if (zoneSelectionModeDisabledForProduct && !usesZoneOfferEndpointForProduct) return undefined;
 
   return tariffZonesSummary(fromTariffZone, toTariffZone, language, t);
 };
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -155,7 +155,7 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
   const fromPlaceName = getPlaceName(fromPlace, language);
 
   const toPlaceName = getPlaceName(toPlace, language);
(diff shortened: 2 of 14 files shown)


src/ticketing/use-get-fare-products-query.tsx
```diff
@@ -0,0 +1,13 @@
+import {getFareProducts} from '@atb/ticketing';
+import {useQuery} from '@tanstack/react-query';
+
+const ONE_HOUR_MS = 1000 * 60 * 60;
+
+export const useGetFareProductsQuery = () => {
+  return useQuery({
+    queryKey: ['getProducts'],
+    queryFn: getFareProducts,
```
hanne-at-atb, line 9: Could maybe use initData from `const {preassignedFareProducts} = useFirestoreConfiguration();`
See: https://tanstack.com/query/v4/docs/framework/react/guides/initial-query-data
Then no need for `useEffect` or `useState` in `TicketTabNav_PurchaseTabScreen`

hanne-at-atb (commented):
One though below 😄

src/ticketing/use-get-fare-products-query.tsx
reidzeibel: Updated, thank you for the suggestion!

src/fare-contracts/utils.ts
```diff
@@ -172,7 +172,11 @@ export const useTariffZoneSummary = (
   );
   const zoneSelectionModeDisabledForProduct =
     fareProductTypeConfig?.configuration.zoneSelectionMode === 'none';
-  if (zoneSelectionModeDisabledForProduct) return undefined;
+
+  const usesZoneOfferEndpointForProduct =
+    fareProductTypeConfig?.configuration.offerEndpoint === 'zones';
```
gorandalum, line 177: I don't think this is correct. If there is zone-selection-mode for the product, then it should probably have a tariff zone summary.

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -119,12 +119,9 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
 
-  const offerEndpoint =
-    zoneSelectionMode === 'none'
-      ? 'authority'
-      : zoneSelectionMode === 'multiple-stop-harbor'
-      ? 'stop-places'
-      : 'zones';
+  const offerEndpoint = params.fareProductTypeConfig.configuration.offerEndpoint
+    ? params.fareProductTypeConfig.configuration.offerEndpoint
+    : 'zones';
```
gorandalum, line 124: Let's just make offerEndpoint required in Firestore, and skip this fallback.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-defaults.ts
```diff
@@ -25,13 +26,13 @@ export function useOfferDefaults(
...
-  const {tariffZones, userProfiles, preassignedFareProducts} =
-    useFirestoreConfiguration();
+  const {data: fareProducts} = useGetFareProductsQuery();
+  const {tariffZones, userProfiles} = useFirestoreConfiguration();
   const {customerProfile} = useTicketingState();
 
   // Get default PreassignedFareProduct
   const productType = preassignedFareProduct?.type ?? selectableProductType;
-  const selectableProducts = preassignedFareProducts
+  const selectableProducts = fareProducts!
```
gorandalum, line 35: Should avoid `!`.

gorandalum (changes requested):

src/fare-contracts/utils.ts
reidzeibel: Updated to check if the offer endpoint is authority instead

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-defaults.ts
reidzeibel: `!` no longer needed, updated

src/fare-contracts/utils.ts
gorandalum: The change we do now is to decouple what endpoint to use from the zone-selection-mode. Thus the tariff-zone summary shouldn't be affected by what endpoint is configured. 

What is the origin of this change? The hjem-jobb-hjem will have zone-selection-mode none, and will return `undefined` as tariff-zone summary even without this change?

reidzeibel: It is to display the zone in the fare contract item : https://www.figma.com/file/zdZwvobgpEWSagKt0tderx/App?type=design&node-id=24257-23134&mode=design&t=4VLBkHRJxSL2KJRF-4

Since the zone selection is `none`, the `zone` part is not shown in the farecontract, that's why I added another check, because this hjemjobbhjem is a bit weird, the zone selection is none, but it is only valid for zone A, therefore we need to show the zone on the ticket.

image

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
reidzeibel: Now updated in app

src/fare-contracts/utils.ts
gorandalum: Aight. We should find another way to handle it, so there isn't any coupling between zone selection mode/zone presentation, and what offer endpoint is used. Let me take a look.

gorandalum (changes requested):
- Need to tackle how to decide if we should present the zones on the purchased fare contract. Let's do it together with Christian.
- What zone to send in for HjemJobbHjem should be configured through Firestore. Now it is just sending in the first zone in the list from Firestore. Although it works out for HjemJobbHjem since zone A is first in the list now it isn't a declarative enough solution.
- Should products always be fetched through the products-api, or should we still use Firestore? I see when presenting fare contracts then we fetch products directly from Firestore. Should have a talk with Martin about his thoughts regarding this.

mjenssen:
> * Should products always be fetched through the products-api, or should we still use Firestore? I see when presenting fare contracts then we fetch products directly from Firestore.

Ideally we should fetch the same data from a single source, so I'm leaning towards not fetching products from Firestore for presenting fare contracts (or anything else). However, if this is the blocker for getting HJH out, it should be fine to skip it for this version of the app since the product endpoint is a very thin layer over the Firestore data.

src/stacks-hierarchy/utils.ts
```diff
@@ -69,3 +69,17 @@ export const useDefaultTariffZone = (
     return {...tariffZones[0], resultType: 'zone'};
   }, [tariffZones, tariffZoneFromLocation]);
 };
+
+export const useFilterTariffZone = (
+  tariffZones: TariffZone[],
+  allowedTariffZones: string[],
```
gorandalum, line 80: I would prefer if `allowedTariffZones` had suffix "refs" or "ids". Based on the naming `tariffZones` and `allowedTariffZones` both seems to be an array of `TariffZone`.

gorandalum (approved):

src/stacks-hierarchy/utils.ts
reidzeibel: done :D

PR #4414, opened 2024-04-04
fix: notification translations for OMS partners

mortennordseth:
I've added a couple of org specific translations to have the correct org name in the notification settings. 

Fixes https://github.com/AtB-AS/kundevendt/issues/17505

### DIFF
--- a/src/translations/screens/Profile.ts
+++ b/src/translations/screens/Profile.ts
@@ -341,6 +341,28 @@ export default orgSpecificTranslations(ProfileTexts, {
               'Bruk billett på reisekort / mobil',
             ),
           },
+          notifications: {
+            pushToggle: {
+              subText: _(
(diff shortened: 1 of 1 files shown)


rosvik (approved):

gorandalum (commented):
Could we use `APP_NAME` from env for these situations instead? I'm sure this will be forgotten for Troms, Innlandet, etc.

mortennordseth:
> Could we use `APP_NAME` from env for these situations instead? I'm sure this will be forgotten for Troms, Innlandet, etc.

Not a bad idea. I think @rosvik had a good suggestion about using a dictionary for similar cases in the webshop (https://github.com/AtB-AS/webshop2/pull/422#pullrequestreview-1919503602). Might be worth looking into the same for the app. This would not couple the app name to the translations, which I think makes sense since the app name might differ from their "contact name"

PR #4420, opened 2024-04-08
fix: remove activation button on sent carnet tickets

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17263

add check for sent tickets when showing the activation button.

### DIFF
--- a/src/ticketing/utils.ts
+++ b/src/ticketing/utils.ts
@@ -172,7 +172,11 @@ function isActiveNowOrCanBeUsedFareContract(f: FareContract, now: number) {
   return false;
 }
 
-export function isCanBeConsumedNowFareContract(f: FareContract, now: number, currentUserId: string | undefined) {
+export function isCanBeConsumedNowFareContract(
+  f: FareContract,
+  now: number,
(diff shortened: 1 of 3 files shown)


rosvik (commented):
Could we make isCanBeConsumedNowFareContract return false when the fare contact is sent instead?

src/ticketing/utils.ts
```diff
@@ -172,7 +172,8 @@ function isActiveNowOrCanBeUsedFareContract(f: FareContract, now: number) {
   return false;
 }
 
-export function isCanBeConsumedNowFareContract(f: FareContract, now: number) {
+export function isCanBeConsumedNowFareContract(f: FareContract, now: number, currentUserId: string | undefined) {
```
rosvik, line 175: Is your linter set up correctly? 🤔 This is expanded to 

```
export function isCanBeConsumedNowFareContract(
  f: FareContract,
  now: number,
  currentUserId: string | undefined,
) {
```

for me.

rosvik (approved):

PR #4427, opened 2024-04-11
fix: avoid ambiguous messages about whether correspondence can or cannot be guaranteed

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17624

A user has asked us a question regarding ambiguous messages. The messages at the top of the app say that correspondence cannot be guaranteed, while the message a little further down says that there is correspondence between 340 and 341 (picture 1). This is confusing. 

### Proposed solution
Remove the last sentence "Kan ikke garantere korrespondanse.", if the message box "Korrespondanse mellom 340 og 341 på Volda sjukehus" appears.   

### Illustration Solution
<details>
<Summary>screenshots/video</Summary>

Image

Image

</details>


gorandalum (changes requested):
This is ment to be handled by `shortWaitTimeAndNotGuaranteedCorrespondence`. Could you look into why that isn't working instead? I think it has something to do with the intermediate walking leg between the corresponding departures.

jonasbrunvoll:
> This is ment to be handled by `shortWaitTimeAndNotGuaranteedCorrespondence`. Could you look into why that isn't working instead? I think it has something to do with the intermediate walking leg between the corresponding departures.

Yes,  as I have understood the problem, `timeIsShort` inside the function `shortWaitTimeAndNotGuaranteedCorrespondence` returns true beacuse the waiting time between bus changes is too short. This makes the sentance "Kan ikke garantere korrespondanse." being included, although the correspondence is guaranteed.

src/travel-details-screens/utils.ts
```diff
@@ -160,6 +161,7 @@ export function hasShortWaitTime(legs: Leg[]) {
 }
 
 export function hasShortWaitTimeAndNotGuaranteedCorrespondence(legs: Leg[]) {
+  if (hasGuaranteedCorrespondence(legs)) return false;
```
gorandalum, line 164: If there is a non-corresponding short wait time between two legs of the trip, then this method should return `true`. I think this now may be masked by some other corresponding legs, which overrides the answer to `false`.

gorandalum (changes requested):

PR #4430, opened 2024-04-12
feat: only allow consumption on non-activated ticket for carnets.

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17595

Don't show carnet consumption button on refunded/cancelled tickets, so basically only show it on a non-activated carnet ticket.

### DIFF
--- a/src/ticketing/utils.ts
+++ b/src/ticketing/utils.ts
@@ -178,6 +178,7 @@ export function isCanBeConsumedNowFareContract(
   currentUserId: string | undefined,
 ) {
   if (f.customerAccountId !== currentUserId) return false;
+  if (f.state !== 1) return false;
   if (!isCarnet(f)) return false;
   const travelRights = f.travelRights.filter(isCarnetTravelRight);
 
--- a/src/ticketing/utils.ts
+++ b/src/ticketing/utils.ts
@@ -178,7 +178,7 @@ export function isCanBeConsumedNowFareContract(
   currentUserId: string | undefined,
 ) {
   if (f.customerAccountId !== currentUserId) return false;
-  if (f.state !== 1) return false;
+  if (f.state !== FareContractState.NotActivated) return false;
   if (!isCarnet(f)) return false;
   const travelRights = f.travelRights.filter(isCarnetTravelRight);
 


src/ticketing/utils.ts
```diff
@@ -178,6 +178,7 @@ export function isCanBeConsumedNowFareContract(
   currentUserId: string | undefined,
 ) {
   if (f.customerAccountId !== currentUserId) return false;
+  if (f.state !== 1) return false;
```
gorandalum, line 181: Could you use the `FareContractState` enum instead? That makes this more readable.

reidzeibel: I totally forgot I could do that... updated 🤦

marius-at-atb (approved):

gorandalum (commented):
What about state `ACTIVATED`, when does that occur?

reidzeibel:
@gorandalum when the ticket is activated, it should change to state `ACTIVATED` and then the button will be hidden by the 
`    !hasActiveCarnetTravelRight(travelRights, now)`

```
export function isCanBeConsumedNowFareContract(
  f: FareContract,
  now: number,
  currentUserId: string | undefined,
) {
  if (f.customerAccountId !== currentUserId) return false;
  if (f.state !== FareContractState.NotActivated) return false;
  if (!isCarnet(f)) return false;
  const travelRights = f.travelRights.filter(isCarnetTravelRight);

  // @TODO: This is a temporary limitation because of issues with consumption of
  // carnets with more than one travel right. Should be removed once there is a
  // solution for this in the backend.
  if (travelRights.length > 1) return false;

  return (
    hasUsableCarnetTravelRight(travelRights, now) &&
    !hasActiveCarnetTravelRight(travelRights, now)
  );
}
```

gorandalum:
Is that an assumption or tested? I don't understand why we have the more complex `hasActiveCarnetTravelRight` evaluating `usedAccesses` if we could just check the fare contract state instead.

reidzeibel:
Hmm, for that part I have to bring @rosvik into the discussion. 😄

reidzeibel:
Noooo, then I think the logic in this PR is wrong, because the state can be `ACTIVATED` even when you don't have an active carnet.

reidzeibel:
@gorandalum @rosvik is this a better logic for checking the expiration status? we want to disable it on tickets that are not on the active ticket page. 

```
  if (
    f.state === FareContractState.Cancelled ||
    f.state === FareContractState.Refunded ||
    f.state === FareContractState.Unspecified
  )
    return false;
```

gorandalum:
> @gorandalum @rosvik is this a better logic for checking the expiration status? we want to disable it on tickets that are not on the active ticket page.
> 
> ```
>   if (
>     f.state === FareContractState.Cancelled ||
>     f.state === FareContractState.Refunded ||
>     f.state === FareContractState.Unspecified
>   )
>     return false;
> ```

What if a new state is introduced from Entur backend, how would we like to handle that?

reidzeibel:
@gorandalum I looked around on the utils, and I think this existing function should be the one that we use right now `isActiveNowOrCanBeUsedFareContract(f: FareContract, now: number)`, since it is used for filtering of expired (or refunded) ticket that is shown in "expired ticket" page.

```
export const filterExpiredFareContracts = (
  fareContracts: FareContract[],
  now: number,
) => {
  const isRefunded = (f: FareContract) =>
    f.state === FareContractState.Refunded;
  const isExpiredOrRefunded = (f: FareContract) =>
    !isActiveNowOrCanBeUsedFareContract(f, now) || isRefunded(f);
  return fareContracts.filter(isExpiredOrRefunded);
};
```

If there is a new state introduced from Entur backend, then I think we would have to rework the checking process. 🤔

PR #4431, opened 2024-04-12
fix: use correct check for disabling carnet consumption

reidzeibel:
closes : https://github.com/AtB-AS/kundevendt/issues/17595

old PR and discussion : https://github.com/AtB-AS/mittatb-app/pull/4430

Use existing function for checking ticket status.

### DIFF
--- a/src/ticketing/utils.ts
+++ b/src/ticketing/utils.ts
@@ -178,7 +178,7 @@ export function isCanBeConsumedNowFareContract(
   currentUserId: string | undefined,
 ) {
   if (f.customerAccountId !== currentUserId) return false;
-  if (!isActiveNowOrCanBeUsedFareContract(f, now)) return false;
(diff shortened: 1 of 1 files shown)


gorandalum (commented):
I don't think `isActiveNowOrCanBeUsedFareContract` should be used for this, as it checks quite a bit more. But `isOrWillBeActivatedFareContract` seems appropriate.

> If there is a new state introduced from Entur backend, then I think we would have to rework the checking process. 

Regarding this, if there is a new state introduced from Entur backend, we won't be able to patch old app versions. So with that in mind it is better to whitelist states (like saying it is allowed for activated and not activated) instead of blacklisting states.

reidzeibel:
@gorandalum I'll keep that in mind (whitelist > blacklist) 👍

gorandalum (approved):

PR #4432, opened 2024-04-15
chore: update SystemFragment type

marius-at-atb:
Update the SystemFragment type after merging https://github.com/AtB-AS/atb-bff/pull/330 in the BFF.

Also included a change to the type readme as it was a bit confusing that these files "are generated" and should "not be edited by hand", while you actually have to copy paste it in..

### DIFF
--- a/src/api/types/generated/readme
+++ b/src/api/types/generated/readme
@@ -1,3 +1,3 @@
 These files are generated or manually imported.
-These files should not be edited by hand.
-Overrides, extensions or extractions should be done in parent directory
+These files should not be edited by typing. Instead, copy the generated type and paste it in.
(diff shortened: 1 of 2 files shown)


src/api/types/generated/readme
```diff
@@ -1,3 +1,3 @@
 These files are generated or manually imported.
-These files should not be edited by hand.
-Overrides, extensions or extractions should be done in parent directory
+These files should not be edited by typing. Instead, copy the generated type and paste it in.
```
rosvik, line 2:
```suggestion
These files should not be edited by typing. Instead, copy the generated type from atb-bff and paste it in.
```

gorandalum (approved):

rosvik (approved):

PR #4433, opened 2024-04-15
fix: Remove screen with restriction about do not allow user to buy ticket on fallback state

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/17026

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
@@ -9,9 +9,7 @@ import {RecentFareContracts} from './Components/RecentFareContracts/RecentFareCo
 import {TicketTabNavScreenProps} from '../navigation-types';
 import {UpgradeSplash} from './Components/UpgradeSplash';
 import {useRecentFareContracts} from './use-recent-fare-contracts';
-import {
-  FareProductTypeConfig,
-} from '@atb/configuration';
+import {FareProductTypeConfig} from '@atb/configuration';
 import {RecentFareContract} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/types';
 import {useTicketingAssistantEnabled} from '@atb/stacks-hierarchy/Root_TicketAssistantStack/use-ticketing-assistant-enabled';
 import {TicketAssistantTile} from '@atb/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Assistant/TicketAssistantTile';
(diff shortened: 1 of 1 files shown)


marius-at-atb:
I think this was the only use case for this screen, so unless there is a plan to use it for something else, we should delete the screen as well.

rosvik:
I think we should still show the screen if fallback isn't active, but the user has not selected phone. E.g. for t:card users.

jorelosorio:
> I think this was the only use case for this screen, so unless there is a plan to use it for something else, we should delete the screen as well.

The screen is still used in other cases, but we can remove the message if is not used in other places.

jorelosorio:
> I think we should still show the screen if fallback isn't active, but the user has not selected phone. E.g. for t:card users.

So the screen is presented, but with the option to skip it? according to the description and what I talked to Sebastian it should work as bus ticket is working, no intermediate screens on it (if something is wrong with mobile token)

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
```diff
@@ -72,25 +72,6 @@ export const TicketTabNav_PurchaseTabScreen = ({navigation}: Props) => {
         });
         return;
       }
-    } else {
-      if (
-        fareProductTypeConfig.configuration.requiresTokenOnMobile &&
-        !hasInspectableMobileToken
```
rosvik, line 78: Pseudocode:

```
fareProductTypeConfig.configuration.requiresTokenOnMobile &&
!token.fallback && 
token.type !== "mobile"
```

jorelosorio:
> I think this was the only use case for this screen, so unless there is a plan to use it for something else, we should delete the screen as well.

We keep the same view now but adding an extra case to show it, so now it is only shown when not fallback.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
```diff
@@ -72,25 +71,24 @@ export const TicketTabNav_PurchaseTabScreen = ({navigation}: Props) => {
...
-            nextScreen: {
-              screen: 'Root_PurchaseOverviewScreen',
-              params: {
-                fareProductTypeConfig,
-                mode: 'Ticket',
-              },
+    } else if (
+      fareProductTypeConfig.configuration.requiresTokenOnMobile &&
+      !hasInspectableMobileToken &&
+      !isFallbackState
```
marius-at-atb, line 77: Should it also check other MobileTokenStatus states here?

marius-at-atb (approved):

rosvik (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
gorandalum: Not completely sure about this. I saw it is now added an additional check regarding if mobile token status is `error`, but in that case it means that fallback isn't enabled. 

@jorelosorio At least add this situation to the acceptance criteria so it is up in the light, and notify Sebastian about the acceptance criteria change, to see if he agrees.

PR #4434, opened 2024-04-15
feat: remove on-behalf-of for night ticket anonymous purchase

reidzeibel:
adds missing check on https://github.com/AtB-AS/kundevendt/issues/17206#issuecomment-2045145986

While we're at it, should we add `isLoggedInWithPhone` inside `AuthContext.State` as well? Then we can use `{isLoggedInWithPhone}` directly from `useAuthState()` directly, or is it too specific to add that?

The result should look like this: 

```
  ...
  ...
  const authenticationType = mapAuthenticationType(state.user); // added
  const isLoggedInWithPhone = authenticationType === 'phone';  // added

  return (
    <AuthContext.Provider
      value={{
        ...
        signOut: useCallback(async () => {
          await auth().signInAnonymously();
        }, []),
        authenticationType: authenticationType, // modified from mapAuthenticationType(state.user);
        isLoggedInWithPhone: isLoggedInWithPhone, // added
        ...
      }}
    >
      {children}
    </AuthContext.Provider>
  );
```

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
@@ -32,6 +32,7 @@ import {HoldingHands} from '@atb/assets/svg/color/images';
 import {ContentHeading} from '@atb/components/heading';
 import {isUserProfileSelectable} from './utils';
 import {useOnBehalfOfEnabled} from '@atb/on-behalf-of';
+import { useAuthState } from '@atb/auth';
 
 type Props = RootStackScreenProps<'Root_PurchaseOverviewScreen'>;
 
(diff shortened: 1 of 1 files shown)


rosvik (approved):
Looks good!

> While we're at it, should we add isLoggedInWithPhone inside AuthContext.State as well? 

I think it's fine as is. When (if?) we add email logins, we need to update the checks, but until then I think it's kind of nice that we explicitly use `authenticationType === 'phone'`.

marius-at-atb (approved):
Lgtm. If `isLoggedIn` will soon require `authenticationType === 'phone' || authenticationType === 'email'`, I would prefer `isLoggedIn` in the context provider, rather than the more specific `isLoggedInWithPhone`

reidzeibel:
We'll decide how to do the `isLoggedIn` part in the next episode! Stay tuned!

PR #4437, opened 2024-04-15
fix: avoid ambiguous messages about whether correspondence can or cannot be guaranteed

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17624

A user has asked us a question regarding ambiguous messages. The messages at the top of the app say that correspondence cannot be guaranteed, while the message a little further down says that there is correspondence between 340 and 341 (picture 1). This is confusing. 

### Proposed solution
Remove the last sentence "Kan ikke garantere korrespondanse.", if the message box "Korrespondanse mellom 340 og 341 på Volda sjukehus" appears.   

### Illustration Solution
<details>
<Summary>screenshots/video</Summary>

Image

Image

</details>

### DIFF
--- a/src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
+++ b/src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
@@ -1,0 +1,87 @@
+//import {addMinutes} from 'date-fns';
+import {hasShortWaitTimeAndNotGuaranteedCorrespondence} from '../utils';
+import {Leg} from '@atb/api/types/trips';
+
+describe('hasShortWaitTimeAndNotGuaranteedCorrespondence', () => {
+  it('should be false if not short wait time', () => {
+    const legs = [
+      {expectedEndTime: '2024-10-31T14:55:00Z'},
+      {expectedStartTime: '2024-10-31T15:00:00Z'},
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(false);
+  });
+
+  it('should be true if short wait time and no guarenteed correspondance', () => {
+    const legs = [
+      {expectedEndTime: '2024-10-31T14:59:00Z'},
+      {expectedStartTime: '2024-10-31T15:00:00Z'},
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(true);
+  });
+
+  it('should be false if short wait time, but correspondance is guarenteed', () => {
+    const legs = [
+      {
+        expectedEndTime: '2024-10-31T14:59:00Z',
+        interchangeTo: {guaranteed: true},
+      },
+      {expectedStartTime: '2024-10-31T15:00:00Z'},
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(false);
+  });
+
+  it('should be false if not short wait time and correspondance is guarenteed', () => {
+    const legs = [
+      {
+        expectedEndTime: '2024-10-31T14:55:00Z',
+        interchangeTo: {guaranteed: true},
+      },
+      {expectedStartTime: '2024-10-31T15:00:00Z'},
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(false);
+  });
+
+  it('should be true if wait time between two legs are short and correspondance is not guranteed.', () => {
+    const legs = [
+      {
+        expectedStartTime: '2024-10-31T14:59:00Z',
+        expectedEndTime: '2024-10-31T15:00:00Z',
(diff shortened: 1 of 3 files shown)


gorandalum (changes requested):
I think this will return `false` whenever two legs contains a correspondence between them, even though there might be a non-corresponding short-wait-time between some other legs?

jonasbrunvoll:
> If there is a non-corresponding short wait time between two legs of the trip, then this method should return true. I think this now may be masked by some other corresponding legs, which overrides the answer to false.

`hasGuaranteedCorrespondence(filteredLegs)`, uses the same conditions as the component 'TripSection.tsx, line 352', to determine whether it should return true or false. As I understand it, correspondence is guaranteed if this box is displayed, even when shorWaitTime kicks in. Or maybe this only applies to FRAM?

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        expectedEndTime: '2024-10-31T15:05:00Z',
+      },
+      {
+        expectedStartTime: '2024-10-31T15:6:00Z',
+        expectedEndTime: '2024-10-31T15:20:00Z',
+      },
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(s7);
+  });
```
gorandalum, line 108: The test is intended to test that even if there is no wait time between bus and walking (first and second leg in this example), the function returns false. To actually test that the last leg should start something like 15:10.

As the test input legs are written now with the end time of the walking leg at 15:05 and the start time of the last bus leg is 15:06, the function should actually return true. So there is a error in the implementation somewhere.

gorandalum (changes requested):

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
```diff
@@ -0,0 +1,139 @@
+import {hasShortWaitTimeAndNotGuaranteedCorrespondence} from '../utils';
+import {Leg} from '@atb/api/types/trips';
+
+describe('hasShortWaitTimeAndNotGuaranteedCorrespondence', () => {
+  const s1 = false;
```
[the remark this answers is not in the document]
jonasbrunvoll: Okey, I´ll change the tests to make them more readable👍

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
```diff
@@ -115,17 +127,119 @@ describe('hasShortWaitTimeAndNotGuaranteedCorrespondence', () => {
...
+      {
+        expectedStartTime: '2024-10-31T15:21:00Z',
+        expectedEndTime: '2024-10-31T15:30:00Z',
+      },
+    ] as unknown as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(true);
+  });
+
+  it('should be true because of short wait time between third and forth leg.', () => {
```
gorandalum, line 219: Shouldn't have two tests with the same name

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
```diff
@@ -115,17 +127,119 @@ describe('hasShortWaitTimeAndNotGuaranteedCorrespondence', () => {
...
+        expectedEndTime: '2024-10-31T15:00:00Z',
+      },
+      {
+        expectedStartTime: '2024-10-31T15:00:00Z',
+        expectedEndTime: '2024-10-31T15:05:00Z',
+      },
+    ] as Leg[];
+
+    expect(hasShortWaitTimeAndNotGuaranteedCorrespondence(legs)).toBe(true);
+  });
```
gorandalum, line 191: I think this one is wrong, and the test should be 'should be false when short wait time after first foot leg' or something like it. 

When there is just walking as the first leg, the journey planner algorithm always make it be a the bus stop "just in time". This means that with this logic, we would get the short wait time message on every trip starting with a foot leg.

gorandalum (changes requested):

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
jonasbrunvoll: Okey, i´ll change the name👍

src/travel-details-screens/__tests__/has_short_wait_time_and_not_guaranteed_correspondance.test.ts
jonasbrunvoll: Okey, I´ll update the logic👍

gorandalum (approved):
Great! :+1:

PR #4439, opened 2024-04-16
feat: add error grouping hash to bugsnag reporting

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/5352

This PR will try to improve the error grouping in Bugsnag dashboard.

### Problem
Currently, if we open an error in the Bugsnag dashboard, it will show the screen below (example):

We can see that while the error name is the same: `MainActivity`, the list of error shows that it can originate from different source, in this case we have both `network error` and `token loading timeout` in the same error category.

Check these links for more examples:
- https://app.bugsnag.com/atb-as/mittatb/errors/6615013360afa8000813aa21
- https://app.bugsnag.com/atb-as/mittatb/errors/6269b17a4d18f800088b523a
- https://app.bugsnag.com/atb-as/mittatb/errors/657bef1e901ded000850a86f 

We can also see that the list of error types in main `Inbox` page have lots of similar issues which could be grouped together. There are multiple `auth/network-request-failed` that could have been grouped together, but instead there are several different items with the same issue.

### Why did the problem occur?
These grouping is the default behavior from Bugsnag, see their [official documentation about issues grouping](https://docs.bugsnag.com/product/error-grouping/#default-grouping). 

### Solution in PR
Bugsnag has provided a [custom grouping](https://docs.bugsnag.com/product/error-grouping/#custom-grouping) solution that can be used to group issues together. They provided 2 premade solution: using `Error class` and `Error context`, while this might work for a smaller app, we usually have similar `Error class` across issues (using the `Error` class). For `Error context` it might work but still unsure about how Bugsnag would group the error(s).

Therefore I opted to use the `Custom grouping hash` solution provided by Bugsnag. This will allow us to set the grouping of errors together, and therefore it should be easier to track them and pinpoint where the issue occurs.

In this PR I used the `notifyBugsnag()` function from `bugsnag-utils.ts`, which I think should be used for all bugsnag reporting going forward. I added the `errorGroupHash` parameter to specify the grouping. Currently, there are several grouping which can be accessed in the document in this PR https://github.com/AtB-AS/docs-private/pull/39.

### Review notes

Please let me know what do you think about the grouping, particularly errors that might need to be grouped together, but may have different error specification, such as the `GetSignedTokenError`, should we put the error type on the group (to differentiate them), or should we put the error type on the metadata instead (to group all of them together).

Also, if there are any suggestions for what other errors should be logged and grouped, feel free to notify and I will add them to the code (preferably also suggest the group hash).

### DIFF
--- a/src/mobile-token/hooks/useGetSignedTokenQuery.tsx
+++ b/src/mobile-token/hooks/useGetSignedTokenQuery.tsx
@@ -22,7 +22,7 @@ export const useGetSignedTokenQuery = () => {
       try {
         return await mobileTokenClient.encode(nativeToken);
       } catch (err: any) {
-        notifyBugsnag(err, `GetSignedTokenError - ${err}`, {
+        notifyBugsnag(err, 'GetSignedTokenError', {
           tokenId: nativeToken.tokenId,
           description: 'Error encoding signed token',
         });
--- a/src/mobile-token/hooks/useGetSignedTokenQuery.tsx
+++ b/src/mobile-token/hooks/useGetSignedTokenQuery.tsx
@@ -22,7 +22,7 @@ export const useGetSignedTokenQuery = () => {
       try {
         return await mobileTokenClient.encode(nativeToken);
       } catch (err: any) {
-        notifyBugsnag(err, {
+        notifyBugsnag(err, `GetSignedTokenError - ${err}`, {
           tokenId: nativeToken.tokenId,
           description: 'Error encoding signed token',
         });
--- a/src/loading-screen/__tests__/use-notify-bugsnag-on-timeout-status.test.ts
+++ b/src/loading-screen/__tests__/use-notify-bugsnag-on-timeout-status.test.ts
@@ -8,8 +8,8 @@ import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
 let mockBugsnagNotification: Parameters<typeof notifyBugsnag> | undefined;
 
 jest.mock('@atb/utils/bugsnag-utils', () => ({
-  notifyBugsnag: (error: string, metadata: any) => {
-    mockBugsnagNotification = [error, metadata];
+  notifyBugsnag: (error: string, errorGroupHash: string, metadata: any) => {
+    mockBugsnagNotification = [error, errorGroupHash, metadata];
   },
 }));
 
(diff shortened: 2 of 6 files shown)


src/mobile-token/hooks/useGetSignedTokenQuery.tsx
```diff
@@ -22,7 +22,7 @@ export const useGetSignedTokenQuery = () => {
       try {
         return await mobileTokenClient.encode(nativeToken);
       } catch (err: any) {
-        notifyBugsnag(err, {
+        notifyBugsnag(err, `GetSignedTokenError - ${err}`, {
```
jorelosorio, line 25: I am just asking a bit here, since this parameter is a hash should it not include ${err} since it can make the hash to change and not group against GetSignedTokenError?

reidzeibel: That is what I asked on the `Review Notes` above, 

> Please let me know what do you think about the grouping, particularly errors that might need to be grouped together, but may have different error specification, such as the GetSignedTokenError, should we put the error type on the group (to differentiate them), or should we put the error type on the metadata instead (to group all of them together).

It is open for discussion and I think `${err}` should be in metadata instead, but I'm up for suggestions 😄

jorelosorio: I wrote a comment in the task :)

gorandalum:
Just a note: To do a quick test I updated the default grouping in BugSnag now too see if it affects something.

We have probably not updated it before.

> Updates are released to make improvements to the grouping algorithm, but they are not automatically applied because they can lead to changes in the way that existing errors are grouped. It is always recommended that you update to the latest grouping algorithm when it is available.

reidzeibel:
@gorandalum I was thinking about doing that on monday but unsure about how it will affect the grouping, thanks for testing it 😆

gorandalum:
Lets test it for a couple days before merging this, just to see if there is any effect.

reidzeibel:
Ready for another round of review, @tormoseng already agreed to try this.

rosvik (approved):

rosvik:
@reidzeibel This broke in several places after we merged the release branch. Could you take a look? See `use-fetch-id-token-with-custom-claims.ts` and `TicketingContext.tsx`

reidzeibel:
@rosvik thanks for pointing it out, will look into it!

reidzeibel:
https://github.com/AtB-AS/mittatb-app/pull/4439

PR #4448, opened 2024-04-18
feat: add action to global messages when there are links and screenreader is enabled

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/2459

This PR will allow the screenreader to "read" global message links, and allows the global message to be activated to follow the URL to the specified link. For example, [this message in firestore](https://console.firebase.google.com/u/1/project/atb-mobility-platform-staging/firestore/databases/-default-/data/~2FglobalMessagesV2~2FMessage%20with%20scheduled%20timer).

~~When the screenreader is on, the action text that shows at the bottom (like the one in anonymous ticket purchase message), is hidden. Instead, the message text is replaced with the parsed text from the link markdown. This ensures that the message is read correctly by the screen reader and there are no redundant information being read aloud. The `onPressAction` will only be activated when the screenreader is active.~~

Following the suggestion from @gorandalum , I added `link` field to Global Message firestore, and use it as the action for the `MessageInfoBox onPressConfig`. I also modified the a11y hint a bit to differentiate between action and link, since it is weird to listen "Activate to https://atb.no/avvik". Also updated the documentation here https://github.com/AtB-AS/docs-private/pull/40.

There's also a fix for issue in metro where it complaints about missing `key`.

Open to suggestions if there are better implementation. 😊

### DIFF
--- a/src/global-messages/GlobalMessage.tsx
+++ b/src/global-messages/GlobalMessage.tsx
@@ -60,8 +60,10 @@ const GlobalMessage = ({
         .filter((gm) => isWithinTimeRange(gm, now))
         .map((globalMessage: GlobalMessageType) => {
           const message = getTextForLanguage(globalMessage.body, language);
-          const onPressAction = globalMessage.link
-            ? {text: globalMessage.link, url: globalMessage.link}
+          const link = getTextForLanguage(globalMessage.link, language);
+          const linkText = getTextForLanguage(globalMessage.linkText, language);
+          const onPressAction = link
+            ? {text: linkText ? linkText : link, url: link}
             : undefined;
           if (!message) return null;
 
--- a/src/global-messages/GlobalMessage.tsx
+++ b/src/global-messages/GlobalMessage.tsx
@@ -63,11 +63,8 @@ const GlobalMessage = ({
         .filter((gm) => isWithinTimeRange(gm, now))
         .map((globalMessage: GlobalMessageType) => {
           const message = getTextForLanguage(globalMessage.body, language);
+          const link = globalMessage.link;
           if (!message) return null;
-          const displayMessageAndAction = getMessageAndAction(
-            message,
-            isScreenReaderEnabled,
-          );
 
           return (
             <>
@@ -89,7 +86,7 @@ const GlobalMessage = ({
                     globalMessage.title ?? [],
                     language,
                   )}
-                  message={displayMessageAndAction.message}
+                  message={message}
                   type={globalMessage.type}
                   isMarkdown={true}
                   onDismiss={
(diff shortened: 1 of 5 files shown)


gorandalum (changes requested):
I think it's a little too hackish. Generally solutions like parsing free-text is last effort solution if we can't find better ones.

For this I would rather like adding a new field for a single link and accompanying text on global messages in Firestore, which populates the `onPressConfig` for the message box. At the same time we should document that using links in markdown body is not advised.

reidzeibel:
@gorandalum makes sense and looks cleaner as well, just need to be documented and posted to kundevent slack, will let you know after I made some changes.

src/global-messages/GlobalMessage.tsx
```diff
@@ -60,39 +60,37 @@ const GlobalMessage = ({
...
-                    globalMessage.isDismissable && !includeDismissed
-                      ? () => dismissGlobalMessage(globalMessage)
-                      : undefined
-                  }
-                  testID="globalMessage"
-                />
-              )}
-            </>
+
+          return globalMessage.subtle ? (
```
reidzeibel, line 68: I removed the <> </> tag and the console no longer complaints about missing `key`

src/global-messages/GlobalMessage.tsx
```diff
@@ -60,39 +60,37 @@ const GlobalMessage = ({
         .filter((gm) => isWithinTimeRange(gm, now))
         .map((globalMessage: GlobalMessageType) => {
           const message = getTextForLanguage(globalMessage.body, language);
+          const onPressAction = globalMessage.link
+            ? {text: globalMessage.link, url: globalMessage.link}
+            : undefined;
```
gorandalum, line 65: I think we should allow to set both link url and text for it through Firestore. The link might be long and non-presentable.

src/global-messages/types.ts
```diff
@@ -23,6 +23,7 @@ export type GlobalMessageRaw = {
   active: boolean;
   title?: LanguageAndTextType[];
   body: LanguageAndTextType[];
+  link?: string;
```
gorandalum, line 26: May have different link urls for different languages, so Firestore should support that.

gorandalum (changes requested):
Looks nice! Just some minor adjustments necessary.

src/global-messages/GlobalMessage.tsx
```diff
@@ -60,39 +60,37 @@ const GlobalMessage = ({
...
+              title={getTextForLanguage(globalMessage.title ?? [], language)}
+              message={message}
+              type={globalMessage.type}
+              isMarkdown={true}
+              onDismiss={
+                globalMessage.isDismissable && !includeDismissed
+                  ? () => dismissGlobalMessage(globalMessage)
+                  : undefined
+              }
+              onPressConfig={onPressAction}
```
gorandalum, line 91: Remember to add to documentation that links only works for non-subtle global messages.

reidzeibel: Documentation updated

gorandalum (approved):

PR #4450, opened 2024-04-19
feat: disable popover

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17628

comment out the code that shows the popover, the rest of the code will be removed if popover is no longer used, which will be decided in the future.

Example issue with popover: https://github.com/AtB-AS/kundevendt/issues/17475


marius-at-atb (approved):
Lgtm. Not sure exactly which errors we have seen for this, but I guess it could be a good idea to also comment out the `PopOver` inside `PopOverContext.Provider`

reidzeibel:
@marius-at-atb example issue : https://github.com/AtB-AS/kundevendt/issues/17475

marius-at-atb:
If only that kind of issue, then all good as is :+1:

gorandalum (approved):

jorelosorio (approved):

PR #4451, opened 2024-04-19
refactor: Upgrade mobile token 2.1.4

gorandalum:
Closes https://github.com/AtB-AS/kundevendt/issues/16909

- Add handling of error resolutions in errors from sdk. As of now
only handling the reset resolution, which signals that the sdk is
stuck. The rest of the errors doesn't have any specific handling
for now and are mainly treated as before.
- The new SDK version has the possibility for both local logger
and remote logger. For now, the previously used logger is passed in
as local logger, and for remote logger we just send the errors
as notify events to Bugsnag. We will see how this turns out in
Bugsnag and if it needs to be revised.
- Added some more libs in the Podfile, which is required
dependencies for the SDK, even if we don't use them yet.

Depends on: https://github.com/AtB-AS/abt-token-server/pull/68


hanne-at-atb (approved):
Things are looking good to me, but probably nice to have another pair of eyes on this since I didn't work too much with the app lately 🌟

src/mobile-token/MobileTokenContext.tsx
```diff
@@ -121,11 +122,12 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
     } else {
       // We've finished with remote tokens. Cancel timeout notification.
       cancelTimeoutHandler?.();
+      setIsTimout(false);
     }
   }, [enabled, nativeTokenStatus, token_timeout_in_seconds]);
 
   useInterval(
-    () => checkRenewMutate(nativeToken),
+    () => checkRenewMutate({token: nativeToken, traceId: 'uuid'}),
```
jorelosorio, line 130: should 'uuid' be a variable instead?

src/mobile-token/hooks/use-load-native-token-query.tsx
```diff
@@ -36,6 +41,7 @@ export const useLoadNativeTokenQuery = (
       }
     },
     enabled,
+    retry: 0,
```
jorelosorio, line 44: No retries?

jorelosorio:
@gorandalum I am not sure if there are breaking changes by doing an update to:

```
New version of the mobile client SDK: 2.1.1
* Make device info optional for barcodes in the beginInspection function
  * Opting out of device info reduces the data size in the barcodes by around 40%
  * The recommended config for device info is to set the application and os fields to false
* Added iOS privacy manifest
```

Seems like there are some improvements there.

src/mobile-token/hooks/use-preemptive-renew-token-mutation.tsx
```diff
@@ -43,5 +52,22 @@ export const usePreemptiveRenewTokenMutation = (userId?: string) => {
         );
       }
     },
+    onError: async (err: any, {token, traceId}) => {
+      notifyBugsnag(err, {
+        errorGroupHash: 'AuthError',
```
jorelosorio, line 57: @reidzeibel Could you please confirm this is the right group hash for mobile-token? :)

reidzeibel: We have these:

`TokenLoadingTimeoutError`
`GetSignedTokenError`

on this PR : https://github.com/AtB-AS/mittatb-app/pull/4439/files#diff-fd9df4f15f1da5e1656210163837e5dd7992c4790f04cd721596106c8813a7f4

We have no grouping yet for this one, but I think it shouldn't be `AuthError`

reidzeibel (changes requested):

src/mobile-token/hooks/use-preemptive-renew-token-mutation.tsx
jorelosorio: Fixed

reidzeibel (approved):
Looks good, in the future we might want to revisit this error group hash for mobile token.

PR #4454, opened 2024-04-23
build: change register-app-version script to use new host + audience

mortennordseth:
This PR updates the register-app-version.sh script to accommodate recent changes made by Entur. Specifically, the host and audience for the ABT publish clients have changed, and the ones in use now are deprecated. Since Troms only has the new client, we need to make this change now to register the app with Entur for Troms.

I've also simplified how we handle secrets on GitHub. Instead of having two secrets for each client in each environment, we now just use one JSON object for both staging and production. This is in line with what we do in the GCP secret manager. 

If these changes are approved we should update the Github secrets before merging the PR. I'll also create a PR to update the [docs](https://github.com/AtB-AS/docs-private/blob/main/secrets.md).

Closes https://github.com/AtB-AS/kundevendt/issues/17778

### DIFF
--- a/scripts/android/register-app-version.sh
+++ b/scripts/android/register-app-version.sh
@@ -3,12 +3,11 @@
 # This script will register a specific Android application version with Entur's
 # MobileApplicationRegistryService API.
 # a base64 encoded SHA256 fingerprint as bytes of the signing certificate for the APK must be provided as an environment variable,
-# in addition to client ID and client secret for Entur ABT OAuth login.
+# in addition to a Entur ABT OAuth client configuration JSON object.
 
 # Check for secrets from env vars
 if [[
-  -z ${ENTUR_CLIENT_ID}
-  || -z ${ENTUR_CLIENT_SECRET}
+  -z ${ENTUR_PUBLISH_CLIENT}
   || -z ${APP_ENVIRONMENT}
   || -z ${AUTHORITY}
   || -z ${ANDROID_APPLICATION_ID}
@@ -19,8 +18,7 @@ if [[
 ]]; then
   echo "Argument error!"
   echo "Expected environment variables:
-  - ENTUR_CLIENT_ID
-  - ENTUR_CLIENT_SECRET
+  - ENTUR_PUBLISH_CLIENT
   - APP_ENVIRONMENT
   - AUTHORITY
   - ANDROID_APPLICATION_ID
@@ -35,7 +33,7 @@ fi
 # Get values based on environment
 case ${APP_ENVIRONMENT} in
   staging)
-    token_url="https://partner-abt.staging.entur.org/oauth/token"
+    token_url="https://partner.staging.entur.org/oauth/token"
     abt_url="https://core-abt-abt.staging.entur.io"
     ;;
   store)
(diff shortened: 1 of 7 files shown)


gorandalum (commented):
@mjenssen Can have a say on this, but on my part I think it is a little cleaner to have client-id, client-secret and audience in GitHub secrets, than to have a token which needs to be parsed.

Whatever this ends up in, it should be documented in [docs-private](https://github.com/AtB-AS/docs-private/blob/main/secrets.md).

scripts/android/register-app-version.sh
```diff
@@ -35,7 +33,7 @@ fi
 # Get values based on environment
 case ${APP_ENVIRONMENT} in
   staging)
-    token_url="https://partner-abt.staging.entur.org/oauth/token"
+    token_url="https://partner.staging.entur.org/oauth/token"
```
mjenssen, line 36: This should just be fetched from the secret, it's provided in the `tokenUrl` field.

scripts/android/register-app-version.sh
```diff
@@ -35,7 +33,7 @@ fi
...
     abt_url="https://core-abt-abt.staging.entur.io"
```
mjenssen, line 37: This is outdated, it should be just `https://core-abt.staging.entur.io` now

scripts/ios/register-app-version.sh
```diff
@@ -29,7 +27,7 @@ fi
 # Get values based on environment
 case $APP_ENVIRONMENT in
   staging)
-    token_url="https://partner-abt.staging.entur.org/oauth/token"
+    token_url="https://partner.staging.entur.org/oauth/token"
```
mjenssen, line 30: This should just be fetched from the secret, it's provided in the `tokenUrl` field.

scripts/ios/register-app-version.sh
```diff
@@ -29,7 +27,7 @@ fi
...
     abt_url="https://core-abt-abt.staging.entur.io"
```
mjenssen, line 31: This is outdated, it should be just `https://core-abt.staging.entur.io` now

mjenssen (changes requested):

mortennordseth:
Good catch, @mjenssen . Should be fixed now 👍

mortennordseth:
I think we might want to do this differently. I took a closer look at this and [GitHub says](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#naming-your-secrets) to avoid creating secrets in the JSON format. I guess one way to handle this is base64 encoding the JSON and decoding it in the script, but then again it might be cleaner to store the client ID and secret in separate variables instead. What do you think, @mjenssen @gorandalum ? 

Quote from GitHub the docs:
> To help ensure that GitHub redacts your secret in logs, avoid using structured data as the values of secrets. For example, avoid creating secrets that contain JSON or encoded Git blobs.

mortennordseth:
We looked a bit more into this and to be sure that secret redacting works as expected we should base64 encode the client. I've updated to reflect this

mjenssen (approved):

PR #4456, opened 2024-04-23
feat: add global rule for checking authentication type

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17781

Adds capability to set rules for `authenticationType`, the usage is similar to the existing rule system.

### DIFF
--- a/src/global-messages/GlobalMessage.tsx
+++ b/src/global-messages/GlobalMessage.tsx
@@ -41,7 +41,7 @@ const GlobalMessage = ({
 
   if (!globalMessageContext) return null;
 
-  const globalRuleVariables = {
+  const globalRuleVariables: RuleVariables = {
     authenticationType,
   };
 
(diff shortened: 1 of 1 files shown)


marius-at-atb (approved):

src/global-messages/GlobalMessage.tsx
```diff
@@ -39,10 +41,14 @@ const GlobalMessage = ({
 
   if (!globalMessageContext) return null;
 
-  const globalMessages = findGlobalMessages(
-    globalMessageContext,
-    ruleVariables,
-  );
+  const globalRuleVariables = {
```
rosvik, line 44: Makes sense to include the typing here, so it's clear that it should be on the RuleVariable format in the future.

```suggestion
  const globalRuleVariables: RuleVariables = {
```

PR #4457, opened 2024-04-23
feat: add app-login and app-login-phone global message contexts

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/17783

Related documentation changes: https://github.com/AtB-AS/docs-private/pull/41

<div>

</div>

### DIFF
--- a/src/global-messages/GlobalMessagesContext.tsx
+++ b/src/global-messages/GlobalMessagesContext.tsx
@@ -63,11 +63,6 @@ const GlobalMessagesContextProvider: React.FC = ({children}) => {
       firestore()
         .collection<GlobalMessageRaw>('globalMessagesV2')
         .where('active', '==', true)
-        .where(
-          'context',
-          'array-contains-any',
-          Object.values(GlobalMessageContextEnum),
-        )
         .onSnapshot(
           async (snapshot) => {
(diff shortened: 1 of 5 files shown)


src/global-messages/GlobalMessagesContext.tsx
```diff
@@ -63,11 +63,6 @@ const GlobalMessagesContextProvider: React.FC = ({children}) => {
       firestore()
         .collection<GlobalMessageRaw>('globalMessagesV2')
         .where('active', '==', true)
-        .where(
-          'context',
-          'array-contains-any',
-          Object.values(GlobalMessageContextEnum),
-        )
```
rosvik, line 70: Event though Firebase docs says they support up to 30 values in an
array-contains-any check, we get an error when we try to filter global
messages on more than 10 contexts.

https://cloud.google.com/firestore/docs/query-data/queries#limitations

We already do filtering based on context in `mapToGlobalMessages`, so
removing the snapshot filtering doesn't actually break anything, but it
probably does mean we use a bit more data and CPU loading and filtering
messages that won't be shown.

marius-at-atb (approved):
LGTM. Nice that you also remembered to update docs-private!

jorelosorio (approved):

PR #4459, opened 2024-04-23
chore: Recaptcha with encoded app id

gorandalum:


ios/atb/Info.plist
```diff
@@ -47,6 +47,7 @@
 				<string>com.googleusercontent.apps.793301954236-2umj9ofp91902ra1p816dtlh15gf7lev</string>
 				<string>app-1-312905478211-ios-a4f79f22f778f506efa4f6</string>
 				<string>app-1-973624729382-ios-930c02d2ab71e75ee12be7</string>
+				<string>app-1-939812594010-ios-5d15fdef3f42e289069363</string>
```
jorelosorio, line 50: I was wondering if it is possible to move these values to the .env file and encrypt then, since now seems like they are publicly visible.

PR #4460, opened 2024-04-23
fix: only render buy button if ticket is valid until the last interchange.

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/17325

### Background
When planning trips with several interchanges within the same zone, and with longer traveling time than the ticket is valid,
users are getting mislead to belive that they can buy one ticket for the whole trip. The users are chaught of guard when they have to buy a new ticket, when the validity period of their initial ticket has expired before the last interchange.

The users should not be able to buy tickets to a trip, if the tickets is not valid until the last interchange.

#### Illustrations
<details>
<summary>screenshots/video/figma</summary>

Before             | After    
Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-23 at 16 24 51 |  Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-23 at 16 26 58

</details>

### Proposed solution

Extend useGetTicketInfoFromTripc() with a condition to only render buy ticket button if one singel ticket covers the intire trip.

### Acceptance Criteria
- [ ] The button 'Buy ticket ', should not appear if the ticket is not valid until the last bus change on the trip. Otherwise, the render of the buy button should act work as usual.   

### Test input

-  Trip Steinshavn - Moa Trafikkterminal. Buy button should not render if ticket is invalid (1 hour and 30 minutes), before the last bus inter change.
- Kristiansund - Molde. Buy button should render.

### DIFF
--- a/src/travel-details-screens/TripDetailsScreenComponent.tsx
+++ b/src/travel-details-screens/TripDetailsScreenComponent.tsx
@@ -169,7 +169,7 @@ function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
     useGetFirstTariffZoneWeSellTicketFor(toTariffZones);
 
   const hasTooLongWaitTime = totalWaitTimeIsMoreThanAnHour(tripPattern.legs);
-  const ticketCoverEntirePeriod = ticketCoversEntireTrip(tripPattern.legs);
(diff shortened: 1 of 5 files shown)


gorandalum (changes requested):
Can't be done like this. Duration needs to come from offer-response. Also different transport companies have different rules regarding how long ticket needs to be valid.

PR #4462, opened 2024-04-24
feat: highlight Vipps login on LoginOptionsScreen

rosvik:
ref. https://github.com/AtB-AS/kundevendt/issues/17792

### DIFF
--- a/src/stacks-hierarchy/Root_LoginOptionsScreen.tsx
+++ b/src/stacks-hierarchy/Root_LoginOptionsScreen.tsx
@@ -186,9 +186,13 @@ export const Root_LoginOptionsScreen = ({
         </ThemeText>
 
         <View style={styles.buttonContainer}>
+          <VippsLoginButton
+            onPress={authenticateUserByVipps}
+            disabled={isLoading}
+          />
           <Button
-            interactiveColor="interactive_0"
-            mode="primary"
+            mode="secondary"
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Looks good! Has a designer taken a look, and updated Figma?

rosvik:
@gorandalum 
> Looks good! Has a designer taken a look, and updated Figma?

Yes, this is based on sketches from @oyvindmikkelsenatb. We thought it didn't make sense to update Figma, since this should be a temporary change, and be reverted after May.

reidzeibel:
Will this affect the OMS as well @rosvik ?

gorandalum:
> We thought it didn't make sense to update Figma, since this should be a temporary change, and be reverted after May.

Aha, I missed the part that this is temporary. On my part I don't see why it should be temporary if Vipps is the more reliable option.

rosvik:
SMS _shouldn't_ be a less reliable method, so I think it makes sense to keep it as a primary option once we're not concerned about the quota. But we can come back to that for the next release.

mortennordseth:
> @mortennordseth / @jonasbrunvoll / @mikaelbr Is this OK behavior for you for one app release? As mentioned above, this will be reverted after May.

Ok 👍

HanneLH:
Tested ok ✅

PR #4467, opened 2024-04-26
chore: update ios-devices for FRAM

mortennordseth:
I've added @marius-at-atb to the ios-devices list.


marius-at-atb (approved):
Nice! It's an iPhone 14 btw, but that's not important 😛

PR #4470, opened 2024-04-26
Fix: Android wrong Bugsnag api merge policy

jorelosorio:


reidzeibel (approved):

HanneLH:
Tested that Android is not crashing with fresh install ✅ 
Also tested ok on iOS.

PR #4472, opened 2024-04-30
refactor: Improve error handling and retry logic when fetching id token

gorandalum:
Now started to use useQuery for fetching id token to get more control
over the flow. Now we:
- Only retry 3 times
- Both errors and custom claims not found lead to retry
- Have a linear backoff starting a 3 seconds before retries. That
means that the first retry (after 3 seconds) and second retry (after
6 more seconds) will both be done before getting the loading boundary
timeout.

### DIFF
--- a/src/auth/use-fetch-id-token-with-custom-claims.ts
+++ b/src/auth/use-fetch-id-token-with-custom-claims.ts
@@ -1,43 +1,67 @@
 import {Dispatch, useEffect} from 'react';
 import {AuthReducerAction} from './types';
-import Bugsnag from '@bugsnag/react-native';
 import {AuthReducerState} from '@atb/auth/AuthContext';
+import {useQuery} from '@tanstack/react-query';
+import {logToBugsnag, notifyBugsnag} from '@atb/utils/bugsnag-utils';
 
-/*
+const RETRY_COUNT = 3;
+
+/**
+ * Variable signalling whether the next fetch id token request should be force
+ * refreshed or not. The first request should not be force refreshed, but
+ * retries should.
+ */
+let shouldForceRefresh = false;
+
+/**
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):
LGTM
The comments in the code regarding shouldForceRefresh and RETRY_COUNT make it easier to understand :+1:

tormoseng:
@gorandalum Testing this build-id I see several retries of the id-token when logging out, i.e creating a new anonymous account. And it's in the marked request that the following info is part of the id-token. It seems to me that this is the same pattern as before? Or am I misunderstanding something?
```
"abt_id": "ATB:CustomerAccount:BXixMx29z5eQjH1M7eB159JHIb72",
"customer_number": 3400096,
```

Screenshot 2024-05-02 at 10 24 58

tormoseng:
After some more testing, this issue seems fine! The retry of the idToken for /notification/v1/config was a but confusing - and still is an issue that generates more requests than necessary. However, this is handled in https://github.com/AtB-AS/mittatb-app/pull/4476.

Below is a test where the idToken-request returns HTTP 502. The idToken was requested at
- 11:51:35
- 11:51:38 (+ 3 sek)
- 11:51:45 (+ 6 sek)
- 11:51:54 (+ 9 sek)
And Bugsnag then sendt `[auth\/internal-error] An internal error has occurred, please try again.: No id token with custom claims received after 3 retries`. The "normal" Bugsnag error is `Token doesn't contain custom claims: No id token with custom claims received after 3 retries`. 

So, for Prod we should check for, at least, `Token doesn't contain custom claims` with `metadata.description` with `No id token with custom claims received after 3 retries`.

<details>
<summary>Requests when idToken gives HTTP 502</summary>

Screenshot 2024-05-02 at 11 53 12

</details>

PR #4473, opened 2024-05-02
fix: remove unnecessary ScreenHeader from FullScreenErrorView

reidzeibel:
part of https://github.com/AtB-AS/kundevendt/issues/17839

On the announcement crash issue, the app doesn't show the usual crash screen. If we run the debug version of the app and trigger the crash, the issue shown would be `useAuthState must be used within a AuthContextProvider`, which is not the actual issue. 

After some investigation, it is because the `FullScreenErrorView` was calling the `ScreenHeader` component, which contains a `GlobalMessage` component. This `GlobalMessage` component reads `AuthState`, after it got updated recently on this issue https://github.com/AtB-AS/mittatb-app/issues/4456. Since the `ErrorBoundary` is the parent of `AuthContextProvider`, the app crashes due to `useAuthState` being called while the current context is the `ErrorBoundary`.

Therefore the solution was to remove `ScreenHeader` from the crash screen, it doesn't really have any purpose when I compared the before & after. There was no title, no icons, and cannot be interacted with, so it should be safe to remove.

Comparison below (before vs after):

### DIFF
--- a/src/error-boundary/FullScreenErrorView.tsx
+++ b/src/error-boundary/FullScreenErrorView.tsx
@@ -57,7 +57,6 @@ export function FullScreenErrorView({onRestartApp, errorCode}: ErrorProps) {
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   safearea: {
     flex: 1,
-    paddingTop: 44,
     backgroundColor: theme.static.background.background_2.background,
   },
   svgContainer: {
--- a/src/error-boundary/FullScreenErrorView.tsx
+++ b/src/error-boundary/FullScreenErrorView.tsx
@@ -58,6 +58,7 @@ export function FullScreenErrorView({onRestartApp, errorCode}: ErrorProps) {
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   safearea: {
     flex: 1,
+    paddingTop: 44,
     backgroundColor: theme.static.background.background_2.background,
   },
   svgContainer: {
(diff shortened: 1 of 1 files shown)


src/error-boundary/FullScreenErrorView.tsx
```diff
@@ -59,6 +57,7 @@ export function FullScreenErrorView({onRestartApp, errorCode}: ErrorProps) {
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   safearea: {
     flex: 1,
+    paddingTop: 44,
```
jorelosorio, line 60: it is possible to remove this? seems like a hack, maybe wrap it in <safearea> component?

jorelosorio (changes requested):

src/error-boundary/FullScreenErrorView.tsx
reidzeibel: We could but then the screen will be moved up by 44 units up, I guess it's okay since we don't expect users to see this screen too often.

jorelosorio: I wrote this link https://reactnative.dev/docs/safeareaview in my previous message but seems like it was not published.

jorelosorio: There is one already, but perhaps not for that view.

reidzeibel: What if I move the padding to the `container` and not `safeArea`?

reidzeibel: Comparison without extra padding

jorelosorio (approved):

reidzeibel:
@tormoseng this is the PR

HanneLH:
Tested ok on release branch! ✅

PR #4474, opened 2024-05-02
feat: add converter to set default values and prevent unwanted crash

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17839

The destructuring of `actionButton` from firestore causes the app to crash when the values are invalid (in this case, `actionType`). This PR adds type-checking during the conversion to add handling of invalid data. With this PR, issues related to `actionType` in the announcements (for example, mistyping), will result in the announcement to not be shown to users.

I added the enum to provide a centralized place for typing, and to help reducing typo during development. In the future if we need another type of action (maybe Snackbar?), then we can just add them to the enum, create another button type or just add it to existing button type, and add correct handling on the `Announcement.tsx`.

Documentation is also updated on https://github.com/AtB-AS/docs-private/pull/47

### DIFF
--- a/src/announcements/converters.ts
+++ b/src/announcements/converters.ts
@@ -1,5 +1,5 @@
 import {FirebaseFirestoreTypes} from '@react-native-firebase/firestore';
-import {AnnouncementRaw, AnnouncementType, ActionButton} from './types';
+import {AnnouncementRaw, AnnouncementType, ActionButton, ActionType} from './types';
 import {mapToLanguageAndTexts} from '@atb/utils/map-to-language-and-texts';
 import {APP_VERSION} from '@env';
 import {AppPlatformType} from '@atb/global-messages/types';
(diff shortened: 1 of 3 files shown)


src/announcements/converters.ts
```diff
@@ -44,6 +44,7 @@ export const mapToAnnouncement = (
   if (!isAppPlatformValid(platforms)) return;
   if (appVersionMin && appVersionMin > APP_VERSION) return;
   if (appVersionMax && appVersionMax < APP_VERSION) return;
+  if (!actionButton) return;
```
marius-at-atb, line 47: Are we sure that we never want to show announcements without an actionButton?

reidzeibel: I have discussed it with @gorandalum and to prevent user error or fallbacks, we should just remove them entirely, similar to global messages, the documentation is also updated to reflect these changes, and adding/removing fields should be trivial since we can edit it directly through firestore. https://github.com/AtB-AS/docs-private/pull/47/

marius-at-atb: In that case, LGTM 😄

marius-at-atb (approved):

PR #4475, opened 2024-05-03
feat: Basic translations related to Troms/Svipper app

tft-kaare:
The hard-coded in-app texts are ready for testing

### DIFF
--- a/src/translations/components/ServiceDisruptions.ts
+++ b/src/translations/components/ServiceDisruptions.ts
@@ -41,4 +41,18 @@ export default orgSpecificTranslations(ServiceDisruptionsTexts, {
       text: _('Åpne i nettleser', 'Open in browser', 'Opne i nettlesar'),
     },
   },
+  troms: {
+    body: _(
+      'Oversikt over endringer, forsinkelser og avvik finner du på svipper.no:',
+      'Get an overview of changes, delays and service disruptions at svipper.no (Norwegian only):',
+      'Oversikt over endringar, forseinkingar og avvik finn du på svipper.no:',
+    ),
+    button: {
+      text: _(
+	'fylkestrafikk.no/trafikkmeldinger',
+	'fylkestrafikk.no/trafikkmeldinger',
+	'fylkestrafikk.no/trafikkmeldinger',
+      ),
+    },
+  },
 });
--- a/src/translations/components/ServiceDisruptions.ts
+++ b/src/translations/components/ServiceDisruptions.ts
@@ -50,8 +50,8 @@ export default orgSpecificTranslations(ServiceDisruptionsTexts, {
     button: {
       text: _(
 	'Trafikkmeldinger',
-	'Trafikkmeldinger',
-	'Trafikkmeldinger',
+	'Customer information',
+	'Trafikkmeldingar',
       ),
     },
   },
--- a/src/translations/components/ServiceDisruptions.ts
+++ b/src/translations/components/ServiceDisruptions.ts
@@ -49,9 +49,9 @@ export default orgSpecificTranslations(ServiceDisruptionsTexts, {
     ),
     button: {
       text: _(
-	'fylkestrafikk.no/trafikkmeldinger',
-	'fylkestrafikk.no/trafikkmeldinger',
-	'fylkestrafikk.no/trafikkmeldinger',
+	'Trafikkmeldinger',
+	'Trafikkmeldinger',
+	'Trafikkmeldinger',
       ),
     },
   },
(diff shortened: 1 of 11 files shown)


gorandalum (approved):
Look alright. Looking at this, we should preferably get something like `APP_PRESENTATION_NAME` in the env-files, so we don't need to replace texts throughout the app.

src/translations/screens/ExtendedOnboarding.ts
```diff
@@ -124,14 +124,14 @@ export default orgSpecificTranslations(OnboardingTexts, {
...
-        'Velkommen til Troms-appen',
+        'Velkommen til Svipper-appen',
+        'Welcome to the Svipper app',
+        'Velkommen til Svipper-appen',
       ),
       titleA11yLabel: _(
-        'Velkommen til Troms appen',
-        'Welcome to the Troms app',
-        'Velkommen til Troms-appen',
+        'Velkommen til Svipper appen',
```
marius-at-atb, line 132:
```suggestion
        'Velkommen til Svipper-appen',
```

src/translations/screens/subscreens/MobileTokenOnboarding.ts
```diff
@@ -310,4 +310,29 @@ export default orgSpecificTranslations(MobileTokenOnboardingTexts, {
       },
     },
   },
+  troms: {
+    ticketSafetyInfo: {
+      description: _(
+	'Billetten din er trygt lagret på **Min profil**. Dermed vil du aldri miste den — selv om du mister bytter mobil.',
```
marius-at-atb, line 316:
```suggestion
	'Billetten din er trygt lagret på **Min profil**. Dermed vil du aldri miste den — selv om du bytter mobil.',
```

src/translations/components/ServiceDisruptions.ts
```diff
@@ -41,4 +41,18 @@ export default orgSpecificTranslations(ServiceDisruptionsTexts, {
...
+    body: _(
+      'Oversikt over endringer, forsinkelser og avvik finner du på svipper.no:',
+      'Get an overview of changes, delays and service disruptions at svipper.no (Norwegian only):',
+      'Oversikt over endringar, forseinkingar og avvik finn du på svipper.no:',
+    ),
+    button: {
+      text: _(
+	'fylkestrafikk.no/trafikkmeldinger',
+	'fylkestrafikk.no/trafikkmeldinger',
+	'fylkestrafikk.no/trafikkmeldinger',
```
marius-at-atb, line 54: Normally URLs should not be hard coded. However here it is only used as a label, so I guess it is acceptable, even though it could cause a mismatch at some point in the future.

marius-at-atb (approved):
Looks good :+1:

tft-kaare:
Agree. Had the same thoughts when doing this, but as we stress with getting a desent version of the app up and running, this was a quick-fix. Follow-up issue seems like a natural next step

src/translations/screens/subscreens/MobileTokenOnboarding.ts
tft-kaare: Good catch

src/translations/components/ServiceDisruptions.ts
```diff
@@ -49,9 +49,9 @@ export default orgSpecificTranslations(ServiceDisruptionsTexts, {
     ),
     button: {
       text: _(
-	'fylkestrafikk.no/trafikkmeldinger',
-	'fylkestrafikk.no/trafikkmeldinger',
-	'fylkestrafikk.no/trafikkmeldinger',
+	'Trafikkmeldinger',
+	'Trafikkmeldinger',
+	'Trafikkmeldinger',
```
marius-at-atb, line 54: Remember to adapt English and Nynorsk as well 😉

rosvik:
@tft-kaare Style-nitpick on this change: I noticed you've used tabs for indentation some places here, instead of the 2 spaces we use elsewhere. I can recommend setting up Prettier to format on save, so you don't have to think about it 😊

tft-kaare:
Aaahh! It's like spinach between the teeth of the speaker: I cannot get the essense of the topic as I only focus on the green goo.

As this is already committed to master, what is the convention for cleaning this up? New issue? Continue on this PR?

rosvik:
> It's like spinach between the teeth of the speaker: I cannot get the essense of the topic as I only focus on the green goo.

🤣

> As this is already committed to master, what is the convention for cleaning this up? New issue? Continue on this PR?

For follow up changes like that, make a new PR, then add a link to the original PR or issue in the description. Just so it's easy for reviewers to find the reasons behind the change.

PR #4476, opened 2024-05-03
refactor: Clear queries on user change

gorandalum:
### Background
Inspecting network requests we see that some requests are triggered more times than expected on user change, and some are also triggered before the user account is fully created, giving 401 responses.

The reason for this is that there are race conditions between the use-query hooks and the invalidate queries statement. Invalidating queries on user change was done in the RootStack, meaning that it happened after contexts had triggered based on user change, but before components had triggered based on user change.

This means that when the RootStack triggered invalidation of queries, existing queries was rerun for the old user, since the user isn't updated in the child-components yet. But it is injected with the new id-token in the axios client, which gives 401 since the request is for the old user, but with the id-token for the new user.

### Solution
Instead of invalidating queries, which makes the immediately refetch,
the query cache is instead cleared on user change. Clearing queries
doesn't automatically refetch them, so queries that should be
refetched after user change needs to have user id in query key.

Also we enable queries after auth status is 'authenticated', as doing
the requests before this gives 401 from the backend.

### Acceptance criteria
- [ ] There should be no requests which responds with 401 after user change
- [ ] Requests which require auth should not be done before the user is authenticated (has id token with custom claims)

### Things to test
- [ ] Touched GET-endpoints:  `/tokens/v4/list`, `/mobility/v1/benefits/${productId}`, `/mobility/benefits`, `/mobility/code/${operatorId}`, `/notification/v1/config`, `/profile/v1`, `product/v1`.
- [ ] All above-mentioned endpoints should be refetched after new user is authenticated. Either immediately, or when opening the appropriate component which needs the data.

### DIFF
--- a/src/auth/AuthContext.tsx
+++ b/src/auth/AuthContext.tsx
@@ -3,6 +3,7 @@ import React, {
   PropsWithChildren,
   useCallback,
   useContext,
+  useEffect,
   useReducer,
 } from 'react';
 import auth, {FirebaseAuthTypes} from '@react-native-firebase/auth';
@@ -25,6 +26,8 @@ import {useFetchIdTokenWithCustomClaims} from './use-fetch-id-token-with-custom-
 import Bugsnag from '@bugsnag/react-native';
 import isEqual from 'lodash.isequal';
 import {mapAuthenticationType} from '@atb/auth/utils';
+import {useQueryClient} from '@tanstack/react-query';
+import {useClearQueriesOnUserChange} from "@atb/auth/use-clear-queries-on-user-change";
 
 export type AuthReducerState = {
   authStatus: AuthStatus;
@@ -130,6 +133,7 @@ export const AuthContextProvider = ({children}: PropsWithChildren<{}>) => {
   useFetchIdTokenWithCustomClaims(state, dispatch);
 
   useUpdateAuthLanguageOnChange();
+  useClearQueriesOnUserChange(state);
 
   const retryAuth = useCallback(() => {
     if (state.authStatus === 'fetch-id-token-timeout') {
--- a/src/auth/AuthContext.tsx
+++ b/src/auth/AuthContext.tsx
@@ -3,7 +3,6 @@ import React, {
   PropsWithChildren,
   useCallback,
   useContext,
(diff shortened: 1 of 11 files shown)


reidzeibel:
Looks ok, just one question, why don't we use the `refetchType:none` instead? It says that we can use that to prevent refetching during invalidation?

```
queryClient.invalidateQueries

The invalidateQueries method can be used to invalidate and refetch single or multiple queries in the cache
based on their query keys or any other functionally accessible property/state of the query.
By default, all matching queries are immediately marked as invalid and active queries are
refetched in the background.

- If you do not want active queries to refetch, and simply be marked as invalid, 
  you can use the refetchType: 'none' option.
- If you want inactive queries to refetch as well, use the refetchType: 'all' option
```

https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientinvalidatequeries

gorandalum:
> Looks ok, just one question, why don't we use the `refetchType:none` instead? It says that we can use that to prevent refetching during invalidation?

I took a look at that, and also `resetQueries` and `removeQueries`. Mainly I think `clear` seems like the most correct option, when changing user we want a full clear of the cached data. But I'm a little unsure what the practical difference between `clear` and `invalidateQueries` with `refetchType: 'none'` would be, or if it would just behave the same 🤔

marius-at-atb (approved):
Looks good

reidzeibel (approved):

tormoseng:
@gorandalum Looks very good. This is a change I like!

However, on one instance, when logging out, I got a lot of 401 from `/profile/v1/lookup` (see screenshot). Tried to re-create it, but the next 3 attempts on logging out I didn't see this request.

<details>
<summary>Requests</summary>

Screenshot 2024-05-08 at 10 33 59

</details>

gorandalum:
I can take a look at this. It probably went under the radar since I didn't have any tickets received from others.

gorandalum:
> I can take a look at this. It probably went under the radar since I didn't have any tickets received from others.

Created a [PR](https://github.com/AtB-AS/mittatb-app/pull/4495) for this.

PR #4477, opened 2024-05-03
refactor: Add more logging to ticketing Firestore subscriptions

gorandalum:
### Background
This is done because of the error Nisha experiences, where the tokens from previous account is shown in the app even after logging out.

### Solution
Log to Bugsnag on the `onError` callback when setting up Firestore.

Also removed some unused reducer actions/state regarding errors.
I'm not really sure how many errors this will give in Bugsnag, but
for now I think it is better to log too much than too little, and
we'll rather reduce it some later if necessary.

### Acceptance criteria
Shouldn't be any functional changes to test. The issue is sent through testing so that testers are aware of the new logging.


reidzeibel (approved):

marius-at-atb (approved):
Lgtm
Something else I noticed reviewing this, is that `isRefreshingFareContracts` is never set to `true`?

PR #4479, opened 2024-05-06
feat: add auth retry when there is no customer number

rosvik:
Made while investigating https://github.com/AtB-AS/kundevendt/issues/17733

- Added message section item if there is no customer number, with action to retry auth.
- Added a red notification dot if there the message box is there, to make it more discoverable. 
- Retries are logged to posthog

ref. https://mittatb.slack.com/archives/CSFC6PG23/p1714987182087399

### Acceptance criteria

Not sure how easy this is to test, but

- [ ] For an account without claims / customer number, the error message shows up on my profile
- [ ] The message doesn't normally show up when logging in and out
- [ ] The red dot shows up at the same time as the error message
- [ ] Retries are logged to posthog

### DIFF
--- a/src/translations/screens/Profile.ts
+++ b/src/translations/screens/Profile.ts
@@ -47,9 +47,9 @@ const ProfileTexts = {
         customerNumber: _('Kundenummer', 'Customer number', 'Kundenummer'),
         phoneNumber: _('Telefonnummer', 'Phone number', 'Telefonnummer'),
         claimsError: _(
-          'Det oppstod et problem med lasting av kontoen din. Trykk for å prøve igjen, eller kontakt kundeservice hvis problemet vedvarer.',
-          'There was a problem loading your account. Tap to try again, or contact customer service if the problem persists.',
-          'Det oppstod eit problem med lasting av kontoen din. Trykk for å prøve igjen, eller kontakt kundeservice hvis problemet blir verande.',
+          'Det oppstod et problem med lasting av kontoen din. Trykk for å laste inn på nytt, eller kontakt kundeservice hvis problemet vedvarer.',
+          'There was a problem loading your account. Tap to reload, or contact customer service if the problem persists.',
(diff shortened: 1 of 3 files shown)


marius-at-atb (approved):
Lgtm :+1:

src/translations/screens/Profile.ts
```diff
@@ -46,6 +46,11 @@ const ProfileTexts = {
       infoItems: {
         customerNumber: _('Kundenummer', 'Customer number', 'Kundenummer'),
         phoneNumber: _('Telefonnummer', 'Phone number', 'Telefonnummer'),
+        claimsError: _(
+          'Det oppstod et problem med lasting av kontoen din. Trykk for å laste inn på nytt, eller kontakt kundeservice hvis problemet vedvarer.',
```
hanne-at-atb, line 50:
```suggestion
          'Det oppstod et problem med lasting av kontoen din.',
```
Together with @Aliaaen we suggest the above as we think the suggested solution might break with WCAG 1.3.3. Also the second sentence does not add any valuable information to the user - given that customer service can't really do anything ▶️  according to @Sebstorvik .

rosvik: https://github.com/AtB-AS/mittatb-app/pull/4483

gorandalum (commented):
Looks good. Only comment from me is that I think in theory the checks should be on whether `authStatus` is `authenticated` or not, instead of checking if `customerNumber` is `undefined`. But in practice it ends up the same.

rosvik:
> Only comment from me is that I think in theory the checks should be on whether `authStatus` is `authenticated` or not, instead of checking if `customerNumber` is `undefined`. But in practice it ends up the same.

@gorandalum That seems to work the same way, so I changed it in https://github.com/AtB-AS/mittatb-app/pull/4483 now 👍

tormoseng:
👍

PR #4484, opened 2024-05-07
fix: fix 1.50 crash and update intercom dependency version

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17968
also closes https://github.com/AtB-AS/kundevendt/issues/2269

So the crash had to do something with Intercom unable to find/call the correct method due to obfuscation on Android. I tried adding several exclusion to the proguard rules, but unable to find the right exclusion rules. Then I tried updating the Intercom dependency version, we have been using a [deprecated library](https://github.com/tinycreative/react-native-intercom) and it wasn't the official version either, so I added the [official intercom library](https://github.com/intercom/intercom-react-native) and removed the old one.

There are some minor changes needed, but the functionality shouldn't change at all. I have tested on both Android (on obfuscated and non-obfuscated version) and iOS.


rosvik (approved):
Looks good 👍 

Remember conventional PR message (`fix: ...`)

jorelosorio:
I see it is merged now, but then this means that the content inside `Intercom.plist` is no longer required. It is possible then to remove this file?

reidzeibel:
@jorelosorio true, we'll see after this issue passed the test

PR #4488, opened 2024-05-08
fix: init intercom user creation regardless of login status

reidzeibel:
followup for issue in https://github.com/AtB-AS/kundevendt/issues/2269

If we use anonymous user, the `registerChatUser()` function doesn't get called, seems like the `userCreation` onboarding was not called for whatever reason :

```
       loadedOnboardingSections.find(
          (lOS) => lOS.onboardingSectionId === 'userCreation',
        )?.isOnboarded
```

so the solution is just to create the chat user regardless of whether the onboarding is done or not, since we don't use any data from the user anyways. Here's the snippet of the `registerChatUser()` function: 

```
export async function register() {
  await Intercom.loginUnidentifiedUser();
  await Intercom.setBottomPadding(Platform.OS === 'ios' ? 40 : 80);

  const installId = await storage.get('install_id');
  const buildNumber = DeviceInfo.getBuildNumber();
  const deviceId = DeviceInfo.getDeviceId();
  const isLocationEnabled = await DeviceInfo.isLocationEnabled();
  const appLocationStatus = await checkGeolocationPermission();
  const {width, height} = Dimensions.get('window');

  await updateMetadata({
    'AtB-Install-Id': installId ?? 'unknown',
    'AtB-Build-Number': buildNumber,
    'AtB-Device-Type': deviceId,
    'AtB-Device-Location-Enabled': isLocationEnabled,
    'AtB-App-Location-Status': appLocationStatus,
    'AtB-Platform-OS': Platform.OS,
    'AtB-OS-Font-Scale': PixelRatio.getFontScale(),
    'AtB-Screen-Size': `${width}x${height}`,
  });
}
```

We can see there that nothing is really tied down to the login status, so it should be safe to init a chat user regardless of the onboarding status.

### DIFF
--- a/src/onboarding/OnboardingContext.tsx
+++ b/src/onboarding/OnboardingContext.tsx
@@ -8,7 +8,6 @@ import React, {
   useReducer,
 } from 'react';
 import RNBootSplash from 'react-native-bootsplash';
-import {register as registerChatUser} from '../chat/user';
 import {storage} from '@atb/storage';
 
 import {
--- a/src/onboarding/OnboardingContext.tsx
+++ b/src/onboarding/OnboardingContext.tsx
@@ -134,13 +134,7 @@ export const OnboardingContextProvider: React.FC = ({children}) => {
         ),
       );
 
-      if (
-        loadedOnboardingSections.find(
-          (lOS) => lOS.onboardingSectionId === 'userCreation',
-        )?.isOnboarded
-      ) {
-        registerChatUser();
-      }
+      registerChatUser();
 
       dispatch({
         type: 'LOAD_ONBOARDING_SETTINGS',
--- a/src/onboarding/OnboardingContext.tsx
+++ b/src/onboarding/OnboardingContext.tsx
@@ -134,8 +134,6 @@ export const OnboardingContextProvider: React.FC = ({children}) => {
         ),
       );
 
-      registerChatUser();
-
       dispatch({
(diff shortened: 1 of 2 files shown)


reidzeibel:
@tormoseng no, as you can see on the code, we only send these values to Intercom: 

```
    'AtB-Install-Id': installId ?? 'unknown',
    'AtB-Build-Number': buildNumber,
    'AtB-Device-Type': deviceId,
    'AtB-Device-Location-Enabled': isLocationEnabled,
    'AtB-App-Location-Status': appLocationStatus,
    'AtB-Platform-OS': Platform.OS,
    'AtB-OS-Font-Scale': PixelRatio.getFontScale(),
    'AtB-Screen-Size': `${width}x${height}`,
```

We don't have any user-identifying information other than Install ID

marius-at-atb (approved):
Lgtm :+1:
Note that the `userCreation` `isOnboarded` flag is updated, but the useEffect isn't re-triggered.

reidzeibel:
@marius-at-atb any suggestions where should we move it?

marius-at-atb:
I think a new useEffect in `RootStack` can be a good place

jorelosorio:
I agree with @marius-at-atb to move it outside the onboarding, seems like it does not is longer required to be there, but I guess the intention was not to create it on an early state but, we might then create another PR to have some refactor around intercom and also evaluate new possibilities with the new version?

jorelosorio (approved):

PR #4490, opened 2024-05-08
fix: only call mobility/code for single and free unlock

marius-at-atb:
It was discovered that the app sends requests to /mobility/code/ even when the benefit code is "free-use".
However it should not.

To avoid this, the new `benefitRequiresValueCodeToUnlock` is returned from `useIsEligibleForBenefit`, and the `useValueCodeQuery` has been moved inside the new `OperatorActionButtonWithValueCode` which is only rendered when `benefitRequiresValueCodeToUnlock` is true. 

Related issue: https://github.com/AtB-AS/team-platform/issues/572

Acceptance criteria:
- [ ] A value code should be requested (`/mobility/code/${operatorId}`) when benefit id is one of
   - [ ] `free-unlock`
   - [ ] `single-unlock`
- [ ] A value code should NOT be requested (`/mobility/code/${operatorId}`) when benefit id is one of
  - [ ] `free-use`
 Note for testing: this request is triggered when opening a shared mobility map item


src/mobility/use-is-eligible-for-benefit.tsx
```diff
@@ -1,5 +1,11 @@
 import {useUserBenefitsQuery} from '@atb/mobility/queries/use-user-benefits-query';
 import {OperatorBenefitType} from '@atb-as/config-specs/lib/mobility-operators';
+import {isDefined} from '@atb/utils/presence';
+
+const benefitIdsRequiringValueCode: OperatorBenefitType['id'][] = [
+  'single-unlock',
+  'free-unlock',
+];
 
 export const useIsEligibleForBenefit = (
```
marius-at-atb, line 10: Could be a good idea to add tests for `useIsEligibleForBenefit`.
@gorandalum do we have any developer guidelines for considering whether it is worth the time or not to add such tests?

gorandalum: Doesn't really have any guidelines on this for now, but when in doubt just add them imo :)

reidzeibel (approved):

src/mobility/use-is-eligible-for-benefit.tsx
marius-at-atb: Okay, I'll take a look at it before merging then :+1:

src/mobility/use-is-eligible-for-benefit.tsx
```diff
@@ -1,23 +1,35 @@
 import {useUserBenefitsQuery} from '@atb/mobility/queries/use-user-benefits-query';
 import {OperatorBenefitType} from '@atb-as/config-specs/lib/mobility-operators';
+import {isDefined} from '@atb/utils/presence';
+
+const benefitIdsRequiringValueCode: OperatorBenefitType['id'][] = [
+  'single-unlock',
+  'free-unlock',
+];
```
gorandalum, line 8: I don't think we should reference operator benefit types in the code, as it breaks the low coupling between Firestore and the app code. Should get this configuration into Firestore instead.

gorandalum (changes requested):

src/mobility/use-is-eligible-for-benefit.tsx
marius-at-atb: Yes, I should have thought about that! Will update

marius-at-atb: https://github.com/AtB-AS/mittatb-app/pull/4490/commits/31a0831079d24389498f837a69b659b5c4fb2d1f

gorandalum (approved):

tormoseng:
Works as expected!

PR #4494, opened 2024-05-10
chore: remove metro warning for non-serializable values on navigation

reidzeibel:
Not related to any issues.

Found the issue during the debugging of loading boundary and tried to logout-login multiple times to test mobile token.

This PR will remove the console warning when you select "Log in" from profile screen, the warning is: 

`Non-serializable values were found in the navigation state`

I removed the `transitionPreset` from navigation call and the warning no longer shows, and tested on both simulator and emulator, the animation still works fine.

### DIFF
--- a/src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
+++ b/src/stacks-hierarchy/Root_LoginActiveFareContractWarningScreen.tsx
@@ -14,7 +14,6 @@ import {
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy/navigation-types';
 import {useTimeContextState} from '@atb/time';
-import {TransitionPresets} from '@react-navigation/stack';
 import {FullScreenFooter} from '@atb/components/screen-footer';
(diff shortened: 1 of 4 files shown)


gorandalum (changes requested):
I was probably a little unclear in the comment that caused this PR. I didn't mean that all `transitionPreset` applied in the app was redundant and could be removed, I ment that the underlying issue of warnings in the console when applying the should be removed.

One possible way to do that might be to change the type from `TransitionPreset` to something like `'slideUp' | 'slideFromRight'`, and map those values to an actual `TransitionPreset` when using them in the options.

gorandalum:
> I was probably a little unclear in the comment that caused the latest changes. I didn't mean that all `transitionPreset` applied in the app was redundant and could be removed, I ment that the underlying issue of warnings in the console when setting `transitionPreset` should be removed.
> 
> One possible way to do that might be to change the type from `TransitionPreset` to something like `'slideUp' | 'slideFromRight'`, and map those values to an actual `TransitionPreset` when using them in the options.

reidzeibel:
Oooh, got it, thanks for the pointers!

reidzeibel:
closing this for the time being to focus on the 1.51 release

PR #4503, opened 2024-05-13
feat: add activate now button to fare contracts

rosvik:
Mostly copied over code from carnet activation, but with different texts and making the API call to `ticket/v4/start-time`. 

I send no start time to the back end, so that we don't have to trust the client clock. Default seems to be that it starts immediately. (If that's an ok solution, @adriawh?)

https://github.com/AtB-AS/mittatb-app/assets/1774972/6295f59a-2eb8-4f56-95ab-1ab4b47465df

closes https://github.com/AtB-AS/kundevendt/issues/16878

### DIFF
--- a/src/fare-contracts/FareContractView.tsx
+++ b/src/fare-contracts/FareContractView.tsx
@@ -28,9 +28,11 @@ import {
   isCanBeConsumedNowFareContract,
   isSentOrReceivedFareContract,
   FareContract,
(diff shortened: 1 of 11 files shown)


mortennordseth (approved):
🚀

marius-at-atb:
Nice and clean!
I think `isCanBeActivatedNowFareContract` is the kinda function that is good to have tests for as well.

PR #4506, opened 2024-05-14
fix: Clear queries before id token fetch

gorandalum:
Clearing queries on user change was introduced on master branch,
while using react-query for id token fetch was introduced on
release branch. When these two merged, the clearing of queries
happened after id token fetch, which led to errors.


reidzeibel:
@gorandalum I modified the PR a bit with the changes from master (`notifyBugsnag` method update)

reidzeibel (approved):

tormoseng:
👍

PR #4511, opened 2024-05-15
fix: display green dot when filters other than the default filters are selected

jonasbrunvoll:
This PR fixes the green dot issue addressed in
https://github.com/AtB-AS/kundevendt/issues/17913#issuecomment-2110323180


tormoseng (approved):
Seems like there's some tsc errors.

jonasbrunvoll:
> Seems like there's some tsc errors.

Yes, but it does not seems like the tsc errors appers from the change I made🤔

PR #4513, opened 2024-05-15
fix: rework intercom registration for the new package

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/18085

relevant info from official intercom community page : https://community.intercom.com/mobile-sdks-24/error-failed-to-register-user-we-already-have-a-regsitered-user-if-you-are-attempting-to-register-a-new-user-call-logout-before-this-3909

The official library/package from Intercom handles registration differently compared to the old package. In this new package, the library doesn't handle registration in Android quite well, we need to really check for Intercom registration status before registering a new unidentified user. Therefore the solution in this PR is to save the registration status into the async storage, and then only registers if the async storage hasn't been set yet.

Why is the value for the async storage never false? because we never log out the Intercom user. The `logout()` function from Intercom should only be called when registering an _identified user_, since we never registers an _identified user_, we don't need to call the `logout()` function [as mentioned in the guideline](https://developers.intercom.com/installing-intercom/react-native/using-intercom/#how-to-logout-an-identified-user)

### DIFF
--- a/src/chat/use-register-intercom-user.ts
+++ b/src/chat/use-register-intercom-user.ts
@@ -32,10 +32,8 @@ async function register() {
 export function useRegisterIntercomUser() {
   useEffect(() => {
     storage.get(StorageModelKeysEnum.IntercomUserRegistered).then((it) => {
-      console.log('Intercom storage', it);
       if (it !== 'true') {
         register().then(() => {
-          console.log('Intercom registration done');
           storage.set(StorageModelKeysEnum.IntercomUserRegistered, 'true');
         });
       }
--- a/src/chat/use-register-intercom-user.ts
+++ b/src/chat/use-register-intercom-user.ts
@@ -1,11 +1,12 @@
 import Intercom from '@intercom/intercom-react-native';
 import {Dimensions, PixelRatio, Platform} from 'react-native';
 import DeviceInfo from 'react-native-device-info';
-import {storage} from '@atb/storage';
+import {StorageModelKeysEnum, storage} from '@atb/storage';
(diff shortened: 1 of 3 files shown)


rosvik (commented):
I'm wondering if this really is the best approach. I don't know how it will happen, but if the local state at some point no longer reflects the actual registered state, we wont ever retry registration with this solution.

1. Is there any way to check if the user is already registered? If so, could we use that instead of local storage?
2. Or, could we keep the same logic as before, except we silently catch and discard the error, since it's expected? Or does that break something?

reidzeibel:
@rosvik 

1. Based on the documentation and methods available on the `Intercom` package, there's no way to check if a user is already registered, they only provide 3 methods regarding login/logout: `loginUnidentifiedUser`, `loginUserWithUserAttributes`, `logout`. There are no `isLoggedInUser` or similar method.
2. If that is okay, then we can just do a try-catch 🤔 ?

rosvik:
> 1. Based on the documentation and methods available on the `Intercom` package, there's no way to check if a user is already registered, they only provide 3 methods regarding login/logout: `loginUnidentifiedUser`, `loginUserWithUserAttributes`, `logout`. There are no `isLoggedInUser` or similar method.

Thats a bit strange, but ok 🤷 

> 2. If that is okay, then we can just do a try-catch 🤔 ?

Yup, sounds good!

reidzeibel:
@rosvik try catch works, at least it doesn't send new events to bugsnag.... I'll close this one and create a new PR for it.

PR #4520, opened 2024-05-16
fix: Reset is timeout state on loading status change

gorandalum:
This is done both on user change, and on status being set to
success. The error was that the timeout state wasn't reset on user
change, which made the loading for the new user go immediately to
timeout state.

### DIFF
--- a/src/loading-screen/use-loading-state.ts
+++ b/src/loading-screen/use-loading-state.ts
@@ -38,9 +38,7 @@ export const useLoadingState = (timeoutMs: number): LoadingState => {
   useEffect(() => {
     if (status === 'success') {
       setIsTimeout(false);
-    }
-
-    if (status === 'loading') {
+    } else if (status === 'loading') {
       const id = setTimeout(() => setIsTimeout(true), timeoutMs);
       return () => clearTimeout(id);
     }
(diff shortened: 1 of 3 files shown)


marius-at-atb (approved):

src/loading-screen/use-loading-state.ts
```diff
@@ -23,15 +24,22 @@ export const useLoadingState = (timeoutMs: number): LoadingState => {
 
   const status = loadSuccessful ? 'success' : isTimeout ? 'timeout' : 'loading';
 
+  useEffect(() => setIsTimeout(false), [userId]);
```
rosvik, line 27: Won't timeout be set to false on mount here? Is that intended?

marius-at-atb: Yeah should be fine as is. However I think it does mean that the order of the useEffects matter.

rosvik (approved):

src/loading-screen/use-loading-state.ts
marius-at-atb: Not right now due to the `setIsTimeout(true)` being wrapped with a `setTimeout`. However in general, when updating a state variable on mount, if several useEffects set it, the order will matter.

PR #4515, opened 2024-05-15
fix: don't call intercom methods when remote config flag is false

reidzeibel:
fixes the crash that happens on OMS-partners regarding Intercom.

closes https://github.com/AtB-AS/kundevendt/issues/2939 as well.

This PR wraps every Intercom-related call with a check to ensure that Intercom is enabled on remote config, to prevent it from getting called when the config is false.

PR #4521, opened 2024-05-16
fix: defaults intercom remote-config to false

reidzeibel:
Fixes the crash on OMS partners.

followup to https://github.com/AtB-AS/kundevendt/issues/2939 and PR https://github.com/AtB-AS/mittatb-app/pull/4515

The problem here is the Intercom code is called before the remote-config for intercom is fetched, because the default is `true`, the call will fail because Intercom is not initialized. 

In this PR, I changed the default to false, and then tested on both FRAM (without the intercom API keys), and AtB (with intercom API keys, and testing Intercom interface), everything works on debug, but will test in staging to confirm.

### DIFF
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -65,7 +65,7 @@ export type RemoteConfig = {
 
 export const defaultRemoteConfig: RemoteConfig = {
   enable_ticketing: !!JSON.parse(ENABLE_TICKETING || 'false'),
-  enable_intercom: true,
+  enable_intercom: false,
   feedback_questions: '',
   must_upgrade_ticketing: false,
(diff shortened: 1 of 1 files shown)


reidzeibel:
@mortennordseth here's the Intercom issue that needs to be merged for release :)

I have tested both on AtB and FRAM staging and both works as intended, but please do a doublecheck on this one: 1715870919

mortennordseth:
> @mortennordseth here's the Intercom issue that needs to be merged for release :)
> 
> I have tested both on AtB and FRAM staging and both works as intended, but please do a doublecheck on this one: 1715870919

Build 1715870919 works on my Android device 👍

marius-at-atb (approved):

PR #4523, opened 2024-05-18
fix: use "QR code" instead of "Barcode"

adriansberg:
Refactor to use the word "QR code" instead of "Barcode" since the norwegian translation used is not a valid term. Additionally, "QR code" has become a well-known term widely used, and currently a QR code is shown in the app. Aztec codes might be used in the future, which is in fact a barcode, but the common user might not see the difference.

Fixes https://github.com/AtB-AS/kundevendt/issues/18061


tft-kaare (approved):
Looks good.There are some language that could be cleaned up, but this is outside the scope of this

marius-at-atb (approved):

PR #4524, opened 2024-05-21
fix: Cache source maps

jorelosorio:
Related to https://github.com/AtB-AS/kundevendt/issues/16314

When the app is running and there is no build task due to a cache, then the upload of source maps fails because there are no .map files created during the build.

This PR creates a cache folder for the upload source map to be found if no build job was skipped.


reidzeibel (approved):
looks good! well done!

jorelosorio:
@rosvik I made new fixes and a new build without cache seems to work: https://github.com/AtB-AS/mittatb-app/actions/runs/9189385086/job/25271328714

rosvik (approved):

PR #4532, opened 2024-05-23
feat: update badge color and add badge to flex discount section

reidzeibel:

Contents of this PR:

- Add the "label" property to `ExpandableSectionItem`.
- Add "new"/"ny" label to Flex discount section.
- ~~Update the color of both "new"/"ny" and "beta" badge. See below for comparison. [Figma component](https://www.figma.com/design/2QTjAdekdIPuLFovQhVrY3/Components?node-id=8344-45773&t=XZMNfCJIx11Rs0LM-4)~~

Color change is reverted and will be done on separate PR.

Added badge:
image

### DIFF
--- a/src/components/label-info/LabelInfo.tsx
+++ b/src/components/label-info/LabelInfo.tsx
@@ -21,7 +21,7 @@ export const LabelInfo = ({label}: LabelInfoProps) => {
 
   return (
     <View style={linkSectionItemStyle.flag}>
-      <ThemeText color="background_accent_3" type="body__tertiary">
+      <ThemeText color="info" type="body__tertiary">
         {flagTranslated}
       </ThemeText>
     </View>
@@ -30,7 +30,7 @@ export const LabelInfo = ({label}: LabelInfoProps) => {
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   flag: {
-    backgroundColor: theme.static.background.background_accent_3.background,
+    backgroundColor: theme.static.status.info.background,
     marginRight: theme.spacings.medium,
     paddingHorizontal: theme.spacings.small,
     paddingVertical: theme.spacings.xSmall,
--- a/src/components/label-info/LabelInfo.tsx
+++ b/src/components/label-info/LabelInfo.tsx
@@ -21,7 +21,7 @@ export const LabelInfo = ({label}: LabelInfoProps) => {
 
   return (
     <View style={linkSectionItemStyle.flag}>
-      <ThemeText color="info" type="body__tertiary">
+      <ThemeText color="background_accent_3" type="body__tertiary">
         {flagTranslated}
       </ThemeText>
     </View>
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
```diff
@@ -35,7 +35,6 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
   const {appTexts} = useFirestoreConfiguration();
   const {flex_ticket_url} = useRemoteConfig();
 
-  if (!userProfiles.some((u) => u.offer.flex_discount_ladder)) return null;
```
jorelosorio, line 38: I guess this is intentional, but this seems to skip check either if a discount is applicable or have had discounts from before.

jorelosorio (approved):

src/components/label-info/LabelInfo.tsx
```diff
@@ -30,7 +30,7 @@ export const LabelInfo = ({label}: LabelInfoProps) => {
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   flag: {
-    backgroundColor: theme.static.status.info.background,
+    backgroundColor: theme.static.background.background_accent_3.background,
```
marius-at-atb, line 33: The color in Figma:
Skjermbilde 2024-05-23 kl  10 29 05
Is there a mismatch between Figma and design system?

reidzeibel: Yes, there's a bigger design system change that might need to be done soon:

https://github.com/AtB-AS/kundevendt/issues/15379
https://github.com/AtB-AS/kundevendt/issues/18138

I have a commit ready on design system, but I'm waiting until this Flexible ticket is done first, and then I can work on the design system issue. 

There's a type change for status icons, where the new design have Primary & Secondary colors instead of a single color, and we will need to change the usage after updating the new design system. 

Since I'm not sure if any of the OMS-partners wants to update the design system (Troms might need to update some illustrations, for example), I don't want to make the change until I can update the usage on the app repo.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
reidzeibel: Thank you, I missed this part! 😅

reidzeibel: It needs to be changed, but this shouldn't be in this commit, thanks for noticing it!

src/components/label-info/LabelInfo.tsx
marius-at-atb: I see. Perhaps it's best to hold back the color change until DS is ready?
Otherwise we should also check how this looks in the other OMS apps before merging

reidzeibel: I think that is a good idea, I'll discuss with Morten when I start working on the design system :)

reidzeibel: Reverted

gorandalum (approved):

marius-at-atb (approved):

PR #4533, opened 2024-05-23
test: accept travel time variations

tormoseng:
When assessing travel times for bike and foot, there are sometimes small variations in minutes. Added a check that accepts small variations in minutes.

Note! Want to have a check that takes an expected value to not succeed on large variations to the expected one.


gorandalum (approved):

tormoseng:
Tests OK: https://github.com/AtB-AS/mittatb-app/actions/runs/9208780696

PR #4535, opened 2024-05-24
feat: update flexible ticket discount dropdown layout

reidzeibel:

In this PR : 
1. Defaults the flexible discount dropdown to closed state.
2. Show original price when a discount is applied.
3. Hide the flex discount layout when there is no flex discounts available, but show them if there are flex discount available. (the new flex discount allows purchasing other types of user and the flex discount will apply to eligible passengers, f.eks: 1 voksen & 1 honnør, only the voksen gets flex discounted price, total price is discounted price for voksen + normal price for honnør) - not testable yet, the backend isn't ready.
4. ~~Shows multiple rows of passengers that gets a discount on discount information (f.eks 5 voksen = 5 rows of passengers on discount information)~~ **UPDATE**: based on latest design update, this change is reverted.
5. When multiple ticket with multiple passenger types are purchased, passenger types without flex discount will show up on the discount information, but will not have any discount. (cannot be tested during the creation of this PR, the backend isn't ready yet with the new system.)

I have commented some of the important part of the code that you might want to take a look at.

<details>
<summary>Reference screen</summary>

</details>

<details>
<summary>In this PR</summary>

https://github.com/AtB-AS/mittatb-app/assets/1777333/faa1229d-dd3e-4d7a-98fe-fec3d6193312

</details>

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
@@ -3,6 +3,7 @@ import {ThemeText} from '@atb/components/text';
 import {Linking, StyleProp, View, ViewStyle} from 'react-native';
 import {
   getTextForLanguage,
+  Language,
   PurchaseOverviewTexts,
   useTranslation,
 } from '@atb/translations';
@@ -22,21 +23,30 @@ import {
 } from '@atb/components/sections';
 import {ContentHeading} from '@atb/components/heading';
 import {BorderedInfoBox} from '@atb/components/bordered-info-box';
+import {Offer} from '@atb/ticketing';
 
 type Props = {
   userProfiles: UserProfileWithCountAndOffer[];
   style: StyleProp<ViewStyle>;
 };
 
+type FlexProfileList = {
+  userProfileKey: string;
+  userProfileName: string;
+  offer: Offer;
+};
+
 export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
   const {t, language} = useTranslation();
-  const [expanded, setExpanded] = useState(true);
+  const [expanded, setExpanded] = useState(false);
   const styles = useStyles();
   const {appTexts} = useFirestoreConfiguration();
   const {flex_ticket_url} = useRemoteConfig();
 
-  if (!userProfiles.some((u) => u.offer.flex_discount_ladder)) return null;
-  
+  const flexProfileList = mapToFlexProfileList(userProfiles, language);
+
+  if (!flexProfileList.every((u) => u.offer.flex_discount_ladder)) return null;
+
   const description =
     getTextForLanguage(appTexts?.discountInfo, language) ||
     t(PurchaseOverviewTexts.flexDiscount.description);
@@ -54,21 +64,24 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
         />
         {expanded && (
           <GenericSectionItem accessibility={{accessible: true}}>
-            <ThemeText>{description}</ThemeText>
+            <ThemeText type="body__secondary" color="secondary">
+              {description}
+            </ThemeText>
           </GenericSectionItem>
         )}
         {expanded &&
-          [...userProfiles].map((u) => {
+          flexProfileList.map((u) => {
             const ladder = u.offer.flex_discount_ladder;
             const discountPercent = ladder?.steps[ladder.current].discount;
-            if (!discountPercent) return null;
 
-            const userProfileName = getReferenceDataName(u, language);
-            const discountText = t(
-              PurchaseOverviewTexts.flexDiscount.discountPercentage(
-                discountPercent.toFixed(0),
-              ),
-            );
+            const userProfileName = u.userProfileName;
+            const discountText =
+              discountPercent &&
+              t(
+                PurchaseOverviewTexts.flexDiscount.discountPercentage(
+                  discountPercent.toFixed(0),
+                ),
+              );
             const priceText =
               formatDecimalNumber(
                 u.offer.prices[0].amount_float || 0,
@@ -82,7 +95,7 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
                   accessible: true,
                   accessibilityLabel: `${userProfileName}, ${discountText}, ${priceText}`,
                 }}
-                key={userProfileName}
+                key={u.userProfileKey}
               >
                 <View style={styles.userProfileDiscountInfo}>
                   <ThemeText type="body__secondary" color="secondary">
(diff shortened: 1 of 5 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
```diff
@@ -125,6 +138,24 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
   );
 };
 
+const mapToFlexProfileList = (
```
reidzeibel, line 141: Might be a better way to do this instead of doing `for-loop`, please let me know if there are better ways :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Summary.tsx
```diff
@@ -52,6 +56,15 @@ export function Summary({
           {t(PurchaseOverviewTexts.summary.price(formattedPrice))}
         </ThemeText>
       )}
+      {formattedOriginalPrice && (
```
reidzeibel, line 59: This part for showing original price, if it is available (only shown when price is discounted)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
```diff
@@ -93,11 +111,21 @@ const getOfferReducer =
...
+          userProfilesWithCounts,
+          action.offers,
+        );
+
         return {
           ...prevState,
           offerSearchTime: Date.now(),
           isSearchingOffer: false,
           validDurationSeconds: getValidDurationSeconds(action.offers?.[0]),
+          originalPrice: originalPrice !== totalPrice ? originalPrice : undefined,
```
reidzeibel, line 128: checks if `original price` is equal to `total price`, if it is equal, then there are no discounts, thus we leave it undefined. Otherwise, if the `original price` is different to `total price`, then show the original price.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
```diff
@@ -22,21 +23,30 @@ import {
...
+  const [expanded, setExpanded] = useState(false);
   const styles = useStyles();
   const {appTexts} = useFirestoreConfiguration();
   const {flex_ticket_url} = useRemoteConfig();
 
-  if (!userProfiles.some((u) => u.offer.flex_discount_ladder)) return null;
-  
+  const flexProfileList = mapToFlexProfileList(userProfiles, language);
+
+  if (!flexProfileList.every((u) => u.offer.flex_discount_ladder)) return null;
```
reidzeibel, line 48: if none of the passengers have `flex_discount_ladder`, don't show the layout.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
rosvik: Could be solved something like this

```ts
const flexProfileList: FlexProfileList[] = userProfileWithCounts.map(
  (userProfile) => ({
    userProfileKey: userProfile.id,
    userProfileName: getReferenceDataName(userProfile, language),
    offer: userProfile.offer,
  })
);
```

rosvik (commented):
Looks very good, but noticed a couple of details.

Design question @hannahATB: When you purchase several tickets for the same traveler category, do we really want to show a row for each? If so, I find it a bit confusing to say e.g. "Pr. Voksen" several times. I think it should say either "Pr. Voksen" once, or just "Voksen" several times.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
reidzeibel: This doesn't multiply the amount of user profile based on ticket count though, if we buy multiple adult ticket, there's only one user profile added to the list. We want to multiply the user profile by the count of ticket of that type. e.g., if there are 5 adult tickets purchased, we want 5 adult profiles.

reidzeibel:
@rosvik the response from back-end is `original_amount`, therefore I chose `originalPrice` to keep the naming consistent.

As for the view logic to show the original price, that does make sense, thanks!

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
reidzeibel: @rosvik the response from back-end is `original_amount`, therefore I named it `originalPrice` to keep the naming consistent.

As for the view logic to show the original price, that does make sense, thanks!

reidzeibel:
@rosvik I think the rows are meant to show the users, which ticket types get the discount, for example, if you buy 2 voksen & 3 barn, then it will show the user which ticket(s) get the discount. 

@hannahATB

hannahATB:
> Looks very good, but noticed a couple of details.
> 
> Design question @hannahATB: When you purchase several tickets for the same traveler category, do we really want to show a row for each? If so, I find it a bit confusing to say e.g. "Pr. Voksen" several times. I think it should say either "Pr. Voksen" once, or just "Voksen" several times.

That's a good point. It is enough to show one passenger category once in the summary. I will update the Figma file :))

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
reidzeibel: This one no longer needed after the design update.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
reidzeibel: updated

reidzeibel:
PR updated with the latest design update from @hannahATB

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
```diff
@@ -30,13 +30,14 @@ type Props = {
...
-  const [expanded, setExpanded] = useState(true);
+  const [expanded, setExpanded] = useState(false);
   const styles = useStyles();
   const {appTexts} = useFirestoreConfiguration();
   const {flex_ticket_url} = useRemoteConfig();
 
-  if (!userProfiles.some((u) => u.offer.flex_discount_ladder)) return null;
-  
+
+  if (!userProfiles.every((u) => u.offer.flex_discount_ladder)) return null;
```
marius-at-atb, line 39: Since there is no discount when purchasing for others, also return null if `isOnBehalfOfToggle` is `true`?

reidzeibel: On-behalf-of will be done in another PR, there's more to do than just hiding the flex discount information. Mainly I need to handle the display of price, for example, we have to show original price instead of discounted price when buying for others, also need to check how the behavior goes with the new system, etc. That's why I didn't include the reference screen about the on-behalf-of in this PR.

marius-at-atb (approved):
Looks good!
Do we have a feature flag for Flexible Ticket? If not, I think it should be added.

reidzeibel:
@marius-at-atb we have this flag ready (the old Flex ticket pilot), just need to update the condition about minimum version, and then we should be good to go. (so no users on the old app gets the discount)

image

rosvik (approved):
💯

PR #4536, opened 2024-05-24
feat: remove check for several carnets to show consume button

rosvik:
Since the back end now supports consume of carnets with more than one travel right, we can remove the check!


marius-at-atb (approved):
🥳

tormoseng:
- [x] iOS
- [ ] Android

PR #4541, opened 2024-05-27
feat: update purchase confirmation screen

reidzeibel:

in this PR: 

- Update the original price display style.
- Update the ticket name display.

There are some style discrepancies between the design on the Figma link below and other ticket design. For example, passenger font style on the design for flexible ticket is `body--primary`, but the other text - and on the app - is still using `body--secondary`, based on what @hannahATB said, we should still follow the existing styles. 

[Figma link](https://www.figma.com/design/zdZwvobgpEWSagKt0tderx/App?node-id=26058-20433&t=Y2xIKRVKWdo4xIuD-4)

<details>
<summary>Reference screen</summary>

</details>

<details>
<summary>Old-New comparison</summary>

</details>

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -26,7 +26,6 @@ import {
   ActivityIndicator,
   ScrollView,
   StyleProp,
-  Text,
   View,
   ViewStyle,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -319,7 +318,7 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
             <GenericSectionItem>
               <View accessible={true} style={styles.ticketInfoContainer}>
                 <ThemeText>
-                  {getReferenceDataName(preassignedFareProduct, language)}
+                  {getTextForLanguage(fareProductTypeConfig.name, language)}
```
rosvik, line 321: This is a bigger change that it might seem, and I don't think it's how we actually want it.

E.g. this will make 30-day periodic tickets no longer show "30-day ticket", but instead "periodic ticket, bus and tram".

reidzeibel: This one is not as important change, so we can revert it for now :)

reidzeibel: reverted

PR #4542, opened 2024-05-27
fix: use correct icon size in FareContractStatusSymbol

rosvik:
Noticed that none of the icons in FareContractStatusSymbol should have `size="large"`, according to [figma](https://www.figma.com/design/2QTjAdekdIPuLFovQhVrY3/Components?node-id=7415-26269) 🤓


gorandalum (approved):

HanneLH:
Tested ok! 🎉

PR #4544, opened 2024-05-28
feat: enforce app update for app versions below minAppVersion

hanne-at-atb:
* The user will be shown a force update screen when their app version is below the minAppVersion set in remote config
* The user will be asked to update the app and redirected to app store / play store. Added new urls fetched from Firestore `iosStoreListing` and `androidStoreListing` for store redirect
* If minAppVersion is not set, the screen will not show

Closes: https://github.com/AtB-AS/kundevendt/issues/15380


src/remote-config.ts
```diff
@@ -90,6 +91,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_car_sharing_in_map: false,
   enable_geofencing_zones: false,
   enable_vehicle_operator_logo: false,
+  minimum_app_version: '1.52.0',
```
marius-at-atb, line 94: It's best to set it `undefined` by default I think (although the current value also works)

src/translations/screens/ForceUpdate.ts
```diff
@@ -0,0 +1,21 @@
...
+  subText: _(
+    'Denne versjonen av appen er veldig utdatert og fungerer ikke lenger. Oppdater til en nyere versjon for å få tilgang igjen.',
+    'This version of the app is very outdated and is no longer working. Update to a newer version to regain access.',
+    'Denne versjonen av appen er veldig utdatert og fungerer ikkje lenger. Oppdater til ein nyare versjon for å få tilgang igjen.',
+  ),
+  externalButton: _('Oppdater appen', 'Update the app', 'Oppdater appen'),
+  errorMessage: _(
+    'Noe gikk galt. Vennligst prøv på nytt, eller gå inn i app-butikken på enheten din for å oppdatere.',
+    'Something went wrong. Please try again, or go to the store on your device to update the app.',
+    'Noko gikk galt. Ver venleg og prøv på nytt, eller gå inn i app-butikken på einheiten din for å oppdatere.',
```
marius-at-atb, line 18:
```suggestion
    'Noko gjekk gale. Ver venleg og prøv på nytt, eller gå inn i app-butikken på eininga di for å oppdatere.',
```

reidzeibel (approved):
Looks good!

src/force-update-screen/ForceUpdateScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+          </ThemeText>
+          <ThemeText style={styles.subText} color={themeColor}>
+            {t(ForceUpdateTexts.subText)}
+          </ThemeText>
+          <Button
+            rightIcon={{svg: ExternalLink}}
+            onPress={() => {
+              const link = Platform.select({
+                ios: `itms-apps://apps.apple.com/us/app/${IOS_STORE_ID}`,
+                android: `market://details?id=${ANDROID_STORE_ID}`,
```
marius-at-atb, line 49: These strings should ideally come from Firestore, as they could change by the time the feature is activated. E.g. Apple's url used to be of the format `https://itunes.apple.com/app/id0123456789`.
I think configuration->urls is a good place for it, and then it is also possible to adapt language, using `us` for `eng` and `no` for `nob` in Apple's url.

marius-at-atb (commented):
Very nice and clean!
See a couple of suggestions.

Also a design question - can we have a bit more "breathing space" between the elements here?

jorelosorio (approved):
Good work :)

src/force-update-screen/ForceUpdateScreen.tsx
jorelosorio: I suggest to keep itms-app since it calls directly the store app, https goes first to a web-browser then opens the app. about the lang, is not need to pass it since the app store opens in the user's default language. (iOS)

src/force-update-screen/__tests__/utils.test.ts
```diff
@@ -0,0 +1,65 @@
+import {isCurrentAppVersionLowerThanMinVersion} from './../utils.ts';
+
+describe('enforceUpdateOfAppVersion', () => {
```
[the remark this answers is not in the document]
hanne-at-atb: I did add these two: https://github.com/AtB-AS/mittatb-app/pull/4544/files#diff-74fc102f922adc750dc63cdb0d055759533aebd1e19a3c110f17c9add6cfe5fbR48-R53
But it was maybe not clear and could be grouped as their own `test("handle undefined"...`
What do you think @marius-at-atb ?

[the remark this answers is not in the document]
marius-at-atb: Somehow I didn't notice those! I guess undefined is a kind of "missing", so no need to separate it.
However `isCurrentAppVersionLowerThanMinVersion(undefined, undefined)` should be checked to be `false` as well I think.

src/force-update-screen/ForceUpdateScreen.tsx
marius-at-atb: I don't see the benefit in only keeping `itms-app` locally hard coded.
Regarding no/us, it is more of a nice to have :+1:

src/remote-config.ts
hanne-at-atb: It complained about setting it to undefined. So changed to empty string.

src/force-update-screen/ForceUpdateScreen.tsx
jorelosorio: `itms-apps` is defined as Apple URL Schemes, and they are only linked immediately to the app, it is probably that is not going to change. We might end up adding it to Firestore, but never change it again.

marius-at-atb: I agree that it is unlikely to change, and hopefully it won't. However I'm suggesting to keep the entire string `"itms-apps://apps.apple.com/us/app/${IOS_STORE_ID}"` as value in a url in Firestore, so it is no extra work to also include `itms-apps://`, and it gives flexibility if needed.

reidzeibel:
Just a random thought that popped up when looking at firestore today, why does the rule to check `appVersionMin` or `appVersionMax` on Global Messages doesn't use this logic?  

This is the check:
```
  if (appVersionMin && appVersionMin > APP_VERSION) return;
  if (appVersionMax && appVersionMax < APP_VERSION) return;
```

src/force-update-screen/utils.ts
rosvik, line 1: We already have a function that's very similar to this in `src/utils/compare-version.ts` (with tests!), so I think you could reuse it instead of adding this.

rosvik:
@reidzeibel 

> why does the rule to check appVersionMin or appVersionMax on Global Messages doesn't use this logic?

I don't think there's a good reason it doesn't. I guess the current check will work ok-ish until we reach v1.100, so we still have some time to fix it 🏃‍♂️

reidzeibel:
@rosvik yeah, I think we should refactor the rest of version checking to use proper logic 👍🏼

src/force-update-screen/ForceUpdateScreen.tsx
hanne-at-atb: I will move this to firestore :)

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -168,6 +170,16 @@ export const RootStack = () => {
     };
   }
 
+  const isCurrentAppVersionLowerThanMinVersion =
+    compareVersion(APP_VERSION, minimum_app_version) > 0;
+
+  if (
+    APP_VERSION &&
+    minimum_app_version &&
```
marius-at-atb, line 178: Tiny detail:
These two conditions could be moved inside `isCurrentAppVersionLowerThanMinVersion`, because the value of that variable isn't really valid unless we know the inputs are valid.

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -168,6 +170,16 @@ export const RootStack = () => {
     };
   }
 
+  const isCurrentAppVersionLowerThanMinVersion =
+    compareVersion(APP_VERSION, minimum_app_version) > 0;
```
marius-at-atb, line 174:
```suggestion
    compareVersion(APP_VERSION, minimum_app_version) < 0;
```
I think you meant this way 😄

marius-at-atb (changes requested):

src/stacks-hierarchy/RootStack.tsx
hanne-at-atb: Good point 🙈

src/stacks-hierarchy/RootStack.tsx
hanne-at-atb: Indeed, fixed now!

marius-at-atb (approved):
Great!

PR #4546, opened 2024-05-29
refactor: sort remote-config file alphabetically to improve legibility

hanne-at-atb:
I suggest sorting the remote-config file alphabetically to make it easier to read and get an overview when adding new parameters.

It's a suggestions so approve if you agree this could be a good idea! 🌟


rosvik (approved):
Good suggestion, this is much nicer! 👍

reidzeibel (approved):
🤩

tormoseng:
👍  all values are there, so seems OK!

PR #4547, opened 2024-05-29
feat: use non-blocking ActivityIndicator when fetching id token

rosvik:
This allows the user to click reload and access the rest of the profile page, while ID token is loading


marius-at-atb (approved):

reidzeibel (approved):

tormoseng:
This fix seems OK!

Tested 
- if you get an error back, the idToken retries (3 sec, 6 sec, etc) and the loading spinner disappear after 3 retries. Then, choosing "Retry", another round of idTokens come.
- if the response never comes, the idToken is retried after 1 min and the loading spinner disappear after 3 retries Then, choosing "Retry", another round of idTokens come.
- if you choose "Retry" while an idToken request is ongoing (and spinner exists), nothing happens.

PR #4551, opened 2024-05-30
feat: remove display of unknown fare contracts

rosvik:
We previously showed "unknown" fare contacts (fare contracts with a `type` we don't support) like this:

With this PR, I propose that instead we

- Remove the "unknown" validity status
- Remove the SimpleFareContract component, since it no longer have any functionality
- Rely on ErrorBoundary to catch errors related to fare contracts that we're not able to render properly. (E.g. it has no travel rights)

If something fails when rendering a fare contract, this will now be shown instead of the unknown farecontact component above:

<div>

</div>


reidzeibel (approved):
I might miss some react-specific stuff but the PR looks good overall and should work as intended!

HanneLH:
@rosvik 
Changed `type` below `tariffzoneRefs`, get the error message in ticket details, but nothing below my tickets and app crashes when click on expired tickets. 

<details>
<summary>screenshots/video</summary>

Android - ticket details:
Screenshot_20240531-121736_AtB (Staging)

iOS - My ticket
https://github.com/AtB-AS/mittatb-app/assets/104772494/c2baca7e-e98c-43ba-ae1b-a23c847741de

</details>

rosvik:
@HanneLH Ok, could you test if this is better or worse than the previous version? There might be issues elsewhere in the app that I haven't touched in this PR, so to not make this a larger task than it needs to, I think it's enough that it doesn't introduce _new_ issues.

HanneLH:
> @HanneLH Ok, could you test if this is better or worse than the previous version? There might be issues elsewhere in the app that I haven't touched in this PR, so to not make this a larger task than it needs to, I think it's enough that it doesn't introduce _new_ issues.

Older version of the app crashes when have change  `type` below `tariffzoneRefs`, so only different now after this PR it that you get the red message if you have ticket details open. So for My ticket page its the same behavior.

rosvik:
@HanneLH Thanks, I won't look more into that, then 👍 I'm also working on some bigger changes as part of https://github.com/AtB-AS/kundevendt/issues/15528, which might affect this.

PR #4556, opened 2024-06-03
Create useSnackbar

marius-at-atb:
An issue with the Snackbar was discovered - showing the exact same message two or more times in a row, is not possible. This is because the diff check for the textContent says no change.
To solve this, a unique key must be provided for each new textContent to display.

Since managing this key is likely to always be desirable, and since @gorandalum had [a good suggestion](https://github.com/AtB-AS/mittatb-app/pull/4534#pullrequestreview-2080538074) for a `useSnackbar` hook, I have added it here.

Example of how to use:
```ts
// init
const {snackbarProps, showSnackbar, hideSnackbar} = useSnackbar();

// show
const textContent = {title: 'My Title', description: 'Some description here'}
showSnackbar({textContent, position: 'top'});

// if force hiding is needed (otherwise it will hide by itself after the default duration)
hideSnackbar()

// add to DOM
<Snackbar {...snackbarProps} />
```

Resolves https://github.com/AtB-AS/kundevendt/issues/17411

### DIFF
--- a/src/components/snackbar/Snackbar.tsx
+++ b/src/components/snackbar/Snackbar.tsx
@@ -28,7 +28,7 @@ export type SnackbarTextContent = {
 
 export type SnackbarProps = {
   textContent?: SnackbarTextContent;
-  position: SnackbarPosition;
+  position?: SnackbarPosition;
   /** Optional action button, only shown if this is provided */
   actionButton?: ButtonProps;
   /** Whether to show the close x button */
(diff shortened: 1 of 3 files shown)


gorandalum (approved):
Is OK for a quick fix now. I think that if the original Snackbar was implemented to be used with a `show`-function, instead of `useEffect` on text content, then we wouldn't need this hackishy workaround.

tormoseng:
Tested OK with other Snackbar-fix. Ref https://github.com/AtB-AS/kundevendt/issues/17411#issuecomment-2154473987

PR #4559, opened 2024-06-04
fix: rejected reservations show on the correct screen

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17689

This PR will fix rejected reservations showing on both `Expired Tickets` and `Tickets sent to others`, by filtering the reservations based on the `customerAccountId`. Since we are removing the filtering of sent-to-other reservations, I also added filtering for subaccount purchases, so reservations for subaccount shouldn't be shown on the list.

### DIFF
--- a/src/fare-contracts/PurchaseReservation.tsx
+++ b/src/fare-contracts/PurchaseReservation.tsx
@@ -27,7 +27,7 @@ export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
       await Linking.openURL(vippsUrl);
     } catch (err: any) {
       Bugsnag.notify(err);
-    } 
+    }
   }
   const getStatus = () => {
     const paymentStatus = reservation.paymentStatus;
@@ -41,11 +41,13 @@ export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
     }
   };
 
+  const isSubAccountReservation = customerProfile?.subAccounts?.some(
+    (id) => id === reservation.customerAccountId,
+  );
+  
   // filter out reservations for subaccount
-  const subaccounts = customerProfile?.subAccounts;
-  if (subaccounts) {
-    const subaccountReservation = subaccounts.filter((subaccountId) => subaccountId === reservation.customerAccountId);
-    if (subaccountReservation.length > 0) return null;
+  if (isSubAccountReservation) {
+    return null;
   }
 
   const status = getStatus();
--- a/src/fare-contracts/PurchaseReservation.tsx
+++ b/src/fare-contracts/PurchaseReservation.tsx
@@ -10,7 +10,6 @@ import {ValidityLine} from './ValidityLine';
 import {FareContractStatusSymbol} from './components/FareContractStatusSymbol';
 import {formatToLongDateTime} from '@atb/utils/date';
 import {fromUnixTime} from 'date-fns';
-import {useAuthState} from '@atb/auth';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
 
 type Props = {
@@ -20,7 +19,6 @@ type Props = {
 export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
   const styles = useStyles();
   const {theme} = useTheme();
-  const {abtCustomerId} = useAuthState();
   const {t, language} = useTranslation();
 
   async function openVippsUrl(vippsUrl: string) {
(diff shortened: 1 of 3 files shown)


src/fare-contracts/PurchaseReservation.tsx
```diff
@@ -42,9 +40,6 @@ export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
     }
   };
 
-  // Filter out reservations for subaccounts
-  if (reservation.customerAccountId !== abtCustomerId) return null;
```
marius-at-atb, line 46: Originally this prevented reservations for subaccounts to be shown. After the changes introduced here, those reservations will now show in `sent` mode. Is this what's desired?

Observation: we have `fareContracts` and `sentFareContracts`, but no such distinction for `reservations` or `rejectedReservations`.

reidzeibel: Right now, with the new filtering, the subaccount reservations should be shown on tickets sent for others. Wouldn't you think it is enough? or should we separate it as well?

marius-at-atb: I think it's probably enough.
However @Sebstorvik @mjenssen you might have some thoughts about this:
- Should tickets bought for subaccounts be shown in ticket history sent to others? If so, should they look exactly the same as any other ticket bought on-behalf-of?
- Technically, should we also have `sentReservations` and `sentRejectedReservations`?

mjenssen: IMO, I think the purchase history should always be visible for the purchaser, no matter who gets the ticket in the end. The subaccount system was the initial on-behalf-of solution, so I think the faster we merge these two solutions the better.

As for splitting reservations, I think it's fairly easy to filter what's in that collection. Unless someone has a convincing argument for splitting, I propose we keep it as it is for now. (The `sentFareContracts` collection was added to prevent old apps from showing sent fare contracts in the main view.)

Long-term, we should look at data modeling of the whole reservation / orders / fare contracts, as well as how this could look in a non-Firestore world.

marius-at-atb: Sounds good to me, just want to be sure we are making conscious decisions here :+1:
The split only matters if we need the ability to separate tickets bought on-behalf-of and for a sub-account for some reason, but I guess there is no such reason as of today.

marius-at-atb (approved):
Lgtm!

reidzeibel:
@gorandalum I added the acceptance criteria on the original issue, could you take a look?

Also, the pending reservation for subaccounts will now show on the main screen, not sure if it is accepted or not, but in my opinon, it should show to allow the purchaser to check the reservation status. (can simulate pending reservation by selecting vipps as payment and delay the payment).

This is how it will look in the app:
image

The reservation there is from subaccount. Once it is paid/capture, it will be gone, but if it is rejected, then it will move to the `sent to others`.

What do you think @tormoseng ?

tormoseng:
@reidzeibel Was it decided to show the pending reservation for subaccounts? I can see your points, but since this subaccount concept is a "webshop thing", I don't see the benefit of showing the pending reservation in the app (also since it hasn't been there until now). I would guess not many would pay attention to the app when buying for a subaccount in the webshop 🤷  

But it's not something I'll fight hard for 😅

src/fare-contracts/PurchaseReservation.tsx
tormoseng: > The subaccount system was the initial on-behalf-of solution, so I think the faster we merge these two solutions the better.

I couldn't agree more. But that's for 2025, I guess..

reidzeibel:
I'll try filtering out subaccounts then :D

reidzeibel:
@gorandalum @tormoseng I added filtering for subaccount reservations, it shouldn't be shown on the list now. 👍🏼

src/fare-contracts/PurchaseReservation.tsx
```diff
@@ -40,6 +41,13 @@ export const PurchaseReservation: React.FC<Props> = ({reservation}) => {
     }
   };
 
+  // filter out reservations for subaccount
+  const subaccounts = customerProfile?.subAccounts;
+  if (subaccounts) {
+    const subaccountReservation = subaccounts.filter((subaccountId) => subaccountId === reservation.customerAccountId);
+    if (subaccountReservation.length > 0) return null;
+  }
+
```
gorandalum, line 50: Just a tip that `some` can be used in those situations where first `filter` is used and then checked if length is greater than 0.

```ts
if (customerProfile?.subAccounts?.some((id) => id === reservation.customerAccountId) {
  return null;
}
```

gorandalum (approved):
Looks good!

src/fare-contracts/PurchaseReservation.tsx
reidzeibel: Thanks, I will keep this in mind!

PR #4561, opened 2024-06-04
feat: support Snackbar position toggling

marius-at-atb:
During the testing of https://github.com/AtB-AS/mittatb-app/pull/4556, @tormoseng discovered that the `Snackbar` behaves badly when toggling `position` between `top` and `bottom`.

Rather than interfering with the position logic for a Snackbar instance that might be in the middle of a transition, two instances are used, where one is always hidden.

This allows the in-transition to start at the same time as the out-transition, and without any lag or otherwise unusual animation. If using only one element, it would have to wait until the first animation is done before animating in again to avoid a "hard jump", increasing complexity.

~~A performance optimisation here is that the extra `SnackbarInstance` is not created until needed, which for most use cases is never.~~ Edit: removed [for simplicity](https://github.com/AtB-AS/mittatb-app/pull/4561#discussion_r1627649879).

Another fix included in this PR ~~is to set opacity to 0 when it is hidden, originally suggested by @rosvik.~~ Edit: is to flip using top and bottom style.
This became necessary because of an edge case where changing textContent so that the Snackbar became taller, would cause it to be visible on the screen until it had moved to the new hiddenY. After the top/bottom flip, the abrupt change now happens outside of the screen.

Resolves https://github.com/AtB-AS/kundevendt/issues/17411


gorandalum (commented):
I took a look at Figma now, and it says that the Snackbar should always appear at the bottom. Could we scrap the positioning logic all together?

Screenshot 2024-06-05 at 09 18 51

marius-at-atb:
> I took a look at Figma now, and it says that the Snackbar should always appear at the bottom. Could we scrap the positioning logic all together?

@gorandalum No, that's outdated in the Figma file. It has to support both.
@Aliaaen remember to update it in Figma 😉

src/components/snackbar/Snackbar.tsx
```diff
@@ -37,19 +38,64 @@ export type SnackbarProps = {
...
+}) => {
+  // isDisabled means whether the SnackbarInstance should move to the hidden location
+  const isDisabled = snackbarProps.position !== position;
+  // isActivated means whether the SnackbarInstance component should exist
+  const [isActivated, setIsActivated] = useState(!isDisabled);
+  // the SnackbarInstance doesn't need to exist in its position until the position matches
+  // however don't revert it again, as it would ruin the exit animation
+  useEffect(() => {
+    if (!isDisabled) setIsActivated(true);
+  }, [isDisabled]);
```
gorandalum, line 68: Is this really needed? I suspect the performance necessary for the `useEffect` and extra render outweighs the performance benefit of this code.

gorandalum (approved):

src/components/snackbar/Snackbar.tsx
marius-at-atb: Not really needed, just me thinking it would improve performance.
So I can remove it if you think it's not worth including?

marius-at-atb: Removed here https://github.com/AtB-AS/mittatb-app/pull/4561/commits/0ad33acf777e4d35465d2d8354d65a653d233132

src/components/snackbar/use-snackbar-vertical-position-animation.ts
```diff
@@ -70,10 +73,27 @@ export const useSnackbarVerticalPositionAnimation = (
     viewHeightIncludingShadow,
   ]);
 
+  // when the Snackbar is disabled, and textContent changes to make the Snackbar taller,
+  // it takes a moment for it to move into the new hiddenY,
+  // however the Snackbar should not be visible during this transition.
+  // isHidden remains true during this transition.
+  const [isHidden, setIsHidden] = useState(!snackbarIsVisible);
```
gorandalum, line 80: Is this state necessary? Can't it just be derived from `snackbarIsVisible` and `isInHiddenY` during render?

marius-at-atb: Yes I think it's necessary. During the transition triggered by the height change while hidden, `snackbarIsVisible` is `false` and `isInHiddenY` is `false`. This happens to be exactly the same state as when it exits normally, in which case it should show..

marius-at-atb:
For future reference if simplification is desired:
It could be that the new `isHidden` can be used to remove the Snackbar's need for both `messageKey` and `useStablePreviousValue`.

marius-at-atb:
Update: now reverted back to before the isHidden stuff since the deduction of that state wasn't ideal. Then solved the problem by switching usage of top/bottom instead [here](https://github.com/AtB-AS/mittatb-app/pull/4561/commits/571a19b955c5cd346a3bdff4966ab2b7c0a56358).

src/components/snackbar/Snackbar.tsx
```diff
@@ -37,24 +38,46 @@ export type SnackbarProps = {
...
   isDismissable,
   customVisibleDurationMS,
-}: SnackbarProps) => {
+  isDisabled,
+}: SnackbarInstanceProps) => {
   const styles = useStyles();
   const {t} = useTranslation();
 
-  const stableTextContent = useStableValue(textContent); // avoid triggering useEffects if no text has been changed
+  const stableTextContent = useStableValue(textContent, isDisabled); // avoid triggering useEffects if no text has been changed
```
marius-at-atb, line 72: Tried replacing `!stableTextContent` with `(!stableTextContent || isDisabled)` instead, which seems to work, but caused a render "blink". Locking `textContent` while `isDisabled` doesn't have this problem.

src/utils/use-stable-value.ts
gorandalum, line 1: Just move this hook under /components/snackbar. With the `lockValue` I think it is pretty use-case specific.

src/utils/__tests__/use-stable-value.test.ts
```diff
@@ -56,6 +56,23 @@ describe('useStableValue', () => {
...
+      {
+        initialProps: initialValue,
+      },
+    );
+
+    rerender({a: 1}); // new object with the same content
+    expect(result.current).toEqual(initialValue);
+
+    rerender({a: 2}); // change content
+    expect(result.current).toEqual(initialValue); // but shouldnt change since lockValue is true
```
gorandalum, line 73: Remember to test the `lockValue` going back from `true` to `false`.

gorandalum (approved):

tormoseng:
Tested OK! Ref: https://github.com/AtB-AS/kundevendt/issues/17411#issuecomment-2154473987

PR #4564, opened 2024-06-05
feat: delete payment card on local storage when payment card is removed from profile

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17817

This PR will remove the card saved in local storage when it is removed from profile. Removing other cards won't affect the saved card. 

Note that removal in webshop isn't handled since it doesn't have access to local storage in the app. As discussed with @tormoseng


gorandalum (changes requested):
I think this is too superficial as it doesn't handle this requirement in the issue:
> Note! If stored payment cards are removed in the webshop, this should also be reflected in the app

I think the proper solution would be to fetch the stored cards from remote in the confirmation screen, which may be used to catch this situation where the card is deleted, also in webshop. With caching from the react-query library this shouldn't introduce much overhead.

reidzeibel:
@gorandalum I discussed about that possibility with @tormoseng and concluded that it is not needed now, but if you think that it's the better solution, then I'll close this PR and work on the new one.

gorandalum:
Yeah, I think that is the proper solution. Stored cards is a remote thing connected to the users account, and not validating the stored card on the device up against the actual stored cards online will at some point cause some issue.

PR #4566, opened 2024-06-06
refactor: remove the concept of "pre activated" travel rights

rosvik:
Excuse the large PR, but there were a few things that needed to happen at once to make this work. The high line count is mostly due to the addition of lots of test fixtures.

- Removes the concept of pre activated travel rights (previously this was just a arbitrary set of travelright types).
- Replaces it with "normal travel rights". This is any travel right that we assume has been populated with expected data. 
	- NOTE: We probably want more thorough validation of the input data here, but that should happen as we fetch them from firestore, and is an issue for another PR.
- Adds travel right tests for `isNormalTravelRight`, `isCarnetTravelRight` and `hasActiveTravelRight`
- Adds fare contract tests for `isCarnet`, `isActiveNowOrCanBeUsedFareContract`
- Moved some util functions from `fare-contracts/utils.ts` to `ticketing/utils.ts` to make the tests run.

closes https://github.com/AtB-AS/kundevendt/issues/15528


gorandalum (approved):
👏

PR #4569, opened 2024-06-07
feat: Only request value code onPress, handle error

marius-at-atb:
In order to not consume unnecessarily many value codes, hold back the request until the user actually presses the app switch button.
Also an error state management is in place now.

### Acceptance Criteria

- [ ] It should request the value code onPress of the AppSwitchButton, and immediately switch app if successful
- [ ] If there is an error, there should be an error message and the retry button should do exactly the same as if you press the AppSwitchButton

### DIFF
--- a/src/mobility/components/OperatorActionButton.tsx
+++ b/src/mobility/components/OperatorActionButton.tsx
@@ -1,7 +1,7 @@
 import {useAnalytics} from '@atb/analytics';
 import {getTextForLanguage, useTranslation} from '@atb/translations';
 import {useAppMissingAlert} from '@atb/mobility/use-app-missing-alert';
-import React from 'react';
+import React, {useCallback} from 'react';
 import {Button} from '@atb/components/button';
 import {MobilityTexts} from '@atb/translations/screens/subscreens/MobilityTexts';
 import {OperatorBenefitType} from '@atb-as/config-specs/lib/mobility-operators';
@@ -9,6 +9,7 @@ import {ExternalLink} from '@atb/assets/svg/mono-icons/navigation';
 import {ActivityIndicator, Linking} from 'react-native';
 import {useValueCodeQuery} from '@atb/mobility/queries/use-value-code-query';
 import {useIsEligibleForBenefit} from '@atb/mobility/use-is-eligible-for-benefit';
+import {MessageInfoBox} from '@atb/components/message-info-box';
 
 type OperatorActionButtonProps = {
   operatorId: string | undefined;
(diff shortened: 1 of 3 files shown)


reidzeibel (approved):
LGTM

src/mobility/queries/use-value-code-query.tsx
```diff
@@ -2,11 +2,14 @@ import {useQuery} from '@tanstack/react-query';
...
 
-export const useValueCodeQuery = (operatorId: string | undefined) => {
+export const useValueCodeQuery = (
+  operatorId: string | undefined,
+  enabled: boolean,
+) => {
   const {userId, authStatus} = useAuthState();
   return useQuery({
     queryKey: ['mobilityValueCode', userId, operatorId],
     queryFn: () => getValueCode(operatorId),
```
reidzeibel, line 12: @marius-at-atb just wondering, should we use `useMutation` here instead of `useQuery`, since it has a `mutationAsync` that can be used, instead of calling `refetch` on the `useQuery`.

marius-at-atb: Yes. Actually @gorandalum also suggested this before leaving for the day.

marius-at-atb: Update: https://github.com/AtB-AS/mittatb-app/pull/4569/commits/997bbfa20d53afcdca2c6f0495383aba1b0cdd24

src/mobility/components/OperatorActionButton.tsx
```diff
@@ -111,17 +112,39 @@ const OperatorActionButtonWithValueCode = ({
...
-  if (isLoadingValueCode) {
+  const {
+    mutate: fetchValueCode,
+    isLoading: isFetchingValueCode,
+    isError: isFetchingValueCodeError,
+  } = useValueCodeMutation(operatorId);
+
+  const appSwitchButtonOnPress = useCallback(() => {
+    fetchValueCode(undefined, {
+      onSuccess: (valueCode) => valueCode && buttonOnPress(valueCode),
```
reidzeibel, line 125: Not sure if using `onSuccess` is better than calling `mutateAsync()` and use `then`. 🤔 

My thought is, since a `Promise` with `then` is built-in JS, it is less prone to error or API changes, compared to using `onSuccess` call through the `useMutation()` API. 

What do you think?

marius-at-atb: Good thinking. Could be more readable as well I guess :+1:

marius-at-atb: https://github.com/AtB-AS/mittatb-app/pull/4569/commits/93bccf37542eaa2fcbfcd1f03fde46a72e46a289

tormoseng:
@marius-at-atb In staging, the endpoint `/mobility/code/YTR:Operator:trondheimbysykkel` gives `HTTP 404`. However, I don't see any error message. I only see a spinner, and that's it. See the screen recording below.

The same for Hyre.

<details>

https://github.com/AtB-AS/mittatb-app/assets/5812433/44705461-8ee0-44b7-9fc8-4d2003c3e7d7

</details>

marius-at-atb:
@tormoseng Good catch!
Fixed in https://github.com/AtB-AS/mittatb-app/commit/2174abfb1cc151ee978bac156f7286e3f35ddc04

tormoseng:
Re-tested OK!

Tested with:
- different benefit-ids
- different response codes and body
- city bikes and hyre
- different scenarios where the error code came on 1st, 2nd, 3rd, etc attempt.

PR #4572, opened 2024-06-10
REFERENCE DRAFT: geofencing zones

marius-at-atb:
This is the full implementation with geofencing zones.
The merging will be done with multiple PRs, this one is just for reference where one can see the big picture.


marius-at-atb:
Closing this since this has served its purpose as a reference draft

PR #4573, opened 2024-06-10
fix: Return error onError in getValueCode

marius-at-atb:
The error should be returned now that it is handled by the OperatorActionButtonWithValueCode component.

Fixes https://github.com/AtB-AS/mittatb-app/pull/4569#issuecomment-2158279620

### DIFF
--- a/src/mobility/api/api.ts
+++ b/src/mobility/api/api.ts
@@ -44,7 +44,6 @@ export const getValueCode = (
     )
     .then((response) => String(response.data.code))
     .catch((error) => {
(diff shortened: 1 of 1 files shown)


jorelosorio (approved):

rosvik (commented):
Do you need to have the `catch` block if it throws the error anyway? (I think it can just be removed)

marius-at-atb:
No, removing that as well :+1:

rosvik (approved):

PR #4574, opened 2024-06-10
feat: get and use geofencing zones

marius-at-atb:
Create hook `useGeofencingZonesQuery`, using `getGeofencingZones` and `GetGeofencingZonesQuery`.

This hook returns the geofencing zones data using the bff endpoint.

This is the first of a series of PRs on the way to support geofencing zones. You can take a look at https://github.com/AtB-AS/mittatb-app/pull/4572/files if you want to see how it is used.

### DIFF
--- a/src/mobility/queries/use-geofencing-zones.tsx
+++ b/src/mobility/queries/use-geofencing-zones.tsx
@@ -2,7 +2,7 @@ import {useQuery} from '@tanstack/react-query';
 import {getGeofencingZones} from '@atb/api/mobility';
 import {useGeofencingZonesEnabled} from '../use-geofencing-zones-enabled';
 
-const ONE_MINUTE = 1000 * 60;
+const TWELVE_HOURS_MS = 1000 * 60 * 60 * 12;
 
 export const useGeofencingZonesQuery = (systemIds: string[]) => {
   const [geofencingZonesEnabled, geofencingZonesEnabledDebugOverrideReady] =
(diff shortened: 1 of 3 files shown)


rosvik (approved):

reidzeibel (approved):

src/mobility/queries/use-geofencing-zones.tsx
```diff
@@ -0,0 +1,20 @@
+import {useQuery} from '@tanstack/react-query';
+import {getGeofencingZones} from '@atb/api/mobility';
+import {useGeofencingZonesEnabled} from '../use-geofencing-zones-enabled';
+
+const ONE_MINUTE = 1000 * 60;
```
gorandalum, line 5: Could we cache this longer? I'm guessing geofencing zones doesn't change that often?

src/mobility/queries/use-geofencing-zones.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const useGeofencingZonesQuery = (systemIds: string[]) => {
```
gorandalum, line 7: Is there ever a use case for multiple system ids? I can't wrap my head around situations where we present geofencing zones for multiple providers at the same time, without it being messy for the user.

src/mobility/queries/use-geofencing-zones.tsx
marius-at-atb: Yes, how long would you suggest?

src/mobility/queries/use-geofencing-zones.tsx
marius-at-atb: Not really, this was only used because the mobility api takes in an array. We can change it to just `systemId` here if you want?

src/mobility/queries/use-geofencing-zones.tsx
gorandalum: 12 hours?

src/mobility/queries/use-geofencing-zones.tsx
gorandalum: Yeah, I think that is better. Also works better with the caching, since we don't have array in the query key.

PR #4576, opened 2024-06-10
feat: geofencing zone explanations

marius-at-atb:
As can be read in the [new docs](https://github.com/AtB-AS/docs-private/pull/53), there are 4 types of geofencingZone features: `Allowed`, `Slow`, `NoParking` and `NoEntry`.
This adds an explanation for each of them.

Also explanations are added for `Unspecified`, for the case when a user presses on a location without a feature.
`GeofencingZoneExtraExplanations` is an extra explanation that is added if `isStation` is set to `true` for that feature. (The rule for `isStation` is also described in the docs).

<table>
  <tr>
    <td></td>
    <td></td>
     <td></td>
      <td></td>
      <td></td>
       <td></td>
  </tr>
  <tr>
    <td>Allowed</td>
    <td>Slow</td>
    <td>No Parking, isStation=true</td>
    <td>No Parking, isStation=false</td>
    <td>No Entry</td>
    <td>Unspecified</td>
  </tr>
</table>

### DIFF
--- a/src/components/map/types.ts
+++ b/src/components/map/types.ts
@@ -20,6 +20,9 @@ import {z} from 'zod';
 import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
 import {Line} from '@atb/api/types/trips';
 
+import {TranslatedString} from '@atb/translations';
+import {GeofencingZoneKeys} from '@atb-as/theme';
+
 /**
  * MapSelectionMode: Parameter to decide how on-select/ on-click on the map
  * should behave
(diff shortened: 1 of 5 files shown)


src/components/map/types.ts
```diff
@@ -182,3 +185,12 @@ export type ParkingType = {
   parkingVehicleTypes: ParkingVehicleTypes;
   totalCapacity: number;
 };
+
+type GeofencingZoneExplanationType = {
+  title: TranslatedString;
+  description: TranslatedString;
+};
+
+export type GeofencingZoneExplanationsType = {
```
marius-at-atb, line 194: To be even more precise, this could be called `GeofencingZoneFeatureExplanationsType`, but it feels long enough as is 😅

gorandalum (approved):

src/components/map/types.ts
rosvik: I agree, long enough as is 😅

rosvik (approved):

PR #4577, opened 2024-06-11
feat: Use 'telefon' instead of 'mobil' in Svipper/Troms

tfk-kaare:
Our marketing team insist on the use of 'telefon' in norwegian instead of 'mobil'.

### DIFF
--- a/src/translations/screens/Profile.ts
+++ b/src/translations/screens/Profile.ts
@@ -442,8 +442,8 @@ export default orgSpecificTranslations(ProfileTexts, {
           travelToken: {
             label: _(
               'Bruk billett på reisekort / telefon',
-              'Use ticket on reisekort / telefon',
-              'Bruk billett på reisekort / relefon',
+              'Use ticket on travel card / phone',
+              'Bruk billett på reisekort / telefon',
             ),
             labelWithoutTravelcard: _(
               'Bruk billett på telefon',
--- a/src/translations/screens/Profile.ts
+++ b/src/translations/screens/Profile.ts
@@ -439,6 +439,18 @@ export default orgSpecificTranslations(ProfileTexts, {
               'Standard billettkategori',
             ),
           },
+          travelToken: {
+            label: _(
+              'Bruk billett på reisekort / telefon',
+              'Use ticket on reisekort / telefon',
+              'Bruk billett på reisekort / relefon',
+            ),
+            labelWithoutTravelcard: _(
+              'Bruk billett på telefon',
+              'Use ticket on phone',
+              'Bruk billett på telefon',
+            ),
+          },
           notifications: {
             pushToggle: {
               subText: _(
--- a/src/translations/screens/subscreens/TravelToken.ts
+++ b/src/translations/screens/subscreens/TravelToken.ts
@@ -710,9 +710,9 @@ export default orgSpecificTranslations(SelectTravelTokenTexts, {
     },
     toggleToken: {
       title: _(
-        'Bytt mellom reisekort/telefon',
-        'Switch between travel card/phone',
-        'Byt mellom reisekort/telefon',
+        'Bytt mellom reisekort / telefon',
+        'Switch between travel card / phone',
+        'Byt mellom reisekort / telefon',
       ),
       radioBox: {
         tCard: {
--- a/src/translations/screens/subscreens/TravelToken.ts
+++ b/src/translations/screens/subscreens/TravelToken.ts
@@ -495,7 +495,7 @@ export default orgSpecificTranslations(SelectTravelTokenTexts, {
             'Kan eg reise utan mobil?',
           ),
           answer: _(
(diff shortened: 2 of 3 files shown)


src/translations/screens/subscreens/TravelToken.ts
```diff
@@ -495,7 +495,7 @@ export default orgSpecificTranslations(SelectTravelTokenTexts, {
             'Kan eg reise utan mobil?',
           ),
           answer: _(
-            'Du må alltid ha med deg det mobilen du har valgt å bruke billetten på.',
+            'Du må alltid ha med deg den mobilen du har valgt å bruke billetten på.',
```
tft-abel, line 498:
```suggestion
            'Du må alltid ha med deg den telefonen du har valgt å bruke billetten på.',
```

src/translations/screens/Profile.ts
```diff
@@ -439,6 +439,18 @@ export default orgSpecificTranslations(ProfileTexts, {
               'Standard billettkategori',
             ),
           },
+          travelToken: {
+            label: _(
+              'Bruk billett på reisekort / telefon',
+              'Use ticket on reisekort / telefon',
```
tft-abel, line 445: Should we translate "reisekort"?

```suggestion
              'Use ticket on travel card / phone',
```

src/translations/screens/subscreens/TravelToken.ts
```diff
@@ -563,34 +563,44 @@ export default orgSpecificTranslations(SelectTravelTokenTexts, {
     travelToken: {
       header: {
         title: _(
-          'Bruk billett på reisekort / mobil',
+          'Bruk billett på reisekort / telefon',
           'Use ticket on travel card / phone',
-          'Bruk billett på reisekort / mobil',
+          'Bruk billett på reisekort / telefon',
```
tft-abel, line 568: Same as above
```suggestion
          'Bruk billett på travel card / phone',
```

src/translations/screens/subscreens/TravelToken.ts
```diff
@@ -700,9 +710,9 @@ export default orgSpecificTranslations(SelectTravelTokenTexts, {
     },
     toggleToken: {
       title: _(
-        'Bytt mellom reisekort / mobil',
-        'Switch between travel card / phone',
-        'Byt mellom reisekort / mobil',
+        'Bytt mellom reisekort/telefon',
+        'Switch between travel card/phone',
+        'Byt mellom reisekort/telefon',
```
tft-abel, line 715: In other places, we have a space between the /

```suggestion
        'Bytt mellom reisekort / telefon',
        'Switch between travel card / phone',
        'Byt mellom reisekort / telefon',
```

src/translations/screens/subscreens/TravelToken.ts
tfk-kaare: Actually, this is an outside context fix on behalf of 'fram'. It's grammar correction and they should use 'mobilen' :-) If anything, this change should be left out.

src/translations/screens/Profile.ts
tfk-kaare: Good catch. Fixed.

src/translations/screens/subscreens/TravelToken.ts
tfk-kaare: Yeh.. This is an overwhelming issue/bug in the entire translation tree. Språkrådet says there should be no [space between single words in listing of single word alternatives](https://sprakradet.no/godt-og-korrekt-sprak/rettskriving-og-grammatikk/tegn/skrastrek/). Though, to have this typo being consistent, I'll 'fix' this

gorandalum:
Isn't this something the marketing teams in the OMS can sit down together and agree on?

I am pretty certain that this will at some point cause problems, since there will probably be situations going forward places where we write "mobilen" while forgetting that Troms should use "telefonen".

tfk-kaare:
I whole-heartedly agree. I see this translation as a short-term fix to please our customer base, but I think the society is overall aligned with the use of 'telefon' to mean 'mobiltelefon' or 'smarttelefon'.

So.

Let's just create a new ticket to raise this suggestion to the entire OMS?

Aliaaen:
We should of course strive to use consistent language and wording, so using the same word everywhere and for everyone sounds nice :) However, I think that these words are pretty much interchangeable at this point, so I'm curious as to why your marketing department is so adamant about it? What is their reasoning?

tfk-kaare:
I suspect our marketers reasons are along these lines:
1) The difference between 'telefon' to mean house phone has completely faded out, as everyone now has mobile phones. So there is no longer need to differentiate these two.
2) Even though the terms are interchangeable to mean the same, I suspect there is a shift/lean towards the use of 'telefon'
3) Consistency. Our language teacher always said that you are free to chose the word/spelling you'd like when there was options, but stick to it.

PR #4578, opened 2024-06-11
feat: use direction config to show boat validity texts

rosvik:
Implementation of config from https://github.com/AtB-AS/fare-product-type-configs/pull/50


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -246,30 +246,10 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
-              t(
-                PurchaseConfirmationTexts.validityTexts.harbor
-                  .messageInHarborZones,
-              ),
-            )}
-          </>
+        return summary(
+          t(
+            PurchaseConfirmationTexts.validityTexts.harbor.messageInHarborZones,
+          ),
```
rosvik, line 252: This message (Which says "Applies for bus/tram in departure and destination zones") can probably be a global message. But that's outside the scope of this PR.

PR #4582, opened 2024-06-11
feat: geofencing zones shape source

marius-at-atb:
This is the part of the map that actually renders the features in the geofencingZones data.

Depends on https://github.com/AtB-AS/mittatb-app/pull/4580
Changelog for this PR after that one has been merged [here](https://github.com/AtB-AS/mittatb-app/pull/4582/files/9449c60d1af9dcf5b26a556c8882515df4603e9b..0e7c79790ea983422a4f6095c7b2e10f46eca81e) (filtered by commits)


src/components/map/components/mobility/GeofencingZones.tsx
```diff
@@ -0,0 +1,83 @@
...
+              lineOpacity,
+            }}
+            aboveLayerID="water-point-label"
+          />
+        </MapboxGL.ShapeSource>
+      );
+    },
+  );
+
+  return <>{mappedGeofencingZones}</>;
```
rosvik, line 82: I think it would be more clean to make a `GeofencingZone` component here and return 

```tsx
return (<>
	{preProcessedGeofencingZones.map(geofencingZone =>
		<GeofencingZone geofencingZone={geofencingZone} />
	)}
</>);
```

Not sure where I heard it, but I think React is better at making optimizations if it's done like that as well.

marius-at-atb: Agreed, updated :+1:

rosvik (approved):

marius-at-atb:
Final adjustment: https://github.com/AtB-AS/mittatb-app/pull/4582/commits/43e2942fd7de0e1858a71803dcfcb1b13edc6585
This avoids some react query warnings

src/components/map/components/mobility/GeofencingZones.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const {
+    data: vehicle,
+    isLoading,
+    isError,
+  } = useVehicleQuery(selectedVehicleId);
+
+  if (!vehicle || !selectedVehicleId || isLoading || isError) {
+    return <></>;
+  }
+  return <GeofencingZonesForVehicle vehicle={vehicle} />;
```
gorandalum, line 27: Just a FYI about react-query: When using `isSuccess`, then you also get type safety that `data` is present. So the check `!vehicle || !selectedVehicleId || isLoading || isError` is a little more complicated than it needs to be.

I think this would do it:
```tsx
const vehiclesQuery = useVehicleQuery(selectedVehicleId);

return vehiclesQuery.isSuccess
  ? <GeofencingZonesForVehicle vehicle={vehiclesQuery.data} />
  : null;
```

marius-at-atb: You would think so, but this still gives "Type 'VehicleExtendedFragment | undefined' is not assignable to type 'VehicleExtendedFragment'."

gorandalum: Then the success type of `useVehicleQuery` can be `undefined`, which is [not allowed by react-query](https://tanstack.com/query/v4/docs/framework/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries) and will give runtime error. You can take a quick look into that.

PR #4584, opened 2024-06-12
fix: Not resubscribe auth on enable_intercom change

gorandalum:
### Background
Since we now check for enable_intercom before updating metadata,
the enable_intercom value was in the dependency list for the
useEffect setting up the auth description. On app startup the
enable_intercom value is initially false, before chaning to true
after RemoteConfig loads, which ends up triggering an unneccesary
auth resubscribe at startup.

### Solution 
Moved the updating of Intercom metadata outside of the auth
subscription to mitigate this.

### Acceptance criteria
- [ ] The breadcrumb "Subscribing to auth user changes" isn't logged twice at app startup (check on both existing app install and fresh app install)
- [ ] The fields 'AtB-Firebase-Auth-Id' and 'AtB-Auth-Type' in Intercom is updated after user change.


tormoseng (approved):
LGTM!

jorelosorio (approved):

tormoseng:
👍

PR #4586, opened 2024-06-13
feat: add identity service SMS login

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/18127

### DIFF
--- a/src/auth/use-is-backend-sms-auth-enabled.tsx
+++ b/src/auth/use-is-backend-sms-auth-enabled.tsx
@@ -4,12 +4,8 @@ import {StorageModelKeysEnum} from '@atb/storage';
 
 export const useIsBackendSmsAuthEnabled = () => {
   const {enable_backend_sms_auth} = useRemoteConfig();
-  const [debugOverride, _, debugOverrideReady] =
-  useBackendSmsAuthEnabledDebugOverride();
-  return [
-    debugOverride !== undefined ? debugOverride : enable_backend_sms_auth,
-    debugOverrideReady,
-  ];
+  const [debugOverride] = useBackendSmsAuthEnabledDebugOverride();
+  return debugOverride !== undefined ? debugOverride : enable_backend_sms_auth;
 };
 
 export const useBackendSmsAuthEnabledDebugOverride = () => {
--- a/src/api/identity.ts
+++ b/src/api/identity.ts
@@ -11,20 +11,26 @@ export const authenticateWithSms = async (
   });
 };
 
+/**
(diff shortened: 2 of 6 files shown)


reidzeibel:
@rosvik does it make sense to use react-query mutation for the login, or should we do it in the next iteration?

rosvik:
@reidzeibel The API calls here are triggered by clicking submit buttons only, so there's no need for dependency arrays and the other features useQuery provides. Just unnecessary complexity.

gorandalum:
I would prefer if it was using mutations, but as we want this out this week I think this is ok and we don't invest time on refactoring it now.

In general I think using react-query mutations gives code which is easier to read and is more maintainable. So on my part I don't think using react-query is adding complexity, it is more reducing it since part of the complexity is handled by react-query. Also note that mutations are not triggered by dependency arrays like queries are, so triggering them by user actions like a submit button is the way they are ment to be used.

src/auth/use-is-backend-sms-auth-enabled.tsx
```diff
@@ -4,12 +4,8 @@ import {StorageModelKeysEnum} from '@atb/storage';
...
 export const useIsBackendSmsAuthEnabled = () => {
   const {enable_backend_sms_auth} = useRemoteConfig();
-  const [debugOverride, _, debugOverrideReady] =
-  useBackendSmsAuthEnabledDebugOverride();
-  return [
-    debugOverride !== undefined ? debugOverride : enable_backend_sms_auth,
-    debugOverrideReady,
-  ];
+  const [debugOverride] = useBackendSmsAuthEnabledDebugOverride();
+  return debugOverride !== undefined ? debugOverride : enable_backend_sms_auth;
```
gorandalum, line 8: Should we have something in Firestore also for the feature-toggle? If we at some point see there is something really wrong with the new login flow, we probably want to be able to toggle it instantly, not letting the user wait up to 6 hours?

gorandalum (approved):

src/auth/use-is-backend-sms-auth-enabled.tsx
rosvik: Talked about it in the office now, and we concluded that a 6 hour delay for this is fine. We have Vipps as alternative and global messages for info if needed.

rosvik:
@gorandalum Ah, right. I'm mixing up useQuery and useMutation here. Still not sure what value a useMutation would add here, but maybe a good idea to look into. Keeping as is for now, so we can start testing this.

PR #4589, opened 2024-06-18
feat: extend travel search to also use id-based search when the location category is group_of_stop_places

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/18317

Some travel searches give different results in the app compared to the travel planner web. For instance, when searching for a trip from Geiranger to Dalsnibba in the travel planner web, it shows a direct bus trip. However, when performing the same search in app, this trip is displayed as a longer and unnecessary trip, where the final leg is the same as displayed in the travel planner web. 

I suggest that we first investigate the reason behind the difference in trip results between the app and the planner web. Once we understand the cause, we can discuss possible solutions to address this issue.

### How to replicate:
Search after Geiranger - Dalsnibba in both the app and the travel planner to view the difference. 
  
### Illustration
<details>
<Summary>screenshots/before</Summary>

| Travel planner web     | App |
| ----------- | ----------- |
| Image | Image |
| Image |  Image|

</details>

<details>
<Summary>screenshots/after</Summary>

| App | App |
| ----------- | ----------- |
| Screenshot_20240618_104908 | Screenshot_20240618_105027 |

</details>

### Proposed solution

Extend createQuery() to also use id-based search when the location category is GroupOfStopPlaces

### Test criteria 
- [ ] The app should provide the same search results as Entur and Travel Planner web


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
```diff
@@ -123,3 +123,14 @@ export const areDefaultFiltersSelected = (
   if (!transportModes || transportModes.length === 0) return false;
   return transportModes.every((tm) => tm.selectedAsDefault === tm.selected);
 };
+
+export const isVenue = (location: Location): boolean => {
+  return location.resultType === 'search' && location.layer === 'venue';
```
jorelosorio, line 128: Result type might not be related to a venue, perhaps only `location.layer === 'venue'` must be in the utility function.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
```diff
@@ -123,3 +123,14 @@ export const areDefaultFiltersSelected = (
...
+};
+
+export const isGroupOfStopPlaces = (location: Location): boolean => {
+  return (
+    location.resultType === 'search' &&
+    location.category[0] === 'GroupOfStopPlaces'
```
jorelosorio, line 134: Consider to have some checks done before accessing the first category object.

marius-at-atb, line 134: `'GroupOfStopPlaces'` should be replaced by the enum `GROUP_OF_STOP_PLACES` directly

marius-at-atb (commented):
After simplifying `isVenue` and `isGroupOfStopPlaces` as noted by @jorelosorio, I think we also no longer need the tests, as it is will essentially be just an if condition.

Perhaps it is better to create a function like
```ts
const getSearchPlace = (location: Location) =>
  location.resultType === 'search' &&
  (isVenue(location) || isGroupOfStopPlaces(location))
    ? location.id
    : undefined;
```
And reuse for both `from` an `to`

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
jonasbrunvoll: Will do👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
jonasbrunvoll: Okey, will look at it👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
jonasbrunvoll: Ok, will fix👍

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
```diff
@@ -123,3 +118,21 @@ export const areDefaultFiltersSelected = (
...
+  return (
+    !!location.category?.length &&
+    location.category[0] === FeatureCategory.GROUP_OF_STOP_PLACES
+  );
+};
+
+export const getSearchPlace = (location: Location) =>
+  location.resultType === 'search' &&
+  (isVenue(location as SearchLocation) ||
+    isGroupOfStopPlaces(location as SearchLocation))
```
marius-at-atb, line 136: We usually want to avoid using `as`.
What you can do instead, is let `isVenue` and `isGroupOfStopPlaces` also accept `Location`, and then check if the location has the property you're looking for.

```ts
const isVenue = (location: Location): boolean => {
  return 'layer' in location && location.layer === 'venue';
};
```
And then the same for `isGroupOfStopPlaces`, but looking for `'category'` instead of `'layer'`

jonasbrunvoll: Nice to know. I´ll update the code👍

marius-at-atb (approved):
Nice!

PR #4592, opened 2024-06-18
refactor: create hook for updating Intercom Metadata

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/18137

Converts the metadata update into a react hook, so that we don't need to check for `enable_intercom` flag everytime we want to update user metadata. The check is now built into the hook itself, so we won't miss it whenever we need to update the metadata.


src/GeolocationContext.tsx
```diff
@@ -258,14 +257,12 @@ export const GeolocationContextProvider: React.FC = ({children}) => {
...
-          if (enable_intercom) {
-            await updateChatUserMetadata({'AtB-App-Location-Status': status});
-          }
+          updateMetadata({'AtB-App-Location-Status': status});
         }
       }
     }
     checkPermission();
-  }, [appStatus, state.status, enable_intercom]);
+  }, [appStatus, state.status, updateMetadata]);
```
jorelosorio, line 265: Looks good in general, my only concern might be to have this function in the dependencies, it might create some complications

reidzeibel: any suggestion how to deal with that dependency? 🤔

jorelosorio: I remember @gorandalum and I did a trick to remove it from dependencies, perhaps hr could refresh my memory here or if it is fine to do in that way.

gorandalum: Depends on what we want. Do we want the `checkPermission` function to rerun when the `enable_intercom`/`updateMetadata` is changed? Should some effects run when it is changed, and some not? Deciding this will a hint on the best way forward.

In general if we wan't to use something, but it shouldn't trigger rerenders, then it should be refs instead of state.

src/chat/metadata.ts
```diff
@@ -20,12 +22,21 @@ type Metadata = {
   'AtB-Beta-TripSearch': string;
 };
 
-export async function updateMetadata(metadata: Partial<Metadata>) {
-  const custom_attributes = pickBy(metadata, (v) => v !== undefined) as {
-    [key: string]: string;
-  };
+export const useIntercomMetadata = () => {
```
marius-at-atb, line 25: The file name should also be updated to `use-intercom-metadata.ts`

src/chat/user.ts
```diff
@@ -3,24 +3,25 @@ import {Dimensions, PixelRatio, Platform} from 'react-native';
...
   try {
     await Intercom.loginUnidentifiedUser();
-  } catch (error: any){
+  } catch (error: any) {
     // do nothing
   }
   await Intercom.setBottomPadding(Platform.OS === 'ios' ? 40 : 80);
+}
 
+export async function setIntercomUserData() {
```
marius-at-atb, line 16: Should be called `getIntercomUserData`

marius-at-atb (approved):
Nice!
Btw something to also consider for a future issue, is to remove the AtB-specific naming of all the data props.

reidzeibel:
@gorandalum I made the change as suggested, ready for another review.

jorelosorio (approved):
I guess it was ok to keep functions inside the useEffect deps.

gorandalum (approved):

PR #4594, opened 2024-06-20
feat: add ticket price to ticket detail screen

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17139

Should be a pretty straightforward PR, just adding the price from fareContract firestore to show on the app.

### DIFF
--- a/src/fare-contracts/details/OrderDetails.tsx
+++ b/src/fare-contracts/details/OrderDetails.tsx
@@ -16,11 +16,9 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
     FareContractTexts.details.orderId(fareContract.orderId),
   );
   const firstTravelRight = fareContract.travelRights[0];
-  const priceString = formatDecimalNumber(
-    parseFloat(fareContract.totalAmount),
-    language,
-    2,
-  );
+  const priceString =
+    fareContract.totalAmount !== '' &&
+    formatDecimalNumber(parseFloat(fareContract.totalAmount), language, 2);
   return (
     <View accessible={true}>
       <ThemeText type="body__secondary" color="secondary">
@@ -67,13 +65,15 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
         {t(FareContractTexts.details.paymentMethod)}
         {_.capitalize(fareContract?.paymentType?.join(', '))}
       </ThemeText>
-      <ThemeText
-        type="body__secondary"
-        color="secondary"
-        style={style.marginTop}
-      >
-        {t(FareContractTexts.details.totalPrice(priceString))}
-      </ThemeText>
+      {priceString && (
+        <ThemeText
+          type="body__secondary"
+          color="secondary"
+          style={style.marginTop}
+        >
+          {t(FareContractTexts.details.totalPrice(priceString))}
+        </ThemeText>
+      )}
       <ThemeText style={style.marginTop}>{orderIdText}</ThemeText>
     </View>
   );
--- a/src/fare-contracts/details/OrderDetails.tsx
+++ b/src/fare-contracts/details/OrderDetails.tsx
@@ -7,6 +7,7 @@ import {fromUnixTime} from 'date-fns';
 import _ from 'lodash';
 import React from 'react';
 import {StyleSheet} from '@atb/theme';
+import {formatDecimalNumber} from '@atb/utils/numbers';
 
 export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
   const style = useStyles();
@@ -15,6 +16,11 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
     FareContractTexts.details.orderId(fareContract.orderId),
   );
   const firstTravelRight = fareContract.travelRights[0];
+  const priceString = formatDecimalNumber(
+    parseFloat(fareContract.totalAmount),
+    language,
+    2,
+  );
   return (
     <View accessible={true}>
       <ThemeText type="body__secondary" color="secondary">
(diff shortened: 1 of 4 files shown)


src/ticketing/types.ts
```diff
@@ -70,6 +70,7 @@ export type FareContract = {
   minimumSecurityLevel: number;
   travelRights: TravelRight[];
   qrCode: string;
+  totalAmount: string;
```
rosvik, line 73: Do we know if `totalAmount` can be undefined in some cases?

Nisha1306: Dont think so. I checked some of my Unknown tickets also that surprises all the time, it's an empty string.

image

reidzeibel: then it looks like I will need to guard against empty strings, thanks @Nisha1306 !

marius-at-atb: Historically, is there a reason why we don't have these types in config-specs? @gorandalum

src/fare-contracts/details/OrderDetails.tsx
```diff
@@ -61,6 +67,13 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
         {t(FareContractTexts.details.paymentMethod)}
         {_.capitalize(fareContract?.paymentType?.join(', '))}
       </ThemeText>
+      <ThemeText
+        type="body__secondary"
+        color="secondary"
+        style={style.marginTop}
+      >
+        {t(FareContractTexts.details.totalPrice(priceString))}
```
jorelosorio, line 75: In case of empty prices, it is better to show an error message here? Or it is ok to ignore it?

reidzeibel: That's what @Sebstorvik might want to look at, but I think for the current iteration, we could write "0" if it is empty string.

rosvik: If it's an empty string, I think we should hide the text entirely. E.g. I don't think it makes sense to show anything about price for school tickets, since the student didn't purchase anything.

reidzeibel: @rosvik makes sense, I'll update it

src/ticketing/types.ts
gorandalum: The reason is that config-specs originally was specs for our configuration, and fare contracts aren't part of the configuration for the app.

src/fare-contracts/details/OrderDetails.tsx
gorandalum: Agree. We save ourselves a bit of trouble if we our presentation is data-driven, then it isn't so fragile to future changes or unexpected data.

reidzeibel:
Updated the handling of empty string on price.

gorandalum:
> Updated the handling of empty string on price.

I think we can just handle `undefined` as well while we are at it. The field `totalPrice` isn't important, it is just for basic presentation, so we don't really need to safe guard that it is present by having it required in the typing.

src/ticketing/types.ts
marius-at-atb: So atm, do we have a bunch of duplicate types that need to be manually updated to be kept in sync e.g. in https://github.com/AtB-AS/mittatb-app/blob/master/src/ticketing/types.ts and https://github.com/AtB-AS/webshop2/blob/main/src/tickets/types.ts?
I think we should consider to have a repository for all types in Firestore.

reidzeibel:
Added handling of undefined values, tested against empty string and undefined values.

src/fare-contracts/details/OrderDetails.tsx
```diff
@@ -15,6 +16,11 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
     FareContractTexts.details.orderId(fareContract.orderId),
   );
   const firstTravelRight = fareContract.travelRights[0];
+  const priceString =
+    fareContract.totalAmount &&
+    fareContract.totalAmount !== '' &&
+    formatDecimalNumber(parseFloat(fareContract.totalAmount), language, 2);
```
marius-at-atb, line 22: When `totalAmount` is an empty string, this will return the empty string.
You can e.g. use a ternary operator instead with `undefined` in the else case.

marius-at-atb (approved):
Lgtm. Just one comment in case the intention was to treat an empty string as undefined.

src/fare-contracts/details/OrderDetails.tsx
reidzeibel: I tested with empty string, and this is the result, it doesn't show as empty string as shown below. Did I miss something here? 🤔

marius-at-atb: An empty string is falsy, so it still works as is. However just wanted to note that the `priceString` variable will be an empty string rather than undefined when `totalAmount === ''` with this code. Strictly speaking I think it is better that the `priceString` is explicitly undefined rather than empty, but mainly I mentioned this just to be aware of how the chaining of `&&` works here.

reidzeibel: I see, then I'll update it to prevent the falsy values, since it might be confusing in the future.

jorelosorio (approved):

src/fare-contracts/details/OrderDetails.tsx
```diff
@@ -15,6 +16,11 @@ export const OrderDetails = ({fareContract}: {fareContract: FareContract}) => {
     FareContractTexts.details.orderId(fareContract.orderId),
   );
   const firstTravelRight = fareContract.travelRights[0];
+  const priceString =
+    fareContract.totalAmount && fareContract.totalAmount !== ''
```
rosvik, line 20: Since the empty string is falsy, the last check is redundant here

```suggestion
    fareContract.totalAmount
```

rosvik (approved):
Looks good 👍

PR #4595, opened 2024-06-20
fix: make skoleskyss not resolve to NormalTravelRight

rosvik:
- Fixes a bug in `isNormalTravelRight` that doesn't catch `startDateTime: undefined`
- Filters out fare contracts that doesn't have any "normal" travel rights. This was needed to make some "firstTravelRight" logic work which assumes every fare contract has a travel right.
- Adds a test for a skoleskyss travelright.

fixes https://github.com/AtB-AS/kundevendt/issues/15528#issuecomment-2176504199

### DIFF
--- a/src/fare-contracts/utils.ts
+++ b/src/fare-contracts/utils.ts
@@ -256,9 +256,6 @@ export const useDefaultPreassignedFareProduct = (
   return preAssignedFareProducts[0];
 };
 
-export const getFareProductRef = (fc: FareContract) =>
-  fc.travelRights[0]?.fareProductRef;
-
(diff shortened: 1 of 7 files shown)


src/fare-contracts/utils.ts
```diff
@@ -256,9 +256,6 @@ export const useDefaultPreassignedFareProduct = (
   return preAssignedFareProducts[0];
 };
 
-export const getFareProductRef = (fc: FareContract) =>
-  fc.travelRights[0]?.fareProductRef;
```
rosvik, line 260: This wasn't in use

PR #4597, opened 2024-06-20
fix: Add includeSnackbar prop to Map

marius-at-atb:
https://mittatb.slack.com/archives/C02EEG7D8EL/p1718869819941549

Quick fix for when the `Snackbar` container is not at the top.

For the future, the `Snackbar` should probably be put inside a ContextProvider just like `BottomSheet` is in `BottomSheetProvider` instead.


reidzeibel (approved):
LGTM!

marius-at-atb:
@tormoseng Different issue, fixed in https://github.com/AtB-AS/mittatb-app/pull/4598

tormoseng:
Re-tested OK!

PR #4599, opened 2024-06-20
chore: update @atb-as/theme and @atb-as/generate-assets

adriansberg:
Updated to latest version to get the color changes for FRAM.

FRAM has got a "new"-ish color palette, basically a set of base colors similar to how AtB has. Designer Petter has gone through the colors and set up a palette based on the main theme colors already used, and looked at some contrasts etc.

Fixes https://github.com/AtB-AS/kundevendt/issues/15463


mortennordseth (approved):

marius-at-atb (approved):

src/components/label-info/LabelInfo.tsx
```diff
@@ -21,7 +22,13 @@ export const LabelInfo = ({label}: LabelInfoProps) => {
 
   return (
     <View style={linkSectionItemStyle.flag}>
-      <ThemeText color="info" type="body__tertiary">
+      <ThemeText
+        // Setting color="info" uses the secondary text color, so we need to set
+        // the primary text color manually for it to work with the label
+        // background.
```
gorandalum, line 28: Just remove this comment, comments clutter the code and this isn't important enough to have a comment on it IMO

gorandalum (approved):

src/components/label-info/LabelInfo.tsx
adriansberg: Will do!

PR #4600, opened 2024-06-21
refactor: Add more logging to auth and mobile token requests

gorandalum:
Mainly some more logging in the fetch id token flow, and changed
the notify Bugsnag event to always be the same message, and the
actual error message is set in the description.

Also added Bugsnag breadcrumb when listing tokens for a user.


marius-at-atb (approved):
Lgtm

rosvik (approved):
Nice, LGTM 👍

rosvik:
@gorandalum 🔔 If we're following the new flow for adding changes to release branches, this should be merged to master, then cherry picket to the release.

PR #4601, opened 2024-06-21
refactor: Redirect user to my tickets if wrong payment state

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/18133

I wrote a comment in the issue about the conclusions of why this.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
@@ -16,7 +16,7 @@ import {WebViewTerminal} from '@atb/stacks-hierarchy/Root_PurchasePaymentScreen/
 import {usePurchaseCallbackListener} from '@atb/stacks-hierarchy/Root_PurchasePaymentScreen/use-purchase-callback-listener';
 import {LoadingOverlay} from '@atb/stacks-hierarchy/Root_PurchasePaymentScreen/LoadingOverlay';
 import {PaymentProcessorStatus} from '@atb/stacks-hierarchy/types';
-import {PaymentType} from '@atb/ticketing';
+import {PaymentType, useTicketingState} from '@atb/ticketing';
 import {useOpenVippsAfterReservation} from '@atb/stacks-hierarchy/Root_PurchasePaymentScreen/use-open-vipps-after-reservation';
 
 type Props = RootStackScreenProps<'Root_PurchasePaymentScreen'>;
@@ -26,6 +26,7 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
   const {t} = useTranslation();
   const {offers, destinationAccountId, paymentMethod} = route.params;
   const analytics = useAnalytics();
+  const {fareContracts, sentFareContracts} = useTicketingState();
   const [paymentProcessorStatus, setPaymentProcessorStatus] =
     useState<PaymentProcessorStatus>('loading');
 
@@ -50,6 +51,23 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
     paymentMethod,
     reserveMutation.data?.recurring_payment_id,
   );
+
+  useEffect(() => {
+  }, [reserveMutation.data]);
+
+  useEffect(() => {
+    const orderId = reserveMutation.data?.order_id;
+    if (orderId) {
+      const fareContract = [...(fareContracts ?? []), ...(sentFareContracts ?? [])].map((fc) => {
+        return {
+          orderId: fc.orderId,
+        };
+      }).find((contract) => contract.orderId === orderId);
+      if (fareContract) {
+        navigateToActiveTicketsScreen();
+      }
+    }
+  }, [fareContracts, sentFareContracts, reserveMutation.data, navigateToActiveTicketsScreen]);
 
   const reserveOffer = reserveMutation.mutate;
   useEffect(() => {
--- a/src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
@@ -52,16 +52,17 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
     reserveMutation.data?.recurring_payment_id,
   );
 
+  const orderId = reserveMutation.data?.order_id;
+  const fareContractReceived = React.useMemo(() => {
+    const allPossibleFareContracts = [...(fareContracts ?? []), ...(sentFareContracts ?? [])];
+    return allPossibleFareContracts.some((fc) => fc.orderId === orderId && orderId !== undefined);
+  }, [fareContracts, orderId, sentFareContracts]);
+
   useEffect(() => {
-    const orderId = reserveMutation.data?.order_id;
-    if (orderId) {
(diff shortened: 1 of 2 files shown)


gorandalum (dismissed):
You can talk with @rosvik about this but as I understand the issue it isn't regarding having a timeout, but to redirect to the my tickets screen if the purchased ticket is received from Firestore while the web view is open.

jorelosorio:
> You can talk with @rosvik about this but as I understand the issue it isn't regarding having a timeout, but to redirect to the my tickets screen if the purchased ticket is received from Firestore while the web view is open.

I am going to ask. However, this solution seems to works for instance if it was successful or not, on my tickets it reflects the real status of the transaction.

gorandalum:
We discussed having a timeout in a refinement meeting, and ended up concluding that we should not have it for now. If it hangs for too long the user can cancel. But what we want to fix is when the purchase is actually successful, but the redirect callback doesn't reach the app.

src/stacks-hierarchy/Root_PurchasePaymentScreen/use-purchase-callback-listener.ts
```diff
@@ -11,19 +11,33 @@ export const usePurchaseCallbackListener = (
...
+      recurringPaymentId,
+    );
+    onCallback();
+  }, [userId, paymentMethod, recurringPaymentId, onCallback]);
+
   useEffect(() => {
-    const {remove: unsub} = Linking.addEventListener('url', async (event) => {
+    const timeout = setTimeout(async () => {
+      completeListener();
+    }, 30000);
```
marius-at-atb, line 27: Are we sure about 30s, or should it be remotely configurable?

marius-at-atb (approved):
Lgtm!

marius-at-atb:
Edit: I was on a non-refreshed state and didn't see the other comments until now.
As long as the user can cancel manually, triggering it based on the ticket being received is better :+1:
One might wonder from a UX perspective though, whether there should be a max time in the loading state.

gorandalum:
> Edit: I was on a non-refreshed state and didn't see the other comments until now. As long as the user can cancel manually, triggering it based on the ticket being received is better 👍 One might wonder from a UX perspective though, whether there should be a max time in the loading state.

Yeah, we might add a timeout as a UX improvement. It exists [an issue](https://github.com/AtB-AS/kundevendt/issues/18060) for it, it just isn't prioritised into the sprint backlog.

jorelosorio:
> triggering it based on the ticket being received is better

It might also happen that is not `ticket being received` at all, so a timeout is required for any solution taken.

rosvik:
I agree with Gøran here. This might be a fine solution/improvement but I don't think it resolves https://github.com/AtB-AS/kundevendt/issues/18133. What I had in mind, was something like "If the ticket context has a fare contract with the same ID as the order you're currently paying for: Navigate away from the purchase flow, and show the ticket"

jorelosorio:
> I agree with Gøran here. This might be a fine solution/improvement but I don't think it resolves [AtB-AS/kundevendt#18133](https://github.com/AtB-AS/kundevendt/issues/18133). What I had in mind, was something like "If the ticket context has a fare contract with the same ID as the order you're currently paying for: Navigate away from the purchase flow, and show the ticket"

It solves it but in a different way, however, I will change my code to listen for `ticket context` instead until it finds an Ok there. No timer on it. 🤷

jorelosorio:
@rosvik @gorandalum I made fixes to the PR and changed it accordingly to what we talked about.

src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
```diff
@@ -51,6 +52,23 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
     reserveMutation.data?.recurring_payment_id,
   );
 
+  useEffect(() => {
+  }, [reserveMutation.data]);
```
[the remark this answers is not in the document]
jorelosorio: Nope, thanks for the catch!

src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
```diff
@@ -51,6 +52,20 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
...
   );
 
+  useEffect(() => {
+    const orderId = reserveMutation.data?.order_id;
+    if (orderId) {
+      const fareContract = [...(fareContracts ?? []), ...(sentFareContracts ?? [])].map((fc) => {
+        return {
+          orderId: fc.orderId,
+        };
+      }).find((contract) => contract.orderId === orderId);
```
marius-at-atb, line 62: This can be simplified by using `.some`
```ts
const allPossibleFareContracts = [...(fareContracts ?? []), ...(sentFareContracts ?? [])]
const fareContractReceived = allPossibleFareContracts.some((fc) => fc.orderId === orderId);
```

src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
```diff
@@ -51,6 +52,20 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
...
+      if (fareContract) {
+        navigateToActiveTicketsScreen();
+      }
+    }
+  }, [fareContracts, sentFareContracts, reserveMutation.data, navigateToActiveTicketsScreen]);
```
marius-at-atb, line 67: I think `reserveMutation.data` should be `reserveMutation.data?.order_id`

marius-at-atb (approved):
Lgtm!

rosvik (approved):
Nice 👍

reidzeibel (approved):

src/stacks-hierarchy/Root_PurchasePaymentScreen/Root_PurchasePaymentScreen.tsx
```diff
@@ -51,6 +52,17 @@ export const Root_PurchasePaymentScreen = ({route, navigation}: Props) => {
...
+  useEffect(() => {
+    const orderId = reserveMutation.data?.order_id;
+    if (orderId) {
+      const allPossibleFareContracts = [...(fareContracts ?? []), ...(sentFareContracts ?? [])];
+      const fareContractReceived = allPossibleFareContracts.some((fc) => fc.orderId === orderId);
+      if (fareContractReceived) {
+        navigateToActiveTicketsScreen();
+      }
+    }
+  }, [fareContracts, sentFareContracts, reserveMutation.data?.order_id, navigateToActiveTicketsScreen]);
```
gorandalum, line 64: One general tip about `useEffect` is that they are often a little simpler to read if transformation/calculation is done outside of it. This also gives fewer dependencies, which decreases the chance of bugs.

Here it would be something like:
```tsx
const allFareContracts = [...(fareContracts ?? []), ...(sentFareContracts ?? [])];
const fareContractReceived = allFareContracts.some((fc) => 
    fc.orderId === reserveMutation.data?.order_id);

useEffect(() => {
  if (fareContractReceived) {
    navigateToActiveTicketsScreen();
  }
}, [fareContractReceived, navigateToActiveTicketsScreen])
```

Regarding doing the `.some` on every render, one can consider to use `useMemo`. But in most cases, a [simple iteration is not expensive](https://react.dev/learn/you-might-not-need-an-effect#how-to-tell-if-a-calculation-is-expensive).

jorelosorio: Good suggestion and made fixes based on those.

PR #4606, opened 2024-06-24
fix: mobile token onboarding will only show on authenticated users

reidzeibel:
closes: https://github.com/AtB-AS/kundevendt/issues/18445

Attempt fix the mobile token onboarding that appears after logout. I added the `authenticationType === 'phone` to `shouldShowPredicate` function, so that the onboarding will only show for authenticated users.


marius-at-atb (approved):
Interesting bug that might have a different root cause, but I think this will work :+1:

PR #4608, opened 2024-06-26
fix: Wrong state for Abt Entur SDK when initializing timer

jorelosorio:
Related to: https://github.com/AtB-AS/kundevendt/issues/18509

TimeContextProvider requires AbtClient to be initialized, this fix import some configurations as well initializes the AbtClient.

### DIFF
--- a/src/mobile-token/abtClientLogger.ts
+++ b/src/mobile-token/abtClientLogger.ts
@@ -10,7 +10,7 @@ export const localLogger: ClientConfig['localLogger'] = {
   },
   warn: (msg, err, metadata?) => {
     const onError = toOnErrorCallback('warning', msg, metadata);
-    Bugsnag.notify(err, onError);
+    if (err) Bugsnag.notify(err, onError);
   },
   error: (msg, err, metadata?) => {
     const onError = toOnErrorCallback('error', msg, metadata);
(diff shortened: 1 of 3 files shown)


src/mobile-token/abtClientLogger.ts
```diff
@@ -10,7 +10,7 @@ export const localLogger: ClientConfig['localLogger'] = {
   },
   warn: (msg, err, metadata?) => {
     const onError = toOnErrorCallback('warning', msg, metadata);
-    Bugsnag.notify(err, onError);
+    if (err) Bugsnag.notify(err, onError);
```
jorelosorio, line 13: A fix for https://app.bugsnag.com/atb-as/mittatb/errors/66600166df36580008bb5349?filters[user.id]=57f8e086-e0c7-49ea-b6d9-974838db6cb0&event_id=667bcf1f00efc9de65c30000

It uses the same check as in `error`.

gorandalum: So we no longer want to notify every warn event to Bugsnag?

gorandalum (approved):

src/mobile-token/abtClientLogger.ts
tormoseng: What kind of warnings is this? Do you have any examples?

jorelosorio: If error is undefined, we got a wrong error with no warning info.

tormoseng: Aha. That one

PR #4612, opened 2024-07-02
chore: updated design system dependencies

tft-abel:


ios/Podfile.lock
reidzeibel, line 1: I think this change should not be here, since it shouldn't be related to the changes in `packages.json` and `yarn.lock`

reidzeibel (changes requested):

ios/Podfile.lock
jorelosorio: The podfile yes, seems to be wrong.

tft-abel: I had a feeling

reidzeibel (approved):
🚀

PR #4613, opened 2024-07-02
chore: update network security config

hanne-at-atb:
For testing purposes we need to use proxyman in staging to intercept network calls. In order for it to work for android in staging we need som additional configuration. This should not affect prod or store. 
From proxyman documentation: https://docs.proxyman.io/debug-devices/android-device#:~:text=Open%20your%20app,files

### DIFF
--- a/android/app/src/staging/AndroidManifest.xml
+++ b/android/app/src/staging/AndroidManifest.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <application
+            android:networkSecurityConfig="@xml/network_security_config">
+    </application>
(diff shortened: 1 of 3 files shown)


reidzeibel (approved):
LGTM! Should be enough to run the proxyman for Android!

android/app/src/staging/AndroidManifest.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <application
+            android:networkSecurityConfig="@xml/network_security_config">
```
reidzeibel, line 4: If later we find that the traffic/network isn't returning proper text, we can add `android:usesCleartextTraffic="true"` in here.

PR #4614, opened 2024-07-02
chore: fix build issue with network security config

reidzeibel:
fix the build issue caused by wrong folder structure on network security config


hanne-at-atb (approved):
👏🏼  Thanks!

PR #4618, opened 2024-07-02
feat: show multimodal stop places in location search

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/16260

Shows the multimodal stop place properly when searching, instead of showing a separate stop place.


rosvik (approved):
That was a smaller change than expected 😅 💯

PR #4619, opened 2024-07-02
build: add update-app-version script

rosvik:
Adds a script to bump version numbers in .env and package.json files, so we don't have to do it manually for every release 🥳 

Screenshot 2024-07-02 at 15 57 34


reidzeibel (approved):
💨 🧠

PR #4621, opened 2024-07-03
feat: prevent login retires for 10s after too many attempts

rosvik:
When the app receives a 429 error (too_many_attempts) from the identity service, buttons to retry requests to /identity are disabled for 10 seconds. The error message "You have tried too many times. Please wait a bit before trying again." is shown as before.

This is just a suggestion for how to add some sort of limit on the amount of times the user can spam the submit buttons, so open for alternative solutions. But I think this is an ok solution because:

- It doesn't affect regular use (no 10s delay unless you're already rate limited by identity/Twilio)
- The logic in the app is simple, and easy to reason about. More complex logic should probably be on the backend.

closes https://github.com/AtB-AS/kundevendt/issues/18555

### DIFF
--- a/src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
+++ b/src/stacks-hierarchy/Root_LoginConfirmCodeScreen.tsx
@@ -43,24 +43,35 @@ export const Root_LoginConfirmCodeScreen = ({route}: Props) => {
   const [isLoading, setIsLoading] = useState(false);
   const focusRef = useFocusOnLoad();
   const {completeOnboardingSection} = useOnboardingState();
+  const [isRateLimited, setIsRateLimited] = useState(false);
 
   const onLogin = async () => {
+    if (isRateLimited) return;
     setIsLoading(true);
     completeOnboardingSection('userCreation');
     const errorCode = await confirmCode(code);
     if (errorCode) {
+      if (errorCode === 'too_many_attempts') {
+        setIsRateLimited(true);
+        setTimeout(() => setIsRateLimited(false), 10000);
+      }
       setError(errorCode);
       setIsLoading(false);
     }
   };
 
   const onResendCode = async () => {
+    if (isRateLimited) return;
     setIsLoading(true);
     setError(undefined);
     setCode('');
     const errorCode = await signInWithPhoneNumber(phoneNumber, true);
     setIsLoading(false);
     if (errorCode) {
+      if (errorCode === 'too_many_attempts') {
+        setIsRateLimited(true);
+        setTimeout(() => setIsRateLimited(false), 10000);
+      }
       setError(errorCode);
     }
   };
@@ -133,7 +144,7 @@ export const Root_LoginConfirmCodeScreen = ({route}: Props) => {
                   interactiveColor="interactive_0"
                   onPress={onLogin}
                   text={t(LoginTexts.confirmCode.mainButton)}
-                  disabled={!code}
+                  disabled={!code || isRateLimited}
                   rightIcon={{svg: ArrowRight}}
                   style={styles.submitButton}
                   testID="submitButton"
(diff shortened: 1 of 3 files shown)


reidzeibel:
Doesn't seem like there's anything wrong, should we show the remaining seconds? (f.eks. "Try again in x second(s)", where x counts down?)

rosvik:
I think that might give the impression that they get a new attempt in 10s. It's really around 10 minutes, so trying again will give the same message and a new 10s timeout. There might be a better way to say "come back later, you're locked out for now" however.

tormoseng:
Agree with @rosvik. The sentence `You have tried too many times. Please wait a bit before trying again` is OK minus - it's the `wait a bit`-part that could be rephrased.

rosvik:
So something like 

```
"You have tried too many times. Please wait a few minutes before trying again"
"Du har prøvd for mange ganger. Vennligst vent noen minutter før du prøver igjen."
```

And go for the 10s timeout in this PR?

reidzeibel (approved):
Can't comment on the wording part, but the code looks OK!

marius-at-atb:
I think this keeps it simple and will work :+1:
However I don't like the code duplication. Perhaps create a simple hook?
Also not a big deal here, but the timeouts can potentially interfere with each other, so best practice is to cancel other timeouts before starting a new one. With a hook this can also be handled in a simple way.

rosvik:
@marius-at-atb 

> However I don't like the code duplication. Perhaps create a simple hook?

I don't see how this can be de-duplicated significantly. The submit handlers can probably be simplified with an useMutation, but I think that's outside of the scope of this change.

We could do something like this, but that turns two lines into one, and is less explicit, so not sure I see the benefit. 

```ts
function toggleTimeout(time: number, set: (value: boolean) => void) {
  set(true);
  return setTimeout(() => set(false), time);
}

if (errorCode === 'too_many_attempts') {
  toggleTimeout(10000, setIsRateLimited);
}
```

Do you have any specific ideas to improve it in other ways?

> the timeouts can potentially interfere with each other

Yeah, we should be a bit careful with timeouts, but I think this is safe since no timeouts will start before the previous has ended (due to `if (isRateLimited) return;`)

marius-at-atb:
I'm thinking something like this:
```ts
const {isRateLimited, rateLimitIfNeeded} = useRateLimiting();
```

So instead of writing
```ts
if (result === 'too_many_attempts') {
   setIsRateLimited(true);
   setTimeout(() => setIsRateLimited(false), 10000);
}
```
every time, it can be just `rateLimitIfNeeded(errorCode)`.

The hook including timeout clearing:

```ts
const rateLimitDurationMS = 10000 // 10s

const useRateLimiting = () => {
  const [isRateLimited, setIsRateLimited] = useState(false);
  const timeoutRef = useRef<NodeJS.Timeout | undefined>();

  const rateLimitIfNeeded = (errorCode: string) => {
    if (errorCode === 'too_many_attempts') {
      timeoutRef.current && clearTimeout(timeoutRef.current);
      setIsRateLimited(true);
      timeoutRef.current = setTimeout(() => setIsRateLimited(false), rateLimitDurationMS);
    }
  };

  return {
    isRateLimited,
    rateLimitIfNeeded,
  };
};
```

E.g. accepting `rateLimitDurationMS` as a parameter could be an option as well. 

Looking at it now, the `errorCode === 'too_many_attempts'` check is the only specific logic in the hook, so it could be turned into a more general hook as well. (Kinda related in that case: `useDebounce` and `useDelayGate`)

rosvik:
@marius-at-atb I tried to implement something similar to your suggestion in https://github.com/AtB-AS/mittatb-app/pull/4624 now, but made more generic with a `predicate` argument. I think it is an ok solution, but since it is more complex and might not being useful outside of this use case, I'll go with the simple solution for now.

If anyone feel differently after they come back from vacation, the other PR is ready for review as well 😊

PR #4623, opened 2024-07-04
feat: extend travel search to ticketing functionality to include single ticket boat

mortennordseth:
This PR extends the travel search to ticketing functionality by including single ticket boat for boat trips. Similar to bus ticket suggestions, the boat ticket option will be displayed only if the entire journey is composed exclusively of boat legs and there are no significant wait times between legs.

Closes https://github.com/AtB-AS/kundevendt/issues/18556

### DIFF
--- a/src/travel-details-screens/TripDetailsScreenComponent.tsx
+++ b/src/travel-details-screens/TripDetailsScreenComponent.tsx
@@ -230,7 +230,7 @@ function getTicketInfoForBus(
     tariffZones,
   );
 
-  if (!fromTariffZone || !fromTariffZone) return;
+  if (!fromTariffZone || !toTariffZone) return;
 
   const fareProductTypeConfig = fareProductTypeConfigs.find(
     (config) => config.type === 'single',
--- a/src/travel-details-screens/TripDetailsScreenComponent.tsx
+++ b/src/travel-details-screens/TripDetailsScreenComponent.tsx
@@ -150,7 +150,9 @@ export const TripDetailsScreenComponent = ({
   );
 };
 
-function useTicketInfoFromTrip(tripPattern: TripPattern) {
+function useTicketInfoFromTrip(
+  tripPattern: TripPattern,
+): TicketInfoForBus | TicketInfoForBoat | undefined {
   const {enable_ticketing} = useRemoteConfig();
   const isFromTravelSearchToTicketEnabled =
     useFromTravelSearchToTicketEnabled();
@@ -202,13 +204,20 @@ function calculateTicketStartTime(legs: Leg[]) {
     : formatISO(tripStartWithBuffer);
 }
 
+type TicketInfoForBus = {
+  fromPlace: TariffZoneWithMetadata;
+  toPlace: TariffZoneWithMetadata;
+  ticketStartTime: string | undefined;
+  fareProductTypeConfig: FareProductTypeConfig;
+};
+
 function getTicketInfoForBus(
   tripPattern: TripPattern,
   nonHumanLegs: Leg[],
   fareProductTypeConfigs: FareProductTypeConfig[],
   tariffZones: TariffZone[],
   ticketStartTime?: string,
-) {
+): TicketInfoForBus | undefined {
   const canSellCollab = canSellCollabTicket(tripPattern);
   const hasOnlyValidBusLegs = !hasLegsWeCantSellTicketsFor(tripPattern, [
     'cityTram',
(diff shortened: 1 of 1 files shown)


src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -151,73 +151,155 @@ export const TripDetailsScreenComponent = ({
...
-  const ticketStartTime =
-    tripStartWithBuffer.getTime() <= Date.now()
-      ? undefined
-      : formatISO(tripStartWithBuffer);
+  const toTariffZone = getTariffZoneWithMetadata(
+    nonHumanLegs[nonHumanLegs.length - 1].toPlace,
+    tariffZones,
+  );
+
+  if (!fromTariffZone || !fromTariffZone) return;
```
adriansberg, line 233: One of these should probably be `toTariffZone`, I would guess.

```suggestion
  if (!fromTariffZone || !toTariffZone) return;
```

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -133,9 +133,9 @@ export const TripDetailsScreenComponent = ({
             onPress={() => {
               analytics().logEvent('click_trip_purchase_button');
               onPressBuyTicket({
-                fareProductTypeConfig: singleTicketConfig,
-                fromPlace: tripTicketDetails.tariffZoneFrom,
-                toPlace: tripTicketDetails.tariffZoneTo,
+                fareProductTypeConfig: tripTicketDetails.fareProductTypeConfig,
```
adriansberg, line 136: `onPressBuyTicket` does not require the config that is given to be of `type === 'single'`?

adriansberg: I see that `getTicketInfoForBus` returns the config with type 'single', and `getTicketInfoForBoat` returns the 'boat-single', so I guess that works 👍

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -151,73 +151,155 @@ export const TripDetailsScreenComponent = ({
 };
 
 function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
```
adriansberg, line 153: Is `Get` needed in the name when it's a hook? I see that you haven't touched the name, but figured I should ask anyways. 😀

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -151,73 +151,155 @@ export const TripDetailsScreenComponent = ({
...
     'regionalBus',
     'shuttleBus',
   ]);
-  if (!enable_ticketing || (someLegsAreNotSingleTicket && !canSellCollab))
-    return;
 
-  const tripStartWithBuffer = addMinutes(
-    parseISO(nonHumanLegs[0]?.aimedStartTime),
-    -5,
+  if (!hasOnlyValidBusLegs && !canSellCollab) return;
```
adriansberg, line 222: Not 100% sure what `canSellCollab` is, but it will never be `true` if `hasOnlyValidBusLegs` is `false`?

src/travel-details-screens/TripDetailsScreenComponent.tsx
mortennordseth: Yes, the `getTicketInfoForBoat` and `getTicketInfoForBus` functions handles which config is passed to the `onPressBuyTicket` since these differs for single ticket bus and boat 👍

src/travel-details-screens/TripDetailsScreenComponent.tsx
adriansberg: Not that it matters in this case, since the condition looks to be correct either way. 👌

adriansberg: Or... wait. Original question is still valid.

src/travel-details-screens/TripDetailsScreenComponent.tsx
mortennordseth: Good suggestion! I've updated the name to `useTicketInfoFromTrip` 😊

src/travel-details-screens/TripDetailsScreenComponent.tsx
mortennordseth: The `canSellCollab` check is AtB-specific code to account for single ticket bus being valid on train trips in zone A. `canSellCollab` is true if some legs in the trip are by train in zone A (only for AtB). 

It can be `true` if `hasOnlyValidBusLegs` is `false` because still some legs might be by train in zone A, but the trip can also include a boat journey for example

rosvik (commented):
I think this looks like a good solution 👍 Hard to know if there are edge cases that will cause issues for some trip combinations, so we'll have to do some thorough testing on this one.

Had one suggestion on return types

src/travel-details-screens/TripDetailsScreenComponent.tsx
```diff
@@ -150,74 +150,156 @@ export const TripDetailsScreenComponent = ({
   );
 };
 
-function useGetTicketInfoFromTrip(tripPattern: TripPattern) {
-  const fromTripsSearchToTicketEnabled = useFromTravelSearchToTicketEnabled();
+function useTicketInfoFromTrip(tripPattern: TripPattern) {
```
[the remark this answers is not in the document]
adriansberg: Added return types to the functions.

PR #4625, opened 2024-07-05
feat: update translations for flexible ticket

rosvik:

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
@@ -66,7 +66,7 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
 
             const userProfileName = getReferenceDataName(u, language);
             const discountText =
-              (discountPercent !== undefined) &&
+              discountPercent !== undefined &&
               t(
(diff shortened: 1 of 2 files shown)


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/FlexTicketDiscountInfo.tsx
```diff
@@ -65,21 +65,19 @@ export const FlexTicketDiscountInfo = ({userProfiles, style}: Props) => {
...
 
             const userProfileName = getReferenceDataName(u, language);
-            const discountText =
-              (discountPercent !== undefined) &&
-              t(
-                PurchaseOverviewTexts.flexDiscount.discountPercentage(
-                  discountPercent.toFixed(0),
-                ),
-              );
+            const discountLabel =
```
reidzeibel, line 68: @rosvik should we keep this here? since the discount ladder shouldn't be relevant if we only shows the new price.

rosvik: Yes, we still want to show some indication that this ticket is or will be discounted.

reidzeibel (approved):
Nice!

tormoseng:
Looks good!

tormoseng:
Will be reverted back to the original solution after discussions.

PR #4628, opened 2024-07-05
feat: update geolocation library

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/94

I updated the geolocation library from https://github.com/Agontuk/react-native-geolocation-service to https://github.com/michalchudziak/react-native-geolocation.

The old one is already 2 years old and doesn't look like it will get updated, also the dialog which keeps showing is from the library itself. With the new library, Google Location Accuracy is not mandatory, and the functionality should be the same between both library. 

Configuration is set as below:

```
  Geolocation.setRNConfiguration({
    locationProvider: 'playServices',
    skipPermissionRequests: true,
  });
```

We use `playServices` to keep the same functionality as before on Android (uses `FusedLocationProvider` from google), and we set `skipPermissionRequests` to `true` because we handle the permission request already.


rosvik (approved):

src/GeolocationContext.tsx
```diff
@@ -135,6 +135,10 @@ export const GeolocationContextProvider: React.FC = ({children}) => {
   const geoLocationName = t(dictionary.myPosition); // TODO: Other place for this fallback
   const currentCoordinatesRef = useRef<Coordinates | undefined>();
   const {updateMetadata} = useIntercomMetadata();
+  Geolocation.setRNConfiguration({
+    locationProvider: 'playServices',
+    skipPermissionRequests: true,
+  });
```
gorandalum, line 141: Should this be set outside of the `GeolocationContextProvider`? Now this is getting called every render.

PR #4636, opened 2024-07-17
fix: remove unwanted widget padding for iOS 17 and above

rosvik:
This was caused by not using the containerBackground API (introduced in iOS 17.0), and instead relying on padding inside DepartureTimesView.

See https://developer.apple.com/documentation/widgetkit/preparing-widgets-for-additional-contexts-and-appearances

fixes https://github.com/AtB-AS/kundevendt/issues/18640

### iOS 15

Screenshot 2024-07-17 at 11 08 32

### iOS 17

Screenshot 2024-07-17 at 11 08 43


adriawh (approved):
😍

PR #4637, opened 2024-07-17
feat: add Xcode Preview of DepartureWidget

rosvik:
Enables live-ish editing of departureWidget

Screenshot 2024-07-17 at 11 56 08

Follow up to https://github.com/AtB-AS/mittatb-app/pull/4636


jorelosorio:
> @jorelosorio Is it safe to remove "AtbAppIntent" from DepartureWidget "Target Membership" list?
> 
> Screenshot 2024-07-17 at 14 54 15
> 
> This makes the preview work, but I don't understand what consequences it has, so waiting for feedback from you before I merge. (Or if I'm on vacation when you have time to look at it, feel free to merge it)

Intent for the widget is show if long press the widget then it backflips with more options, if we disable that then it is like we are not using it, the widget breaks or no options are show when long press.

jorelosorio (approved):

PR #4638, opened 2024-07-29
feat: always show Norwegian time even when outside of Norwegian timezone

mikaelbr:
Fixes https://github.com/AtB-AS/kundevendt/issues/9610

This overrides the internal format function from date-fns to by default show all time in CET timezone, except for in some cases on input mode to avoid "double convertion". This should work as before if you already are in CET, but if you are in any other timezone it would behave as if in Norway.

See test scenarios in the original issue (9610).

Avoid marging until after https://github.com/AtB-AS/kundevendt/issues/18441

Possible follow-up PRs:

- Updating date-fns allow for removing iso8601-duration (https://github.com/AtB-AS/kundevendt/issues/7795)
- `language` no longer needed in `formatLocaleTime`


src/utils/date.ts
```diff
@@ -198,17 +217,23 @@ export function isRelativeButNotNow(
...
+  // @TODO: No longer in use. deprecate and remove
+  _language: Language,
+  options: Partial<LocalTimeFormatOptions> = {ignoreTimeZone: false},
+) {
+  const parsed = parseIfNeeded(date);
+  const opts = {
+    ignoreTimeZone: false,
+    ...options,
+  };
+  return format(parsed, 'HH:mm', opts);
```
mikaelbr, line 234: This used language at some point to probably support different formats, but now it seems to only do 24-hour. Haven't updated all references but removed unneeded code

package.json
```diff
@@ -68,7 +68,8 @@
     "axios": "^0.28.0",
     "axios-better-stacktrace": "^2.1.6",
     "axios-retry": "^3.3.1",
-    "date-fns": "^2.29.3",
+    "date-fns": "^3.6.0",
+    "date-fns-tz": "^3.1.3",
```
mikaelbr, line 72: Using external library to support proper timezoning on cross devices. There are some differences in the way JS engines handle timezone converting. date-fns-tz is confirmed and battle tested with the Hermes engine

marius-at-atb: Btw just a note that [this issue](https://github.com/AtB-AS/kundevendt/issues/7795) becomes possible to complete after upgrading to date-fns v3.

mikaelbr: 👍 Only handle necessary breaking changes in this PR, but I guess this can be relevant for in relation to https://github.com/AtB-AS/kundevendt/issues/16313 @jorelosorio ? Or maybe better yet separate PR

src/utils/date.ts
```diff
@@ -1,39 +1,65 @@
...
+function format(
+  date: string | number | Date,
+  formatStr: string,
+  options?: FormatOptions & {ignoreTimeZone?: boolean},
+) {
+  if (options?.ignoreTimeZone) {
+    return fnsFormat(date, formatStr, options);
+  } else {
+    return formatInTimeZone(date, CET, formatStr, {
+      ...(options as FormatOptionsWithTZ),
```
marius-at-atb, line 58: Since `timeZone` is added right after `options` here, and the type name is "WithTZ", maybe more intuitive like this?
```ts
{
   ...options,
   timeZone: CET,
} as FormatOptionsWithTZ
```

marius-at-atb (approved):
Nice :+1:

gorandalum (approved):
Looks great! 🎉

src/utils/date.ts
mikaelbr: Agreed! Fixed

PR #4639, opened 2024-07-30
fix: adds a11y prefix hint for message box criticality

mikaelbr:
Adds simple prefix message to a11y hint. Will be present at all times even if icon is not visible, as all non visually impaired people will get this indication by colour.

Note for future: I see that there isn't any live region or automatic reading of messages here. These components typically have some assertive/polite reading to them to make sure people with screen readers notice them. Should test and investigate the consequences of introducing this now.

Note: I don't have phone connected and can't test actual screen reader now only debug inspection. So would be nice to have this actually tested on device.

Fixes https://github.com/AtB-AS/kundevendt/issues/18138

### DIFF
--- a/src/translations/components/MessageBox.ts
+++ b/src/translations/components/MessageBox.ts
@@ -2,7 +2,11 @@ import {translation as _} from '../commons';
 
 const MessageBoxTexts = {
   a11yHintActionPrefix: _('Aktiver for å ', 'Activate to ', 'Aktiver for å'),
-  a11yHintUrlPrefix: _('Aktiver for å åpne lenke ', 'Activate to open link ', 'Aktiver for å opne lenke '),
+  a11yHintUrlPrefix: _(
+    'Aktiver for å åpne lenke ',
+    'Activate to open link ',
+    'Aktiver for å opne lenke ',
+  ),
   dismiss: {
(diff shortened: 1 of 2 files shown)


gorandalum (approved):

src/translations/components/MessageBox.ts
```diff
@@ -2,10 +2,21 @@ import {translation as _} from '../commons';
...
   dismiss: {
     allyLabel: _('Lukk melding', 'Close message', 'Lukk melding'),
   },
+
+  a11yCriticalityPrefix: {
+    valid: _('Vellykket', 'Success', 'Vellykka'),
+    info: _('Info', 'Info', 'Info'),
+    warning: _('Advarsel', 'Warning', 'Advarsel'),
+    error: _('Feilmelding', 'Error', 'Feilmelding'),
+  },
```
gorandalum, line 19: Coming to think of it, we should probably reuse the `messageTypes` in `dictionary.ts` instead?

mikaelbr: Hmm. Didn't think of those! Good point. Also see this is used by a different message component (🤔(

mikaelbr: That also explains why this is a regression. Prefix is present in other component, but not this. Anyways. fixed

PR #4644, opened 2024-07-31
feat: Global message to inform about flexible tickets for anonymous

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/18591

Simulator Screenshot - iPhone 15 - 2024-07-31 at 12 15 18

### DIFF
--- a/src/global-messages/types.ts
+++ b/src/global-messages/types.ts
@@ -18,6 +18,7 @@ export enum GlobalMessageContextEnum {
   appServiceDisruptions = 'app-service-disruptions',
   appLogin = 'app-login',
   appLoginPhone = 'app-login-phone',
+  appAnonymousFlexibleTickets = 'app-anonymous-flexible-tickets',
 }
 
 export type GlobalMessageRaw = {
--- a/src/global-messages/types.ts
+++ b/src/global-messages/types.ts
@@ -18,7 +18,6 @@ export enum GlobalMessageContextEnum {
   appServiceDisruptions = 'app-service-disruptions',
   appLogin = 'app-login',
   appLoginPhone = 'app-login-phone',
-  appAnonymousFlexibleTickets = 'app-anonymous-flexible-tickets',
 }
 
 export type GlobalMessageRaw = {
--- a/src/rule-engine/rules.ts
+++ b/src/rule-engine/rules.ts
@@ -20,6 +20,7 @@ export enum RuleOperator {
   lessThanOrEqualTo = 'lessThanOrEqualTo',
   contains = 'contains',
   notContains = 'notContains',
+  onlyContains = 'onlyContains',
 }
 
 export const checkRules = (
@@ -111,6 +112,10 @@ const checkRule = (
       return false;
     case RuleOperator.notContains:
       if (Array.isArray(localValue)) return !localValue.includes(ruleValue);
+      return false;
+    case RuleOperator.onlyContains:
+      if (Array.isArray(localValue))
+        return localValue.every((v) => v === ruleValue);
       return false;
     default:
       return false;
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
@@ -181,7 +181,7 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
 
   const focusRefs = useFocusRefs(params.onFocusElement);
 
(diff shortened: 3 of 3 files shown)


src/global-messages/types.ts
```diff
@@ -18,6 +18,7 @@ export enum GlobalMessageContextEnum {
   appServiceDisruptions = 'app-service-disruptions',
   appLogin = 'app-login',
   appLoginPhone = 'app-login-phone',
+  appAnonymousFlexibleTickets = 'app-anonymous-flexible-tickets',
```
reidzeibel, line 21: I think this should be `appTicketingAnonymous` to be consistent with how we name the global message context. From my understanding, the global message can be used on various cases, not only flexible ticket.

gorandalum (changes requested):
Could you elaborate some on why we need a new global message context? We already have a global message context on the same place in the app.

jorelosorio:
> Could you elaborate some on why we need a new global message context? We already have a global message context on the same place in the app.

I do not have an explanation, of why. However, I see that there was a previous global message with different rules validations and context than the one we want to show here. So then I put another one with different rules.

Existing:
```ts
<GlobalMessage
                globalMessageContext={
                  GlobalMessageContextEnum.appPurchaseOverview
                }
                textColor="background_0"
                ruleVariables={{
                  preassignedFareProductType: preassignedFareProduct.type,
                  fromTariffZone: fromPlace.id,
                  toTariffZone: toPlace.id,
                }}
              />
```

Added one:

```ts
              <GlobalMessage
                globalMessageContext={
                  GlobalMessageContextEnum.appAnonymousFlexibleTickets
                }
                textColor="background_0"
                ruleVariables={{
                  isLoggedIn: isLoggedIn,
                }}
              />
```

Any better suggestion?

gorandalum:
What about adding the rule variable `isLoggedIn` to the existing context?

gorandalum:
I took a closer look at this in the code and what we have before. Can we use the global rule variable `authenticationType` for this? 🤔

jorelosorio:
> I took a closer look at this in the code and what we have before. Can we use the global rule variable `authenticationType` for this? 🤔

`isLoggedIn` is basically a `authenticationType == 'phone'`, so we can switch to `authenticationType` if it wants a more detailed rule.

gorandalum:
But isn't `authenticationType` already there, so we might not need any code changes?

jorelosorio:
> But isn't `authenticationType` already there, so we might not need any code changes?

No, the code as is right now, does not have any rule for `authenticationType`. I see is inside a conditional, `isFree` but that one does not depend on the user login state.

gorandalum:
> No, the code as is right now, does not have any rule for `authenticationType`. I see is inside a conditional, `isFree` but that one does not depend on the user login state.

Seems like it is there as a global rule variable, applicable to all rules:
https://github.com/AtB-AS/mittatb-app/blob/9c6e29ad013ba95253cacb51cbc25994a3745f2d/src/global-messages/GlobalMessage.tsx#L45

jorelosorio:
We found a global variable definition for all rules that includes authenticationType, so this PR is being closed.

image

gorandalum (commented):
And have you tested it with a rule in global messages? I'm a little concerned regarding the `userTypeString` list not being filtered 🤔

jorelosorio:
> And have you tested it with a rule in global messages? I'm a little concerned regarding the `userTypeString` list not being filtered 🤔

Yes, it loosk like:

image

jorelosorio:
> > And have you tested it with a rule in global messages? I'm a little concerned regarding the `userTypeString` list not being filtered 🤔
> 
> Yes, it loosk like:
> 
> image

I also added a new rule called `onlyContains` if we want to make sure only that element appears. @gorandalum

gorandalum:
> > And have you tested it with a rule in global messages? I'm a little concerned regarding the `userTypeString` list not being filtered 🤔
> 
> Yes, it loosk like:
> 
> image

Doesn't seem to work as intended for me, here the global message comes even though I haven't selected adult as traveller. Think it needs some filtering.

jorelosorio:
> > > And have you tested it with a rule in global messages? I'm a little concerned regarding the `userTypeString` list not being filtered 🤔
> > 
> > 
> > Yes, it loosk like:
> > image
> 
> Doesn't seem to work as intended for me, here the global message comes even though I haven't selected adult as traveller. Think it needs some filtering. 

Use the new rule, contains Only instead.

gorandalum (approved):
Good. Remember to update the acceptance criteria so they actually target the two new features:
- Possibility to use the `userTypes` rule in the `appPurchaseOverview` context.
- Possibility to use the new rule operator `onlyContains` on all rules where the variable is a list.

PR #4645, opened 2024-07-31
feat: Update buy-for-other screen according to Figma

gorandalum:
Figma sketch:  

Before / After in app (without the save recipient checkbox yet)
<div>
  

</div>


marius-at-atb (approved):
Looks good!

PR #4646, opened 2024-08-01
feat: adds feature toggle for boat ticket from trip details

mikaelbr:
Adds feature toggle `enable_from_trip_details_to_ticket_boat` that defaults to `false`. This will enable/disable the button to buy tickets in trip details for express boats.

Fixes issue after feedback in https://github.com/AtB-AS/kundevendt/issues/18556 by @hanne-at-atb

### DIFF
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -36,6 +36,7 @@ export type RemoteConfig = {
   enable_show_valid_time_info: boolean;
   enable_ticket_information: boolean;
   enable_ticketing: boolean;
+  enable_from_trip_details_to_ticket_boat: boolean;
   enable_ticketing_assistant: boolean;
   enable_tips_and_information: boolean;
   enable_token_fallback: boolean;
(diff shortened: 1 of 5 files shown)


gorandalum (changes requested):
We also add a possibility to override feature toggles in our debug menu, so that our testers can enable it in prod without enabling it for everybody. Take a look at other feature toggles and how they are used in `Profile_DebugInfoScreen`.

mikaelbr:
I see! There is an issue with non-updating existing view after overriding (some shared state issue it looks like), but that seems to be an existing problem so not looking much more into it here. It is overridable now at least!

gorandalum:
> fix: updates to use debug override pattern



> I see! There is an issue with non-updating existing view after overriding (some shared state issue it looks like), but that seems to be an existing problem so not looking much more into it here. It is overridable now at least!

Yeah, debug overrides most often needs app restart, testers are aware of this.

gorandalum (approved):

PR #4647, opened 2024-08-02
feat: Save recipient name

gorandalum:
Added a checkbox and text input for saving recipient name in the on
behalf of purchase process. The recipient name is sent to the
backend when reserving the offer.

Figma sketches:
Screenshot 2024-08-02 at 13 43 41

Screenshots from app:
<div>

</div>


jorelosorio (approved):
Nice implementation and refactoring for naming!

PR #4648, opened 2024-08-02
feat: Consolidate radio button group

gorandalum:
### Background
The new sketches for saving recipient has a radio section group with the standard radio selection circle instead of using checkmarks as we did before. 

### Solution
To not have two different radio groups in the app I took the liberty to just rewrite the existing one, making every radio button group use the new design.

Originally this was done as part of https://github.com/AtB-AS/kundevendt/issues/17226, but I added the PR to the board as an individual issue so it can be tested on its own.

<details>
<summary>Screenshots of the change</summary>

**Figma**:
<div>

</div>

**Some before/after examples from the app**:
<div>

</div>
<div>

</div>
<div>

</div>
<div>

</div>
</details>

### Acceptance criteria
- [ ] Radio groups in the app should have a selection circle to the left instead of checkmark to the right.
- [ ] Radio groups in the app should behave as before, and work just as good with screen reader and large text. 

The places in the app where we have used the `RadioGroupSection`. Should be enough to test just some of them.
- Single traveller selection
- Product selection by product (as with AtB night ticket)
- Default start screen selection
- Default traveller selection
- App language selection
- T-card/Mobile token selection
- Date now/departure/arrival selection in trip search
- Mobile token selection in the "need-token-on-mobile" screen which shows up when trying to buy boat tickets while having t-card as inspectable token

### DIFF
--- a/src/components/sections/RadioGroupSection.tsx
+++ b/src/components/sections/RadioGroupSection.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {ActionSectionItem} from './items/ActionSectionItem';
+import {RadioSectionItem} from './items/RadioSectionItem.tsx';
 import {Section, SectionProps} from './Section';
 import {HeaderSectionItem} from './items/HeaderSectionItem';
 import {InteractiveColor} from '@atb/theme/colors';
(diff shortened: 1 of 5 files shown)


gorandalum:
FYI @ckbilstad. Some sketches in Figma probably needs to be updated after this change.

ckbilstad:
👍🏼

jorelosorio (approved):
nice work

hanne-at-atb:
While testing I found that voiceOver does not announce that the radio button is interact-able when choosing departure time in travel search. Was not a problem in the other screens.  See screenshot below to see what voiceover reads out. It's also a bit different between iOS and android as on android it tells you that you can click on the button. 

It's the same also for:

- Default start screen selection
- Default traveller selection




This might be something to fix as you are at this @gorandalum ? 🌱

gorandalum:
I found this myself during development, that the accessibility announcements aren't that good on iOS anymore. Seems like it has something to do with the update to react-native 0.73:
https://github.com/facebook/react-native/issues/43266

I think we either should change role to button and handle the selected-announcement manually, or just wait it out until there is a fix in React-native.

hanne-at-atb:
I think it would be good to change role to button and handle the selected-announcement manually, and not wait for a fix in React-native as we don't know how long that will be.

gorandalum:
> I think it would be good to change role to button and handle the selected-announcement manually, and not wait for a fix in React-native as we don't know how long that will be.

Fix for the RadioSectionGroup and the checkbox in select recipient flow:
https://github.com/AtB-AS/mittatb-app/pull/4663

hanne-at-atb:
Better now, tested ok for Android and iOS.

PR #4649, opened 2024-08-05
update: Update non react-dependencies in app

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/16313


src/mobility/utils.ts
```diff
@@ -101,8 +102,10 @@ export const needsReload = (
     shownArea.range,
   );
 
+  if (!prevAreaFeature || !newAreaFeature) return true;
```
gorandalum, line 105: Can these be `undefined`?

gorandalum (commented):
You mentioned something about upgrading everything with non-breaking-changes, could you mention the dependencies (if there are any) which had breaking-changes and are not included in this PR?

src/mobility/utils.ts
jorelosorio: The new function does not allow undefined as before, instead requires a valid object, but those could be undefined yes.

jorelosorio:
> You mentioned something about upgrading everything with non-breaking-changes, could you mention the dependencies (if there are any) which had breaking-changes and are not included in this PR?

@turf/boolean-point-in-polygon (Adds a new function that expect different parameters)
@babel/core, @babel/preset-env, @babel/preset-typescript, @babel/runtime for those the new babel is 7.25, I tried to update them but got a lot of complications while "compiling" code, we need to introduce new rules and not sure even if react-native support it yet.
@marked is also breaking the app, but requires extra configurations to be setup, and I did not manage to make it work.
@svg-cli after that yarn setup worked but errors while compiling code
@react-native-camera-kit is up to date, upgrade it send it back to version 13 that breaks the app.
@axios breaks the app, needs extra work
@posthog-react-native needs some types (compiles missing functions definitions)

For this other, it requires look one by one because some instruduces some erros and requires implementations:

"react-native-bootsplash": "^4.4.1",
    "react-native-camera-kit": "14.0.0-beta12",
    "react-native-compressor": "^1.8.12",
    "react-native-device-info": "^10.3.0",
    "react-native-flipper": "^0.212.0",
    "react-native-gesture-handler": "^2.15.0",
    "react-native-get-random-values": "^1.8.0",
    "react-native-inappbrowser-reborn": "^3.7.0",
    "react-native-kettle-module": "1.2.0",
    "react-native-linear-gradient": "^2.6.2",
    "react-native-localize": "^2.2.4",
    "react-native-pager-view": "^6.1.2",
    "react-native-permissions": "^3.10.0",
    "react-native-popover-view": "^5.1.7",
    "react-native-portalize": "^1.0.7",
    "react-native-safe-area-context": "^4.7.4",
    "react-native-screens": "^3.29.0",
    "react-native-svg": "^13.7.0",
    "react-native-tab-view": "^3.3.4",
    "react-native-webview": "^13.8.1",
    
    those can be evaluated in a separated but they are somehow related to the react-native version.

src/mobility/utils.ts
gorandalum: By looking at the code I can't see how they can be `undefined`? Seems like `extend` can't return `undefined`, and if it does then the method signature is wrong.

gorandalum (approved):
> those can be evaluated in a separated but they are somehow related to the react-native version.

Aight. This PR looks good, but remember that the other dependencies need to be done in the issue as well, even if they are major versions. If there is something hindering us from updating a specific dependency, then create an issue for those.

src/mobility/utils.ts
jorelosorio: This is what the lint returns after remove the condition:

```
src/mobility/utils.ts:106:46 - error TS2322: Type 'Feature<Polygon | MultiPolygon, GeoJsonProperties> | undefined' is not assignable to type 'Feature<Polygon | MultiPolygon, GeoJsonProperties>'.
  Type 'undefined' is not assignable to type 'Feature<Polygon | MultiPolygon, GeoJsonProperties>'.

106   const diff = difference(featureCollection([newAreaFeature, prevAreaFeature]));
                                                 ~~~~~~~~~~~~~~

src/mobility/utils.ts:106:62 - error TS2322: Type 'Feature<Polygon | MultiPolygon, GeoJsonProperties> | undefined' is not assignable to type 'Feature<Polygon | MultiPolygon, GeoJsonProperties>'.
  Type 'undefined' is not assignable to type 'Feature<Polygon | MultiPolygon, GeoJsonProperties>'.

106   const diff = difference(featureCollection([newAreaFeature, prevAreaFeature]));
                                                                 ~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/mobility/utils.ts:106
```

gorandalum: Aha, I see the signature from `@turf/buffer` was changed in latest version!

jorelosorio:
@gorandalum I made more updates to libraries, here more details about what is missing and some information why it is failing: https://github.com/AtB-AS/kundevendt/issues/16313#issuecomment-2283663699 ✌️

gorandalum (approved):

PR #4650, opened 2024-08-07
fix: Troms privacy policy and customer service urls

mortennordseth:
I've changed the privacy policy and customer service urls for Troms from the old fylkestrafikk.no site to the new svipper.no site.


tft-abel (approved):
Tested URLs: they link to the correct pages ✅

PR #4653, opened 2024-08-09
feat: Add support for right side action on radio sections

gorandalum:
Necessary for the delete recipient flow.

Screenshot from design-system in app:
Screenshot 2024-08-09 at 11 12 40

### DIFF
--- a/src/components/sections/items/RadioSectionItem.tsx
+++ b/src/components/sections/items/RadioSectionItem.tsx
@@ -12,7 +12,6 @@ import {PressableOpacity} from '@atb/components/pressable-opacity';
 import {RadioIcon} from '@atb/components/radio';
 import {PressableOpacityOrView} from '@atb/components/touchable-opacity-or-view';
 
-type ActionModes = 'check';
 type Props = SectionItemProps<{
   text: string;
   subtext?: string;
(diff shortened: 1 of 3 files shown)


src/components/sections/items/RadioSectionItem.tsx
```diff
@@ -22,7 +23,13 @@ type Props = SectionItemProps<{
   mode?: ActionModes;
   accessibility?: AccessibilityProps;
   color?: InteractiveColor;
+  rightAction?: {
+    icon: (props: SvgProps) => JSX.Element;
+    onPress: () => void;
+    isLoading?: boolean;
+  };
```
marius-at-atb, line 30: Can this type be extracted into a variable and be reused both here and in RadioGroupSection?

marius-at-atb (approved):
Lgtm!

src/components/sections/items/RadioSectionItem.tsx
marius-at-atb: Btw is it a bit strange that in one place rightAction is a function, and in another an object? Perhaps the names should be slightly different?

gorandalum: Good suggestion on the `itemToRightAction`, I'll do that!

PR #4656, opened 2024-08-09
feat: Delete ticket recipient

gorandalum:
Recording showing deletion, with the first delete-attempt failing (on purpose):

delete_recipient


src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/components/ExistingRecipientsList.tsx
```diff
@@ -38,6 +52,16 @@ export const ExistingRecipientsList = ({
...
   }, [recipientsQuery.status, recipientsQuery.data, onErrorOrEmpty]);
 
+  const onDelete = ({accountId}: ExistingRecipientType) => {
+    if (recipient?.accountId === accountId) {
+      onSelect(undefined);
+    }
+    mutation.mutate(accountId);
+  };
+
+  const isAtMaxRecipients =
```
marius-at-atb, line 62: When `state.settingName` is `false`, the max limit shouldn't be relevant?

marius-at-atb (approved):
Looks good! Just a couple of suggestions in the comments you can consider.

src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/components/ExistingRecipientsList.tsx
gorandalum: It is implemented as it is in Figma, where it is shown below the recipient list, disabling the possibility to send to other. This is the Figma sketch:
Screenshot 2024-08-12 at 11 33 55

But based on your question, I realise that maybe the user should be able to send to a another phone number (without saving) even if the recipient list is at max. I'll ask designer.

gorandalum: At least now things are fixed accoring to Figma.

Figma:


App (max recipients at 2 just for the screenshot):
<div>


</div>

@ckbilstad may comment on whether it should be a max recipients limit or not.

ckbilstad: I dont have a strong opinion on the exact number, but i do think we should have some kind of max amount. If the list becomes too long, it might be difficult for users to keep track of everything. A lengthy list can also slow down navigation and make it harder to find the CTA button.

marius-at-atb: Perhaps a "recipient group" could be a new feature request, intended for such use cases as an entire football team?

rosvik: @ckbilstad I'm sure the current solution will work great for 99,9% of users, so I don't think it's important to change how it works now, but I have a couple of thoughts:

> difficult for users to keep track of everything

I think a long list is easier to manage than not being able to add more users (and having to type phone numbers manually). Also the user needs to put a fair bit of effort into adding more than 10 recipients, so they are probably fine with scrolling a bit afterwards. If they think the list is too long, items are easy to remove.

> A lengthy list can also slow down navigation and make it harder to find the CTA button.

Maybe the CTA button should be fixed to the bottom of the screen then? Today it can be hidden behind the keyboard even when the list is quite short.

PR #4659, opened 2024-08-12
refactor: replace custom rounding of minutes with date-fns

rosvik:
After [the upgrade of date-fns](https://github.com/AtB-AS/mittatb-app/pull/4638), a bug in roundToNearestMinutes was fixed, allowing us to use it instead of rolling our own.

The existing tests on formatToClock, that failed with the old version of date-fns, now passes!


marius-at-atb (approved):
🎉

PR #4660, opened 2024-08-12
feat: add condition to check whether passengers should stay seated between leg interchanges. 

jonasbrunvoll:
Closes https://github.com/AtB-AS/kundevendt/issues/18758

### Background 
In https://github.com/AtB-AS/kundevendt/issues/18588, travel planner web was updated with a feature to clarify when the travellers should stay seated during a bus line number change.  Currently, the app does not distinguish between how a bus line change and an actual bus interchange is displayed. This can be confusing for the users and might lead some to get of the bus when they are supposed to stay seated.

#### Illustrations

<details>
<summary>screenshots/video/reis-nordland</summary>

| Trip view  | Detailed view |
| -------- | ------- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-08-12 at 14 25 36  |  Simulator Screenshot - iPhone SE (3rd generation) - 2024-08-12 at 14 27 29 |

</details>

### Proposed solution

Update the app to display bus line number changes the same way as currently displayed in travel planner web, to avoid confusion and to maintain equal view in app and travel planner web. 

### Acceptance Criteria
- [ ] This update should not affect how any trips are being displayed other than the one where staySeated: true.

### Test input
- [ ] Some trips from Mosjøen sentrum to Søvik are lines where staySeated: true.

### DIFF
--- a/src/api/types/generated/TripsQuery.ts
+++ b/src/api/types/generated/TripsQuery.ts
@@ -127,7 +127,12 @@ export type TripFragment = {
         notices: Array<{id: string; text?: string}>;
         journeyPattern?: {notices: Array<{id: string; text?: string}>};
       };
-      interchangeTo?: {guaranteed?: boolean; toServiceJourney?: {id: string}; maximumWaitTime?: number};
+      interchangeTo?: {
+        guaranteed?: boolean;
+        toServiceJourney?: {id: string};
+        maximumWaitTime?: number;
+        staySeated?: boolean | undefined;
+      };
       pointsOnLink?: {points?: string; length?: number};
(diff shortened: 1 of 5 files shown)


gorandalum (changes requested):
I think this looks good! You can just try to fix the minor presentation glitch on the trip search result item:

Screenshot 2024-08-12 at 15 09 04

jonasbrunvoll:
> I think this looks good! You can just try to fix the minor presentation glitch on the trip search result item:
> 
> Screenshot 2024-08-12 at 15 09 04

Will fix!

gorandalum (approved):

PR #4661, opened 2024-08-13
feat: Show recipient name

gorandalum:
Show it on the confirmation screen and on the confirmation screen
and on sent-to-others tickets.

Needed a refactor where the fetch recipients query was moved to a
place it could be reused, so it was rebranded to an
on-behalf-of-accounts-query, which is more inline with the api
naming.

<div>

</div>


rosvik (approved):
Looks good! 

Is TicketRecipientType in use now, or can it be removed completely in favor of OnBehalfOfAccountType?

gorandalum:
> Looks good!
> 
> Is TicketRecipientType in use now, or can it be removed completely in favor of OnBehalfOfAccountType?

`TicketRecipientType` is in use as typing for the recipient param we send to the purchase confirmation screen and the payment screens. It is almost the same as `OnBehalfOfAccountType`, just that the `name` is not required. I thought about making one type be derived from the other, but ended up concluding that the coupling it introduces between API types and screen params wasn't worth the duplication it avoided.

PR #4663, opened 2024-08-13
feat: Improve a11y for radio and checkbox

gorandalum:
Improves a11y for checkbox and radioboxes in select recipient flow.

Reason why this is now necessary:
https://github.com/facebook/react-native/issues/43266

### DIFF
--- a/src/components/sections/RadioGroupSection.tsx
+++ b/src/components/sections/RadioGroupSection.tsx
@@ -39,10 +39,6 @@ export function RadioGroupSection<T>({
     <Section {...props} accessibilityRole="radiogroup">
       {headerText && <HeaderSectionItem text={headerText} mode="subheading" />}
       {items.map((item: T, index) => {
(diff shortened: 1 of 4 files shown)


rosvik (approved):
Looks like a good solution to the RN bug, but I guess we should change the accessibilityRoles back to proper values once it's fixed. Will we remember to do so whenever that is?

gorandalum:
> Looks like a good solution to the RN bug, but I guess we should change the accessibilityRoles back to proper values once it's fixed. Will we remember to do so whenever that is?

Maybe, maybe not 😅 But the react-native handlings wasn't really that good either when it was working, for example it didn't have translations for the checkbox accessibility states so it said "checked"/"unchecked" even if Norwegian VoiceOver. All in all I think this is a good enough solution to not need to create an issue for improving it later. If at a later point the thought of improving this comes up we can tackle it then.

PR #4665, opened 2024-08-14
fix: allow navigation to same favorite chips in departure

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/17342

This PR will fix the issue where user couldn't navigate to the same favorite chip on the departure screen (and anywhere else). 

Other existing function should still function normally.

Also I added an offset for the refresh loading progress so it shows properly on the screen.

A question, can I use `useRef` to fix the issue on the other eslint issue in the same file?

https://github.com/user-attachments/assets/d0cbcf92-8ba2-408d-9dc1-7f856859babb

### DIFF
--- a/src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
+++ b/src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
@@ -76,6 +76,8 @@ export const NearbyStopPlacesScreenComponent = ({
     [data],
   );
 
+  const openLocationSearch = () => onPressLocationSearch(location);
+
   useEffect(() => {
     if (
       (location?.resultType == 'search' ||
@@ -85,8 +87,6 @@ export const NearbyStopPlacesScreenComponent = ({
       onSelectStopPlaceRef.current(location);
     }
   }, [location]);
-
-  const openLocationSearch = () => onPressLocationSearch(location);
 
   function setCurrentLocationAsFrom() {
     onUpdateLocation(
--- a/src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
+++ b/src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
@@ -76,6 +76,16 @@ export const NearbyStopPlacesScreenComponent = ({
     [data],
   );
 
+  useEffect(() => {
+    if (
+      (location?.resultType == 'search' ||
+        location?.resultType === 'favorite') &&
+      location?.layer === 'venue'
+    ) {
+      onSelectStopPlaceRef.current(location);
+    }
+  }, [location]);
+
   const openLocationSearch = () => onPressLocationSearch(location);
 
   function setCurrentLocationAsFrom() {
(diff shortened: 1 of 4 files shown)


src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
```diff
@@ -167,6 +162,7 @@ export const NearbyStopPlacesScreenComponent = ({
         // Quick fix for iOS to fix stuck spinner by removing the RefreshControl when not focused
         isFocused || Platform.OS === 'android' ? (
           <RefreshControl
+            progressViewOffset={90}
```
rosvik, line 165: What was the issue here, without the offset?

rosvik (approved):
Nice!

src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
reidzeibel: Hmm, I checked again and it's weird, without the offset, the refresh circle doesn't show up properly on dev builds, but in Prod it works properly.

reidzeibel: It looks like this without the offset

https://github.com/user-attachments/assets/26e01c90-6995-490a-9a90-d2e220572868

reidzeibel: Reverting this change

src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
```diff
@@ -58,6 +58,8 @@ export const NearbyStopPlacesScreenComponent = ({
 
   const screenHasFocus = useIsFocused();
 
+  const onSelectStopPlaceRef = useRef(onSelectStopPlace);
```
gorandalum, line 61: Should rather use `useCallback` on the input functions instead of `useRef` here? I know `useRef` feels safer, but I think we should rather do it the more correct React way.

src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
```diff
@@ -82,16 +84,9 @@ export const NearbyStopPlacesScreenComponent = ({
...
     }
-    /*
-     Refreshing (by dragging down) doesn't work since the location id doesn't
-     change. Take a look at fixing this when looking at the exchaustive-deps
-     error. May also consider using react-query instead of the custom reducer
-     with side effect flow.
-    */
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [location?.id]);
+  }, [location]);
```
gorandalum, line 89: I'm a little scared of this change 😬 Might be ok, but remember to test it thoroughly with saved location, search results location and my position location.

src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
reidzeibel: refactored to use `useCallback`

src/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
reidzeibel: no longer relevant after refactoring

reidzeibel:
@rosvik might want to do another look after another refactor

PR #4667, opened 2024-08-14
fix: update condition in useStations

adriansberg:
If city bikes was not enabled the useEffect getting and setting the stations is not run. The hook would also return undefined if city bikes was not enabled.

Updated the condition to check if either city bikes or car sharing is enabled.

Fixes https://github.com/AtB-AS/kundevendt/issues/18797

### DIFF
--- a/src/mobility/use-stations.tsx
+++ b/src/mobility/use-stations.tsx
@@ -41,6 +41,8 @@ export const useStations: (
 
   const [stations, setStations] = useState<StationFeatures>(emptyStations);
 
+  const enableStations = isCityBikesEnabled || isCarSharingEnabled;
+
   useEffect(() => {
     if (!initialFilter) {
       // Initial filter === undefined, use user's filter from store.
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):
LGTM 👍🏼

marius-at-atb (approved):
Nice!
Btw did you check if there are any other places in the app which also only uses one of `isCityBikesEnabled` and `isCarSharingEnabled`?

adriansberg:
Did a search now and by the usages I see this was the only place `isCityBikesEnabled` was used where `isCarSharingEnabled` should be checked alongside it. I only saw one other usage and that was in the mobility filters where it used to decide if the form factor filter should be shown or not.

gorandalum:
Remember to update the acceptance criteria in the original issue. The acceptance criteria is not really that Otto should be shown in the Reis 0app, but that car-sharing stations can be shown even if bike-stations not enabled, and vice versa.

adriansberg:
@gorandalum, updated the acceptance criterias accordingly.

PR #4668, opened 2024-08-15
Initial implementation new variable structure

tft-abel:
Shows the changes that will be made to incorporate the new design token/Figma variables structure.

Let's review these changes briefly now and decide if this is the right direction, or if we should pivot to a new method.

We can also discuss if/how we can split up this PR, since you can imagine there are a lot of files that will be touched this way. I would almost suggest merging this into a `master-new-vars` branch where we merge these changes into, and when we are `done` and have tested it, that we merge that branch into actual `master`. Would love to hear your thoughts on this.

Pros and cons with this approach:

**Advantages**
- Direct use of the `theme` object means that there is no "magic" happening anywhere anymore

**Disadvantages**
- We need to use `useTheme` in a lot of places, this might have (performance) side effects

Other options we can consider
- Allows colors being entered as `Background.0` and later use that key to access the actual color of that key (using for example `lodash.get`). This way the consumer of components does not need to have knowledge of `theme`, but can just send in string (with TypeScript to know that it is a path of keys of the `theme` object)

_______________

Changelog

- `<ThemeIcon>`: Aligned fill with `ThemeText`s `color` prop, with extension of `Statuses`. Removed `ColorType` and `fill` props.
- `<ThemeText>`: Changed type of `color` prop


src/anonymous-purchase-consequences-screen/AnonymousPurchaseConsequencesScreenComponent.tsx
```diff
@@ -29,9 +26,12 @@ export const AnonymousPurchaseConsequencesScreenComponent = ({
   leftButton,
 }: Props) => {
   const styles = useStyle();
+  const { theme } = useTheme();
   const {t} = useTranslation();
   const focusRef = useFocusOnLoad();
 
+  const themeColor = theme.Background.Accent[0];
```
marius-at-atb, line 33: In cases like this, `themeColor` should be passed in as an argument to `useStyle`. See e.g. `useStyles` in `TravelTokenBox.ts` for reference.

marius-at-atb (commented):
I like that there is more explicit usage and less magic :+1:
Using the theme hook in more places shouldn't cause noticeable performance changes I reckon.
Also since there will be many lines changed, I agree it is a good idea to create a new branch to merge PRs into, and then finally merge that into master.

gorandalum (commented):
I think this looks good! Nothing major to comment on.

Some minor/optional comments:
- Not using camel case hurts my eyes a little, and might at some times be confusing as capital letters are mostly used for components/classes. Could it be possible to lower case the first letter when mapping the response from Figma REST API?
- Maybe we should use `contrastColor` everywhere for the color field name to components. Now we have `color`, `colorType`, `themeColor` going around, and I think it would be nice to unite them. 
- Regarding `background[0], I kinda have thought of "background_0" as a proper noun for that background, and not the first background in a list of backgrounds. I think having proper nouns is slightly better if we want to do some syntactic sugar at a later point to make it easier/shorter to specify colors like we had from before.
- It would still be nice if we could specify the theme color as a const in the top of a file, so it is just one place to specify it and it is changed throughout the components/styles in the file.

tft-abel:
> I think this looks good! Nothing major to comment on.
> 
> Some minor/optional comments:
> 
> * Not using camel case hurts my eyes a little, and might at some times be confusing as capital letters are mostly used for components/classes. Could it be possible to lower case the first letter when mapping the response from Figma REST API?
> * Maybe we should use `contrastColor` everywhere for the color field name to components. Now we have `color`, `colorType`, `themeColor` going around, and I think it would be nice to unite them.
> * Regarding `background[0], I kinda have thought of "background_0" as a proper noun for that background, and not the first background in a list of backgrounds. I think having proper nouns is slightly better if we want to do some syntactic sugar at a later point to make it easier/shorter to specify colors like we had from before.
> * It would still be nice if we could specify the theme color as a const in the top of a file, so it is just one place to specify it and it is changed throughout the components/styles in the file.

- CamelCase: I agree, that felt wrong. I'll change the script
- ContrastColor everywhere, yes!
- I can change the leafs to be `neutral0`, `neutral1`, `accent0` etc., that might be a good middle ground not having numbers, but not repeating a key that's nested a few levels higher either
- The thing with the variables at the top of the file is that we do not have access to the `theme` object there., but using the `neutral0` approach, we can at least reference all background colors locally in components using a string

gorandalum:
We can skip the point about statically referencing a color at the top of the file, this will probably be easier to do if we at some point add some syntactic sugar for this. It could be possible with a const func at the top of the file (like `theme => theme.background.neutral0`), but probably not worth it for now.

I like the approach with `neutral0`, `accent0` and `accent1` as leaf names. As they currently are presented in Figma I think the neutral and accent colors belong side by side (not with sub-collections `neutral` and `accent`) as they all irect leafs of static background colors.

tft-abel:
Commit changes for the already changed files to now use `camelCase`, and a function/const to define the `themeColor` for a component at the top of the file.

Discussing the changing of leaf nodes in `background` with @ckbilstad

tft-abel:
We now created a compatibility mode, such that these changes are not required

PR #4669, opened 2024-08-15
fix: Wrong error message when no recipient name

gorandalum:
This was caused by retrieving the recipients directly with the
getQueryData function, which doesn't apply the select function on
the returned data. Instead changed this to use the actual query.

https://github.com/AtB-AS/kundevendt/issues/17226#issuecomment-2290957485

Resolves https://github.com/AtB-AS/kundevendt/issues/17226


marius-at-atb (approved):
Lgtm

PR #4672, opened 2024-08-16
build: Split update devices job to be run in serial

jorelosorio:
There was an issue while update devices ran at the same time to update provisioning profiles, it was separated in order to make it serialized, first development then adhoc.

Before:

image

Now:

image

### DIFF
--- a/.github/actions/update-devices/action.yml
+++ b/.github/actions/update-devices/action.yml
@@ -28,9 +28,11 @@ runs:
         env:
           GIT_CRYPT_KEY: ${{ inputs.git-crypt-key }}
       - name: Copy Development files
+        shell: bash
         if: ${{ inputs.app-environment=='development' }}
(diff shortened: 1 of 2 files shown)


reidzeibel:
Changes looks fine, but shouldn't this be a `chore` instead of `fix`? because this is not a fix for the app but a chore to fix the build flow.

jorelosorio:
> `chore` instead of `fix`

I named it fix because even that the job ends successful, it was not able to push the changes in the git repo for new devices, so when a build was made the new device was not found because the provisioning profiles do not have the updated device.

reidzeibel (approved):

rosvik:
There's also `build` 🤓 https://github.com/AtB-AS/mittatb-app/blob/master/CONTRIBUTING.md#not-included-in-changelog

tormoseng:
Nice! Test: https://github.com/AtB-AS/mittatb-app/actions/runs/10417698159

PR #4679, opened 2024-08-19
feat: support multiple PreassignedFareProducts per offer search

rosvik:
This adds support for a new field in reference data, `productAliasId`. When doing offer searches, ids for all PreassignedFareProducts with the same productAliasId will be provided (named `preassignedFareProductAlternatives` in the code).

This does in theory allow more than one offer to be returned from the offer search, but in practice tickets should be configured to not overlap (e.g. for Troms, the Saturday boat ticket should apply only to Saturday, and the regular ticket should apply outside of Saturday). If there for some reason is more than one result, we use the cheapest option as a fallback.

closes https://github.com/AtB-AS/kundevendt/issues/18693

### DIFF
--- a/src/configuration/utils.ts
+++ b/src/configuration/utils.ts
@@ -59,3 +59,12 @@ export const isProductSellableInApp = (
 
   return product.distributionChannel.some((channel) => channel === 'app');
 };
+
+export const removeProductAliasDuplicates = (
+  val: PreassignedFareProduct,
+  i: number,
+  arr: PreassignedFareProduct[],
+): boolean => {
+  if (val.productAliasId === undefined) return true;
+  return arr.map((p) => p.productAliasId).indexOf(val.productAliasId) === i;
+};
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
@@ -59,13 +59,13 @@ const getOfferForTraveller = (offers: Offer[], userTypeString: string) => {
   const offersForTraveller = offers.filter(
     (o) => o.traveller_id === userTypeString,
   );
+
+  // If there are multiple offers for the same traveller, use the cheapest one.
+  // This shouldn't happen in practice, but it's a sensible fallback.
   const offersSortedByPrice = offersForTraveller.sort(
     (a, b) =>
       getCurrencyAsFloat(a.prices, 'NOK') - getCurrencyAsFloat(b.prices, 'NOK'),
   );
-
-  // If there are multiple offers for the same traveller, use the cheapest one.
-  // This shouldn't happen in practice, but it's a sensible fallback.
   return offersSortedByPrice[0];
 };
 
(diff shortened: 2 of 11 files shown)


gorandalum (commented):
Good work! Some questions comes up when looking at this:
- There probably needs to be some code hindering products with same product alias to show up individually in product selector?
- There probably needs to be talk about how recents tickets will work with this. For example pressing buy again on a "Saturday single ticket" won't get you to a screen prefilled with Saturday as travel date. It would be good to mention this to those interested (designers, product owner, OMS) so it doesn't come as a surprise when it goes into production.
- Tickets which have specific prices on specific times probably shouldn't have the activate now mechanism, or at least it should be up to the transport company if they want to take that risk. Since users can then buy discounted tickets and activate them at times there would normally be no discount. So we probably need it configurable on each product whether it has the "activate now" functionality.

rosvik:
> There probably needs to be some code hindering products with same product alias to show up individually in product selector?

Good point! Added now, for ProductSelectionByProducts and ProductSelectionByAlias.

> There probably needs to be talk about how recents tickets will work with this.

Needed a small fix to make it work: 8ddedad97fbe87a1ac6c297c06a7042ec52f8289

Discussion here: https://mittatb.slack.com/archives/CSFC6PG23/p1724157465707539

> Tickets which have specific prices on specific times probably shouldn't have the activate now mechanism

This is already the case today, (you can buy a Saturday ticket and activate it in advance) so I think it's something for another task/PR. But this change will probably make the problem more visible, so it's a good time to figure out a solution.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
```diff
@@ -55,16 +55,28 @@ const getValidDurationSeconds = (offer: Offer): number | undefined =>
...
     : undefined;
 
+const getOfferForTraveller = (offers: Offer[], userTypeString: string) => {
+  const offersForTraveller = offers.filter(
+    (o) => o.traveller_id === userTypeString,
+  );
+
+  // If there are multiple offers for the same traveller, use the cheapest one.
+  // This shouldn't happen in practice, but it's a sensible fallback.
+  const offersSortedByPrice = offersForTraveller.sort(
```
jorelosorio, line 65: What if the offer I am looking for is the expensive one? It might not be part of this PR but would be nice to have a selector for complex cases.

jorelosorio (approved):
Nice work :)

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
rosvik: Shouldn't happen in practice, so it shouldn't matter. It's there just because it probably makes "more" sense to pick the cheap one if we get two offers by mistake :)

src/configuration/utils.ts
```diff
@@ -59,3 +59,12 @@ export const isProductSellableInApp = (
 
   return product.distributionChannel.some((channel) => channel === 'app');
 };
+
+export const removeProductAliasDuplicates = (
```
gorandalum, line 63: May be able to use `onlyUniquesBasedOnField` with slight modification

rosvik: Could solve it like this, I guess: https://github.com/AtB-AS/mittatb-app/pull/4690

PR #4683, opened 2024-08-20
chore: Update Release.md with version number checking

reidzeibel:
So that people don't make the same mistake that I did....


gorandalum (approved):

marius-at-atb (approved):
:+1:

PR #4684, opened 2024-08-21
fix: improve handling of redirection when adding payment card

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/18722

Relevant Slack thread : https://mittatb.slack.com/archives/C02EEG7D8EL/p1723717797085529

This PR will change how we handle the callback URL after finishing the payment flow.

The old flow:
1. Submit card information
2. Auth from the card provider (if any, e.g. Bank ID, Mastercard, etc.)
3. Fetch the callback URL once auth successful
4. Redirect to Profile page

The new flow:
1. Submit card information
2. Auth from the card provider (if any, e.g. Bank ID, Mastercard, etc.)
3. Fetch the callback URL once auth successful
4. Add `window.location = ` to the URL, then save it to a variable called `callbackUrl`.
5. Send the `callbackUrl` to the `WebView` when `appState === 'active'`.

Video of the new flow: https://mittatb.slack.com/archives/C02EEG7D8EL/p1724166760528719?thread_ts=1723717797.085529&cid=C02EEG7D8EL

### DIFF
--- a/src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
+++ b/src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
@@ -1,6 +1,6 @@
 import {FullScreenHeader} from '@atb/components/screen-header';
-import React, {useEffect, useState} from 'react';
-import {View} from 'react-native';
+import React, {useEffect, useRef, useState} from 'react';
+import {AppState, View} from 'react-native';
 import {StyleSheet} from '@atb/theme';
 import {useAddPaymentMethod} from '@atb/stacks-hierarchy/Root_AddPaymentMethodScreen/use-add-payment-method';
 import {dictionary, useTranslation} from '@atb/translations';
@@ -10,6 +10,8 @@ import AddPaymentMethodTexts from '@atb/translations/screens/subscreens/AddPayme
 import WebView from 'react-native-webview';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy';
 import {Processing} from '@atb/components/loading';
+import {WebViewNavigationEvent} from 'react-native-webview/lib/RNCWebViewNativeComponent';
+import {useAppStateStatus} from '@atb/utils/use-app-state-status';
 
 type Props = RootStackScreenProps<'Root_AddPaymentMethodScreen'>;
 
@@ -17,6 +19,10 @@ export const Root_AddPaymentMethodScreen = ({navigation}: Props) => {
   const styles = useStyles();
   const {t} = useTranslation();
   const [showWebView, setShowWebView] = useState<boolean>(true);
+  const [callbackUrl, setCallbackUrl] = useState<string>('');
+  const webViewRef = useRef<WebView>(null);
+  const appState = useAppStateStatus();
+
   const {
     terminalUrl,
     onWebViewLoadStart,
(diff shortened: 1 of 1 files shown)


gorandalum (commented):
Is this a problem only when adding payment card, and not a problem when doing purchases?

reidzeibel:
@gorandalum yes, only when adding payment card, the webview component is different

gorandalum:
When I look at the differences between the purchase callback handling and the add payment method callback handling, it seems that the former uses `Linking.addEventListener`, while the latter uses linking set up in the NavigationContainer. Could we try using `Linking.addEventListener` also on the add payment method callback and see if that works? 🤔

reidzeibel:
Seems to be working on dev, will try on prod with the release candidate test

reidzeibel:
The fix based on your suggestion is here @gorandalum : https://github.com/AtB-AS/mittatb-app/commit/b10e957419b2fd7381e2e7ef938e998158557236

src/stacks-hierarchy/Root_AddPaymentMethodScreen/Root_AddPaymentMethodScreen.tsx
```diff
@@ -31,6 +37,56 @@ export const Root_AddPaymentMethodScreen = ({navigation}: Props) => {
...
   );
 
+  useEffect(() => {
+    /**
+     * Listens to app state change, if app comes back from backgrounded/inactive
+     * state, it will then trigger the webview to load the {@link callbackUrl}, 
+     * which will redirect the user to the main payment card screen, and triggers
+     * a reload, so the user can see their added card(s). 
+     */
+    const subscription = AppState.addEventListener('change', (nextAppState) => {
```
jorelosorio, line 47: Because it is a listener, I think it could be nice to move out of the useEffect, and add as a dependency the app state to then inject the JS. What do you think?

reidzeibel: move out this part from `useEffect`?

```
const subscription = AppState.addEventListener('change', (nextAppState) => {
      if (appState.match(/inactive|background/) && nextAppState === 'active') {
        webViewRef.current?.injectJavaScript(callbackUrl);
      }
    });
```

jorelosorio: Yes, it's listening for events without need for a useEffect

reidzeibel: Hmm, should it though?

Since, if we create the const inside `useEffect`, then it will only be re-created when the `useEffect` is triggered, instead of every render. Also, many of the similar usages (the `addEventListener` part) in our code is created inside a `useEffect`. Moreover the official documents for `AppState` usage in https://reactnative.dev/docs/appstate, shows that in their sample code, the `const` part is created inside a `useEffect`. 

I'm not entirely sure, I think creating it inside the `useEffect` makes more sense because of the reason I mentioned above. Maybe @gorandalum could elaborate more on this? 😁

gorandalum: All subscriptions/listeners need to be created inside a `useEffect`, or else they will be resubscribed every render.

gorandalum (approved):

reidzeibel:
Use this PR instead https://github.com/AtB-AS/mittatb-app/pull/4687

PR #4690, opened 2024-08-22
refactor: rely on onlyUniquesBasedOnField for productAliasId

rosvik:
https://github.com/AtB-AS/mittatb-app/pull/4679#discussion_r1726884593


reidzeibel (approved):
More intuitive in a glance!

gorandalum (approved):
Also an option to do a change in the `onlyUniquesBasedOnField` function to skip the `removeProductAliasDuplicates` all together. But then we should have some unit tests, so we can wait until we see the need for this more places in the code. 
```ts
export const onlyUniquesBasedOnField =
  <T>(field: keyof T, treatUndefinedAsUnique?: boolean) =>
  (element: T, index: number, array: T[]) =>
    array.findIndex(
      (el) =>
        (treatUndefinedAsUnique && !el[field]) || el[field] === element[field],
    ) === index;
```

rosvik:
@gorandalum That makes sense! I updated it now, with some slight tweaks to the logic to make it work.

__tests__/utils.test.ts
```diff
@@ -153,6 +153,32 @@ describe('Function onlyUniqueBasedOnField', () => {
...
+    const deduped2 = testData.filter(onlyUniquesBasedOnField('field1', false));
+    expect(deduped2.length).toEqual(3);
+  });
+
+  it('does not remove undefined if treatUndefinedAsUnique is true', () => {
+    const testData = [
+      {field1: 'id-1'},
+      {field1: undefined},
+      {field1: undefined},
+    ];
```
gorandalum, line 178: Could add a `{field1: 'id-2'}` here for good measure

gorandalum (approved):

PR #4693, opened 2024-08-26
fix: add boat-carnet ticket illustration type

rosvik:
This (in addition to changing fare-product-type-config) makes boat carnets show the boat illustration instead of bus.

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/TicketingTileIllustration.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/TicketingTileIllustration.tsx
@@ -68,6 +68,8 @@ const getIllustrationFileName = (
       return 'Ticket';
     case 'boat-period':
       return 'PeriodTicket';
+    case 'boat-carnet':
+      return 'TicketMultiple';
     case 'ticketMultiple':
       return 'TicketMultiple';
     case 'city':


rosvik:
This should probably be part of the release (cc. @hanne-at-atb / @tormoseng)

reidzeibel (approved):

marius-at-atb (approved):

gorandalum (commented):
Note that there really isn't any need to add a new illustration key `boat-carnet` for this. The boat carnet could just use `carnet` as `illustration`, and would get the correct image.

rosvik:
@gorandalum I realize I didn't do a great job explaining the issue here. The problem was the inspection symbol, that uses `fareProductTypeConfig?.illustration?.includes('boat')` to show the correct transport icon.

https://github.com/atB-AS/mittatb-app/blob/master/src/fare-contracts/components/InspectionSymbol.tsx#L101

gorandalum:
> @gorandalum I realize I didn't do a great job explaining the issue here. The problem was the inspection symbol, that uses `fareProductTypeConfig?.illustration?.includes('boat')` to show the correct transport icon.
> 
> https://github.com/atB-AS/mittatb-app/blob/master/src/fare-contracts/components/InspectionSymbol.tsx#L101

Whoa! Then that is what should be fixed instead. Should instead rely on `transportModes`.

hanne-at-atb:
Tested ok in staging.

PR #4697, opened 2024-08-28
feat: remove TermsInformationScreen

rosvik:
Follow up to https://github.com/AtB-AS/kundevendt/issues/18748


marius-at-atb:
Instead of an external link, could we use an in-app browser instead?
For some new users, their first click will take them out of the app, which isn't great onboarding imo 😅

marius-at-atb:
Also I'm not a big fan of the gap between the text and icon here. Perhaps the icon can be omitted if using in-app browser?

rosvik:
@marius-at-atb Good idea! I'll look into it

rosvik:
@marius-at-atb Updated now!

https://github.com/user-attachments/assets/539b2a65-e2c1-42de-bf5f-724aaf0e6b38

rosvik:
@AtB-AS/design (Does tagging designers like this work?)

I'm a bit uncertain about how we should mark links/buttons that navigate to a in-app-browser like in the recording above. 

1. Should we keep the "external link" icon?
2. Should screen readers treat it as a "link" or "button"?

marius-at-atb (approved):
Nice!

ckbilstad:
> @AtB-AS/design (Does tagging designers like this work?)

No 😅

rosvik:
Hmm, let me try again, then.

🔔⚠️🚨 @ckbilstad @ckbilstad @ckbilstad IMPORTANT QUESTIONS ABOVE 🚨⚠️🔔

ckbilstad:
The in app browser is 🔥

Should we keep the "external link" icon?
- In my opinion, If you're using an in-app browser, you don't necessarily need an external icon on the link since users aren't leaving the app. 

Should screen readers treat it as a "link" or "button"?
- @Aliaaen probably has knowledge about this after doing the accessibility statement recently

I would love to have an underline or something else that shows the string is actually clickable, but i guess thats a task for another day.

ckbilstad:
> Hmm, let me try again, then.
> 
> 🔔⚠️🚨 @ckbilstad @ckbilstad @ckbilstad IMPORTANT QUESTIONS ABOVE 🚨⚠️🔔

Now, this is best practice 🥇

Aliaaen:
In-app browser is a definite improvement! 🔥 

The weird, anonymous link button thing is one of my biggest pet peeves with the app right now 😅 
Technically speaking it should be a link, as it leads to content and does not perform an action. We don't have a functional link component or guidelines right now though so we keep using tertiary "buttons".

WCAG-wise this is probably in breach unless we **at least** use an icon. Ideally we want icon+underline+colorful text, but Christian is right - that's another issue. I'll make one.

Icon-wise I think the "external link" one is a good fit because the content is from somewhere else and opening an in-app browser is still a context switch.

rosvik:
@Aliaaen Thanks! Would be really nice for development to have a consistent way to do links as well. 

For now, I'll keep the icon and screen reader "link" role, then.

jorelosorio (approved):

PR #4698, opened 2024-08-29
chore: Patch Mobile Token SDK iOS to improve error messages

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/18559

More details about why this, in the issue messages.


gorandalum (approved):
OK as a quick fix now. But remember to also add PRs for the mobiletoken sdk, so it is fixed there.

PR #4701, opened 2024-09-03
refactor: refactor usePreviousPaymentMethods

rosvik:
By passing `recurringPayments` as a prop to SelectPaymentMethodSheet, we don't have to do an extra API call to fetch the list of saved payments. This removes the loading state from the page, and means we can remove some state logic.

Also refactored the usePreviousPaymentMethods hook a bit.

related to https://github.com/AtB-AS/kundevendt/issues/18885

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
@@ -1,12 +1,12 @@
 import React, {useEffect, useState} from 'react';
-import {ActivityIndicator, ScrollView, View} from 'react-native';
-import {StyleSheet, useTheme} from '@atb/theme';
+import {ScrollView, View} from 'react-native';
+import {StyleSheet} from '@atb/theme';
 import {Button} from '@atb/components/button';
 import {PurchaseConfirmationTexts, useTranslation} from '@atb/translations';
 import {ArrowRight} from '@atb/assets/svg/mono-icons/navigation';
 import {ThemeText} from '@atb/components/text';
 import SelectPaymentMethodTexts from '@atb/translations/screens/subscreens/SelectPaymentMethodTexts';
-import {listRecurringPayments, PaymentType} from '@atb/ticketing';
+import {PaymentType, RecurringPayment} from '@atb/ticketing';
 import {PaymentMethod, SavedPaymentOption} from '../../types';
 import {useAuthState} from '@atb/auth';
 import {BottomSheetContainer} from '@atb/components/bottom-sheet';
@@ -20,6 +20,7 @@ import {PressableOpacity} from '@atb/components/pressable-opacity';
 type Props = {
   onSelect: (value: PaymentMethod) => void;
   previousPaymentMethod?: SavedPaymentOption;
+  recurringPayments?: RecurringPayment[];
 };
 
 function getSelectedPaymentMethod(
@@ -65,19 +66,13 @@ function isRecurring(option: PaymentMethod): option is {
   );
 }
 
-const remotePaymentOptions: SavedPaymentOption[] = [];
-
 export const SelectPaymentMethodSheet: React.FC<Props> = ({
   onSelect,
   previousPaymentMethod,
+  recurringPayments,
 }) => {
   const {t} = useTranslation();
-
-  const [loadingRecurringOptions, setLoadingRecurringOptions] =
-    useState<boolean>(true);
-
   const {authenticationType} = useAuthState();
-  const {theme} = useTheme();
   const {paymentTypes} = useFirestoreConfiguration();
 
   const defaultPaymentOptions: SavedPaymentOption[] = paymentTypes.map(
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -92,30 +87,28 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
-        recurringCard: {
-          id: option.id,
-          masked_pan: option.masked_pan,
-          expires_at: option.expires_at,
-          payment_type: option.payment_type,
-        },
-      };
-    });
-    return [...recurringOptions.reverse()];
+    const recurringOptions: Array<SavedPaymentOption> | undefined =
```
marius-at-atb, line 94:
```suggestion
    const recurringOptions: Array< RecurringPaymentOption > | undefined =
```
Since it never returns `DefaultPaymentOption`, it is more precise to return `RecurringPaymentOption`.

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -92,30 +87,28 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
+      recurringPayments?.map((option) => {
+        return {
+          savedType: 'recurring',
+          paymentType: option.payment_type,
+          recurringCard: {
+            id: option.id,
+            masked_pan: option.masked_pan,
+            expires_at: option.expires_at,
+            payment_type: option.payment_type,
+          },
```
marius-at-atb, line 104:
```suggestion
          recurringCard: {...option},
```
Can be simplified

src/stacks-hierarchy/saved-payment-utils.ts
```diff
@@ -4,42 +4,31 @@ import {storage} from '@atb/storage';
...
-    }
-
     if (!userId) {
-      setPaymentMethod(undefined);
-      return;
+      setPreviousPaymentMethod(undefined);
     } else {
-      run(userId);
+      getPreviousPaymentMethodByUser(userId).then((method) => {
+        setPreviousPaymentMethod(method);
```
marius-at-atb: Right here just before setting it, might be a good place to ensure that this is still a valid option? I guess this is what @reidzeibel has already commented on.

src/stacks-hierarchy/saved-payment-utils.ts
```diff
@@ -4,42 +4,31 @@ import {storage} from '@atb/storage';
...
-    SavedPaymentOption | undefined
-  >(undefined);
+export function usePreviousPaymentMethods(): {
+  recurringPayments: RecurringPayment[] | undefined;
+  previousPaymentMethod: SavedPaymentOption | undefined;
+} {
   const {userId} = useAuthState();
-
   const {data: recurringPayments} = useListRecurringPaymentsQuery();
+  const [previousPaymentMethod, setPreviousPaymentMethod] =
```
marius-at-atb, line 15: I find it confusing that this is called method, when it is an option.

rosvik:
@reidzeibel Good catch! It was a bit hard to understand what happened there, so added a comment for the next person that reads through this 😄

src/stacks-hierarchy/saved-payment-utils.ts
marius-at-atb: Edit: I see this has just been addresed in the new commit

rosvik:
@marius-at-atb There are lots of confusing names/types and logic in this part of the app, and I have a PR in the works that will resolve much of it here: https://github.com/AtB-AS/mittatb-app/compare/rosvik/update-recurring-payments...rosvik/refactor-payment-types?expand=1

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
rosvik: This will change in https://github.com/AtB-AS/mittatb-app/compare/rosvik/update-recurring-payments...rosvik/refactor-payment-types?expand=1

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
rosvik: It will all be `PaymentOption` after https://github.com/AtB-AS/mittatb-app/compare/rosvik/update-recurring-payments...rosvik/refactor-payment-types?expand=1

marius-at-atb:
Ahh, great! Looking forward to that 😄

src/stacks-hierarchy/saved-payment-utils.ts
rosvik: Yup, I missed a bit here. This will _also_ change in https://github.com/AtB-AS/mittatb-app/compare/rosvik/update-recurring-payments...rosvik/refactor-payment-types?expand=1 😄

marius-at-atb (approved):

reidzeibel (approved):
Nice, I tested the behaviour for saved cards and it works as intended.

src/stacks-hierarchy/saved-payment-utils.ts
marius-at-atb: Btw just to mention it - the current validation ignores Vipps, which means that right now you never get Vipps preselected? Would be a nice feature to add.

rosvik: True! Good catch, me fix.

PR #4703, opened 2024-09-03
refactor: unify payment types to PaymentMethod

rosvik:
There was a lot of redundant logic in the purchase flow that could be removed by having more sensible types for payment options.

See the changes in src/stacks-hierarchy/types.ts for my proposed solution to this.
- Combines SavedPaymentOption, RecurringPaymentOption, DefaultPaymentOption, and PaymentMethods into a shared type "PaymentMethod"

Should be no functional changes

related to https://github.com/AtB-AS/kundevendt/issues/18885 / follow up to PR https://github.com/AtB-AS/mittatb-app/pull/4701


src/stacks-hierarchy/types.ts
```diff
@@ -1,36 +1,23 @@
...
   payment_type: number;
 };
 
-export type DefaultPaymentOption = {
-  savedType: 'normal';
+export type PaymentOption = {
+  savedType: 'normal' | 'recurring';
   paymentType: PaymentType;
+  recurringCard?: SavedRecurringPayment;
 };
```
gorandalum, line 20: I'm not completely sold 😅  I prefer verbosity if it gives more precision in the typing. With this change an object like this is possible:
```ts
{
  savedType: 'recurring',
  paymentType: 'Vipps',
  recurringCard: undefined,
}
```

rosvik: I would prefer to remove `savedType` as well, but since it's in local storage, I'm hesitant to touch it. We could maybe mark it as deprecated or something like that.

> I'm not completely sold 

Let me try to sell it, then 😄 

Let's compare it to fare products. What do you think is better?

```ts

type NormalFareProduct = {
  type: 'single' | 'period';
  ...
};
type CarnetFareProduct = {
  id: 'carnet';
  ...
  usedAccesses: Access[];
};
type FareProduct = CarnetFareProduct | NormalFareProduct;
```

or 


```ts
type FareProduct = {
  id: string;
  ...
  usedAccesses: Access[] | undefined;
};
```

One is more data driven. 

- If it has a usedAccesses, we show usedAccesses related UI, otherwise they are all fare products.
- If it has a recurringCard, we show recurringCard related UI, otherwise they are all payment options.

I also appreciate the intention of `SavedPaymentOption = DefaultPaymentOption | RecurringPaymentOption`, but in practice it leads to unnecessary checks for `if ('recurringCard' in paymentOption)` when it could have been a `paymentOption.recurringCard?`.

marius-at-atb: Is a middle ground possible, simply adding `recurringCard: undefined;` to the definition of `DefaultPaymentOption`? So that invalid states remain impossible, but the need for extra `'recurringCard' in paymentOption` checks is removed.

src/stacks-hierarchy/navigation-types.ts
```diff
@@ -86,7 +86,8 @@ type Root_PurchaseAsAnonymousConsequencesScreenParams = {
 type Root_PurchasePaymentScreenParams = {
   offers: ReserveOffer[];
   preassignedFareProduct: PreassignedFareProduct;
-  paymentMethod: PaymentMethod;
+  paymentMethod: PaymentOption;
+  savePaymentMethod: boolean;
```
marius-at-atb, line 90: Can `savePaymentMethod` be prefixed with `should` to distinguish it from a function that saves it?

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -461,11 +417,11 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
                       'Ticketing',
                       'Pay with previous payment method clicked',
                       {
-                        paymentMethod: previousMethod?.paymentType,
+                        paymentMethod: previousPaymentOption?.paymentType,
```
marius-at-atb, line 420: I think this line shows why it is easy to get confused here.
`method`, `option` and `type` - what is the current relationship between these terms, and what should it be? 🤔

src/stacks-hierarchy/Root_PurchasePaymentScreen/use-reserve-offer-mutation.ts
```diff
@@ -1,42 +1,34 @@
...
 import {useAuthState} from '@atb/auth';
-import {PaymentMethod, TicketRecipientType} from '@atb/stacks-hierarchy/types';
+import {PaymentOption, TicketRecipientType} from '@atb/stacks-hierarchy/types';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import {FETCH_ON_BEHALF_OF_ACCOUNTS_QUERY_KEY} from '@atb/on-behalf-of/queries/use-fetch-on-behalf-of-accounts-query.ts';
 
 type Args = {
   offers: ReserveOffer[];
-  paymentMethod: PaymentMethod;
+  paymentMethod: PaymentOption;
```
marius-at-atb, line 10: Method or option? 😅

src/stacks-hierarchy/types.ts
rosvik: Huh, I think that will work! 💡 This gives no typescript errors on my branch:

```ts
type CardPaymentOption = {
  savedType: 'recurring' | 'normal';
  paymentType: PaymentType.Mastercard | PaymentType.Visa;
  recurringCard?: RecurringCard;
};
type VippsPaymentOption = {
  savedType: 'normal';
  paymentType: PaymentType.Vipps;
  recurringCard?: undefined;
};
export type PaymentOption = CardPaymentOption | VippsPaymentOption;
```

src/stacks-hierarchy/saved-payment-utils.ts
```diff
@@ -1,55 +1,85 @@
...
 } {
   const {userId} = useAuthState();
   const {data: recurringPayments} = useListRecurringPaymentsQuery();
-  const [previousPaymentMethod, setPreviousPaymentMethod] =
-    useState<SavedPaymentOption>();
+  const [previousPaymentOption, setPreviousPaymentOption] =
+    useState<PaymentOption>();
+  const {paymentTypes} = useFirestoreConfiguration();
+
+  const isValidPaymentOption = useCallback(
```
marius-at-atb, line 20: Really nice function 😎

marius-at-atb (commented):
I like what this is doing!
Still some confusion on my end though. Do we still need to use the term `Methods` at all?

src/stacks-hierarchy/navigation-types.ts
rosvik: Sure, updated now 👍

rosvik:
@marius-at-atb Good input! I previously thought we needed both a PaymentOption and a PaymentMethod, but now that we have one type to rule them all, it does make sense to use `PaymentMethod`, since that's what used in API calls, and in the local storage key.

src/stacks-hierarchy/types.ts
gorandalum: Nice, we got bag and backpack!

gorandalum (approved):

src/stacks-hierarchy/saved-payment-utils.ts
```diff
@@ -1,51 +1,81 @@
...
+        isValidPaymentMethod(storedMethod) ? storedMethod : undefined,
+      );
+    });
+  }, [userId, isValidPaymentMethod]);
+
+  const recurringPaymentMethods = recurringPayments?.map(
+    (recurringPayment): PaymentMethod => ({
+      savedType: 'recurring',
+      paymentType: recurringPayment.payment_type,
+      recurringCard: recurringPayment,
```
marius-at-atb, line 64: While we're at it - should we do something about this one as well?

Skjermbilde 2024-09-06 kl  13 04 27
Skjermbilde 2024-09-06 kl  13 04 44

src/stacks-hierarchy/types.ts
```diff
@@ -1,36 +1,23 @@
...
   expires_at: string;
   masked_pan: string;
   payment_type: number;
 };
 
-export type DefaultPaymentOption = {
-  savedType: 'normal';
-  paymentType: PaymentType;
+type CardPaymentMethod = {
+  savedType: 'recurring' | 'normal';
```
marius-at-atb, line 11: Could have a `savedType` as an enum, with explanations as well ("normal" isn't the most self-explanatory type ever 😛)

marius-at-atb (approved):
Nice and clean 🧹 🎉

src/stacks-hierarchy/saved-payment-utils.ts
rosvik: Yes, nice! Getting rid of RecurringCard 🔥

src/stacks-hierarchy/types.ts
rosvik: Sure, does something like this work? 

```ts
export enum SavedPaymentMethodType {
  /**
   * Contains only the type of payment, and no payment details. When used, the
   * user has to enter the payment details manually or pay through a third party
   * like Vipps.
   */
  Normal = 'normal',
  /**
   * Contains the type of payment and information to enable automatic payment
   * without entering more payment details. BankID might still be required.
   */
  Recurring = 'recurring',
}
```

marius-at-atb: Great!

PR #4704, opened 2024-09-03
feat: dynamically set transport modes for harbors query based on fare…

mortennordseth:
… product type

To let Troms be able to sell express boat tickets for ferry lines, we need to add the option to configure which harbors are listed. This configuration can be determined by the fare product type, defaulting to the existing harbors if none is specified. This PR implements this functionality.

### TODO
- [x] Ensure the correct harbors are set when using the repeat purchase functionality and navigating to the purchase screen.

<details>

image

image

image

image

image

</details>

Closes https://github.com/AtB-AS/kundevendt/issues/18966

### DIFF
--- a/src/fare-contracts/FareContractInfo.tsx
+++ b/src/fare-contracts/FareContractInfo.tsx
@@ -4,6 +4,7 @@ import {
   getReferenceDataName,
   PreassignedFareProduct,
   TariffZone,
+  useFirestoreConfiguration,
   UserProfile,
 } from '@atb/configuration';
 import {StyleSheet} from '@atb/theme';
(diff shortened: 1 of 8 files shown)


mortennordseth:
I've decided to drop the TODO regarding the repeat purchase functionality, as I couldn't find a suitable solution at this time. Since this is an edge case and we would rather prioritize releasing the core functionality, I've marked this PR as ready for review

mortennordseth:
I've updated the PR to include a fix to the repeat purchase functionality so that harbors found using transport mode on the fare product config are used. Ready for review

gorandalum (approved):

PR #4706, opened 2024-09-09
fix: Id token state not updated throughout the app

jorelosorio:
Closes:

https://github.com/AtB-AS/kundevendt/issues/18830

Authentication comes from Firebase & token ID comes from the mobile token SDK, so if the token is not set for an issue like no Wi-Fi but then Wi-Fi returns and updates the token ID and the user is in their profile, the app is not checking the token Id to be set, only checks for authentication, which leads into an issue that still shows the wrong auth.

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -43,7 +43,7 @@ declare module 'axios' {
   }
 }
 
-function shouldRetry(error: AxiosError): boolean {
+export function shouldRetry(error: AxiosError): boolean {
   const isRateLimited = error.response?.status === 429;
   if (isRateLimited) return false;
 
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -116,8 +116,8 @@ async function requestIdTokenHandler(config: InternalAxiosRequestConfig) {
 }
 
 function responseIdTokenHandler(error: AxiosError) {
-  if (error?.config && error.config.authWithIdToken) {
-    error.config.forceRefreshIdToken = true;
+  if (error?.config && error.response?.status) {
+    error.config.forceRefreshIdToken = error.config.authWithIdToken && [401, 403, 500, 502, 503].includes(error.response?.status);
   }
   throw error;
 }
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -76,7 +76,7 @@ export function createClient(baseUrl: string | undefined) {
   });
   client.interceptors.request.use(requestHandler, undefined);
   client.interceptors.request.use(requestIdTokenHandler);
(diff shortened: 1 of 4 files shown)


src/api/client.ts
```diff
@@ -103,7 +104,12 @@ function requestHandler(config: InternalAxiosRequestConfig): InternalAxiosReques
 async function requestIdTokenHandler(config: InternalAxiosRequestConfig) {
   if (config.authWithIdToken) {
     const user = auth().currentUser;
-    const idToken = await user?.getIdToken(config.forceRefreshIdToken);
+    const tokenResult = await user?.getIdTokenResult(config.forceRefreshIdToken);
+    if (config.forceRefreshIdToken && tokenResult?.claims['customer_number']) {
```
jorelosorio, line 108: @gorandalum @rosvik I was looking at and testing this change, however, if I remove `config.forceRefreshIdToken` to test the logic of the timer, it works it forces a retryAuth every 10 seconds. But `config.forceRefreshIdToken` is in the condition and testing on staging, the issue seems like the retryAut is not called, it could be because the `config.forceRefreshIdToken` is never set to true because the response did not get a 401 as a response instead might get something different because no internet.

gorandalum: You need to use a proxy to introduce some 401 errors to test it appropriately.

jorelosorio: Yes, working on testing it on different types of responses.

rosvik (approved):
This looks like a good solution to me! Would be nice to have some monitoring ref. my comment bellow.

src/api/client.ts
```diff
@@ -49,11 +50,21 @@ function shouldRetry(error: AxiosError): boolean {
...
+      isIdempotentRequestError(error))
+  );
+}
+
+// Implements a custom axios-retry to force a refresh if 401 error is found
+function handleUnauthorizedError(error: AxiosError) {
+  if (error.config && error.response?.status === 401) {
+    error.config.forceRefreshIdToken = true;
+    SHOULD_FORCE_REFRESH.value = true;
+    return client(error.config);
```
gorandalum, line 65: Not really sure about returning a client invocation here 🤔 Could you sum up some reasoning behind it?

gorandalum (changes requested):
I started this locally and when using the test account with missing custom claims (+4794444444) it ends up refetching id token from server in an endless loop. I'm not really sure what or why it happens, but this can't be merged as it is now, as it would hit the Firebase Auth usage limits quickly.

I'm still not sure it can be solved in a clean manner using Axios interceptors, and making react-query be the handler of retries might be the way to go.

PR #4707, opened 2024-09-09
feat: confirm payment method selection in purchase flow

rosvik:
related to https://github.com/AtB-AS/kundevendt/issues/18885

The current flow, where selecting a payment method completes a purchase, can be counter intuitive at best, and can lead to wrong purchases at worst. It also breaks with how other bottom sheets work in the app, where making a selection doesn't also submit the parent screen. So after a talk with designers, it was decided to change the flow to have the bottom sheet apply changes, but not complete the purchase.

[Updated sketches](https://www.figma.com/design/zdZwvobgpEWSagKt0tderx/App?node-id=28048-99792&node-type=INSTANCE&t=EkVYakSquEsRJVYY-0)

https://github.com/user-attachments/assets/7e29c38c-5696-4fe0-a1a0-5815edeabb3b

### Acceptance criteria

- [ ] The purchase flow is never completed directly from the bottom sheet, but by clicking the "Pay with [method]" button.
- [ ] Saving of payment method still work as before. (Should test both iOS and Android)
- [ ] Payment options are not shared between accounts.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -417,7 +417,7 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
                   interactiveColor="interactive_0"
                   disabled={!!error}
                   rightIcon={{
-                    svg: getPaymentOptionSvg(paymentMethod.paymentType),
+                    svg: getPaymentTypeSvg(paymentMethod.paymentType),
                   }}
                   onPress={() => {
                     analytics.logEvent(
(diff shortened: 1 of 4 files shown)


gorandalum (approved):
Nice! 👏 

Could consider cleaning up the different variations of representing the `paymentMethod` and `shouldSavePaymentMethod` combo, sometimes they are treated individually (state) and sometimes as a combined object (`currentOptions`). My intuition is that it will look slightly cleaner if generally treating it as a combined object.

rosvik:
@gorandalum So something like this?

```ts
type PaymentMethodSelection = PaymentMethod & { shouldSavePaymentMethod?: boolean }
```

Then we're closer to the solution from before, where some places use the PaymentMethod type, and some use the PaymentMethodSelection type. Using it in the bottom sheet and in the data flow after it, means half the purchase flow use a different type again.

There might be some solution that makes the difference and boundaries clear, but I think separating them makes more sense for now. We could update `PaymentMethod` to include shouldSavePaymentMethod globally, but it shouldn't be saved to local storage, so I'm again hesitant to do so.

reidzeibel (approved):

PR #4711, opened 2024-09-12
feat: use in app browser for payment

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/18885

- Removes the PurchasePaymentScreen
- Moves reserve offer logic to PurchaseConfirmationScreen
- Uses in app browser instead of webview for payment
- Adds a wrapper function for in app browser


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -93,26 +101,89 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
+    if (
+      reserveMutation.isSuccess &&
+      paymentMethod?.paymentType !== PaymentType.Vipps &&
+      reserveMutation.data.url
+    ) {
+      try {
+        const oldStyle = StatusBar.pushStackEntry({
+          barStyle: 'dark-content',
+        });
+        InAppBrowser.open(reserveMutation.data.url, {
```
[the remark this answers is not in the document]
rosvik: Update: Made a PR for it here https://github.com/AtB-AS/mittatb-app/pull/4713

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -61,45 +61,29 @@ export const Profile_PaymentMethodsScreen = () => {
...
-    };
+    }
+  };
 
-    // Listen to deep link redirects back here after
-    // the add card process completes.
-    const eventSubscription = Linking.addEventListener(
-      'url',
+  const onAddRecurringPayment = async () => {
+    const callbackUrl = `${APP_SCHEME}://payment-method-callback`;
```
reidzeibel, line 79: should this be "://profile" to match the one in `RootStack.tsx`? 🤔

particularly this line on `RootStack.tsx` : 

```
     TabNav_ProfileStack: 'profile',
``` 

Although based on the new flow, I don't think it matters. Just thinking out loud here.

gorandalum (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
rosvik: I found it a bit strange to use the `://profile` link when it doesn't link to the profile root screen. But then again, they both work, so maybe no point in changing it in this PR. 

I'll revert the change 👍

rosvik: @gorandalum Made the argument that it's better to have an url that can only trigger one thing, so there's no confusion where it is handled.

Reverted the revert 😅

PR #4713, opened 2024-09-13
refactor: add in app browser util functions

rosvik:
follow up to https://github.com/AtB-AS/mittatb-app/pull/4711#discussion_r1756891669

### DIFF
--- a/src/in-app-browser/in-app-browser.tsx
+++ b/src/in-app-browser/in-app-browser.tsx
@@ -1,13 +1,10 @@
 import {StatusBar} from 'react-native';
-import InAppBrowser, {
-  InAppBrowserOptions,
-} from 'react-native-inappbrowser-reborn';
+import InAppBrowser from 'react-native-inappbrowser-reborn';
 import {notifyBugsnag} from '../utils/bugsnag-utils';
 
 export async function openInAppBrowser(
   url: string,
   dismissButtonStyle: 'cancel' | 'close' | 'done',
-  options?: InAppBrowserOptions,
 ) {
   const previousStatusBarStyle = StatusBar.pushStackEntry({
     barStyle: 'dark-content',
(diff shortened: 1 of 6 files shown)


src/in-app-browser/in-app-browser.tsx
```diff
@@ -0,0 +1,40 @@
...
+  } catch (error: any) {
+    notifyBugsnag(error);
+  }
+  StatusBar.popStackEntry(previousStatusBarStyle);
+}
+
+/**
+ * @returns true if the InAppBrowser was available, false otherwise
+ */
+export async function closeInAppBrowser() {
```
gorandalum, line 34: Maybe not return anything? It doesn't seem we use the return value for anything now. And if we don't return anything we also don't need to have it async.

I personally find "fire-and-forget" functions a delight to use 😅 Just invoke it and no more troubles. When a function returns `Promise<boolean>` I'm like "Doh, here is some stuff that must be handled".

src/in-app-browser/in-app-browser.tsx
```diff
@@ -0,0 +1,40 @@
+import {StatusBar} from 'react-native';
+import InAppBrowser, {
+  InAppBrowserOptions,
+} from 'react-native-inappbrowser-reborn';
+import {notifyBugsnag} from '../utils/bugsnag-utils';
+
+export async function openInAppBrowser(
+  url: string,
+  dismissButtonStyle: 'cancel' | 'close' | 'done',
+  options?: InAppBrowserOptions,
```
gorandalum, line 10: I really like having `dismissButtonStyle` as its own parameter, enabling users of this util function not needing to send in a full options object. I think I would prefer just removing the `options` param all together (YAGNI).

src/in-app-browser/in-app-browser.tsx
rosvik: Good point, let's to that 😄

rosvik:
Due to an issue with deep link redirects on Android, we had to use openAuth with a set destination URL that triggers a callback. Wrapped this in a new openInAppBrowserWithCallback. Since this is closely related to https://github.com/AtB-AS/mittatb-app/pull/4711, I'll merge it into that branch now, and we can do a review on it all together.

PR #4714, opened 2024-09-16
Create Shared Mobility endpoint hooks + test usage

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/18891

The endpoints are as described in the issue.
One extra added for getting a booking based on id instead of just the active one: `/mobility/v1/bookings/${bookingId}`.
This is needed because the summary of a booking should be shown when it is no longer active.

A new test feature has been added behind the `shmoDeepIntegrationEnabled` flag, which can be turned on in the debug menu and restarting the app.

### DIFF
--- a/src/api/types/mobility.ts
+++ b/src/api/types/mobility.ts
@@ -1,3 +1,5 @@
+import {z} from 'zod';
+
 export type ViolationsReportingInitQuery = {
   lng: string;
   lat: string;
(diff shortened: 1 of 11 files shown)


src/components/map/components/mobility/ShmoTesting.tsx
```diff
@@ -0,0 +1,257 @@
...
+  ShmoBookingEventType,
+} from '@atb/api/types/mobility';
+import {StyleSheet} from '@atb/theme';
+import {ThemeText} from '@atb/components/text';
+import {useActiveShmoBookingQuery} from '@atb/mobility/queries/use-active-shmo-booking-query';
+import {useGetIdsFromQrCodeMutation} from '@atb/mobility/queries/use-get-ids-from-qr-code-mutation';
+import {useInitShmoOneStopBookingMutation} from '@atb/mobility/queries/use-init-shmo-one-stop-booking-mutation';
+import {useSendShmoBookingEventMutation} from '@atb/mobility/queries/use-send-shmo-booking-event-mutation';
+import {useShmoBookingQuery} from '@atb/mobility/queries/use-shmo-booking-query';
+// eslint-disable-next-line no-restricted-imports
```
marius-at-atb, line 13: This whole file is just for test purposes

marius-at-atb:
Update:
Replaced the `preferredLanguage` option with `Accept-Language` headers for each shmo call, as this change has been made both by Entur and in AtBs Mobility service.

src/api/use-accept-language.ts
```diff
@@ -0,0 +1,8 @@
+import {useTranslation} from '@atb/translations';
+
+export const useAcceptLanguage = () => {
```
marius-at-atb, line 3: A different approach which can be considered, is to add a useEffect which updates the `Accept-Language` axios default header when `language` changes. However this would affect all axios requests, unless we also add an option for a handler in `createClient`.

Another difference would be that since `acceptLanguage` would not be in the query key, it will not be refetched when `language` is changed.

So for now at least I think it is easiest to use this hook instead.

src/api/types/mobility.ts
```diff
@@ -46,3 +48,147 @@ export type ViolationsReportQuery = {
...
+    .string()
+    .optional()
+    .describe('This is the same id as vehicleId from the mobility API'),
+});
+
+export type InitShmoOneStopBookingRequestBody = z.infer<
+  typeof InitShmoOneStopBookingRequestBodySchema
+>;
+
+export enum ShmoBookingEventType {
```
reidzeibel, line 160: Interesting to not have an "Unknown", "Init", "Start booking" event here, is that intended or how does it work?

marius-at-atb: Nice observation!

For now we get everything needed to start a trip with the init call, and after that only support ending a trip (which is done in two steps: 1. `START_FINISHING` to end the part you pay for, and 2. `FINISH` to also be able to send a parking photo).

As the TOMP API standard is designed to support all sorts of use cases for shared mobility, there are several more [there](https://app.swaggerhub.com/apis-docs/TOMP-API-WG/transport-operator_maas_provider_api/1.5.0#/trip%20execution/post_legs__id__events): `PREPARE`, `ASSIGN_ASSET`, `SET_IN_USE`, `TIME_EXTEND`, `TIME_POSTPONE`, `PAUSE`, `OPEN_TRUNK`.

Especially `PAUSE` AND `SET_IN_USE` are likely to be used by us in the future to support pausing and resuming a trip.

marius-at-atb: And btw this is already supported in the [Entur API](https://petstore.swagger.io/?url=https://api.dev.entur.io/api-docs/shared-mobility#/Booking/oneStopEvent) with slightly easier to understand event names (AtB doesn't use TOMP directly):
Skjermbilde 2024-09-19 kl  14 35 35

reidzeibel (approved):

src/api/types/mobility.ts
reidzeibel: Got it, so there are possibilities to expand the events in the future, after the integration are properly implemented.

marius-at-atb: Indeed 😄

PR #4715, opened 2024-09-17
ci: remove 'environment', and combine lint and tsc actions

rosvik:
https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment

- Removes the `environment` field from tsc.yml
- Combines lint.yml and tsc.yml to reduce the amount of jobs that run for each commit

Now PRs will have these 4 jobs, instead of the 6 from before:

Screenshot 2024-09-17 at 15 15 59


reidzeibel (commented):
So this will make lints and tsc run a single instance instead of one for each tenants?

reidzeibel (approved):

PR #4717, opened 2024-09-19
fix: save payment method whenever payment is completed

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/18885#issuecomment-2357887414

This fixes an issue introduced after I started using the openAuth in-app-browser and forgot to call saving of payment methods. To make this harder to miss in the future, this PR gathers saving payment method and closing the in-app-browser into one `onPaymentCompleted` callback function. This also means payment method is saved when fare contracts appear in the background, and the regular payment flow isn't completed.

I also moved the saveLastUsedPaymentMethod into saved-payment-utils, and renamed it to `saveLastUsedRecurringPaymentOrType` as I thought that made more sense together with `savePreviousPaymentMethodByUser`.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -20,7 +20,10 @@ import {addMinutes} from 'date-fns';
 import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import {ActivityIndicator, ScrollView, View} from 'react-native';
 import {useOfferState} from '../Root_PurchaseOverviewScreen/use-offer-state';
-import {usePreviousPaymentMethods} from '../saved-payment-utils';
+import {
+  saveLastUsedRecurringPaymentOrType,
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -137,29 +151,25 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
   ]);
 
   // When deep link {APP_SCHEME}://purchase-callback is called, save payment
   // method and navigate to active tickets.
-  usePurchaseCallbackListener(
-    navigateToActiveTicketsScreen,
-    paymentMethod?.paymentType,
-    reserveMutation.data?.recurring_payment_id,
-  );
+  usePurchaseCallbackListener(onPaymentCompleted);
```
gorandalum, line 166: Do we really need this listener at all anymore? 🤔

rosvik: Yes, Vipps payment still uses it.

PR #4719, opened 2024-09-20
feat: Only stop places checkbox in location search

gorandalum:
### Background
When looking at geocoder request logs, we see that only ~8% of
location searches end up in the user selecting an address. They
instead use my position or a stop place.

It was decided to do a proof of concept by swiftly adding a checkbox,
and get quick feedback on whether this is a feature that users will use
or not.

### Solution
To make searching for stop places easier we add a checkbox the user
can check so only stop places are returned when searching for a
location. We will enable this to a subset of the users, and see if
it is a feature which will be used.

### Acceptance criteria
- [ ] If the is not checked, both stop places and other locations are returned.
- [ ] If the checkbox is checked, then only stop places are returned.
- [ ] When toggling the checkbox, a PostHog event is sent ('Location search'/'Only stop places checkbox pressed').
- [ ] The checkbox is only shown if `enable_only_stop_places_checkbox` evaluates to `true` in RemoteConfig
- [ ] When using debug override, the setting in RemoteConfig is overridden

### DIFF
--- a/src/components/checkbox/CheckboxWithLabel.tsx
+++ b/src/components/checkbox/CheckboxWithLabel.tsx
@@ -1,0 +1,48 @@
+import {
+  screenReaderPause,
+  ThemeText,
+  ThemeTextProps,
+} from '@atb/components/text';
+import {StyleProp, TouchableOpacity, ViewStyle} from 'react-native';
+import {dictionary, useTranslation} from '@atb/translations';
+import {Checkbox} from '.';
+import React from 'react';
+import {StyleSheet} from '@atb/theme';
+
+export const CheckboxWithLabel = ({
+  label,
+  checked,
+  onPress,
+  color,
+  style,
+}: {
+  label: string;
+  checked: boolean;
+  onPress: (v: boolean) => void;
+  color?: ThemeTextProps['color'];
+  style: StyleProp<ViewStyle>;
+}) => {
+  const {t} = useTranslation();
+  const styles = useStyles();
+
+  const a11yLabel =
+    label +
+    screenReaderPause +
+    t(checked ? dictionary.checked : dictionary.unchecked);
+
+  return (
+    <TouchableOpacity
+      style={[styles.container, style]}
+      accessible={true}
+      accessibilityLabel={a11yLabel}
+      onPress={() => onPress(!checked)}
+    >
+      <Checkbox checked={checked} />
+      <ThemeText color={color}>{label}</ThemeText>
+    </TouchableOpacity>
+  );
+};
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  container: {flexDirection: 'row', gap: theme.spacings.small},
+}));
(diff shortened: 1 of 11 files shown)


src/components/checkbox/CheckboxWithLabel.tsx
```diff
@@ -0,0 +1,48 @@
...
+    t(checked ? dictionary.checked : dictionary.unchecked);
+
+  return (
+    <TouchableOpacity
+      style={[styles.container, style]}
+      accessible={true}
+      accessibilityLabel={a11yLabel}
+      onPress={() => onPress(!checked)}
+    >
+      <Checkbox checked={checked} />
```
marius-at-atb, line 40: Should we use the Toggle component instead here?

src/stacks-hierarchy/Root_LocationSearchByTextScreen/components/LocationSearchContent.tsx
```diff
@@ -141,6 +149,22 @@ export function LocationSearchContent({
             onAddFavorite={onAddFavorite}
           />
         )}
+        {onlyStopPlacesCheckboxEnabled && (
+          <CheckboxWithLabel
+            label={t(LocationSearchTexts.onlyStopPlacesCheckbox)}
+            checked={onlyStopPlaces}
+            onPress={(v) => {
```
marius-at-atb, line 156: Not sure what `v` stands for here, `checked` might be a better name?

marius-at-atb (approved):
Lgtm!

src/components/checkbox/CheckboxWithLabel.tsx
gorandalum: I like the checkbox better! 😅 (And @Mariblaas said it was okay 😇)

src/stacks-hierarchy/Root_LocationSearchByTextScreen/components/LocationSearchContent.tsx
gorandalum: Short answer:
`v` is short for value, and I think it is pretty common to use, much like using `i` for item or `ev` for event. 

Long answer:
I actually had named it `checked` originally, but it wasn't really precise since it isn't actually checked yet. So I changed it to `newChecked`, but felt it was unnecessarily verbose, so I just thought "f* it,  let's just use `v`" 💯💪

marius-at-atb: Fair enough, though not great 😅

tormoseng:
Note to after release of 1.57:
- Make PostHog dashboard for event `Location search: Only stop places checkbox pressed` with `cheked: true | false`
- Look on Grafana for requests with `layers=venue` like `/bff/v1/geocoder/features?lat=63.43119807409924&layers=venue&`

tormoseng:
✅  Tested OK!

tormoseng:
- Grafana dashboard: https://mittatb.grafana.net/d/de09a8f9zd7uoa/geocoder?from=now-2d&to=now&timezone=browser

PR #4722, opened 2024-09-24
test: create summary after android performance test

tormoseng:
The Android performance tests ([ref](https://github.com/AtB-AS/mittatb-app/actions/workflows/test-performance-android.yml)) creates results in a JSON that can be used to create an HTML file to view locally with `$ flashlight report <file-name>`. To be able to see any changes from test to test, this PR creates a JSON summary of the test in terms of:
- Score
- CPU overall
- CPU per thread (i.e. JS, UI and Render)
- FPS
- RAM

I will use the summary to compare against previous test runs, and visualise it probably in https://github.com/AtB-AS/e2e-test-results. Probably the results from running on GitHub will be worse than running them locally, but the important part is deviations and trends.


tormoseng:
Tests OK: https://github.com/AtB-AS/mittatb-app/actions/runs/11013496870

with summary
```
{
  "version": "1.57-1726831636",
  "cpu": {
    "avg": 35.1,
    "threads": {
      "RN JS Thread": 11.6,
      "Render Thread": 14,
      "UI Thread": 2.4
    }
  },
  "fps": 53.9,
  "ram": 456.2,
  "score": 92
}
```

jorelosorio (approved):
Looks good and nice to have this kind of measures, is there something like this for iOS?

reidzeibel:
How do we find the bottleneck if there is something weird with the results? 🤔

tormoseng:
> Looks good and nice to have this kind of measures, is there something like this for iOS?

Havn't seen it. At least, `flashlight` only supports Android so far, but are working to support iOS. Their view was that performance differ more on Android so therefore the focus was there.

tormoseng:
@reidzeibel 
> How do we find the bottleneck if there is something weird with the results? 

First of all, the same test creates a JSON with all results that can locally generate an HTML with `flashlight report <file>` so it's possible to see when there was problem. So far, the test cases are pretty straight forward but with a pause of 10 seconds between each test case that can somewhat differentiate them. But future work include to look into running a performance test and create metric trends per use case - or "area" (flashlight doesn't support labelling the measurement to even better distinguish between the test cases).

PR #4727, opened 2024-09-27
feat: select vehicle with qr scanning

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/18913
Closes https://github.com/AtB-AS/kundevendt/issues/19097

This adds a button for going to the scan qr screen, and the scan screen itself.
Also some minor tweaks to smooth out the UX.

Note: The state management in the map needs a rewrite and is really tricky to work with atm. We should refactor it. In this PR, a MapContext is created, which should probably be used a lot more soon.

Changes since recordings were made below:
- No longer zooming in, as this triggered a performance bug.
- Some text changes.
- GeofencingZones showing.
- Changed from `back` to `close` back button

When found:

https://github.com/user-attachments/assets/bfa35956-b9ad-4284-9fdb-88d656c40b54

When not found:

https://github.com/user-attachments/assets/9edae325-b49f-4562-953d-85789036797d


src/stacks-hierarchy/RootStack.tsx
```diff
@@ -303,6 +304,10 @@ export const RootStack = () => {
                 component={Root_LocationSearchByMapScreen}
                 options={screenOptions(TransitionPresets.SlideFromRightIOS)}
               />
+              <Stack.Screen
+                name="Root_ScanQrCodeToSelectVehicleScreen"
```
marius-at-atb, line 308: Todo: I'll rename this to something more general then just "vehicle"

src/components/map/Map.tsx
```diff
@@ -55,25 +60,44 @@ export const Map = (props: MapProps) => {
...
+    isScooter(selectedFeature) || isBicycle(selectedFeature);
 
   const [geofencingZonesEnabled, geofencingZonesEnabledDebugOverrideReady] =
     useGeofencingZonesEnabled();
 
   const showGeofencingZones =
     geofencingZonesEnabled &&
     geofencingZonesEnabledDebugOverrideReady &&
-    (isScooter(selectedFeature) || isBicycle(selectedFeature));
+    (selectedFeatureIsAVehicle || aVehicleIsAutoSelected);
```
marius-at-atb, line 93: Having both a selected feature and auto selected feature in state is not great as it produces duplicate logic and state management.
To unify them requires a deeper refactor, which we need to come back to.

src/components/map/hooks/use-control-styles.ts
```diff
@@ -29,8 +29,7 @@ export function useControlPositionsStyle() {
       },
       controlsContainer: {
         position: 'absolute',
-        bottom:
-          bottom + bottomPaddingIfBottomSheetIsOpen + theme.spacings.medium,
+        bottom: bottomPaddingIfBottomSheetIsOpen + theme.spacings.medium,
```
marius-at-atb, line 32: @tormoseng @hanne-at-atb This is the change we need to check if also works for Android phones with the Android NavBar covering the bottom of the screen.

marius-at-atb: Update: turns out it works fine

src/stacks-hierarchy/Root_ScanQrCodeScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+        },
+      ],
+    );
+  }, [navigation, t]);
+
+  const idsFromQrCodeReceivedHandler = useCallback(
+    (idsFromQrCode: IdsFromQrCodeResponse) => {
+      // NB come back and fix this once the Shmo QR endpoint is updated!
+      // the AutoSelectableBottomSheetType should then be decided based on FormFactor(s)
+      const isBysykkelOperator = idsFromQrCode.operatorId.includes('bysykkel');
```
marius-at-atb, line 52: https://github.com/AtB-AS/kundevendt/issues/19215

rosvik (approved):
I think this looks good, although I find the structure of the map and it's interactions a bit confusing at this point. There's probably a better solution to auto select bottom sheets than the MapContext if the structure was different. But that's probably for another time. 

Would be nice to have a look at all Mapbox usage in the app, and see if there are improvements we can make. (Maybe it belongs on the tech debt list?) 

Also: Not sure if it happened in this PR, but the padding of map controls have changed. It didn't look great before either to be fair, but maybe have a look at `Root_LocationSearchByMapScreen` before we release this 😄

Screenshot 2024-10-07 at 09 35 02

src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
```diff
@@ -33,17 +33,26 @@ export const useUpdateBottomSheetWhenSelectedEntityChanges = (
...
     closeBottomSheet();
     closeCallback();
-  };
+  }, [closeBottomSheet, closeCallback]);
+
+  const onReportParkingViolation = useCallback(() => {
+    closeWithCallback();
+    analytics.logEvent('Mobility', 'Report parking violation clicked');
+    navigation.navigate('Root_ParkingViolationsSelectScreen');
+  }, [closeWithCallback, analytics, navigation]);
```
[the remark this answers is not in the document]
marius-at-atb: I agree it would be better, but unfortunately since `onReportParkingViolation` depends on `closeWithCallback`, that won't work with the current setup. We should address this as well in the upcoming refactor I think.

```ts
const onReportParkingViolation = () => {...} // depends on closeWithCallback
const {
    ...,
    closeWithCallback,
  } = useMapSelectionChangeEffect(
    ...,
    onReportParkingViolation
  );
```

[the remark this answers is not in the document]
rosvik: Aha, ok 👍

src/stacks-hierarchy/Root_ScanQrCodeScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+      t(MapTexts.qr.notFound.description),
+      [
+        {
+          text: t(MapTexts.qr.notFound.ok),
+          style: 'default',
+          onPress: navigation.goBack,
+        },
+      ],
+    );
+  }, [navigation, t]);
```
gorandalum, line 46: There is something a little iffy here. Now if the language changes, the `alertResultError` will change, which might trigger the Alert on unexpected times. It is a corner case, but the thought experiment exposes that something should be looked into.

src/mobility/utils.ts
```diff
@@ -70,7 +70,7 @@ export const getAvailableVehicles = (
     .reduce((sum, count) => sum + count, 0) ?? 0;
 
 export const getRentalAppUri = <T extends {rentalUris?: RentalUrisFragment}>(
-  t: T | undefined,
+  t: T | undefined | null,
```
gorandalum, line 73: In general we use `undefined` throughout the app, not `null`. If there is some logic depending on whether the return value is `undefined` or `null` it should preferably be modelled in a return type instead.

src/mobility/use-vehicle.tsx
```diff
@@ -1,9 +1,11 @@
 import {useSystem} from '@atb/mobility/use-system';
 import {getRentalAppUri} from '@atb/mobility/utils';
 import {useVehicleQuery} from '@atb/mobility/queries/use-vehicle-query';
+import {useMemo} from 'react';
 
 export const useVehicle = (id: string) => {
-  const {data: vehicle, isLoading, isError} = useVehicleQuery(id);
+  const {data: vehicleData, isLoading, isError} = useVehicleQuery(id);
+  const vehicle = useMemo(() => vehicleData, [vehicleData]);
```
gorandalum, line 8: Does this really do anything? 🤔

src/MapContext.tsx
gorandalum, line 1: What does 'AutoSelected' mean here? I didn't intuitively understand it, so the term could be explained by some JSdoc for the `MapContext`.

gorandalum (approved):
Some comments, but in general it looks good :+1: But as you say, there is about time for some refactoring of the Map as a whole.

src/utils/use-do-once-on-truthy.ts
gorandalum, line 1: I don't see the difference from `useDoOnceWhen` on first glance. If it needs its own hook, then add some unit tests.

src/stacks-hierarchy/Root_ScanQrCodeScreen.tsx
marius-at-atb: Hmm, it's only triggered by the `onCapture` event, not used e.g. in a `useEffect`. So I don't think it's a problem?

src/mobility/use-vehicle.tsx
marius-at-atb: Probably not tbh. I see this remains after debugging this [performance problem](https://github.com/AtB-AS/mittatb-app/blob/marius/select-vehicle-with-qr-scanning/src/components/map/hooks/use-auto-select-map-item.tsx#L57-L59). I'll revert it.

src/MapContext.tsx
marius-at-atb: Updated with jsdoc on `useMapState` :+1:

src/utils/use-do-once-on-truthy.ts
marius-at-atb: This seems to work:
```ts
const useOnItemReceived = (
  action: ((item: any) => void) | undefined,
  item: any,
) => {
  useDoOnceWhen(() => item && action?.(item), !!item);
};
```

```ts
useOnItemReceived(onVehicleReceived, vehicle);
```

One difference is that if the `item` goes from truthy to falsy to truthy, the action will re-run. However not sure if that's a bug or a feature 😛

marius-at-atb: Update – replaced with:
```ts
export const useDoOnceOnItemReceived = <T>(
  action: ((item: T) => void) | undefined,
  item: T | undefined | null,
) => {
  useDoOnceWhen(() => action?.(item!), !!item, true);
};
```
and `onlyOnce` support in `useDoOnceWhen`

src/utils/use-do-once-when.ts
gorandalum, line 1: No need to do it now, just a comment that the change to `useDoOnceWhen` and accompanying tests are a good example of something that could have been its own small PR.

src/utils/use-do-once-when.ts
```diff
@@ -1,19 +1,27 @@
...
     }
+
     return () => {
-      firstTimeRef.current = true;
+      if (!onlyOnce) {
+        isFirstTimeRef.current = true;
+      }
     };
-  }, [condition]);
+  }, [condition, onlyOnce]);
```
gorandalum, line 26: Also add some tests on what happens when `onlyOnce` changes, since it is used in the dependency array for the main `useEffect` in the util.

gorandalum (approved):

src/utils/__tests__/use-do-once-when.test.ts
```diff
@@ -78,11 +78,55 @@ describe('useDoOnceWhen', () => {
...
+    expect(functionRunCount).toBe(0);
+
+    hook.rerender({condition: true, onlyOnce: true});
+    expect(functionRunCount).toBe(1);
+
+    hook.rerender({condition: false, onlyOnce: false});
+    expect(functionRunCount).toBe(1);
+
+    hook.rerender({condition: true, onlyOnce: false});
+    expect(functionRunCount).toBe(2);
```
gorandalum, line 105: What about:
```ts
    hook.rerender({condition: true, onlyOnce: true});
    expect(functionRunCount).toBe(1);

    hook.rerender({condition: true, onlyOnce: false});
    expect(functionRunCount).toBe(1);

    hook.rerender({condition: true, onlyOnce: true});
    expect(functionRunCount).toBe(?);
```

marius-at-atb: The `?` should be `2`, because the "do only once" starts counting from the render where `onlyOnce` is `true`. Extra test added.

marius-at-atb: Note that the count could be very high after `onlyOnce` has been `false`, it is not possible or meaningful to "go back" to insist it should have only ran once previously..

gorandalum: Does it really make sense? Isn't kinda the core of the hook that it only run once when the `condition` stays `true`? 🤔 
- If `onlyOnce` is `false`, then the function is only run once as long as the `condition` stays `true`.
- If `onlyOnce` is `true`, then the function is only run once as long as the `condition` stays `true`.
- But in the case that `onlyOnce` switches between `false` and `true`, then the function is run multiple times even if `condition` always stays `true`.

I'm kinda nitpicking on this one, the actual case here isn't really important. But the subject touches upon something that we have some problems with in the app, that several effects run to often and in unwanted/unnecessary situations.

marius-at-atb: Since switching `onlyOnce` from `false` to `true` is an impossible ask if it has already been run twice or more, I think [this approach](https://github.com/AtB-AS/mittatb-app/pull/4745/files#diff-eebed2193d07efdf49b24bf8d3a425c552998752642ff0cbccaacdb89aa0d491) is preferable.

gorandalum: I would rather do the same with `onlyOnce` as with the given function, transform it to a ref so it doesn't affect the `useEffect`.

marius-at-atb: That works, even though the issues noted before here remain 😅 
New PR created: https://github.com/AtB-AS/mittatb-app/pull/4746

src/stacks-hierarchy/Root_ScanQrCodeScreen.tsx
marius-at-atb: Hopefully the final changes have been made to this PR now - with the updates from master merged into it and updated usage of global variables for both `t` and `coordinates` in this component.
Any further comments before merge?

gorandalum: Go ahead!

PR #4735, opened 2024-10-08
feat: Handle multiple stops in StopPlacesView

gorandalum:
Possible to look into each of the three commits one by one when doing the review:
- The first commit is a refactor which was necessary for this solution, where the `StopPlaceView` is changed to a `StopPlacesView` which can show departures for more than one stop place.
- The second commit is the core changes of this PR, where we stop looking up the multi modal parent id, and instead utilise the `adjacentSites` on the clicked feature.
- The third commit is a smaller refactor improving on `useQuery` usage.


reidzeibel (approved):
LGTM!

Should we remove the BFF function for getting the stop place parent ID?

gorandalum:
> LGTM!
> 
> Should we remove the BFF function for getting the stop place parent ID?

Need to be there for backwards compatibility for the app versions using it.

PR #4738, opened 2024-10-09
Support tGlobal

marius-at-atb:
As @gorandalum discovered [here](https://github.com/AtB-AS/mittatb-app/pull/4727#discussion_r1790061992), the current `t` function can be problematic when you don't want it to re-trigger an action.

This PR introduces a new function `tGlobal`, which will give the correct translation when called, but not re-trigger anything on language changes.

`tGlobal ` is _not_ required as part of dependency arrays.

### DIFF
--- a/src/LocaleProvider.tsx
+++ b/src/LocaleProvider.tsx
@@ -6,8 +6,9 @@ import {
   DEFAULT_LANGUAGE,
   DEFAULT_REGION,
   FALLBACK_LANGUAGE,
+  setStaticLanguage,
 } from '@atb/translations/commons';
-import { AppState } from 'react-native';
+import {AppState} from 'react-native';
 
 export type Locale = {
   language: Language;
@@ -68,6 +69,8 @@ function useLocale(): Locale {
       setLanguage(mapLanguageStringToEnum(userPreferencedLanguage));
     }
   }, [useSystemLanguage, userPreferencedLanguage, systemLocale]);
+
+  useEffect(() => setStaticLanguage(language), [language]);
 
   return {
     language: language,
--- a/src/LocaleProvider.tsx
+++ b/src/LocaleProvider.tsx
@@ -11,7 +11,6 @@ import {
 import {AppState} from 'react-native';
 
 let globalLanguage: Language = DEFAULT_LANGUAGE;
-export const getGlobalLanguage = () => globalLanguage;
 const setGlobalLanguage = (language: Language) => {
   globalLanguage = language;
 };
--- a/src/LocaleProvider.tsx
+++ b/src/LocaleProvider.tsx
@@ -63,14 +63,13 @@ function useLocale(): Locale {
 
   // listen for updates to language settings
   useEffect(() => {
-    if (useSystemLanguage) {
-      setLanguage(systemLocale.language);
-    } else {
-      setLanguage(mapLanguageStringToEnum(userPreferencedLanguage));
-    }
-  }, [useSystemLanguage, userPreferencedLanguage, systemLocale]);
+    const newLanguage = useSystemLanguage
+      ? systemLocale.language
+      : mapLanguageStringToEnum(userPreferencedLanguage);
 
-  useEffect(() => setStaticLanguage(language), [language]);
+    setStaticLanguage(newLanguage);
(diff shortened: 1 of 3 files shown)


gorandalum (commented):
Great! And nice work updating the eslint rule as well.

I know some other variables that also are good candidates for using the same pattern. For example:
- Id token, so it can be used in http requests without retriggering them when id token is changed
- Current location, so it can be used for in http requests, but not retrigger them when the location changes.

Since we know there will be other usages of this, we should make a little extra effort in this one, as it sets precedent for the other situations and establishes the pattern for them.

**Naming**
My intuitive thought was that 'global' was a better term for this than 'static', since static is a reserved keyword in Javascript which has slightly different meaning as it is connected to a class.

As the trend nowadays I asked ChatGPT for some pointers:
- What is a static variable?
> A static variable is a variable that retains its value between function calls and exists for the lifetime of the program or a class, depending on its context.
- What is a global variable?
> A global variable is a variable that is defined outside of any function or block and can be accessed from any part of the program.

All in all I think both 'static' and 'global' fits the description of what we want here and both are good options. If you lean towards 'static' then we'll go with that.

**Encapsulation**
I think we should assert that these global/static variables are not set outside of the contexts where they are ment to be modified. To ensure this here the variable should probably be moved from commons to the `LocaleProvider` and the `setStaticLanguage` should not be exported.

Also we should probably have a getter function for the variable. Something like `getTStatic` (or maybe `getStaticT` flows better?). It is not really necessary for `t` as it already is a `const` function, but it will be necessary for most static variables, and it will be preferable to have the same pattern for all of them. For example we can't export `currentLocationStatic` as a variable since then it can be modified everywhere in the app, so we need to export `getStaticCurrentLocation` instead.

marius-at-atb:
I think both are fine - `tGlobal` is more precise regarding the type of variable as you mention, while `tStatic` makes it more clear that it doesn't re-trigger effects, which is the point of it. We can go with global in case confusion with the static keyword is an issue :+1:

`getTStatic` -> did you mean `getStaticLanguage`, such that we have a getter and a setter for the variable, and that the function is still used directly?

marius-at-atb:
Updated now with renaming and better encapsulation as suggested.

As a side note, this is another thing that [zustand](https://github.com/pmndrs/zustand) can be good for, as this can be achieved directly with 
`const value = useStore.getState().someValue;`

src/LocaleProvider.tsx
```diff
@@ -7,7 +8,19 @@ import {
...
+export const getGlobalLanguage = () => globalLanguage;
+const setGlobalLanguage = (language: Language) => {
+  globalLanguage = language;
+};
+/**
+ * tGlobal can be used instead of the t function for when you don't want
+ * language changes to potentially retrigger an action (such as e.g. an alert box)
+ */
+export const tGlobal: TFunc<typeof Language> = (translatable) =>
+  translatable[globalLanguage];
```
marius-at-atb, line 23: @gorandalum Do you think this should use `getGlobalLanguage()` instead, allowing `tGlobal` to be defined wherever is most natural?

src/LocaleProvider.tsx
```diff
@@ -7,7 +8,19 @@ import {
   DEFAULT_REGION,
   FALLBACK_LANGUAGE,
 } from '@atb/translations/commons';
-import { AppState } from 'react-native';
+import {AppState} from 'react-native';
+
+let globalLanguage: Language = DEFAULT_LANGUAGE;
+export const getGlobalLanguage = () => globalLanguage;
```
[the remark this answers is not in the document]
marius-at-atb: Removed

src/LocaleProvider.tsx
marius-at-atb: I think maybe it's "a bit much" to insist on functions also having to go through getters?
I quite like the last example, thinking that the "Global" naming convention is sufficiently clear.

gorandalum: Lets try without and we'll get a feel for it

gorandalum (approved):

tormoseng:
Will be tested through regression tests of the map and feature test of [QR scanning](https://github.com/AtB-AS/kundevendt/issues/19097).

PR #4740, opened 2024-10-10
feat: Axios client not use Firebase Auth directly

gorandalum:
Use the Auth state values from the FirebaseAuthContext. They are
exported using the global variable pattern, to not make client
dependant on the React state.

### DIFF
--- a/src/auth/AuthContext.tsx
+++ b/src/auth/AuthContext.tsx
@@ -35,7 +35,7 @@ import {useLocaleContext} from '@atb/LocaleProvider';
 export type AuthReducerState = {
   authStatus: AuthStatus;
   user?: FirebaseAuthTypes.User;
-  idToken?: FirebaseAuthTypes.IdTokenResult;
+  idTokenResult?: FirebaseAuthTypes.IdTokenResult;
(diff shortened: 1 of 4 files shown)


src/auth/AuthContext.tsx
```diff
@@ -89,6 +97,7 @@ const authReducer: AuthReducer = (prevState, action): AuthReducerState => {
         customerNumber,
         authStatus,
       });
+      idTokenGlobal = action.idToken.token;
```
marius-at-atb, line 100: Since `idToken` in the state here is of type `FirebaseAuthTypes.IdTokenResult`, the string name for the token on idToken should probably be different?

I guess it is really `idTokenTokenGlobal` atm? 😛

marius-at-atb: Another option is to assign just `action.idToken` here, and then use `getIdTokenGlobal().token` in client.ts

gorandalum: I renamed the state field and action field to `idTokenResult`, since that was the most precise solution.

marius-at-atb (approved):
Lgtm!

PR #4746, opened 2024-10-10
feat: Accept onlyOnce param for useDoOnceWhen and add tests

marius-at-atb:
This allows a parameter to only run an action once, even if the condition keeps toggling between `true` and `false`.

More discussion can be found [here](https://github.com/AtB-AS/mittatb-app/pull/4727#discussion_r1793339837).


gorandalum (approved):

tormoseng:
Will be tested through regression tests of the map and feature test of [QR scanning](https://github.com/AtB-AS/kundevendt/issues/19097).

PR #4747, opened 2024-10-10
test: run app tests on previous ubuntu

tormoseng:
The emulator action `reactivecircus/android-emulator-runner@v2` doesn't support ubuntu 24 yet. Will thus have to use ubuntu 22.04 since ubuntu-latest now has become ubuntu 24.


reidzeibel (approved):
image

PR #4748, opened 2024-10-10
feat: Error msg on ticketing screen if something wrong with account

gorandalum:
This is in addition to the error message in the profile screen, since
we have a theory that isn't always found.

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/Components/AccountWrongMessage.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/Components/AccountWrongMessage.tsx
@@ -1,0 +1,30 @@
+import {useAuthState} from '@atb/auth';
+import {TicketingTexts, useTranslation} from '@atb/translations';
+import {MessageInfoBox} from '@atb/components/message-info-box';
+import {StyleProp, ViewStyle} from 'react-native';
+import {useAnalytics} from '@atb/analytics';
+
+type Props = {style?: StyleProp<ViewStyle>};
+
+export const AccountWrongMessage = ({style}: Props) => {
+    const {authStatus, retryAuth} = useAuthState();
+    const {t} = useTranslation();
+    const {logEvent} = useAnalytics();
+    if (authStatus === 'authenticated') return null;
+
+    return (
+        <MessageInfoBox
+            style={style}
+            type="error"
+            title={t(TicketingTexts.accountError.title)}
+            message={t(TicketingTexts.accountError.message)}
+            onPressConfig={{
+                action: () => {
+                    logEvent('Ticketing', 'Retry auth');
+                    retryAuth();
+                },
+                text: t(TicketingTexts.accountError.actionText),
+            }}
+        />
+    );
+};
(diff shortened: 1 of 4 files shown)


marius-at-atb (approved):
Looks good!

rosvik:
To me, this message makes it sound like the entire ticket purchase system is down, not that there's something wrong with the users account/device specifically. Worst case, it can make users think "oh well, if it doesn't work for anyone, I don't need to by a ticket". 

Maybe it's already been talked about, but it's probably worth clarifying with designers, Sebastian or market if there are ways to improve the text.

gorandalum:
> To me, this message makes it sound like the entire ticket purchase system is down, not that there's something wrong with the users account/device specifically. Worst case, it can make users think "oh well, if it doesn't work for anyone, I don't need to by a ticket".
> 
> Maybe it's already been talked about, but it's probably worth clarifying with designers, Sebastian or market if there are ways to improve the text.

Its been discussed with @Sebstorvik and @Mariblaas, and you pinpoint the exact thing that we were not really happy with. 
- The first iteration of opening phrase was "There is an issue with your account", but we ended up deciding it was to technical. And also the message will come (maybe most often) on anonymous accounts, which for the user may seem strange since they think they don't have any account.
- The second iteration was "You are unable to buy tickets" which Mari found slightly too aggressive saying "you" to the customer like they have done something wrong.

Mari and Sebastian can comment if your input nudges them towards another text, and feel free to throw out some suggestions.

rosvik:
Ok, I can see it could be hard to find a easy way to phrase it 😄

Not sure it's better, but we could use wording along the lines of "We had some issues setting up your device for ticket purchase" or "There were some issues when making your device ready for purchasing tickets"

Sebstorvik:
Could we use your suggestion Johannes as the header and remove the "ticket purchase doesn't work" as the first sentence? Would that work? 

**There were some issues when making your device ready for purchasing tickets**
Check your internett connection and reload the app. If this message persist, contact customer service
_Link text_

gorandalum:
Thoughts on the input @Mariblaas?

Side by side, old/new:
<div>


</div>

Mariblaas:
I see what you mean that it seems that it affects the whole ticket system! To me, the new one gets a bit too technical, but it works to mention "your device". How about: "Your device is not loading tickets" ?

Mariblaas:
**Your device is not loading tickets**
Check your internett connection and reload the app. If this message persist, contact customer service
_Link text_

gorandalum:
But it is actually loading your tickets, or it may them cached, you are just unable to buy new ones.

Mariblaas:
Jaa, I see, but does that really matter that it is not 100 % correct? As long as it is understandable that something is wrong with their device and that they should call customer service? What do you think @Aliaaen ?

**Your device is not loading new tickets**
Check your internett connection and reload the app. If this message persist, contact customer service
_Link text_

Aliaaen:
What about:

Your device is not loading new tickets
Check your internett connection and restart the app. If this does not help, contact customer service
Link text

gorandalum:
What about:
Your device is not able to buy tickets
Check your internet connection and reload the app. If this does not help, contact customer service.
Link text

- 'Able to buy tickets' is more precise than 'load new tickets'. And buying tickets is probably the action they are looking to do, so the phrasing is more relevant to them.
- Reload the app vs restart the app. Reload the app is used since it references the link text which is "reload app". The link text can't really be "Restart app", since that is not what happens.

Sebstorvik:
I think this one is really good☝️

Aliaaen:
Good stuff!

gorandalum:
Fixed here:
https://github.com/AtB-AS/mittatb-app/pull/4764

PR #4749, opened 2024-10-10
feat: Remote Config flag for id token retry count

gorandalum:
Added as an option so we can increase the retry count and see if it
affects how many end up in the missing-id-token state.


src/RemoteConfigContext.tsx
```diff
@@ -66,6 +66,7 @@ export const RemoteConfigContextProvider: React.FC = ({children}) => {
     try {
       await remoteConfig().fetchAndActivate();
       const currentConfig = await getConfig();
+      console.log("CURRENT CONFIG", currentConfig)
       setConfig(currentConfig);
```
gorandalum, line 70:
```suggestion
```

marius-at-atb (approved):

PR #4750, opened 2024-10-10
feat: use form factor to decide AutoSelectableBottomSheetType

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/19215
Completes https://github.com/AtB-AS/kundevendt/issues/19097

Now that Entur's endpoint supports `formFactor`, use that instead of the temporary solution to determine which type of bottom sheet to open after scanning a QR code in the map.

### DIFF
--- a/src/api/types/mobility.ts
+++ b/src/api/types/mobility.ts
@@ -184,7 +184,7 @@ export const IdsFromQrCodeResponseSchema = z.object({
   operatorId: z.string(),
   vehicleId: z.string().nullable().optional(),
   stationId: z.string().nullable().optional(),
-  formFactor: FormFactorSchema,
+  formFactor: FormFactorSchema.optional(),
 });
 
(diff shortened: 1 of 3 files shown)


src/api/types/mobility.ts
```diff
@@ -1,4 +1,5 @@
 import {z} from 'zod';
+import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
```
marius-at-atb, line 2: The FormFactor from Entur's qr-endpoint in Shared Mobility is the same as FormFactor from their Mobility API. However, we may still consider creating a separate type `ShmoFormFactor` just to separate them in case they diverge 🤔 
I'm thinking just keep it like this for now, but feel free to voice any opinion on this!

gorandalum (approved):

PR #4751, opened 2024-10-11
refactor: Use tGlobal to not make Alert functions reactive to language

gorandalum:
There are three alerts fixed in this PR. Can look at each commit individually to inspect them one by one.

### Acceptance criteria
- [ ] Should show alert if delete profile request fails
- [ ] Should show alert if requesting geolocation when rejected it before (set to never share in phone settings)
- [ ] Should show alert if unable to open app store/play store when deep-linking to mobility operator app (not sure how this is replicated, tested locally by simulating an error directly in code)


rosvik (approved):
Nice!

tormoseng:
✅  Two first point OK.
✅  Cannot replicate this after several tries. Will have to do with local testing.

PR #4757, opened 2024-10-14
chore: Fix compilation issues for Xcode 16 & mapbox

jorelosorio:
Compiling the app using the XCode 16 fails, these changes are intented to fix that + some upgrades needed for Mapbox.


marius-at-atb (approved):
Nice!

PR #4759, opened 2024-10-15
refactor: PurchaseSelectionBuilder

gorandalum:
Just a proof of concept of how the purchase selection module can look. Here it exports a hook `usePurchaseSelectionBuilder` which should be used everywhere one want to create or modify a purchase selection. 

Take a look and let me know how the module "feels". The main point is that the outward-facing API to usages outside of the module feels nice, while the secondary point is how the internal implementation in the module looks.

### DIFF
--- a/src/purchase-selection/__tests__/purchase-selection-builder.test.ts
+++ b/src/purchase-selection/__tests__/purchase-selection-builder.test.ts
@@ -10,7 +10,7 @@ describe('purchaseSelectionBuilder', () => {
   it('Selected zone is the first one if no zone specified as default', () => {
     const input = {
       ...DEFAULT_INPUT,
-      tariffZones: [{id: 'A'}, {id: 'B'}, {id: 'A'}],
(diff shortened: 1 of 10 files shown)


src/purchase-selection/__tests__/purchase-selection-builder.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const selection = createEmptyBuilder(input).forType('single').build();
+
+    expect(selection.fromPlace.id).toBe('A');
+  });
+
+  it('Selected zone is the zone specified as default', () => {
+    const input = {
+      ...DEFAULT_INPUT,
+      tariffZones: [{id: 'A'}, {id: 'B', isDefault: true}, {id: 'A'}],
```
gorandalum, line 24:
```suggestion
      tariffZones: [{id: 'A'}, {id: 'B', isDefault: true}, {id: 'C'}],
```

gorandalum:
Implemented in this PR:
https://github.com/AtB-AS/mittatb-app/pull/4796

PR #4760, opened 2024-10-16
feat: Support for better list views in markdown

jorelosorio:
Part of: 
https://github.com/AtB-AS/kundevendt/issues/17887

|Before|After|
|-|-|
|Simulator Screenshot - iPhone 16 Pro - 2024-10-16 at 09 11 21|Simulator Screenshot - iPhone 16 Pro - 2024-10-16 at 08 58 01|

Bullets points are aligned to the left and text to the right, there is a space between elements.


gorandalum (approved):
Nice! You may have done it already, but I would suggest just quickly going through the different places in the app where `isMarkdown={true}` are used, and see that there are no unwanted side effects.

PR #4761, opened 2024-10-16
feat: Support markdown in the toggle component subtext

jorelosorio:
Complement for: https://github.com/AtB-AS/mittatb-app/pull/4760

### DIFF
--- a/src/components/sections/items/ToggleSectionItem.tsx
+++ b/src/components/sections/items/ToggleSectionItem.tsx
@@ -14,6 +14,7 @@ import {SectionTexts, useTranslation} from '@atb/translations';
 type Props = SectionItemProps<{
   text: string;
   subtext?: string;
+  isSubtextMarkdown?: boolean;
   label?: LabelType;
   onValueChange: (checked: boolean) => void;
   value?: boolean;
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Looks good! I think if we start using more and more markdown in the app then we should at some point introduce a better pattern for this. Maybe introduce a new type which can be used throughout the app for this? If the `ThemeText` accepts this new type, then it is automatically handled by all components presenting text just by them also using this type in their params.

Totally pseudocode: 
```tsx
type TextType = string | { value: string, isMarkdown: boolean }

const ThemeText = (text: TextType) => {...}
const OtherComponent = (text: TextType, subText: TextType) => {
  return <>
    <ThemeText text={text}/>
    <ThemeText text={subText} />
  </>
}
```

jorelosorio:
> Looks good! I think if we start using more and more markdown in the app then we should at some point introduce a better pattern for this. Maybe introduce a new type which can be used throughout the app for this? If the `ThemeText` accepts this new type, then it is automatically handled by all components presenting text just by them also using this type in their params.
> 
> Totally pseudocode:
> 
> ```tsx
> type TextType = string | { value: string, isMarkdown: boolean }
> 
> const ThemeText = (text: TextType) => {...}
> const OtherComponent = (text: TextType, subText: TextType) => {
>   return <>
>     <ThemeText text={text}/>
>     <ThemeText text={subText} />
>   </>
> }
> ```

Nice suggestion, I am going to create an issue for this with refactor maintenance.

PR #4763, opened 2024-10-17
refactor: Use LaunchActivity on Android

gorandalum:
### Background
Reopening the app through the app icon on Android closes an open in-app-browser. This is not good when the user needs to do secure customer authentication (SCA) in another app. This only happens when using the app icon, if following a deeplink back to the app or using the app switcher, then this doesn't happen.

Description of the problem:
https://github.com/proyecto26/react-native-inappbrowser/issues/213#issuecomment-878915862

### Solution
The solution with adding a `LaunchActivity` described here:
https://github.com/proyecto26/react-native-inappbrowser/issues/213#issuecomment-2165493514

**Before (with just `MainActivity`) / After (with `LaunchActivity`)**
<div>

</div>

### Acceptance criteria
It is hard to know if there are any unwanted side effects of adding this `LaunchActivity`. I have tried looking it up, but couldn't really find a problem with it outside of that many intents could make the app slower. Since this is a pretty small one I don't think that should be noticable.

- [ ] Reopening backgrounded app through app icon doesn't close in-app-browser
- [ ] When compared to app releases before this one, there should be no negative changes or noticeable performance degrade when opening the app through app icon (both fresh open and reopen), app-switching and deeplinking.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -21,6 +21,14 @@
         android:allowBackup="false"
         android:theme="@style/AppTheme">
         <activity
+            android:exported="true"
+            android:name=".LaunchActivity">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>
+        <activity
             android:name=".MainActivity"
             android:label="@string/app_name"
             android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize|uiMode"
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):
Looks good AFAICT. Probably a good idea for @reidzeibel to look at this as well when he's back.

jorelosorio (approved):
I can not say too much from Android, the only thing is that we could have a look at the deep linking and see if that is affected, in the way react-native works.

gorandalum:
Note: @reidzeibel should take a post-merge look at this when he's back.

tormoseng:
✅  Looks good, and don't observe any performance degradation.

PR #4766, opened 2024-10-17
feat: add first iteration of journey aid screen

rosvik:
First iteration of https://github.com/AtB-AS/kundevendt/issues/19125

Most of the features are missing at this point, this is just to have a screen with journey info visible, then we'll iterate from that.

<div>

</div>

Note: There's some inconsistency between the use of "travel aid" and "journey aid" in the app now. I'll make a separate PR to fix this.

### DIFF
--- a/src/api/serviceJourney.ts
+++ b/src/api/serviceJourney.ts
@@ -1,4 +1,3 @@
-import {formatISO} from 'date-fns';
 import {client} from './client';
 import qs from 'query-string';
 import {stringifyUrl} from './utils';
@@ -7,11 +6,10 @@ import {ServiceJourneyWithEstCallsFragment} from '@atb/api/types/generated/fragm
 
 export async function getServiceJourneyWithEstimatedCalls(
   id: string,
-  date: Date,
+  isoDate: string,
 ): Promise<ServiceJourneyWithEstCallsFragment> {
   const encodedId = encodeURIComponent(id);
-  const formattedDate = formatISO(date, {representation: 'date'});
-  const url = `bff/v2/servicejourney/${encodedId}/calls?date=${formattedDate}`;
+  const url = `bff/v2/servicejourney/${encodedId}/calls?date=${isoDate}`;
   const response = await client.get<{
     value: ServiceJourneyWithEstCallsFragment;
   }>(url);
--- a/src/api/serviceJourney.ts
+++ b/src/api/serviceJourney.ts
@@ -4,12 +4,16 @@ import {stringifyUrl} from './utils';
 import {ServiceJourneyMapInfoData_v3} from '@atb/api/types/serviceJourney';
 import {ServiceJourneyWithEstCallsFragment} from '@atb/api/types/generated/fragments/service-journeys';
 
+/**
+ * @param id Service Journey ID
+ * @param serviceDate Service Journey date on the format "YYYY-MM-DD"
+ */
 export async function getServiceJourneyWithEstimatedCalls(
   id: string,
-  isoDate: string,
+  serviceDate: string,
 ): Promise<ServiceJourneyWithEstCallsFragment> {
   const encodedId = encodeURIComponent(id);
-  const url = `bff/v2/servicejourney/${encodedId}/calls?date=${isoDate}`;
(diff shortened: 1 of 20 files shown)


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -177,6 +181,14 @@ export const DepartureDetailsScreenComponent = ({
                 {title ?? t(DepartureDetailsTexts.header.notFound)}
               </ThemeText>
             </View>
+            {travelAidEnabled && (
```
jorelosorio, line 184: Just a small reminder here that this button is also affected by the toggle in the settings. But that toggle is in a separate PR.

rosvik: Yes, we need to fix that once your PR is merged 👍

src/travel-details-screens/use-departure-data.ts
```diff
@@ -99,7 +100,7 @@ export function useDepartureData(
   return [data, isLoading];
 }
 
-function addMetadataToEstimatedCalls(
+export function addMetadataToEstimatedCalls(
```
gorandalum, line 103: I Couldn't see the new usages. Does this need to be exported?

src/api/serviceJourney.ts
```diff
@@ -7,11 +6,10 @@ import {ServiceJourneyWithEstCallsFragment} from '@atb/api/types/generated/fragm
 
 export async function getServiceJourneyWithEstimatedCalls(
   id: string,
-  date: Date,
+  isoDate: string,
```
gorandalum, line 9: Could consider `string | Date` as type, and do the `formatToIsoDate` here. There is a principle that sounds something like "Don't make the client do what the API can do for them", although I can't remember the exact phrasing, but I think that might be a good fit here. At least I think it is architecturally ok for the api-module to use date utils.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const {themeName} = useTheme();
+
+  const {data: serviceJourney} = useTravelAidDataQuery(
+    serviceJourneyDeparture.serviceJourneyId,
+    formatToISODate(serviceJourneyDeparture.serviceDate),
+  );
+
+  // TODO: Change focused stop over time
+  const focusedEstimatedCall =
+    serviceJourney?.estimatedCalls && serviceJourney.estimatedCalls.length > 0
```
gorandalum, line 47: Could skip these two checks? Just have `serviceJourney?.estimatedCalls.find`?

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const styles = useStyles();
+  const {language, t} = useTranslation();
+  const {themeName} = useTheme();
+
+  const {data: serviceJourney} = useTravelAidDataQuery(
+    serviceJourneyDeparture.serviceJourneyId,
+    formatToISODate(serviceJourneyDeparture.serviceDate),
+  );
+
+  // TODO: Change focused stop over time
```
gorandalum, line 45: Since we have a TODO on this, maybe add one TODO regarding error handling and service journey not found?

gorandalum (approved):
Nice! Finding my inner nitpicker just to have something to comment on 😅

jorelosorio (approved):

src/api/serviceJourney.ts
rosvik: This made me look into this a bit closer, and I think the correct thing to do here is to remove date conversion all together, and rely on the serviceDate string. We don't need the date object at any point, and might even lead to time zone bugs.

Updated in 692676580fb58cd69e62293abd28ce94ea9e9443

rosvik: This fails in `date.test.ts`, so I think we might have had a bug here!

```ts
  describe('date conversion', () => {
    let datestring = '2024-01-25T23:00:00Z';
    let date = new Date(datestring);
    expect(formatISO(date, {representation: 'date'})).toBe('2024-01-25');
  });
```

PR #4768, opened 2024-10-18
refactor: ThemeIcon undefined svg safeguard

gorandalum:
We got some reports in Bugsnag where the app has crashed because
of the svg prop sent to ThemeIcon was undefined. This should really
not happen, but it is a little extreme that the app crashes if it
does. This safeguard is added to prevent app crash, rather there
will be no rendered ThemeIcon.

### Acceptance criteria
Haven't been able to reproduce, we will rely on code review for this one.


rosvik (approved):

marius-at-atb (approved):

tormoseng:
✅  Code review OK + tested quickly that the purchase flow is OK.

PR #4769, opened 2024-10-18
fix: Add refresh logic for id token

gorandalum:
There was no refresh logic for internal id token state in the
auth context. This wasn't really noticable before since the acios
client refetched the id tokens, so it didn't really matter that the
auth state was outdated. Now that we use id token from auth state
also for axios requests, then we need refresh logic in auth context.

### Acceptance criteria
- [ ] Id token refreshed from server when less than 5 minutes until expiry. Should happen only in authenticated state, and while the app is open. Id token is valid for 1 hour.
- [ ] Not invoke unnecessary fetch id token requests. (url `https://securetoken.googleapis.com/v1/token?…`)


rosvik (approved):
LGTM 👍

PR #4770, opened 2024-10-21
test: a small testId fix for estimated calls for departures

tormoseng:
A small testId-fix after https://github.com/AtB-AS/mittatb-app/pull/4765.


rosvik (approved):
🙈

PR #4772, opened 2024-10-21
feat: add travel aid error handling and loading state

rosvik:
part of https://github.com/AtB-AS/kundevendt/issues/19125

Need to merge https://github.com/AtB-AS/mittatb-app/pull/4771 before this one :)

### DIFF
--- a/src/api/serviceJourney.ts
+++ b/src/api/serviceJourney.ts
@@ -17,7 +17,7 @@ export async function getServiceJourneyWithEstimatedCalls(
   const response = await client.get<{
     value: ServiceJourneyWithEstCallsFragment;
   }>(url);
-  return response.data?.value;
+  return response.data.value;
 }
 
 export async function getServiceJourneyMapLegs(


src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -65,47 +62,77 @@ export const TravelAidScreenComponent = ({
...
+            type="error"
+            message={t(TravelAidTexts.error.message)}
+            onPressConfig={{action: refetch, text: t(dictionary.retry)}}
+          />
+        )}
+        {serviceJourney?.estimatedCalls && (
+          <TravelAidSection
+            serviceJourney={serviceJourney}
+            fromQuayId={serviceJourneyDeparture.fromQuayId}
+          />
```
gorandalum, line 77: There is some stuff here that seems unnatural or suboptimal with regards to usage of the react-query status and data. We can have a quick look at it together.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -65,47 +62,77 @@ export const TravelAidScreenComponent = ({
...
+      </GenericSectionItem>
+      <GenericSectionItem>
+        <View style={styles.sectionContainer}>
+          {status === TravelAidStatus.NoRealtime && (
+            <MessageInfoBox
+              type="error"
+              title={t(TravelAidTexts.noRealtimeError.title)}
+              message={t(TravelAidTexts.noRealtimeError.message)}
+            />
+          )}
```
gorandalum, line 119: It is kinda a shift in thought that not having real time data is such a bad thing that it gives a red error message. We generally have treated not having real time data just as normal data, presenting the aimed times instead. 

Seb has also been suggesting some changes that uses and shows realtime data even less than we do now, which would be a change in the other direction than this. I think it might be nice to have a general discussion with Seb and design lead on how central real time data should be in the app going forward, and whether not having realtime data is considered a faulty state or not.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -65,47 +62,77 @@ export const TravelAidScreenComponent = ({
...
+  serviceJourney: ServiceJourneyWithEstCallsFragment;
+  fromQuayId?: string;
+}) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+
+  const {status, focusedEstimatedCall} = getFocusedEstimatedCall(
+    serviceJourney.estimatedCalls!,
+    fromQuayId,
+  );
```
gorandalum, line 97: Nice reuse!

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -65,47 +62,77 @@ export const TravelAidScreenComponent = ({
...
+  fromQuayId,
+}: {
+  serviceJourney: ServiceJourneyWithEstCallsFragment;
+  fromQuayId?: string;
+}) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+
+  const {status, focusedEstimatedCall} = getFocusedEstimatedCall(
+    serviceJourney.estimatedCalls!,
```
gorandalum, line 95: Possible to avoid the `!`? Maybe just `return null` at the start?

gorandalum (approved):

PR #4778, opened 2024-10-25
refactor: `useFigmaStructure: true` for design system

tft-abel:
We are now using the structure for style variables that comes directly from Figma by using the feature flag `useFigmaStructure: true`. 

The interface for `<ThemeIcon />` has been simplified in this process. The properties `colorType`, `fill` and the SVG specific `color` have been replace with a single `color` property, which accepts `ContrastColor`, a text color (`primary`, `secondary` or `disabled`) or a direct `ColorValue` (e.g. hex string).


src/components/sections/Section.tsx
```diff
@@ -31,7 +31,7 @@ export function Section({
 
         const additionalProps: Partial<BaseSectionItemProps> = {
           radius: toRadius(index, lastIndex, firstIndex),
-          radiusSize: 'regular',
+          radiusSize: 'medium',
```
rosvik, line 34: Looks like this should be `'regular'`. Also, shouldn't this cause tsc errors? Could be related to the `Partial<BaseSectionItemProps>` type if it does elsewhere.

```suggestion
          radiusSize: 'regular',
```

tft-abel: Weird that this didn't trigger a `tsc` error, it did trigger errors elsewhere. I'll make the change and investigate the `Partial<BaseSectionItemProps>` type

tft-abel:
> I think this looks good!
> 
> It seems the sections has lost their border, might be because of what Johannes commented.
> 
> Looking at the API of our shared components like `ThemeText` and `ThemeIcon` I feel it might be worth it to be more precise and get rid of the helpers for now. Some suggestions on my part:
> 
> * Lets embrace the `ContrastColor` type! It is the core node of our design system, and I think we make it easier for ourselves if it gets the attention it deserves. So my suggestion is that we rename the `color` parameters on `ThemeIcon` and `ThemeText` to `contrastColor`, then everyone understands at first glance which type of parameter is accepted. Other components, for example those who need to know what background they are placed on, should also use`contrastColor` in their parameter naming, like `bgContrastColor` or `contrastColorBg`.
> * Lets get rid of some of the helper functions for now. It is not obvious that sending in a status color to `ThemeIcon` inverses the foreground and background colors. Let's rather just add a `inverseColors: boolean` parameter to the components for this case, and then it is up to the users of the component, not a hidden implementation detail inside the component themselves.
> 
> These suggestions are not written in stone, just me thinking out loud. Let me know what you think.

Johannes' fix solved the border-radius issue. 

I am afraid making changes to ThemeText and ThemeIcon here will become rather complex and hard to catch errors with this many changes in the PR already. I created a follow-up ticket to [make `<ThemeText/>` and `<ThemeIcon/>` great again](https://github.com/AtB-AS/kundevendt/issues/19412).

How about that?

src/chat/ContactSheet.tsx
```diff
@@ -16,14 +16,19 @@ import {ArrowRight, ExternalLink} from '@atb/assets/svg/mono-icons/navigation';
...
 import {useAnalytics} from '@atb/analytics';
 import {useParkingViolationsReportingEnabled} from '@atb/parking-violations-reporting';
+import {Theme} from '@atb/theme/colors';
 
 type Props = {
   onReportParkingViolation: () => void;
 };
 
+const getInteractiveColor = (theme: Theme) => theme.color.interactive[2]
+const getBackgroundColor = (theme: Theme) => theme.color.background.neutral[0]
```
rosvik, line 26: Considering how many if these there are in this PR, we should probably look for a more ergonomic way to fetch specific contrast and interactive colors from the design system. Not sure how that will work now, when ids like "background_0" isn't a thing anymore. Do you have any thoughts?

But I'm all for merging this PR first, then finding another solution later 👍

rosvik (approved):
ship it 🚀

src/chat/ContactSheet.tsx
tft-abel: The nice thing here is that it is very explicit what variable we are fetching, but it would be nice having a `token` function, where we can just write `token('color.background.neutral.0')` and get the best of both worlds. Perhaps this could even align with the way we access these tokens in CSS (with a custom plugin we talked about last week).

Some ideas :) I will get this merged first!

rosvik: Yeah the explicitness of it is nice! We'll see if some nice solution presents itself, or if the current solution works as we get used to it 😄

PR #4779, opened 2024-10-25
refactor: Add feature toggle context

gorandalum:
This is a refactor to use a new FeaturesToggleContext as the way
toggles work in the app now has been slightly irritating me for some
time, especially the need for restarts and quite a bit of code
necessary for adding new ones.

This way the feature toggles will be loaded at app startup, and
kept in sync while the app is open. So no more need for restart
after changing override in debug menu, and no more need for the
debugOverrideReady flag.

New toggles are added by just modifying the toggleSpecifications
list, with no further code necessary.

The context, especially the types, is a little complex, but this is to
favour simplicity and strict type safety when using the context and
adding new toggles.

**Tips for review: Review the three commits one by one**
- The first commit is the core of the implementation, creation of the context, and utility methods/typing. Review the implementation with regards to ease of use and maintainability.
- The second commit is the work of actually migrating the old feature toggles to the new ones. No major changes are done here, mainly some changing of variable names. Look for simple typos.
- The third commit is adding tests, and some refactoring to increase testability. Take a look if there are something that should be tested that is missing.


src/storage/StorageModel.tsx
rosvik, line 1: Is it better to use "multiple" for naming rather than "multi"? Not that much longer 😄

src/feature-toggles/toggle-specifications.ts
```diff
@@ -0,0 +1,34 @@
...
+  // These are not in use in this commit. Just to show how they will be written.
+  {
+    name: 'isActivateTicketNowEnabled',
+    remoteConfigKey: 'enable_activate_ticket_now',
+  },
+  {
+    name: 'isPosthogEnabled',
+    remoteConfigKey: 'enable_posthog',
+  },
+] as const satisfies readonly FeatureToggleSpecification[];
```
rosvik, line 128: Fancy!

rosvik (approved):
This looks really nice 🙌

src/storage/StorageModel.tsx
gorandalum: I saw your comment and ment to rename this, but forgot it before merging 🙈

tormoseng:
✅

PR #4782, opened 2024-10-28
feat: Add situation / notices messages

jorelosorio:
Closes: https://github.com/AtB-AS/kundevendt/issues/19126

NOTE: Notices / Situations are read on changes, we might need to just announce it once but we need to agree with PO, Designers, Developers.


src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -114,7 +121,21 @@ const TravelAidSection = ({
     fromQuayId,
   );
 
-  useTravelAidAnnouncements({status, focusedEstimatedCall});
+  let notices: NoticeFragment[] = [];
+  if (serviceJourney !== undefined) {
+    notices = getNoticesForServiceJourney(serviceJourney, fromQuayId) ?? [];
+  }
```
rosvik, line 127: Can this be simplified to this?

```suggestion
  const notices = getNoticesForServiceJourney(serviceJourney, fromQuayId);
```

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -114,7 +121,21 @@ const TravelAidSection = ({
...
+
+  const situations =
+    focusedEstimatedCall?.situations.sort((n1, n2) =>
+      n1.id.localeCompare(n2.id),
+    ) ?? [];
```
rosvik, line 132: `focusedEstimatedCall` shouldn't ever be undefined

```suggestion
  const situations = focusedEstimatedCall.situations.sort((n1, n2) =>
    n1.id.localeCompare(n2.id),
  );
```

rosvik:
@jorelosorio With your latest changes, how does this work when both situation messages and stop information changes? Won't the screen reader announcements from useAnnounceSituationOrNotices and useTravelAidAnnouncements trigger at the same time?

jorelosorio:
> @jorelosorio With your latest changes, how does this work when both situation messages and stop information changes? Won't the screen reader announcements from useAnnounceSituationOrNotices and useTravelAidAnnouncements trigger at the same time?

That is a good question, as it is now it announces what first comes updated, however during my tests I saw that at least iOS respect was being said is finished to continue with the new announcement. That is why I do not implement a mechanism to queue all announcements since it could become more complex. At leas the documentation mentions it. https://reactnative.dev/docs/accessibilityinfo#announceforaccessibility but not quite sure what solution we want here.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -159,6 +182,21 @@ const TravelAidSection = ({
...
+                <SituationMessageBox key={situation.id} situation={situation} />
+              ))}
+
+              {noticesForSelected.map(
+                (notice) =>
+                  notice.text && (
+                    <MessageInfoBox type="info" message={notice.text} />
+                  ),
+              )}
+            </View>
```
rosvik, line 198: Should `situationsForFocusedStop` also be displayed here?

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -114,10 +119,33 @@ const TravelAidSection = ({
...
 
-  useTravelAidAnnouncements({status, focusedEstimatedCall});
+  const selectedEstimatedCall = serviceJourney.estimatedCalls.find(
+    (e) => e.quay.id === fromQuayId,
+  );
+  const situationsForSelected =
+    selectedEstimatedCall?.situations.sort((n1, n2) =>
+      n1.id.localeCompare(n2.id),
+    ) ?? [];
+  const noticesForSelected = selectedEstimatedCall?.notices ?? [];
```
gorandalum, line 129: For `situations` it is enough (and advised) too only look at `estimatedCall.situations` since this is a collection of all messages on relevant entities (stop place, quay, service joyney, line).

However, this is not the case for `notices`, and other places in the app where we show `notices` we "manually" retrieve them from the different relevant entities, like we do here:
https://github.com/AtB-AS/mittatb-app/blob/cf7423f955ea07cc8b2ac542ddd128ae3b17fc87/src/travel-details-screens/utils.ts#L71

Here is an example of why we needed to do this in our app, where a critical message regarding ferries for Fram where not shown:
https://github.com/AtB-AS/kundevendt/issues/2881

I also posted some relevant info here:
https://mittatb.slack.com/archives/C02EEG7D8EL/p1729842515032949

@jorelosorio @rosvik It might be that you have already decided to not do this extra effort on `notices` in the travel aid screen, but I'm just mentioning it in case there haven't been done a conscious decision on it. And whatever is decided should be reflected in the acceptance criteria or things to test.

jorelosorio: @gorandalum We did not decide to do it because it took a bit of effort, we intended to be the most precise as possible. So we can resume that for notices we need to use the `getNoticesForServiceJourney` function. Or do we need to do some extra filter?

gorandalum: I think the same code can be used, the question is more whether you have the data readily available or not.

PR #4783, opened 2024-10-28
feat: add travel aid screen reader announcements

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/19125

Adds a useEffect to the travel aid screen that reads out updates every time the a11y label changes.

Since hooks can't be run conditionally and I needed to check if `serviceJourney.estimatedCalls` was undefined before rendering TravelAidSection, I had to move the undefined check, and do some type-trickery to make things work. Added a type to utils to make it easier to discover this method in the future. ~Found one spot where it was used before.~


gorandalum:
> Since hooks can't be run conditionally and I needed to check if `serviceJourney.estimatedCalls` was undefined before rendering TravelAidSection, I had to move the undefined check, and do some type-trickery to make things work. Added a type to utils to make it easier to discover this method in the future. Found one spot where it was used before.

There is another trick for this as well. Take a look at the `ScreenReaderAnnouncement`, where the hook that should be conditionally run is wrapped in a component.

gorandalum (approved):

PR #4785, opened 2024-10-29
refactor: Map new architecture PoC

gorandalum:
PoC for this architecture:
Map refactor


src/components/map_v2/plugins/el-scooter-plugin/use-el-scooters-plugin.tsx
```diff
@@ -0,0 +1,46 @@
...
+  setSharedState,
+}) => {
+  const {open: openBottomSheet, close: closeBottomSheet} = useBottomSheet();
+  const vehicles = useVehicles({SCOOTER: {showAll: true, operators: []}});
+  const elScooterFeatures = vehicles?.vehicles.scooters;
+  const updateRegion = vehicles?.updateRegion;
+
+  useEffect(() => {
+    if (region) updateRegion?.(region);
+  }, [region, updateRegion]);
```
marius-at-atb, line 19: After the tile server + vector source is in place, we can completely remove the updateRegion stuff :)

src/components/map_v2/plugins/el-scooter-plugin/use-el-scooters-plugin.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <ScooterSheet
+              vehicleId={scooterFeature.properties.id}
+              onClose={() => {
+                setSharedState({selectedEntityId: undefined});
+                closeBottomSheet();
+              }}
+              onReportParkingViolation={() => {}}
+            />
+          );
+        }, false);
```
marius-at-atb, line 36: I think there is something very "un-react" about calling functions like this.
I would prefer dispatching an action that would set the app in the desired state 🤔

marius-at-atb:
An interesting approach!
I guess the onPress handling is the main advantage of the plugin setup as it is now.
However I'm not sure if that's worth it compared to using components directly that depend on shared state, and having a unified onPress handler 🤔 
As the complexity of the BottomSheet usage increases (and also its coordination with the map), I suspect reducer state management can be valuable.
I think it's a good idea to also try an example implementation of that approach, and see what works best.

adriansberg:
@marius-at-atb, is this PR still relevant?

marius-at-atb:
@adriansberg Still relevant as reference and as an option if we see others solutions become problematic, but won't be merged, so closing this for now.

PR #4786, opened 2024-10-30
feat: Add nynorsk feature toggle

gorandalum:
### Acceptance criteria
- [ ] Nynorsk should be selectable or not in the language screen based on the feature toggle.

### DIFF
--- a/src/feature-toggles/toggle-specifications.ts
+++ b/src/feature-toggles/toggle-specifications.ts
@@ -66,6 +66,10 @@ export const toggleSpecifications = [
     remoteConfigKey: 'enable_non_transit_trip_search',
   },
   {
+    name: 'isNynorskEnabled',
+    remoteConfigKey: 'enable_nynorsk',
+  },
+  {
     name: 'isOnBehalfOfEnabled',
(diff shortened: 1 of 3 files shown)


marius-at-atb:
@Sebstorvik I don't think it makes any sense to hide a supported language?
And as this PR shows, it adds unnecessary logic to maintain..

mortennordseth:
@marius-at-atb, NFK, and Troms do not want Nynorsk enabled in the app, so a feature flag would be nice.

marius-at-atb (approved):
Approving the code, although still not a fan of the feature 😛

gorandalum:
Note that the feature is **opt-out**. Nynorsk will appear in the language menu unless `enable_nynorsk` is set to `false` in RemoteConfig.

tormoseng:
✅  Added a RemConf in Staging and Prod.

@Sebstorvik @mortennordseth Note that if the user has nynorsk set on his/her phone, then nynorsk will also be in the app.

mortennordseth:
Added `enable_nynorsk: false` to RemoteConfig for NFK and Svipper.

PR #4789, opened 2024-10-31
fix: handle focus for travel aid error and loading states

rosvik:
Sets focusRef on the loading indicator and error message on the travel aid screen, fixing point 4 from comment https://github.com/AtB-AS/kundevendt/issues/19125#issuecomment-2446398675

closes https://github.com/AtB-AS/kundevendt/issues/19125


jorelosorio (approved):

tormoseng:
✅  Works well!

PR #4794, opened 2024-11-04
fix: sort out issues with splash screen

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/19454
related to https://github.com/AtB-AS/mittatb-app/pull/4763

This fix is a hotfix that should be already released on version 1.58.1-rc2.

It seems that the issue occurring on 1.58 was caused by the wrong splash screen placement. In 1.58, the `LaunchActivity` doesn't have any theme or splash screen on it, so when it launches the `MainActivity`, it will then show the splash screen stuffs. While this shouldn't block anything, on some Android devices, the splash screen blocks the user from interacting with the UI.

The fix here is to move all splash screen related functionality to the `LaunchActivity` instead, so when it launches `MainActivity`, it should directly shows the app, since the splash screen is shown on the `LaunchActivity`.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -38,8 +38,7 @@
             android:screenOrientation="portrait"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustPan"
-            android:exported="true"
-            android:theme="@style/BootTheme">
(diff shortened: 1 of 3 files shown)


gorandalum (approved):
LGTM :+1:

PR #4796, opened 2024-11-05
feat: Add purchase selection module

gorandalum:
Add a new module for creation, modification and testing of the
purchase selection domain model. It uses a builder pattern for
creating and modifying a purchase selection, which ensures that
the built purchase selection is valid and referentially stable
when not edited. The builder it self is not referentially stable,
and should not be passed around or used in hooks dependency lists.

The purchase selection module is thoroughly tested, and it is
configured a required coverage of 100% for the purchase selection
module. This is also checked during the app build.

- The first commit is the builder and the tests
- The second commit is the refactor to use the new builder


reidzeibel (approved):
I can't seem to find anything peculiar. Although seeing the builder pattern in React/TS surprised me. I read somewhere that builder pattern isn't something that is commonly found (especially when building components, this one isn't).

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/navigation-types.ts
```diff
@@ -1,5 +1,5 @@
 import {TicketRecipientType} from '@atb/ticketing';
-import {PurchaseSelectionType} from '@atb/stacks-hierarchy/types.ts';
+import {PurchaseSelectionType} from "@atb/purchase-selection";
```
jorelosorio, line 2: before it was a single quote ', now is double ", maybe to follow how it is in other places.

src/purchase-selection/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+};
+
+export const isSelectableZone = (
+  product: PreassignedFareProduct,
+  zone: TariffZone,
+) => {
+  const zoneLimitations = product.limitations.tariffZoneRefs;
+  return (
+    !zoneLimitations?.length ||
```
jorelosorio, line 116: This || here is a bit confusing especially because of a check with some.

Not sure if you want here is that product.limitations.tariffZoneRefs contains the zone, to be considered as selectable.

If !zoneLimitations?.length is true (meaning zoneLimitations is either absent or an empty array), the function immediately returns true without evaluating the some condition.

Is a selectable zone for an empty array? or undefined array?

src/purchase-selection/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  product: PreassignedFareProduct,
+): UserProfile[] =>
+  userProfiles.filter((profile) => isSelectableProfile(product, profile));
+
+export const isSelectableProduct = (
+  input: PurchaseSelectionBuilderInput,
+  selection: PurchaseSelectionType,
+  product: PreassignedFareProduct,
+) => {
+  if (selection.fareProductTypeConfig.type !== product.type) return false;
```
jorelosorio, line 93: It might be clear just by reading the code, but only if you know the context. the parameter selection is preselected in the UI? you do not want to include himself as selectable even it can be selectable? I think this util should only respond to isProductSellableInApp, the decision of selection must be done higher, and this encapsulates a logic that does not belong here.

Also would be worth it to add a small documentation in the method itself to know what selection and input are.

src/purchase-selection/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  product: PreassignedFareProduct,
+): UserProfileWithCount[] => {
+  const selectableProfiles = getSelectableUserProfiles(
+    input.userProfiles,
+    product,
+  );
+  const profile = selectableProfiles.reduce((selected, current) =>
+    current.userTypeString === input.defaultUserTypeString ? current : selected,
+  );
+  return [{...profile, count: 1}];
```
jorelosorio, line 79: If only returns one profile why it expect an array?

src/purchase-selection/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+      compareVersion(product.limitations.appVersionMin, input.appVersion) >
+        0) ||
+    (product.limitations.appVersionMax &&
+      compareVersion(product.limitations.appVersionMax, input.appVersion) < 0)
+  )
+    return false;
+
+  if (
+    product.distributionChannel.some((channel) => channel === 'debug-app') &&
+    input.customerProfile?.debug
```
jorelosorio, line 28: I understand that this is used for testing but should not this belong to the test itself not as part of the utils?

.gitignore
```diff
@@ -108,4 +108,7 @@ yalc.lock
 .cxx/
 
 # Temporary files created by Metro to check the health of the file watcher
-.metro-health-check*
\ No newline at end of file
+.metro-health-check*
+
+# Test coverage reports
+/coverage
```
jorelosorio, line 114: I would like to see the introduction of coverage in a separate PR and if we can use an opensource tool like https://about.codecov.io/ and put a label in our repo.

jorelosorio:
I have small comments spread here and there, they are not strong in a way that blocks this PR but would be nice in general have some extra documentation in the methods to know more context. It might sound like a bit of extra work but it would be nice to have a small talk with the team before introducing new patterns, just to have a common agreement of what to expect at least when doing PRs.

Other than that, nothing else to say, it looks nice to see more tests in the implementations and good work!

src/purchase-selection/utils.ts
reidzeibel: I think this is related to tickets that are sold to the distribution channel of `debug-app`, and also only allowed for users marked with `debug`. This is mainly here to allow testing new tickets on prod, such as when Gunn needs to test something.

src/purchase-selection/utils.ts
gorandalum: A zone is selectable if either there is no limitations (empty `zoneLimitations`) or if the limitations includes the given zone id.

I now created a boolean variable `emptyLimitations` that could make it a little more clear.

```ts
  const zoneLimitations = product.limitations.tariffZoneRefs;
  const emptyLimitations = !zoneLimitations?.length;
  return (
    emptyLimitations ||
    zoneLimitations.some((allowedZoneId) => zone.id === allowedZoneId)
  );
```

src/purchase-selection/utils.ts
gorandalum: > It might be clear just by reading the code, but only if you know the context. the parameter selection is preselected in the UI?

It is the current purchase selection, which should be what is presented to the user. I renamed the parameter to `currentSelection` to make it more clear.

> I think this util should only respond to isProductSellableInApp, the decision of selection types is not the same must be done higher level as a first validation, and this encapsulates a logic that does not belong here.

You are right that there needs to be some logic regarding this higher in the app. This module doesn't cover presentation logic, and what product options to display for a given type config is logic that also needs to be present in the presentation layer. However, it must still be checked here as well, since one of the features of this new module is that it guarantees the validity of the purchase selection it creates, so it can't rely on checks being done outside of the module.

> Also would be worth it to add a small documentation in the method itself to know what selection and input are.

As mentioned I renamed the `selection` parameter to `currentSelection`. Regarding `input`, that parameter is used throughout the whole module. Instead of documenting the param, I added some doc on the `PurchaseSelectionBuilderInput` type itself.

src/purchase-selection/utils.ts
gorandalum: Yeah, this helper could have been `getDefaultUserProfile` and only returned a single profile, and then builder it self could map it to `[{...profile, count: 1}]`. But this is a helper to make the builder as slim/clean as possible, and from the viewpoint of the builder it needs default value for the field `userProfiles`, and then it makes sense to have it as a `getDefaultUserProfiles` function.

src/purchase-selection/utils.ts
gorandalum: Yeah, that's correct. Even though the code mentions 'debug' and its purpose is to allow testing, it it still production code.

gorandalum:
> It might sound like a bit of extra work but it would be nice to have a small talk with the team before introducing new patterns, just to have a common agreement of what to expect at least when doing PRs.

Fully agree with that! A rough outline of the plan was described in the [PR for the `PurchaseSelectionType`](https://github.com/AtB-AS/mittatb-app/pull/4755). There was also created a [draft PR ](https://github.com/AtB-AS/mittatb-app/pull/4759) which had an API-suggestion, with the developers in the team as reviewers, and an accompanying [post on Slack](https://mittatb.slack.com/archives/C02EEG7D8EL/p1729587218910419) asking for feedback.

src/purchase-selection/utils.ts
jorelosorio: It is good to refresh this concept, I was a bit confused.

src/purchase-selection/utils.ts
jorelosorio: I see, good to know and for the details, at first glance it gives different impressions.

jorelosorio:
> > It might sound like a bit of extra work but it would be nice to have a small talk with the team before introducing new patterns, just to have a common agreement of what to expect at least when doing PRs.
> 
> Fully agree with that! A rough outline of the plan was described in the [PR for the `PurchaseSelectionType`](https://github.com/AtB-AS/mittatb-app/pull/4755). There was also created a [draft PR ](https://github.com/AtB-AS/mittatb-app/pull/4759) which had an API-suggestion, with the developers in the team as reviewers, and an accompanying [post on Slack](https://mittatb.slack.com/archives/C02EEG7D8EL/p1729587218910419) asking for feedback.

I saw you made it, there was not so much movement in the topic, but we could make this process better by making more active suggestions and proposing ideas from all of us and communicating them more often. I will do my best to support this and promote more team dialogues.

PR #4798, opened 2024-11-05
feat: Handle 400 NotAvailable error from offer search

gorandalum:
This error comes when the the travel date is outside of the
availability times configured for the product.

This is how the error looks:

With this change the empty offer response is just shown as a generic error:

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
@@ -19,7 +19,7 @@ export type UserProfileWithCountAndOffer = UserProfileWithCount & {
 };
 
 export type OfferError = {
-  type: ErrorType | 'empty-offers';
+  type: ErrorType | 'empty-offers' | 'not-available';
 };
 
 type OfferState = {
@@ -233,8 +233,13 @@ export function useOfferState(
               },
             });
           }
-        } catch (err) {
-          const errorType = getAxiosErrorType(err);
+        } catch (err: any) {
+          const isNotAvailableError =
+            err.response?.status === 400 &&
+            err.response?.data === 'NotAvailable';
+          const errorType = isNotAvailableError
+            ? 'not-available'
+            : getAxiosErrorType(err);
           if (errorType !== 'cancel') {
             console.warn(err);
             dispatch({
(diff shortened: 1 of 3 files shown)


src/translations/screens/subscreens/PurchaseOverview.ts
```diff
@@ -18,12 +18,19 @@ const PurchaseOverviewTexts = {
...
-      _(
-        `${productName} er ikke tilgjengelig akkurat nå.`,
-        `${productName} is not available right now.`,
-        `${productName} er ikkje tilgjengeleg nett no.`,
+    productUnavailable: {
+      title: (productName: string) =>
+        _(
+          `${productName} er ikke gyldig nå`,
+          `${productName} is not valid now`,
+          `${productName} er ikkje gyldig nå`,
```
marius-at-atb, line 26:
```suggestion
          `${productName} er ikkje gyldig no`,
```

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
```diff
@@ -233,8 +233,13 @@ export function useOfferState(
...
           }
-        } catch (err) {
-          const errorType = getAxiosErrorType(err);
+        } catch (err: any) {
+          const isNotAvailableError =
+            err.response?.status === 400 &&
+            err.response?.data === 'NotAvailable';
+          const errorType = isNotAvailableError
+            ? 'not-available'
+            : getAxiosErrorType(err);
```
marius-at-atb, line 242: Do we expect the NotAvailable error to be unique to offers?
If not, it could be an idea to handle this inside `getAxiosErrorType`.

Since `'empty-offers'` isn't being used anymore, that would also make it possible to remove the `OfferError` type and just use `ErrorType` directly.

marius-at-atb (approved):

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
gorandalum: > Do we expect the NotAvailable error to be unique to offers?
If not, it could be an idea to handle this inside getAxiosErrorType.

The 'NotAvailable' is unique to offers endpoint in the ticket legacy service. Platform team will look at a more proper fix when refactoring to the planned sales services.

> Since 'empty-offers' isn't being used anymore, that would also make it possible to remove the OfferError type and just use ErrorType directly.

It could be possible, but it would need the empty-offers error to be "masked" as an axios error, which isn't preferable.

PR #4799, opened 2024-11-06
fix: Remove duplicated messages on Journey Aid screen

jorelosorio:
Related to: https://github.com/AtB-AS/kundevendt/issues/19126

This change fixes duplications and get information of notices from: https://github.com/AtB-AS/mittatb-app/blob/cf7423f955ea07cc8b2ac542ddd128ae3b17fc87/src/travel-details-screens/utils.ts#L71

### DIFF
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -233,12 +233,12 @@ const useTravelAidAnnouncements = (
     if (state.focusedEstimatedCall.quay.id !== previousQuayId.current) {
       previousQuayId.current = state.focusedEstimatedCall.quay.id;
       AccessibilityInfo.announceForAccessibility(
-        message + screenReaderPause + noticies + screenReaderPause + situations,
+        message + screenReaderPause + notices + screenReaderPause + situations,
       );
     } else {
       AccessibilityInfo.announceForAccessibility(message);
     }
-  }, [message, situations, noticies, state.focusedEstimatedCall.quay.id]);
+  }, [message, situations, notices, state.focusedEstimatedCall.quay.id]);
 };
 
 const TimeInfo = ({state}: {state: FocusedEstimatedCallState}) => {
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -18,7 +18,10 @@ import {
   useTranslation,
 } from '@atb/translations';
 import {TravelAidTexts} from '@atb/translations/screens/subscreens/TravelAid';
-import {getLineA11yLabel} from '@atb/travel-details-screens/utils';
+import {
+  getLineA11yLabel,
+  getNoticesForServiceJourney,
+} from '@atb/travel-details-screens/utils';
 import {ThemeIcon} from '@atb/components/theme-icon';
 import {Realtime as RealtimeDark} from '@atb/assets/svg/color/icons/status/dark';
 import {Realtime as RealtimeLight} from '@atb/assets/svg/color/icons/status/light';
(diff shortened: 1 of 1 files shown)


gorandalum (changes requested):

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -123,28 +126,33 @@ const TravelAidSection = ({
...
-      situationsForSelected?.find((s) => s.id !== situation.id),
-    ) ?? [];
 
-  useTravelAidAnnouncements(
-    {status, focusedEstimatedCall},
-    situationsForFocusedStop,
+  const uniqueSituations: SituationType[] = [];
+  const seenIds = new Set();
+
+  [...selectedEstimatedCall, ...focusedEstimatedCall.situations].forEach(
```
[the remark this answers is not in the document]
gorandalum: I didn't really understand that, could you elaborate a little? I don't think global messages are connected to this?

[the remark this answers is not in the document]
jorelosorio: by global messages, I meant notices/situations (messages) in an entire journey, those at the top in the UI, which are different in the sub-sections (legs, trips, stops, etc). The code do not use references to global messages (those in forestore) either.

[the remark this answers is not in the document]
gorandalum: So if I understand correctly, the assumption was that one of `selectedEstimatedCall` and `focusedEstimatedCall` has the higher-level (what you describe as global) situations, while the other one has situations relevant for the departure.

I don't think that is a precise description of it. The selected and focused call are both containing relevant situations for both that specific call **and** the higher-level situations:
`selectedCallSituations` -> Situations for selected call, the quay, the stop place, the service journey and the line
`focusedCallSituations` -> Situations for focused call, the quay, the stop place, the service journey and the line

The quay, stop place, service journey and line are the same for both the focused and estimated call, and thats one reason why there was duplicates.

So the question is, do we need to show the situations for both the focused and the selected estimated call in this screen? When the focused call is a call later in the service journey than the selected call, that means the selected call is passed, and will it then be relevant to show the situations for it?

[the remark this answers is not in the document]
jorelosorio: We did a call @gorandalum @rosvik and I @jorelosorio to clarify concepts and took some decisions on those notifications and situations, we agreed to show situations for the focusedCallSituations.

gorandalum (approved):

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -123,30 +126,24 @@ const TravelAidSection = ({
...
-  const situationsForFocusedStop =
-    focusedEstimatedCall.situations.filter((situation) =>
-      situationsForSelected?.find((s) => s.id !== situation.id),
-    ) ?? [];
 
   useTravelAidAnnouncements(
     {status, focusedEstimatedCall},
-    situationsForFocusedStop,
+    situationsForFocused,
+    noticesForFocused,
```
rosvik, line 138: What did we decide in regards to notices? Will they change for different calls, or can we remove them from the announcements?

gorandalum: I thought that there would be another PR for announcements, handling both `notices` and `situations`, and as such didn't really look into it on this PR.

jorelosorio: Yes, that is the plan to handle it on another PR, but we decided to announce it all once and then filter the new ones based on some kind of announced IDs.

rosvik: If it should be handled in another PR, then I think you should remove the changes to useTravelAidAnnouncements from this PR.

jorelosorio: I added it to have some base structure before starting the other and also to at least announce the notices as well.

rosvik (approved):

PR #4800, opened 2024-11-06
feat: Remove Flipper

gorandalum:
Flipper is deprecated, not used that much by us, and also is not
possible to use together with the newest RN Firebase version.


jorelosorio (changes requested):
Some checks are failing.

reidzeibel (approved):
Seems reasonable to remove flipper since it doesn't work anyway.

[Announcement from Flipper GitHub page](https://github.com/facebook/flipper?tab=readme-ov-file#react-native-support)

PR #4801, opened 2024-11-07
chore: Update Firebase dependencies

gorandalum:
The reason for updating Firebase is that the Android recaptcha lib that Google wants us to update is not a library we use directly, but an inherited dependency from Firebase Auth. By updating Firebase we get the necessary update to a newer Android recaptcha lib version.

To reduce risk we update to version 20.5.0 of Firebase, and not the recent newest version (21.3.0).

Regarding the changes in the Podfile:
We introduce statically linked frameworks, as described as a requirement in the documentation:
https://rnfirebase.io/#altering-cocoapods-to-use-frameworks

The doc also contains a relevant section regarding why Flipper was removed from the app:
> Notes: React-Native-Firebase uses use_frameworks, which has compatibility issues with Flipper & Fabric.
>
>Flipper: use_frameworks [is not compatible with Flipper](https://github.com/reactwg/react-native-releases/discussions/21#discussioncomment-2924919). You must disable Flipper by commenting out the :flipper_configuration line in your Podfile. Flipper is deprecated in the react-native community and this will not be fixed - Flipper and react-native-firebase will never work together on iOS.
>
>New Architecture: Fabric is partially compatible with use_frameworks!. If you enable the bridged / compatibility mode, react-native-firebase will compile correctly and be usable.


reidzeibel (approved):
If the auth flow works properly, then it should be safe to merge! LGTM

PR #4802, opened 2024-11-07
fix: Announce situations and notices

jorelosorio:

This PR allows notices/situations to be selected/focused to be used in accessibility mode, and improve the announcements.

### DIFF
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -3,7 +3,7 @@ import {Button} from '@atb/components/button';
 import {EstimatedCallInfo} from '@atb/components/estimated-call';
 import {StyleSheet, useTheme} from '@atb/theme';
 import {ServiceJourneyDeparture} from '@atb/travel-details-screens/types';
-import React, {Ref, useEffect, useRef} from 'react';
+import React, {Ref, useEffect, useRef, useState} from 'react';
 import {SafeAreaView} from 'react-native-safe-area-context';
 import {useTravelAidDataQuery} from './use-travel-aid-data';
 import {ScrollView} from 'react-native-gesture-handler';
@@ -41,6 +41,7 @@ import {RequireValue} from '@atb/utils/object';
 import {getSituationSummary} from '@atb/situations/utils';
 import {SituationType} from '@atb/situations/types';
 import {isDefined} from '@atb/utils/presence';
+import {onlyUniques} from '@atb/utils/only-uniques';
 
 export type TravelAidScreenParams = {
   serviceJourneyDeparture: ServiceJourneyDeparture;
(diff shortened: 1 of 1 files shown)


src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -216,29 +230,58 @@ const useTravelAidAnnouncements = (
...
+    );
+  }, [
+    message,
+    state.focusedEstimatedCall.quay.id,
+    currentAnnouncedSituationIds,
+    currentAnnouncedNoticeIds,
+    noticesForFocusedStop,
+    situationsForFocusedStop,
+    language,
+  ]);
```
gorandalum, line 284: Such large dependency lists are quite risky, and it also seems that some of the elements here are not referentially stable between renders? Just by reading the code I would be a little scared that something is announced on every render cycle. 

If possible, try to create the message outside of the `useEffect`, and let the effect be stripped down to just what it should actually do: Announce a message.

gorandalum (changes requested):

tormoseng:
Please pay attention to the acceptance criteria created in https://github.com/AtB-AS/kundevendt/issues/19126 when reviewing this PR 😊

src/travel-aid/TravelAidScreenComponent.tsx
jorelosorio: Sure, I going to refactor a bit more this part.

jorelosorio:
> Please pay attention to the acceptance criteria created in [AtB-AS/kundevendt#19126](https://github.com/AtB-AS/kundevendt/issues/19126) when reviewing this PR 😊

I added more code to this PR based on the meeting today and what @tormoseng is pointing here :)

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -126,26 +132,39 @@ const TravelAidSection = ({
...
 
   const quayName = getQuayName(focusedEstimatedCall.quay) ?? '';
 
-  const accessibilityLabel =
-    getNoticesA11yLabel(noticesForFocused) +
-    screenReaderPause +
-    getSituationA11yLabel(situationsForFocused, language) +
-    screenReaderPause +
-    getFocussedStateA11yLabel({status, focusedEstimatedCall}, t, language);
+  const accessibilityLabel = getFocussedStateA11yLabel(
```
rosvik, line 163: Maybe rename to `focussedStateA11yLabel` for clarity

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -126,26 +132,39 @@ const TravelAidSection = ({
...
+      serviceJourney,
+      focusedEstimatedCall.quay.id,
+    ).sort((n1, n2) => n1.id.localeCompare(n2.id)) ?? [];
+
+  const announcedSituationIds = situationsForFocused
+    .map((s) => s.id)
+    .filter(onlyUniques);
+  const announcedNoticeIds = noticesForFocused
+    .map((s) => s.id)
+    .filter(onlyUniques);
```
rosvik, line 150: Looks like this can be moved into `useTravelAidAnnouncements`. That way you don't have to pass them as props.

rosvik (approved):
This looks good 🙌 Some minor suggestions, but nothing affecting functionality.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -97,6 +100,7 @@ export const TravelAidScreenComponent = ({
               ...serviceJourney,
               estimatedCalls: serviceJourney.estimatedCalls,
             }}
+            isTripCancelled={serviceJourneyDeparture.isTripCancelled}
```
gorandalum, line 103: I service journey isn't actually cancelled, it is the individual calls that can be cancelled. So the `isTripCancelled` on the `ServiceJourneyDeparture` is actually that the estimated call from the `fromQuayId` is cancelled.

Because of that I think it would be better to use the `cancelled` param on the `serviceJourney.estimatedCall`. The reason for that is mainly that we can then present the `cancelled` value for the focused call, and also that we get more up to date data as the `serviceJourney` is fetched continuously, while the `serviceJourneyDeparture` is more static input parameter to the screen.

gorandalum (changes requested):

gorandalum (approved):

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -257,9 +256,7 @@ const useTravelAidAnnouncements = (
 
   const message =
     getFocussedStateA11yLabel(state, t, language) +
-    (isTripCancelled
-      ? screenReaderPause + t(CancelledDepartureTexts.message)
-      : '');
+    (cancelled ? screenReaderPause + t(CancelledDepartureTexts.message) : '');
   screenReaderPause +
```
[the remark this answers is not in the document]
jorelosorio: Yes, should not be there, here a fix: https://github.com/AtB-AS/mittatb-app/pull/4806

PR #4805, opened 2024-11-11
chore: upgrade babel and deduplicate dependencies

rosvik:
While working on https://github.com/AtB-AS/kundevendt/issues/19400, I ran into some issues with babel, and conflicting dependencies when I tried to upgrade it. Running `npx yarn-deduplicate yarn.lock` fixed the issue, and deleted about 2GB of my node_modules folder 🔥 (it went from 2.9GB to 1.0GB) This should hopefully make setup and CI faster.

I ran `ncdu` to compare disk usage of node_modules:

| on master | after yarn-deduplicate |
| --- | --- |
| master | after dedupe |


gorandalum (approved):
Great! 🎉

package.json
```diff
@@ -118,11 +118,12 @@
     "uuid": "^10.0.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.22.10",
-    "@babel/plugin-transform-template-literals": "^7.22.5",
-    "@babel/preset-env": "^7.22.10",
-    "@babel/preset-typescript": "^7.22.5",
-    "@babel/runtime": "^7.22.10",
+    "@babel/core": "^7.26.0",
```
reidzeibel, line 121: any reason why we use the `^` on the versioning? From what I understand, that will cause the app to use the latest version of the same major version (in this case >=7.0.0 and 8.0.0<) 🤔

rosvik: I'm not sure there's a reason other than it being default when installing packages with yarn. I guess it can cause issues locally if something breaks between minor versions, but it shouldn't be an issue release, since we use `yarn ci` in github actions, which takes versions from yarn.lock.

PR #4806, opened 2024-11-11
fix: Wrong semicolon breaks text in accessibility for situation messages

jorelosorio:

Fix for https://github.com/AtB-AS/mittatb-app/pull/4802


rosvik (approved):
🎉

PR #4807, opened 2024-11-11
feat: Add stop signal button

gorandalum:
First iteration of the long awaited blue button! It is not
connected to an actual backend endpoint yet, but mocks an api
request by just adding a 3000ms delay.

Some things are not covered by this app and will be done later:
- Actual backend integration
- Config in firestore like applicable transport modes and minutes
  before availability.
- Animation on background color.
- Design more inline with Figma when the button component is up to
  date with what is in Figma.
- Saving sent stop place signals in memory so the state is stored when
  navigating back into the travel aid screen. 

Screenshots:
<div>

</div>
<div>

</div>


src/travel-aid/components/StopSignalButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const shouldShowStopButton = (
+  serviceJourney: ServiceJourneyWithGuaranteedCalls,
+  call: EstimatedCallWithQuayFragment,
+) => {
+  return (
+    !call.actualDepartureTime &&
+    serviceJourney.transportMode &&
+    ['bus', 'tram'].includes(serviceJourney.transportMode) &&
```
jorelosorio, line 83: Is this going to be on firebase? and the AUTHORITY?

gorandalum: The modes are going to be in Firestore. The authority is not going to in Firestore, it is in the env file.

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -370,7 +389,15 @@ const getStopHeader = (
...
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
-  container: {
+  container_notSentSignal: {
+    flex: 1,
+    backgroundColor: theme.color.background.neutral['1'].background,
+  },
+  container_afterSentSignal: {
+    flex: 1,
+    backgroundColor: theme.color.background.accent['3'].background,
```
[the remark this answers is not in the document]
gorandalum: That would be a design-system decision for the designers to make. As of now the different background/accent colors doesn't have any semantic structure like 'primary', 'secondary', 'tertiary'.

src/travel-aid/components/StopSignalButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+    AUTHORITY === serviceJourney.line.authority?.id
+  );
+};
+
+const isStopButtonEnabled = (call: EstimatedCallWithQuayFragment) => {
+  if (!call.realtime) return false;
+
+  const activationWindowStart = addMinutes(call.expectedArrivalTime, -60);
+  const activationWindowEnd = addMinutes(call.expectedArrivalTime, -2);
```
marius-at-atb, line 92: I'm sure this has been discussed, but 2 minutes before sounds quite strict to me 🤔 
Is this something that should be remotely configurable?

Also do we have a max refresh frequency for `expectedArrivalTime`? If not, it is currently possible that the enabled state will jump back and forth rapidly.

src/translations/screens/subscreens/TravelAid.ts
```diff
@@ -15,6 +15,15 @@ export const TravelAidTexts = {
       `Scheduled at ${time}`,
       `Rutetid klokken. ${time}`,
     ),
+  stopButton: {
+    text: _('Send stopp', 'Send stop', 'Send stopp'),
+    textAfterSubmit: _('Stopp sendt', 'Stop sent', 'Stopp sendt'),
+    notAvailable: _(
+      'Ikke mulig å sende stopp signal nå',
+      'Not possible to send stop signal now',
+      'Ikke mogleg å sende stopp signal no',
```
marius-at-atb, line 24:
```suggestion
      'Ikkje mogleg å sende stopp-signal no',
```

src/translations/screens/subscreens/TravelAid.ts
```diff
@@ -15,6 +15,15 @@ export const TravelAidTexts = {
       `Scheduled at ${time}`,
       `Rutetid klokken. ${time}`,
     ),
+  stopButton: {
+    text: _('Send stopp', 'Send stop', 'Send stopp'),
+    textAfterSubmit: _('Stopp sendt', 'Stop sent', 'Stopp sendt'),
+    notAvailable: _(
+      'Ikke mulig å sende stopp signal nå',
```
marius-at-atb, line 22:
```suggestion
      'Ikke mulig å sende stopp-signal nå',
```

marius-at-atb (approved):
LGTM 🟦

jorelosorio:
Seems like, it is possible now to tap on stop sent multiple times, is that correct for this PR?

src/travel-aid/components/StopSignalButton.tsx
gorandalum: It will be configurable, as mentioned in the PR:
> - Config in firestore like applicable transport modes and minutes
before availability.

The service journey, including all estimated calls, is refetched every 10 second. And yeah, you are right, in theory it can jump back and forth from being disabled or not at this refetch. I don't think it will be such a big problem that it needs explicit handling, but might change my mind if we get some reports.

marius-at-atb: Might not be an issue then :+1:
If it comes to it, we could consider using `now` from `useNow` instead of `new Date()`, and return `isEnabled` as a state variable that is only updated in a useEffect with now as a dependency..

jorelosorio:
> > Seems like, it is possible now to tap on stop sent multiple times, is that correct for this PR?
> 
> Good input! I have dialog with design how to solve this. The "active" state is also used for buttons which are clickable, so we can't add the rule that every active button is not clickable. The Button component itself doesn't support setting `disabled={true}` without showing it as disabled state.
> 
> We (Probably me, @hannahATB and @AkselAN) will discuss this when she is back.

Ok, but do you want to handle that in this PR or another PR?

gorandalum:
> Ok, but do you want to handle that in this PR or another PR?

In another PR. That is part of this bullet point mentioned above: 
> - Design more inline with Figma when the button component is up to
date with what is in Figma.

But I did make a quick fix for this, a more hackish one where the `onPress` handler doesn't do anything if mutation status is already success.

PR #4808, opened 2024-11-12
Fix: Situations are not being announced properly

jorelosorio:

The conclusion was:

The component was re-rendered after updating the current announced situations/notifications internally in the hook, creating a loop that left the new IDs empty, because they were already added. To avoid a re-render while updating current situations/notifications we instead used useRef.

### DIFF
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -3,7 +3,7 @@ import {Button} from '@atb/components/button';
 import {EstimatedCallInfo} from '@atb/components/estimated-call';
 import {StyleSheet, useTheme} from '@atb/theme';
 import {ServiceJourneyDeparture} from '@atb/travel-details-screens/types';
-import React, {Ref, useEffect, useRef, useState} from 'react';
+import React, {Ref, useEffect, useRef} from 'react';
 import {SafeAreaView} from 'react-native-safe-area-context';
 import {useTravelAidDataQuery} from './use-travel-aid-data';
 import {ScrollView} from 'react-native-gesture-handler';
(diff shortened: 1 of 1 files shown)


gorandalum (commented):
Refs might be the proper choice, but note that there is a best practice that refs shouldn't be read or written during the render it self.

https://react.dev/learn/referencing-values-with-refs

> Don’t read or write ref.current during rendering. If some information is needed during rendering, use [state](https://react.dev/learn/state-a-components-memory) instead. Since React doesn’t know when ref.current changes, even reading it while rendering makes your component’s behavior difficult to predict. (The only exception to this is code like if (!ref.current) ref.current = new Thing() which only sets the ref once during the first render.)

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -233,57 +246,85 @@ const useTravelAidAnnouncements = (
...
-
-  const [currentAnnouncedSituationIds, setCurrentAnnouncedSituationIds] =
-    useState<string[]>(announcedSituationIds);
-  const [currentAnnouncedNoticeIds, setCurrentAnnouncedNoticeIds] =
-    useState<string[]>(announcedNoticeIds);
+  const [announcedSituationIds, setAnnouncedSituationIds] = useState<string[]>(
+    situationsForFocusedStop.map((s) => s.id).filter(onlyUniques),
+  );
+  const [announcedNoticeIds, setAnnouncedNoticeIds] = useState<string[]>(
+    noticesForFocusedStop.map((s) => s.id).filter(onlyUniques),
```
reidzeibel, line 253:
```suggestion
    noticesForFocusedStop.map((n) => n.id).filter(onlyUniques),
```

tormoseng:
Can confirm that the latest fix announce all messages when they appear on a stop.

jorelosorio:
This PR was moved into draft, there is a different approach in Android and iOS to handle these cases, so for now, this solution is on standby while the other solution is being tested.

PR #4809, opened 2024-11-12
feat: Make buttons in loading state not pressable

gorandalum:
### Background
Doesn't really make sense that you can click a button (even spam
it) while it is loading.

### Acceptance criteria
- [ ] Buttons that are loading should not be pressable. This means no effect when pressing it, and no animation when pressing it. Some buttons that can be tested for this are the buttons in the activate-now/consume-carnet bottom sheets.

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -15,7 +15,7 @@ import {
 import {ThemeIcon, ThemeIconProps} from '@atb/components/theme-icon';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
 import {shadows} from '@atb/components/map';
-import { ContrastColor, InteractiveColor } from '@atb/theme/colors';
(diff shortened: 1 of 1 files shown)


jorelosorio (approved):

ckbilstad (commented):
Agree, LGTM

hanne-at-atb:
When testing this I am not able to press the button while it is in loading state (as expected), but the animation for the spinner keeps spinning and the button is not disabled (at least color stays the same). This goes for both activating carnet button and start now button. I'm a bit unsure if this is as expected from the acceptance criteria @gorandalum ?

gorandalum:
I'm not 100% sure what you mean, maybe add a video? A pressed button should be in loading state until the action is done (or failed). The only change that is ment to come from this PR is that the button is not pressable while loading.

hanne-at-atb:
Ok if the only change is that it is not pressable while loading, then everything works as expected! All is good then ✅

PR #4810, opened 2024-11-12
feat: various travel aid tweaks

rosvik:
## Changes / Acceptance criteria

- Announces the time only if the focused stop hasn't changed since the last announcement
- Enables the travel aid button even when realtime is false
- Updates the no realtime message box wording and type
- Only shows scheduled time on `NotYetArrived`
- Refresh interval is 5 seconds

### DIFF
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -234,6 +234,7 @@ const useTravelAidAnnouncements = (
 ) => {
   const {language, t} = useTranslation();
   const isFirstRender = useRef(true);
+  const previousQuayId = useRef(state.focusedEstimatedCall.quay.id);
 
   const announcedSituationIds = situationsForFocusedStop
     .map((s) => s.id)
@@ -261,6 +262,7 @@ const useTravelAidAnnouncements = (
     getSituationA11yLabel(newSituations, language) +
     screenReaderPause +
     getNoticesA11yLabel(newNotices);
+  const timeInfoMessage = getTimeInfoA11yLabel(state, t, language);
 
   if (newSituations.length > 0) {
     setCurrentAnnouncedSituationIds((prev) => [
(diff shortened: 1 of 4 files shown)


src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -234,6 +234,7 @@ const useTravelAidAnnouncements = (
 ) => {
   const {language, t} = useTranslation();
   const isFirstRender = useRef(true);
+  const previousQuayId = useRef(state.focusedEstimatedCall.quay.id);
```
gorandalum, line 237: Could the hook `useStablePreviousValue` be used for this? Also maybe the `isFirstRender` part can be removed? Since a previous quay id of `undefined` means that it is first render.

gorandalum (approved):

src/travel-aid/TravelAidScreenComponent.tsx
rosvik: I don't immediately see how `useStablePreviousValue` can be used here, since stablePreviousValue uses useState internally, which I don't think will work as intended in our useEffect.

The isFirstRender could indeed be removed 👍

gorandalum: Aha, I didn't look through it properly. Other `usePrevious` hooks that I have seen/used have used refs, and I thought this was a variation of that.

tormoseng:
✅  OK!

tormoseng:
@rosvik I see now on Android (Android 12) that after only the new changed time is announced, it announces everything again. I don't see this on iOS. See the screen recording below.

<details>

https://github.com/user-attachments/assets/eef9a602-0b83-4053-9110-368f5457a118

</details>

tormoseng:
✅  The comment above is re-tested OK in https://github.com/AtB-AS/kundevendt/issues/19525

PR #4812, opened 2024-11-13
feat: Add sent stop signals cache

gorandalum:
A simple in-memory cache that is cleared at app restart.

### DIFF
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -63,9 +63,9 @@ export const TravelAidScreenComponent = ({
   serviceJourneyDeparture,
   goBack,
 }: Props) => {
-  const stopSignalMutation = useStopSignalMutation(() =>
-    sentStopSignalsCache.addSent(serviceJourneyDeparture),
-  );
+  const stopSignalMutation = useStopSignalMutation({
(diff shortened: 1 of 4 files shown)


src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -48,23 +48,36 @@ import {useStopSignalMutation} from '@atb/travel-aid/use-stop-signal-mutation';
...
   const styles = useStyles();
   const {t} = useTranslation();
   const {theme, themeName} = useTheme();
   const focusRef = useFocusOnLoad();
 
+  const hasSentStopSignal = sentStopSignalsCache.hasSent(
+    serviceJourneyDeparture,
+  );
+  const sendStopSignalStatus = hasSentStopSignal
+    ? 'success'
```
jorelosorio, line 78: Does `stopSignalMutation.status` not contain `success` as well?

jorelosorio (approved):
Looks good. 

If there was not rules like expiration time for the cache and things like that maybe a Map/Hash/Keypair of a int and boolean was enough?

src/travel-aid/TravelAidScreenComponent.tsx
gorandalum: Yes it does, but if the sent stop signal is already saved, then we override/force the status value to be 'success'.

PR #4814, opened 2024-11-14
feat: add journey aid posthog events

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/19521

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_TravelAidScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_TravelAidScreen.tsx
@@ -15,6 +15,8 @@ import {usePreferences} from '@atb/preferences';
 import {useRemoteConfig} from '@atb/RemoteConfigContext';
 import Bugsnag from '@bugsnag/react-native';
 import {useFirestoreConfiguration} from '@atb/configuration';
+import {useAnalytics} from '@atb/analytics';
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_TravelAidScreen.tsx
```diff
@@ -52,9 +56,17 @@ export const Profile_TravelAidScreen = () => {
...
-            onValueChange={(checked) =>
-              setPreference({journeyAidEnabled: checked})
-            }
+            onValueChange={(checked) => {
+              analytics.logEvent(
+                'Journey aid',
+                'Journey aid preference toggled',
+                {
+                  enabled: checked,
+                  isScreenReaderEnabled,
```
tormoseng, line 65: You don't need to add `screenreader: isScreenReaderEnabled` to have a parameter name on the event?

rosvik: This is shorthand for `isScreenReaderEnabled: isScreenReaderEnabled` in javascript. But I did want to consistently use `screenReaderEnabled` here, so thanks for making me notice 😄

rosvik: So from now, it should be `screenReaderEnabled: true/false` in posthog

tormoseng (approved):
LGTM

PR #4818, opened 2024-11-15
chore: upgrade storybook to v8.4

rosvik:
This upgrades storybook from 6.5 to 8.4, using the migration guide at https://github.com/storybookjs/react-native/blob/next/MIGRATION.md

There are some issues on master with the recent design system upgrades. Those are not fixed by this PR, so this is mainly to make sure the builds and app work ok after the upgrade, then the storybook functionality can be tested after the design system changes are fixed.

closes https://github.com/AtB-AS/kundevendt/issues/19400


reidzeibel (approved):
Great!

PR #4821, opened 2024-11-15
fix: enable travel aid toggle regardless of stop button config

rosvik:
Issue found by @tormoseng in the release candidate


jorelosorio (approved):

hanne-at-atb:
From what I can see when testing in Staging the toggle should work and be independent of the RemConf variables now. Tested on android ✅

tormoseng:
Verified the same on iOS ✅

PR #4824, opened 2024-11-18
fix: Skip error logging for not available offer search

gorandalum:
Not available response from offer search is a normal flow and not
really an error situation that needs to be logged.

### Acceptance criteria
- [ ] No error in Bugsnag regarding 400 NotAvailable from _ticket/v3/search/authority_.


rosvik (approved):

tormoseng:
✅

PR #4825, opened 2024-11-18
refactor: Better typing of places in purchase selection type

gorandalum:
The type for selecting a place in the purchase selection was not
optimal. This was observed already when creating the
PurchaseSelectionType, but not done anything with to not embrace
too much.

The place selection was typed to either a zone or a StopPlace, and
was mandatory. That ment that every time a stop place was not
selected, then a zone was selected. So for example during a boat
ticket purchase process, if the selected from harbor was undefined,
then in the data structure the from place was actually a tariff
zone.

This was improved by creating two different fields for zones and
stop places, and they can both be undefined if there is no
zoneSelectionMode on the type config.

### Acceptance criteria
- [ ] Purchase flow should work as before. Especially test zone selection variations (map, list, location search) and the harbor selection flow.


reidzeibel (approved):
Looks much better! 🚀

tormoseng:
✅

PR #4826, opened 2024-11-18
feat: use live regions to announce travel aid updates

rosvik:
Refactors travel aid to use [live regions](https://reactnative.dev/docs/accessibility#accessibilityliveregion-android) to announce changes on the page. Since iOS doesn't support live regions, the `useLiveRegionAnnouncement` hook is added, that triggers screen reader announcements on iOS when a11yLabel changes. There's also a wrapper component `LiveRegionWrapper` that can be used for non-interactive components in JSX.

### Example usage of live regions

There are two main ways to implement live regions

```tsx
const a11yProps = useLiveRegionAnnouncement(a11yLabelThatChangesOverTime, true);
return (
	<InteractiveComponentWithAnnoucements
		accessibilityRole="button"
		onPress={() => {}}
		{...a11yProps}
	/>
)
```

or

```tsx
return (
	<LiveRegionWrapper accessibilityLabel={a11yLabelThatChangesOverTime}>
		<NonInteractiveComponent />
	</LiveRegionWrapper>
)
```

closes https://github.com/AtB-AS/kundevendt/issues/19525

### DIFF
--- a/src/components/screen-reader-announcement/use-live-region-announcement.ts
+++ b/src/components/screen-reader-announcement/use-live-region-announcement.ts
@@ -1,0 +1,31 @@
+import {useEffect} from 'react';
+import {AccessibilityInfo, AccessibilityProps, Platform} from 'react-native';
+
+/**
+ * When a11yLabel changes, announce it to screen readers without interrupting
+ * ongoing speach. Returns accessibility props that should be added to the
+ * relevant component. For non-interactive components, consider using
+ * `LiveRegionWrapper` instead.
+ *
+ * Docs: https://reactnative.dev/docs/accessibility#accessibilityliveregion-android
+ */
+export const useLiveRegionAnnouncement = (
+  a11yLabel?: string,
+  enabled?: boolean,
+): AccessibilityProps => {
+  // Since iOS does not support accessibilityLiveRegion, we trigger screen
+  // reader announcements when a11yLabel changes for a equivalent effect.
+  useEffect(() => {
+    if (enabled && a11yLabel && Platform.OS === 'ios') {
+      AccessibilityInfo.announceForAccessibilityWithOptions(a11yLabel, {
+        queue: true,
+      });
+    }
+  }, [a11yLabel, enabled]);
+
+  return {
+    accessible: true,
+    accessibilityLabel: a11yLabel,
+    accessibilityLiveRegion: enabled ? 'polite' : undefined,
+  };
+};
--- a/src/travel-aid/TravelAidScreenComponent.tsx
+++ b/src/travel-aid/TravelAidScreenComponent.tsx
@@ -3,21 +3,15 @@ import {Button} from '@atb/components/button';
 import {EstimatedCallInfo} from '@atb/components/estimated-call';
 import {StyleSheet, useTheme} from '@atb/theme';
 import {ServiceJourneyDeparture} from '@atb/travel-details-screens/types';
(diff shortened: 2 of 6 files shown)


src/components/screen-reader-announcement/use-live-region-announcement.ts
```diff
@@ -0,0 +1,37 @@
...
+  accessible: true;
+  accessibilityLabel?: string;
+  accessibilityLiveRegion?: 'polite';
+};
+
+/**
+ * When a11yLabel changes, announce it to screen readers without interrupting
+ * ongoing speach. Returns accessibility props that should be added to the
+ * relevant component. For non-interactive components, consider using
+ * `AnnouncementWrapper` instead.
```
gorandalum, line 14:
```suggestion
 * `LiveRegionWrapper` instead.
```

src/components/screen-reader-announcement/use-live-region-announcement.ts
```diff
@@ -0,0 +1,37 @@
+import {useEffect} from 'react';
+import {AccessibilityInfo, Platform} from 'react-native';
+
+type LiveRegionAccessibilityProps = {
+  accessible: true;
+  accessibilityLabel?: string;
+  accessibilityLiveRegion?: 'polite';
+};
```
gorandalum, line 8: Maybe deduce this type from `AccessibilityProps`?

src/travel-aid/TravelAidScreenComponent.tsx
```diff
@@ -228,28 +205,36 @@ const TravelAidSection = ({
...
-            accessibilityLabel={focussedStateA11yLabel}
-            style={styles.stopHeaderContainer}
-          >
-            <View>
+          <View accessible style={styles.stopHeaderContainer}>
+            <LiveRegionWrapper
+              accessibilityLabel={
+                getStopHeader(status, t) + screenReaderPause + quayName
+              }
+            >
```
gorandalum, line 232: Should there be two layers of `accessible` here?

gorandalum (approved):
This looks great!

jorelosorio (approved):
We had a talk while exploring this option, but I haven't tested with the cases I covered in my solution to handle multiple messages and priorities. The code looks good based in our discussions.

src/components/screen-reader-announcement/use-live-region-announcement.ts
```diff
@@ -0,0 +1,37 @@
...
+      AccessibilityInfo.announceForAccessibilityWithOptions(a11yLabel, {
+        queue: true,
+      });
+    }
+  }, [a11yLabel, enabled]);
+
+  return {
+    accessible: true,
+    accessibilityLabel: a11yLabel,
+    accessibilityLiveRegion: enabled ? 'polite' : undefined,
```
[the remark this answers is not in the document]
rosvik: I don't think keeping the android specific props is an issue for iOS here. I guess it wouldn't hurt to omit it either, but it's extra code with no effect.

src/travel-aid/TravelAidScreenComponent.tsx
rosvik: The topmost `accessible` here is to better match the sketches, where the time and stop place is one focus area. It is, however, handled differently on iOS and Android. Talkback makes the two children focusable, while VoiceOver focuses the parent. Both platforms have the same announcements. I've clarified with @hannahATB that it's ok to have it this way.

PR #4829, opened 2024-11-19
feat: make delay in LoadingScreenBoundary configurable

rosvik:
Requesting some feedback on a change I want to make, since it's been bugging me for a while 😄

The app currently has a 1s timeout on every cold start, _after_ the initial data is loaded (Firestore config, auth, etc.). This was done as an additional safety measure to prevent the "app freeze" a while back. At that point, the initial render of the app could be very demanding, since several contexts updated data globally, and caused several renders of the entire layout tree. It was probably due to some `setInterval` calls with a never ending update queue that caused the freeze to never stop, which was fixed by the addition of the `useInterval` hook.

But, as far as I know, we have no reason to believe that the 1s delay is necessary today, since no users with slow phones are complaining about the app being slow on boot.

The effect of this is (in my opinion) that the app feels heavy on launch, even on modern phones that usually have <100ms start times for apps. 1s boot is long enough that it feels like something is wrong / not loading correctly. If we reduce it to 200ms, we _might_ see some slowness on boot on the first render for very slow phones, but in return, we won't make app launch slow for everyone.

Maybe now isn't the right time to change it, and maybe 200ms isn't the correct value (maybe it could be less!), but it would be nice to discuss if there's any arguments to keep or change the current delay.

### Acceptance criteria

- [ ] The app loads much faster 🚀 
- [ ] The `loading_screen_delay_ms` remote config flags configures how long the loading screen (gray screen with spinner), stays after inital data is loaded.
- [ ] No app freeze on slow android phones 🥶

### DIFF
--- a/src/loading-screen/LoadingScreenBoundary.tsx
+++ b/src/loading-screen/LoadingScreenBoundary.tsx
@@ -18,8 +18,8 @@ export const LoadingScreenBoundary = ({
   const {status, retry, paramsRef} = useLoadingState(LOADING_TIMEOUT_MS);
   useNotifyBugsnagOnTimeoutStatus(status, paramsRef);
 
-  // Wait one second after load success to let the app "settle".
-  const waitFinished = useDelayGate(1000, status === 'success');
+  // Wait 200ms after load success to let the app "settle".
+  const waitFinished = useDelayGate(200, status === 'success');
 
   if (!isLoadingScreenEnabled) return children;
 
--- a/src/loading-screen/LoadingScreenBoundary.tsx
+++ b/src/loading-screen/LoadingScreenBoundary.tsx
@@ -5,6 +5,7 @@ import {useDelayGate} from '@atb/utils/use-delay-gate';
 import {useLoadingState} from '@atb/loading-screen/use-loading-state';
 import {useNotifyBugsnagOnTimeoutStatus} from '@atb/loading-screen/use-notify-bugsnag-on-timeout-status';
 import {useFeatureToggles} from '@atb/feature-toggles';
+import {useRemoteConfig} from '@atb/RemoteConfigContext';
 
 const LOADING_TIMEOUT_MS = 10000;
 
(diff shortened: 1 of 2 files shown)


marius-at-atb (approved):
Let's go!

rosvik:
Side note: I think it was @gorandalum that suggested a new storage model for the app, where all state from local storage was loaded at app launch instead of every time we needed to read from it. I think that could tie nicely into the loading boundary, by making sure we don't render before all local storage is fetched.

marius-at-atb:
If we are going to introduce a new storage model for the app, I think Zustand should be evaluated as part of it.
Its popularity is on the rise for good reason, and I think it can be valuable for us in the future.
It has built-in support for loading, persisting and migrating locally stored data:
https://github.com/pmndrs/zustand/blob/main/docs/integrations/persisting-store-data.md#migrate

rosvik:
@gorandalum I think that's a good idea! 

Do you think there are any advantages to using remote config vs firestore here? Maybe remote config is more reliable on app launch because of the aggressive caching?

gorandalum:
I think both are cached pretty aggressively.

Regarding this parameter I think it can go either way. Do we envision that we for example would like to:
- Change or remove it for x % of users?
- Try out different values for Android/iOS?
- Have different values for different app versions?

If we don't need this type of configurability the correct place is Firestore.

marius-at-atb:
Regarding evaluation of Zustand, here is a [new issue](https://github.com/AtB-AS/kundevendt/issues/19568) for it.
I reckon there are many more advantages than listed in the issue as well!

rosvik:
> Try out different values for Android/iOS?

I imagine this could be relevant. We never had any freeze issues on iOS, and performance seems to be pretty good with what the loading boundary is already doing, so I imagine zero delay would be ok here.

Looking at it the other way around, I don't think the slow remote config update times (~12h TTL) will be an issue. So remote config it is!

gorandalum (approved):

marius-at-atb (approved):

tormoseng:
✅  The issue itself is approved, while we can agree on values afterwards. Tested on:
- Android 12
- Android 7 (Samsung A5 from 2016)
- iOS 17
- The oldest available Samsung (S- and A-series) from [Samsung Remote Test Lab](https://developer.samsung.com/remote-test-lab)

Will also be tested on different phones and OS versions living in Staging until next release.

rosvik:
@tormoseng Correct, this is a safeguard against render issues 👍 I think that could be a good idea! And maybe keep it at 0 in staging to see if it provokes any problems.

gorandalum:
I think it might be good to leave some room (200ms is a good start) for the app to load as much as it can during loading. Some things (like RemoteConfig, FeatureTogglesContext state) aren't directly waited upon during the loading boundary, but we rely on the loading taking as much time that they are actually loaded when the user enters the app.

tormoseng:
Very well. The value is set to `0 ms` in Staging, while I've created the param in Prod with a value of `200 ms` (for all)

marius-at-atb:
@gorandalum Now that the delay is reduced, maybe even to 0ms, I think we should [do this](https://github.com/AtB-AS/mittatb-app/pull/4773#discussion_r1810690794) after all.

gorandalum:
It is at least more arguments to do it now! Hippety hap (my personal translation of "hipp som happ"!) for me as the debug flags in local storage are not relevant for our users, but you can create an issue and put it up for prioritization.

marius-at-atb:
https://github.com/AtB-AS/kundevendt/issues/19585

PR #4833, opened 2024-11-21
fix: Allow update certs to rebundle the app with cache on

jorelosorio:
It allows the certs always to be updated, this avoids always forcing a cache to update the certs, which means those are going to be up to date.

### DIFF
--- a/.github/actions/update-devices/action.yml
+++ b/.github/actions/update-devices/action.yml
@@ -20,6 +20,10 @@ inputs:
     description: "The ruby version to use"
     default: "2.7.6"
     required: false
+  keychain-name:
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):

jorelosorio:
NOTE:

When looking at the fastlane source code https://github.com/fastlane/fastlane/blob/ec78d7d81a6a66e9e89fd29f6e0616d5ba09840a/fastlane_core/lib/fastlane_core/cert_checker.rb#L55 this message is a generic message when no certs are found, by nature the WWDR is not part of the sign identities, therefore, no listed and the message is printed. I removed the install WWDR in our action because seems to be covered by Fastlane already.

PR #4834, opened 2024-11-21
feat: Remove use of NFC

jorelosorio:
After the message we got from Apple rejecting the app because use of NFC, this PR removes the usage of it on the code and creates a patch towards Entur MobileTokenSDK.


reidzeibel (approved):
we can try this one, and if it still fails, we might need to update the podfile to match the one mentioned on Slack channel https://mittatb.slack.com/archives/C03DTC05VRQ/p1732195272406539?thread_ts=1732182724.707639&cid=C03DTC05VRQ

gorandalum (approved):

tormoseng:
✅  App seems to work as before.

PR #4838, opened 2024-11-22
fix: make journey aid skip cancelled calls

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/19512

### DIFF
--- a/src/translations/screens/subscreens/TravelAid.ts
+++ b/src/translations/screens/subscreens/TravelAid.ts
@@ -28,11 +28,18 @@ export const TravelAidTexts = {
       'Ikke mogleg å sende stopp signal no',
     ),
   },
-  error: {
+  apiError: {
     message: _(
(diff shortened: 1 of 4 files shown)


gorandalum (commented):
A couple questions:
- This will sometimes end up with an `estimatedCalls` list that doesn't include the `fromQuayId`. Can that be a problem?
- If a full service journey is cancelled, that is represented by all stops being cancelled. That will probably look a little strange?

Maybe both of these can be solved by not allowing going to travel aid screen if the call is cancelled?

rosvik:
@gorandalum Good points! I added these changes now:

1. If the current stop is cancelled or not found, default to `estimatedCalls[0]`. (There was a bug that prevented this from happening from before)
2. If there are no estimated calls (because the entire trip is cancelled) show the error "This journey has no upcoming departures."
3. Disable travel aid button from departure details if the call is cancelled.

gorandalum (approved):

PR #4839, opened 2024-11-22
fix: Wrong tests sometimes

jorelosorio:
My theory is that something was messed up with the current date, so as a fix I set a specific time to test, some tests are async so it might be the dates for the tests were not synch at all so then it failed.

Acceptance criteria:

- [ ] Order for tickets must be working as expected


gorandalum (changes requested):
Hint: Try changing the order of the input parameters of the second unit test and see if it still passes.

jorelosorio:
> I don't quite understand how this will fix the problem 🤔 Does the reservation and fare contract in the flaky test need to have the exact same date? Or would something like
> 
> ```ts
> mockupFareContract('1', 'valid', -1),
> mockupReservation('3', 'reserving', 0),
> mockupFareContract('2', 'valid', 1),
> ```
> 
> also fix it? (that is, using -1, 0, 1 instead of -1, 0, 0)
> 
> Maybe there's a different way to test "Reservation should be first if reservation is being processing"?

Time is fixed but not the same, it was more to solve the two with the same time order, I just avoided reserving, valid and current time at the same time, but I understand that changing it to 1 also avoids that. I did the fix.

gorandalum (changes requested):
The sorting function should be deterministic and give the same sorting order no matter the order of the input parameters. Since changing the order of the input parameters affected the result, there is something that should be fixed in the sort function itself.

src/fare-contracts/__tests__/sort-fc-reservation-by-validity-creation.test.ts
```diff
@@ -112,7 +112,7 @@ describe('Sort by Validity', () => {
   it('Reservation should be first if reservation is being processing', async () => {
     const fcOrReservations: (FareContract | Reservation)[] = [
       mockupFareContract('1', 'valid', -1),
-      mockupReservation('3', 'reserving', 0),
+      mockupReservation('3', 'reserving', 1),
```
gorandalum, line 115: It seems to me that it was kinda the point that the reservation and farecontract had the same timestamp, to test that the reserving reservation should be sorted first.

jorelosorio:
> The sorting function should be deterministic and give the same sorting order no matter the order of the input parameters. Since changing the order of the input parameters affected the result, there is something that should be fixed in the sort function itself.

The 0, -1, 1 is how much time is taken out from the current time, the order is also taken by the priority of the `reserving` and others, I do not think it is the order function which is wrong, it was tested before.

gorandalum:
> > The sorting function should be deterministic and give the same sorting order no matter the order of the input parameters. Since changing the order of the input parameters affected the result, there is something that should be fixed in the sort function itself.
> 
> The 0, -1, 1 is how much time is taken out from the current time, the order is also taken by the priority of the `reserving` and others, I do not think it is the order function which is wrong, it was tested before.

But we can agree that it seems strange that it gives different sorting order based on the order of the input parameters? It would be more understandable if there were two fare contracts with identical creation date or two reservations with identical creation date, but that wasn't the case here.

rosvik:
From what I understand, reservations are supposed to be sorted first regardless of their date compared to fare contracts: https://github.com/AtB-AS/mittatb-app/blob/d026702fc4be0487f225c233946acbb98423d503/src/fare-contracts/utils.ts#L100-L101

Based on this, I would expect that something like

```ts
mockupFareContract('1', 'valid', -3),
mockupReservation('3', 'reserving', -2),
mockupFareContract('2', 'valid', -1),

// ...

expect(ids).toEqual(['3', '2', '1']);
```

would be a good way to test it. The reservation is older than one of the fare contracts, but should appear first. If that's not the case, I guess we have a bug?

jorelosorio:
After a more deep analysis:

The reason why it was failing is that `reserving` was returned as 1 as well as `valid` so in the logic then it should order it by date which in that case `reserving` was created slightly first. That is why it comes first in the test, however, if for some reason the mockup creates first the `valid` instead of `reserving` then we fall into the wrong sort.

I made a fix for that in the PR.

So now in the `sort` if both are equal means that need to order them by `date`, this is if there are two `reserving` or `valid`.
The sort function is now interpreted as "place a before b".

This combination of two things makes it work as expected.

tormoseng:
✅  Nice

PR #4840, opened 2024-11-22
feat: Not always announce scheduled time in travel aid

gorandalum:
Requested by PO that the scheduled time shouldn't be read aloud
because of changes in the relative real time.

### Acceptance criteria
- [ ] On changes to the expected arrival time, only the expected arrival time is announced and not the scheduled time.


rosvik (approved):
Just a note: I guess this also means that the realtime and scheduled time will be two different screen reader focus areas, which is different from the sketches. @hannahATB

tormoseng:
✅

PR #4842, opened 2024-11-25
fix: Only enable get active booking when feature flag true

marius-at-atb:
The call to get active shmo booking shouldn't be sent unless the feature flag is enabled.


jorelosorio (approved):

tormoseng:
✅

PR #4843, opened 2024-11-25
chore: Use PreassignedFareProduct id type directly

marius-at-atb:
I noticed that `export type PreassignedFareProductId = string;` was used instead of getting the type from source.
This is fixed in this PR.

### DIFF
--- a/src/configuration/types.ts
+++ b/src/configuration/types.ts
@@ -3,8 +3,6 @@ export * from '@atb-as/config-specs/lib/mobility-operators';
 
 export type FirestoreConfigStatus = 'loading' | 'success';
 
-export type PreassignedFareProductId = string;
-
 export type PointToPointValidity = {
   fromPlace: string;
   toPlace: string;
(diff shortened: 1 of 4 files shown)


src/configuration/types.ts
```diff
@@ -3,8 +3,6 @@ export * from '@atb-as/config-specs/lib/mobility-operators';
 
 export type FirestoreConfigStatus = 'loading' | 'success';
 
-export type PreassignedFareProductId = string;
```
marius-at-atb, line 6: Does anyone know why we have `export * from '@atb-as/config-specs';` in this file?
Seems to me that we could import from `@atb-as/config-specs` directly in the other files 🤔

gorandalum: It is a way of doing encapsulation, making the rest of the app consumers of an internal module instead of directly exposing a third party package around the app. If it really makes sense for config-specs I'm a little unsure, but I think it is a good general pattern for third party dependencies.

gorandalum (approved):

tormoseng:
✅

PR #4846, opened 2024-11-26
feat: remove favorite departures button from profile dev menu

rosvik:
I don't think there's any point of having this button in the developer menu. The same can be done by everyone on the dashboard.


gorandalum (commented):
Could you add some designers as reviewer as well? Not to review the code, but to get traceability that they approve of the change.

Also is there some more screens that can be removed if we remove this functionality? I see reference to `Profile_FavoriteDeparturesScreen` in this PR, and that screen probably drills further down into other screens.

Note that this is a good example of something it would be nice to have analytics on. I also assume users add favorites through the dashboard, but statics are better than assumptions.

rosvik:
@gorandalum This only removes the button from the developer menu (visible with `debug: true`). If it can be removed from my profile entirely is another issue (and I don't think it should).

image

tormoseng:
✅

PR #4847, opened 2024-11-26
fix: Purchase builder products input from API

gorandalum:
The fare products in the PurchaseSelectionBuilderInput needs to
come from the products API for HjemJobbHjem to work.

Added staleTime to not make the app query products API everytime
it tries to open the purchase flow. Also set initialDataUpdateAt: 0
so the products directly from Firestore is considered stale at
query mount.


reidzeibel (approved):

tormoseng:
✅  Tested OK!

PR #4850, opened 2024-11-27
feat: add onCloseFocusRef param to bottom sheet open()

rosvik:
Makes `onCloseFocusRef` a required parameter to every bottom sheet `open()` call. 

Previously the close ref was stored in the bottom sheet state globally, limiting the target element to just one per page. If there were several elements that could have focus after close (e.g. there were two situation messages, both with an `onClick` triggering a bottom sheet), there were no way for the bottom sheet to know which one to focus on close.

With the changes in this PR, the responsibility of providing a onCloseFocusRef is moved to the caller of `open()`, making every opened bottom sheet store the ref that should be focused on the next `close()`. The ref is still stored globally, so there might still be issues if at some point in the future several bottom sheets are opened at once. See the changes to `BottomSheetContext.tsx`.

The parameter is set as required, so that it's hard to forget setting a closeRef when adding the bottom sheet new places in the app.

closes https://github.com/AtB-AS/kundevendt/issues/16840

### DIFF
--- a/src/chat/use-chat-icon.tsx
+++ b/src/chat/use-chat-icon.tsx
@@ -1,7 +1,7 @@
 import {IconButtonProps} from '@atb/components/screen-header';
 import {ThemeIcon} from '@atb/components/theme-icon';
 import {StyleSheet, useTheme} from '@atb/theme';
-import React from 'react';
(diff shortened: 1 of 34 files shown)


gorandalum (approved):
Nice!

PR #4851, opened 2024-11-28
feat: Use PurchaseSelectionType in Traveller components

gorandalum:
- Instead of having multiple params (product, type config, user
profiles) which are sent to the traveller components, we only send
the purchase selection as the only param representing what is
currently selected.
- Ensure that PurchaseSelectionType only includes selected user
profiles, and not user profiles with count 0.
- Removed the unused 'UPDATE_SELECTABLE' action in the
useUserCountState.
- Moved OnBehalfOfToggle component out of Single/Multiple traveller
selection component, and added it to the parent. This way it didn't
need to be duplicated two places.

### Acceptance criteria
- [ ] Ticket purchase should work as before. Especially check traveller stuff.
- [ ] On-behalf-of should be presented on same places as before, and work as before


tormoseng:
✅  Looks good!

PR #4852, opened 2024-11-28
refactor: replace bottomSheet useMemo to fix exhaustive deps

rosvik:
Looking for feedback in case I'm missing something, but I don't think the useMemo had much of an effect here. After replacing the `const bottomSheet = useMemo()` with a `BottomSheetOnBackDrop` component, it still seems performant, and without excessive re-renders.

closes https://github.com/AtB-AS/kundevendt/issues/16267

### DIFF
--- a/src/components/bottom-sheet/BottomSheetContext.tsx
+++ b/src/components/bottom-sheet/BottomSheetContext.tsx
@@ -9,7 +9,6 @@ import React, {
   useRef,
   useState,
 } from 'react';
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Seems ok for me! But I would test this extensively on both iOS/Android, both with and without backdrop, and especially in the map. And that network requests in the bottom sheet isn't run more often because of more `contentFunction` triggers.

PR #4860, opened 2024-12-03
feature: Departure details header

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/17880

Dependencies

- [ ] https://github.com/AtB-AS/mittatb-app/pull/4867

|Before|After|
|-|-|
|Simulator Screenshot - iPhone 16 Pro - 2024-12-06 at 09 46 38|Simulator Screenshot - iPhone 16 Pro - 2024-12-06 at 09 46 12|
|After after||
|Simulator Screenshot - iPhone 16 Pro - 2024-12-09 at 10 39 56||


src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -177,21 +192,21 @@ export const DepartureDetailsScreenComponent = ({
...
               <ThemeText
-                type="heading--medium"
+                type="heading__title"
                 color={themeColor}
-                style={{flexShrink: 1}}
+                style={{flex: 1}}
               >
                 {title ?? t(DepartureDetailsTexts.header.notFound)}
               </ThemeText>
+              {fromQuay && <DepartureTime departure={fromQuay} />}
```
jorelosorio, line 209: @rosvik or @gorandalum how do I get a EstimatedCall from EstimatedCallWithMetadata? fromQuay is EstimatedCallWithMetadata

jorelosorio: Yes, I did that but the function I want to use is getLineAndTimeA11yLabel and internally it access objects that are not part of EstimatedCallWithQuayFragment

jorelosorio: It also subcalls another functions with other properties and I ended up adding | EstimatedCallWithQuayFragment to 5 or more functions but there was one that missed data.

jorelosorio: This conversation was discussed internally at the office. We decided to use a different approach since EstimatedCall and EstimatedCallWithMetadata are two other models.

ios/Podfile.lock
```diff
@@ -2851,7 +2851,7 @@ SPEC CHECKSUMS:
   SocketRocket: f32cd54efbe0f095c4d7594881e52619cfe80b17
   SwiftProtobuf: 7773c4e96a99d7b8ab7cda0fc30a883732ff93b1
   TokenCore: 3235fc5e40d5364f169618de8ba37929607618ff
-  TokenStateReactNativeLib: 7d0ff81bceaa26361ad7fc5e90cf161eb1a4b921
+  TokenStateReactNativeLib: cb1c6fc96795046bd9b69445804dbbc0010828c0
```
reidzeibel: are these intended?

gorandalum:
Could you add before/after picture? Either here or in the issue.

jorelosorio:
> Could you add before/after picture? Either here or in the issue.

Done ;)

gorandalum (changes requested):
Take a look at some text color in light mode:
Screenshot 2024-12-06 at 10 00 01

Also should there be some margin between line name and departure time?

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
```diff
@@ -211,47 +230,46 @@ export const DepartureDetailsScreenComponent = ({
...
-                        mode: mode,
-                        subMode: subMode,
-                      });
-                    }}
-                  />
-                ) : null}
+            {shouldShowMapButton ? (
+              <View style={styles.actionButtons}>
+                <Button
+                  type="small"
```
rosvik, line 236: We don't have the favorite button yet, but it looks like the map button should fill the width of its container, which I guess the `"small"` button doesn't currently support 🤔 Might make more sense to have width as it's own parameter to `Button` instead of being decided by `type`.

Maybe a problem for another issue/PR, but we should figure out if this solution is ok to release if it isn't solved before then.

jorelosorio:
> Take a look at some text color in light mode: Screenshot 2024-12-06 at 10 00 01
> 
> Also should there be some margin between line name and departure time?

I made a fix to that

gorandalum:
> I made a fix to that

Are you sure? No commit after my comment, and seems to still be there when running locally.

src/travel-details-screens/DepartureDetailsScreenComponent.tsx
jorelosorio: Yes, I had it before with a new attribute called "expand" but it seems like it can be messy for the design system and such, this is going to be in a different PR.

src/components/estimated-call/DepartureTime.tsx
```diff
@@ -1,22 +1,24 @@
...
   formatLocaleTime,
   formatToClockOrRelativeMinutes,
   secondsBetween,
 } from '@atb/utils/date';
 import {EstimatedCallWithMetadata} from '@atb/travel-details-screens/use-departure-data.ts';
+import {ContrastColor, TextColor} from '@atb/theme/colors.ts';
 
 type DepartureTimeProps = {
   departure: EstimatedCall | EstimatedCallWithMetadata;
+  color?: ContrastColor | Statuses | TextColor | ColorValue;
```
gorandalum, line 19: I think this is a little overkill. Can't it be just `ContrastColor`?

src/components/estimated-call/DepartureTime.tsx
```diff
@@ -37,7 +39,9 @@ export const DepartureTime = ({departure}: DepartureTimeProps) => {
               ? 'body__primary--strike'
               : 'body__primary--bold'
           }
-          color={departure.cancellation ? 'secondary' : 'primary'}
+          color={
+            color ? color : departure.cancellation ? 'secondary' : 'primary'
+          }
```
gorandalum, line 44: This means that if a color is sent in, then there is no difference in text for `cancellation` anymore? 🤔

jorelosorio: The colour is used in the header which is the only one sending it in, the header as is now remains grey no matter the theme, the colour passed in is the one that contrasts, but it applies to both texts including cancellation, during the test the grey colour did not fit well (cancellations) in light mode.

image

jorelosorio: A PR to fix this was made on https://github.com/AtB-AS/mittatb-app/pull/4867

src/components/estimated-call/DepartureTime.tsx
jorelosorio: Covered in the other PR we did together. I did as I saw in other places.

reidzeibel:
@jorelosorio need to run code formatting on `src/components/estimated-call/DepartureTime.tsx`

gorandalum (approved):

PR #4863, opened 2024-12-04
fix: Reduced borderWidth for InspectionSymbol, and removed to-from zone names

strandlie:
Closes AtB-AS/kundevendt#19517

### DIFF
--- a/src/fare-contracts/components/InspectionSymbol.tsx
+++ b/src/fare-contracts/components/InspectionSymbol.tsx
@@ -5,7 +5,6 @@ import {ThemeText} from '@atb/components/text';
 import React from 'react';
 import {ThemeIcon} from '@atb/components/theme-icon';
 import {
-  getReferenceDataName,
   PreassignedFareProduct,
   ProductTypeTransportModes,
   TariffZone,
@@ -77,7 +76,6 @@ const InspectableContent = ({
   toTariffZone?: TariffZone;
   themeColor: ContrastColor;
 }) => {
-  const {language} = useTranslation();
   const styles = useStyles();
 
   const {fareProductTypeConfigs} = useFirestoreConfiguration();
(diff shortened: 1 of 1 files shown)


reidzeibel (commented):
I think the removal of zone name should be followed by the design update for the ticket stuffs, because if we remove it now, the user has no way to know the valid zone of their ticket. 

We need the zone indicator to be added before we remove the zone name.
image

strandlie:
> I think the removal of zone name should be followed by the design update for the ticket stuffs, because if we remove it now, the user has no way to know the valid zone of their ticket.
> 
> We need the zone indicator to be added before we remove the zone name. image

I will update all designs included in AtB-AS/kundevendt#4220 before I merge this. Hang tight for a couple more PRs coming soon :)

gorandalum:
> I think the removal of zone name should be followed by the design update for the ticket stuffs, because if we remove it now, the user has no way to know the valid zone of their ticket.
> 
> We need the zone indicator to be added before we remove the zone name. image

Just a note that we do show zone names on fare contracts as text as well, not just as part of the inspection symbol.
Screenshot 2024-12-05 at 10 53 46

src/fare-contracts/components/InspectionSymbol.tsx
```diff
@@ -27,8 +26,6 @@ export type InspectionSymbolProps = {
 
 export const InspectionSymbol = ({
   preassignedFareProduct,
-  fromTariffZone,
-  toTariffZone,
```
gorandalum, line 31: Can also remove these from the `InspectionSymbolProps`

strandlie: Done

strandlie:
> > I think the removal of zone name should be followed by the design update for the ticket stuffs, because if we remove it now, the user has no way to know the valid zone of their ticket.
> > We need the zone indicator to be added before we remove the zone name. image
> 
> I will update all designs included in [AtB-AS/kundevendt#4220](https://github.com/AtB-AS/kundevendt/issues/4220) before I merge this. Hang tight for a couple more PRs coming soon :)

Not true any more. Will merge this now, as the mentioned change is very nearly finished.

PR #4864, opened 2024-12-04
fix: use correct selectable traveller count

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/19640

The `userProfileWithCount` is not used to determine whether the user can select travellers or not, therefore I added `selectableUserProfiles` to help determine the number of travellers that can be selected, this will then affect the display of the on-behalf-of layout on the purchase overview screen.

I'm guessing we could also just use the `userProfileRefs` count from the `limitation` field inside the `referenceData`, but then we need to have `UserProfile[]` for the `isSelectableUserProfile` function to work.

### DIFF
--- a/src/purchase-selection/utils.ts
+++ b/src/purchase-selection/utils.ts
@@ -88,21 +88,15 @@ export const getDefaultUserProfiles = (
   input: PurchaseSelectionBuilderInput,
   product: PreassignedFareProduct,
 ): UserProfileWithCount[] => {
-  const selectableProfiles = getSelectableUserProfiles(
-    input.userProfiles,
-    product,
-  );
-  const profile = selectableProfiles.reduce((selected, current) =>
-    current.userTypeString === input.defaultUserTypeString ? current : selected,
-  );
+  const profile = input.userProfiles
+    .filter((profile) => isSelectableProfile(product, profile))
+    .reduce((selected, current) =>
+      current.userTypeString === input.defaultUserTypeString
+        ? current
+        : selected,
+    );
   return [{...profile, count: 1}];
 };
-
-export const getSelectableUserProfiles = (
-  userProfiles: UserProfile[],
-  product: PreassignedFareProduct,
-): UserProfile[] =>
-  userProfiles.filter((profile) => isSelectableProfile(product, profile));
 
 export const isSelectableProduct = (
   input: PurchaseSelectionBuilderInput,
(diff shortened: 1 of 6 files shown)


jorelosorio (approved):

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-defaults.ts
```diff
@@ -59,6 +60,10 @@ export function useOfferDefaults(
     ],
   );
 
+  const selectableUserProfiles = userProfiles.filter((u) =>
+    selection.preassignedFareProduct.limitations.userProfileRefs.includes(u.id),
+  );
```
gorandalum, line 65: Should use the `getSelectableUserProfiles` util from purchase-selection instead which also handles empty `limitations`.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -49,7 +49,11 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
     ? 'isFree' in params.toPlace && !!params.toPlace.isFree
     : false;
 
-  const {selection, preassignedFareProductAlternatives} = useOfferDefaults(
+  const {
+    selection,
+    preassignedFareProductAlternatives,
+    selectableUserProfiles,
+  } = useOfferDefaults(
```
gorandalum, line 56: Rather do this outside of the `useOfferDefaults`. The `useOfferDefaults` will be deleted soon when continuing the refactor utilising the `PurchaseSelectionType`.

gorandalum (changes requested):

src/purchase-selection/utils.ts
```diff
@@ -88,22 +88,16 @@ export const getDefaultUserProfiles = (
...
   product: PreassignedFareProduct,
 ): UserProfileWithCount[] => {
-  const selectableProfiles = getSelectableUserProfiles(
-    input.userProfiles,
-    product,
-  );
-  const profile = selectableProfiles.reduce((selected, current) =>
-    current.userTypeString === input.defaultUserTypeString ? current : selected,
-  );
+  const profile = input.userProfiles
```
reidzeibel, line 91: Not sure how to refactor this one to work with the new hook, any suggestions @gorandalum ?

gorandalum: If we didn't want to touch this code we could let `getSelectableUserProfiles` still exist, and make `useSelectableUserProfiles` use `getSelectableUserProfiles`. But no biggie, this looks ok!

gorandalum (approved):

reidzeibel:
@gorandalum thanks, I'll note it down

PR #4867, opened 2024-12-06
refactor: Allow change theme text to allow a type color definition

jorelosorio:
Part of https://github.com/AtB-AS/kundevendt/issues/17880

This tries to allow ThemeText to be more flexible about the colour to choose, here is a discussion https://github.com/AtB-AS/mittatb-app/pull/4860#discussion_r1873260173 where they did not match the header well because the secondary colour did not match when light mode.

Refer to the link for more details.

### DIFF
--- a/src/storybook/stories/ThemeText.stories.tsx
+++ b/src/storybook/stories/ThemeText.stories.tsx
@@ -16,14 +16,14 @@ const ThemeTextMeta: Meta<ThemeTextMetaProps> = {
   title: 'ThemeText',
   component: ThemeText,
   argTypes: {
-    type: {
+    typography: {
       control: 'select',
       options: textNames,
     },
     ...themedStoryControls,
   },
   args: {
-    type: 'body__primary',
+    typography: 'body__primary',
     isMarkdown: false,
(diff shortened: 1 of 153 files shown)


src/storybook/stories/ThemeText.stories.tsx
```diff
@@ -16,14 +16,14 @@ const ThemeTextMeta: Meta<ThemeTextMetaProps> = {
...
-    type: {
+    typography: {
       control: 'select',
       options: textNames,
     },
     ...themedStoryControls,
   },
   args: {
-    type: 'body__primary',
+    typography: 'body__primary',
```
gorandalum, line 26: Could you also add the new type parameter in Storybook?

tft-abel (approved):
A natural progression of the `<ThemeText/>` component I'd say, integrating even better with the new structure of the design system! It is still unclear when ThemeText takes the foreground color and when it uses the actual background color when passed a ContrastColor, but that is out of scope for this issue. Nice improvement! :)

gorandalum (approved):

PR #4868, opened 2024-12-09
fix: Length checks need to be converted to booleans

gorandalum:
Or else this code 'arr.length && …' will try to render the text "0"
if the array length is zero, which will blow up since the text
isn't rendered inside a Text component.


reidzeibel (approved):
good catch!

strandlie (approved):

PR #4870, opened 2024-12-09
feat: add feature flag for onboarding login

rosvik:
Disables the onboarding login if the `enable_onboarding_login` remote config flag is `false`.

Also adds remote config to the loading boundary, making the app wait for remote config to load before letting users into the app, since the `enable_onboarding_login` flag is needed before onboarding starts. If loading of remote config fails, the default value is true, meaning that the login will be shown after a loading boundary timeout.

Note: There's an issue with the extended onboarding (used by FRAM, and maybe other OMS partners), where if it's the only onboarding screen, it won't be able to navigate past the final extended onboarding page. (It ends up on `navigation.goBack()` in https://github.com/AtB-AS/mittatb-app/blob/master/src/onboarding/use-onboarding-navigation.ts#L42 )

closes https://github.com/AtB-AS/kundevendt/issues/19356

### DIFF
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -30,6 +30,7 @@ export type RemoteConfig = {
   enable_non_transit_trip_search: boolean;
   enable_nynorsk: boolean;
   enable_on_behalf_of: boolean;
+  enable_onboarding_login: boolean;
   enable_parking_violations_reporting: boolean;
   enable_posthog: boolean;
   enable_push_notifications: boolean;
@@ -99,6 +100,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_non_transit_trip_search: true,
   enable_nynorsk: true,
   enable_on_behalf_of: false,
+  enable_onboarding_login: true,
   enable_parking_violations_reporting: false,
   enable_posthog: false,
   enable_push_notifications: false,
@@ -208,6 +210,9 @@ export function getConfig(): RemoteConfig {
   const enable_on_behalf_of =
     values['enable_on_behalf_of']?.asBoolean() ??
     defaultRemoteConfig.enable_on_behalf_of;
+  const enable_onboarding_login =
+    values['enable_onboarding_login']?.asBoolean() ??
+    defaultRemoteConfig.enable_onboarding_login;
   const enable_parking_violations_reporting =
     values['enable_parking_violations_reporting']?.asBoolean() ??
     defaultRemoteConfig.enable_parking_violations_reporting;
@@ -351,6 +356,7 @@ export function getConfig(): RemoteConfig {
     enable_non_transit_trip_search,
     enable_nynorsk,
     enable_on_behalf_of,
+    enable_onboarding_login,
     enable_parking_violations_reporting,
     enable_posthog,
     enable_push_notifications,
--- a/src/RemoteConfigContext.tsx
+++ b/src/RemoteConfigContext.tsx
@@ -21,6 +21,7 @@ const RETRY_INTERVAL_MS_MAX = 30000;
 
 export type RemoteConfigContextState = RemoteConfig & {
   refresh: () => void;
+  isLoaded: boolean;
   feedback_questions: FeedbackConfiguration[];
 };
 
@@ -61,6 +62,7 @@ export const RemoteConfigContextProvider: React.FC = ({children}) => {
   const [config, setConfig] = useState<RemoteConfig>(defaultRemoteConfig);
   const [fetchError, setFetchError] = useState(false);
   const [retryInterval, incrementRetryInterval] = useRetryIntervalWithBackoff();
+  const [isLoaded, setIsLoaded] = useState(false);
 
   const fetchConfig = useCallback(async () => {
     try {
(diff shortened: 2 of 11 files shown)


gorandalum (commented):
Should this be in the `FeatureTogglesContext` to be able to test with debug override?

rosvik:
@gorandalum Right, I guess it's possible to test locally by setting debug override, then restarting onboarding sections 👍 I'll add it

src/remote-config.ts
```diff
@@ -99,6 +100,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_non_transit_trip_search: true,
   enable_nynorsk: true,
   enable_on_behalf_of: false,
+  enable_onboarding_login: true,
```
gorandalum, line 103: Could we get a race condition between Remote config and the onboarding? After app install the user needs Remote config to be loaded to get the `false` value, but RemoteConfig isn't part of things we wait for during app loading.

rosvik: Oh, then maybe 🤔 I'll try to test for it, but we _should_ probably wait for remote config on app launch then?

gorandalum: Yeah, might be good to add that logic now. Note that we can't always wait for RemoteConfig on app start, since if the existing data is not outdated it won't be fetched.

gorandalum:
> Note: There's an issue with the extended onboarding

If I understand correctly, the new flag `enable_onboarding_login` can't be used in combination with the flag `enable_extended_onboarding`?

rosvik:
> > Note: There's an issue with the extended onboarding
> 
> If I understand correctly, the new flag `enable_onboarding_login` can't be used in combination with the flag `enable_extended_onboarding`?

Correct.

> Regarding the extended onboarding issue: For a screen that is the last that has `shouldShowBeforeUserCreated=true` in `onboarding-config.ts` and returns `true` from `shouldShowPredicate`, it probably needs to use `completeUserCreationOnboardingAndEnterApp` instead of just `completeOnboardingSection` to continue.

I guess that would make sense. I guess it's ok to complete the user creation onboarding even though it wasn't actually shown. Again, the code that causes the bug is this in [`continueFromOnboardingSection`](https://github.com/AtB-AS/mittatb-app/blob/master/src/onboarding/use-onboarding-navigation.ts#L42), where calling `goBack` doesn't help with completing the onboarding:

```ts
if (nextOnboardingSection?.initialScreen?.name) {
  goToScreen(true, nextOnboardingSection?.initialScreen);
} else {
  navigation.goBack();
}
```

@marius-at-atb Do you know for what case the `navigation.goBack()` is useful? Could it be changed to this?

```ts
if (nextOnboardingSection?.initialScreen?.name) {
  goToScreen(true, nextOnboardingSection?.initialScreen);
} else {
  completeUserCreationOnboardingAndEnterApp()
}
```

src/remote-config.ts
rosvik: Added now. Done by setting a `isLoaded` flag in RemoteConfigContext after config has been fetched. Seems to works well when testing locally on fresh installs, and `isLoaded` becomes true right after default values are overwritten.

marius-at-atb:
@rosvik E.g. when the stack is `Root_TabNavigatorStack` + `Root_LocationWhenInUsePermissionScreen` , the go back returns to `Root_TabNavigatorStack`.
In Root_TabNavigatorStack, you see the line 
```ts
nextOnboardingScreen?.name && goToScreen(false, nextOnboardingScreen);
```
basically this can push an onboarding screen, and the go back pops it away again, so that you return to exactly where you were. If using `completeUserCreationOnboardingAndEnterApp` there, it would reset the stack, and might not return to the same place.

Side notes:
- Of all the screens to remove from the onboarding, `Root_LoginOptionsScreen` was the one we didn't expect 😅 
- The need for `shouldShowBeforeUserCreated` was also meant to be removed, but we had to move on.
- The loading boundary 'intervenes' in the switch between an anonymous user and a phone user, but doesn't when just using an anonymous user (as is the case without `Root_LoginOptionsScreen`). This is why from before, `completeUserCreationOnboardingAndEnterApp` is used for anon login, but not phone.

So ideally we'd have no need for `shouldShowBeforeUserCreated`, and also no `LoadingBoundary` intervention. Not sure if that could be within scope here?

gorandalum (approved):
Nice!

rosvik:
@marius-at-atb Ok, thanks! I don't think I'll dig more into this for now since it's unlikely we'll use the extended onboarding in combination with no login onboading any time soon. 

But a thought: Would some of these issues be fixed by moving the extended onboarding to after login? The screens might not have to be in the current order.

marius-at-atb:
Makes sense :+1:
Yes, that should work!
Actually, even just switching `shouldShowBeforeUserCreated` to `false` as well might be sufficient now that i come to think of it?

rosvik:
> Actually, even just switching shouldShowBeforeUserCreated to false as well might be sufficient now that i come to think of it?

Not as easy as that, it seems. Still have the same issues locally. Maybe I'm missing something, but I'll leave it at that 😄

PR #4871, opened 2024-12-09
style: reduce size of Aztek

adriansberg:
Now that FRAM will be going from using static QR code to using Aztek, the testers have experienced that the Aztek is too large for FLV+, and that FLV+ sometimes has trouble reading. By reducing it to a maximum width of 200 (up for discussion), it will be easier for the FLV+ to read the code.

Co-authored-by: Morten Nordseth <morten@nordseth.net>
Co-authored-by: Mathias Oterhals Myklebust <mom@variant.no>

### DIFF
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -257,9 +257,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     aspectRatio: 1,
     padding: theme.spacing.large,
     backgroundColor: '#FFFFFF',
-    marginLeft: 'auto',
-    marginRight: 'auto',
-    maxWidth: 200,
+    maxHeight: 275,
   },
   staticBottomContainer: {
     flex: 1,
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -257,6 +257,9 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
     aspectRatio: 1,
     padding: theme.spacing.large,
     backgroundColor: '#FFFFFF',
+    marginLeft: 'auto',
+    marginRight: 'auto',
+    maxWidth: 200,
   },
   staticBottomContainer: {
     flex: 1,
--- a/src/fare-contracts/details/Barcode.tsx
+++ b/src/fare-contracts/details/Barcode.tsx
@@ -106,15 +106,13 @@ const MobileTokenAztec = ({fc}: {fc: FareContract}) => {
   }
 
   return (
-    <View style={{alignItems: 'center'}}>
-      <View
-        style={styles.aztecCode}
-        accessible={true}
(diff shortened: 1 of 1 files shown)


gorandalum:
The code changes looks ok, but @tormoseng will test this with our inspection app before we approve.

tormoseng:
@adriansberg Me and @gorandalum tested the new size with the inspection app from Entur. From simple research it seems that we need to decrease the distance to the code on the screen some before it's read. That might make it worse for the inspectors. Is it possible to increase to size `300` for a test - or do you have an upper limit for the FLV+ ?

Also, the code is not centralized on my device (see screenshot).

<details>
<summary>Screenshot</summary>

Screenshot_20241210_122954_AtB (Dev)

</details>

adriansberg:
@tormoseng Looks like iOS and Android behaves differently for the centering, is centered on iOS. Will fix that. 

`250` is the largest that have been used for the static QR. I can increase the size to `275` and test that?



<details><summary>275</summary>
<p>

image


</p>
</details> 

<details><summary>Screenshots</summary>
<p>
Before:

2024-12-10 at 13 01 39@2x

After:
2024-12-10 at 12 57 34@2x


</p>
</details>

mikaelbr:
In the FLV+ there is a camera. That camera has a field of view based on distance to device, so given that the device is optimized to be about 2cm away from the screen (this is a calibration that can be done on FLV+') there are some sizes that will be cut off. E.g. iPhone Pro Max with the size as of now, the Aztec will be so large only a part of the barcode will be observable from the camera it self. Aztec has fault tolerance of something like 26% so it should be somewhat robust, but in testing we have a hard time getting it to work seamlessly.

So it is hard to say specific upper limit, and it is kind of hard to translate actual real world sizes to device sizes, but say it shouldn't be much taller than FLV+ glass size. (4cm?).

Regarding inspection app: I think that it is easier for inspectors to learn new usage patterns than having all public customers having to compensate for this while onboarding bus with lots of travellers etc. This is something we have issue with today that reading barcodes takes too long and we have worked _a lot_ with Fara, Entur and MT Buss to reduce the time, but having size larger when going over to Aztec would potentially set us noticeable back on the work for reducing validation time.

tormoseng:
@adriansberg 275 works well with the inspection app. If that works for you, I say go for it.

@mikaelbr Thanks for the background information. I really see your point.

@Sebstorvik Guess this issue is good for you to know about also. Maybe mention to Stig & Co.

adriansberg:
@tormoseng We'll go ahead with 275 and will test it with various devices and distances in staging.

gorandalum (approved):

adriansberg:
Centering verified on both Android and iOS. To me it seemed that the `View` wrapping the other `View` in `MobileTokenAztec` was unnecessary, since the Aztec is centered without it. Could you double check that that is true, @gorandalum, so that nothing breaks by removing it?

tormoseng:
Tested it locally, and it looks good in the AtB app.

PR #4872, opened 2024-12-09
chore: Add eslint rules for not leaking values to jsx

gorandalum:
In react-native the app crashes if trying to render text outside of
Text component. This has happened in our production app more than
once because of checks on length, and to be fair it will probably
happen more if we don't do anything about it.

This change adds an eslint rule that gives error when using in
combination with && inside jsx.

The eslint rule used is not that popular, but seems well created
and safe to use when reading through its code.

### Acceptance criteria
- [ ] Should be no changes for the app.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,7 +29,9 @@ module.exports = {
   },
   parserOptions: {
     sourceType: 'module',
+    project: './tsconfig.json',
   },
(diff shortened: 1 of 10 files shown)


rosvik (approved):
Nice!

strandlie (approved):
Good idea. SImple way to handle a common source of bugs.

tormoseng:
✅

PR #4874, opened 2024-12-11
chore: Eslint rule to enforce no file endings on imports

gorandalum:
Remember to update your IDE to not auto-include file endings on
imports, this should be an option in most IDEs.


reidzeibel (approved):
VSCode settings location

image

marius-at-atb (approved):

jorelosorio (approved):

PR #4876, opened 2024-12-11
refactor: Add more clear and tested fare contract availability status

gorandalum:
The logic whether a fare contract is upcoming, valid, expired, etc., 
was a little all over the place and hard to really get a grasp on. This
PR is made as an effort to be more precise on what these terms
actually are, and to make it easier to retrieve fare contracts with
specific availability status.

The new utils and hooks are not taken in use in this PR. I found
that taking them in use gave a lot of changes, but I wanted the PR
to be standalone to make possible for better reviews of the actual
logic, which is important to be correct and understandable. After
this is merged it will be another PR to replace old util functions
with usage of these new ones.

This will make the code inline with the terminology we have
described for fare contracts:
https://github.com/AtB-AS/docs-private/blob/main/terminology.md


rosvik (commented):
Some feedback on the terminology: https://github.com/AtB-AS/docs-private/pull/67#issuecomment-2536307974

gorandalum:
Updated the PR after the latest [terminology discussion](https://github.com/AtB-AS/docs-private/pull/67).

src/ticketing/utils.ts
```diff
@@ -47,8 +47,11 @@ export function hasValidRightNowTravelRight(
   );
 }
 
-function isValidRightNowTravelRight(travelRight: TravelRight, now: number) {
-  // Unknown travel rights are not active
+export function isValidRightNowTravelRight(
```
rosvik, line 50: I guess we don't need the "RightNow" parts of names anymore, if "valid" is always relative to `now`

rosvik (approved):
Great! Makes sense to me 💯

src/ticketing/utils.ts
gorandalum: This wasn't ment to be included for now, so I removed it. But you are right, we don't need to "RightNow" part, and hopefully we don't need this function at all after we start using `getAvailabilityStatus` around the app.

PR #4878, opened 2024-12-11
feat: Store only stop places checkbox state on device

gorandalum:
This is decided to be in place for next round of testing of the
only stop places checkbox. In last round of testing the usage of
the checkbox was quite high at the start, but then dropped of. The
theory is that users don't use the extra click to check off the
checkbox for every search.

### Acceptance criteria
- [ ] Only stop places checkbox state persists when opening and closing location search screen
- [ ] Only stop places checkbox state persists when restarting app
- [ ] Only stop places checkbox state is shared between all places one can do location search in app

### DIFF
--- a/src/stacks-hierarchy/Root_LocationSearchByTextScreen/components/LocationSearchContent.tsx
+++ b/src/stacks-hierarchy/Root_LocationSearchByTextScreen/components/LocationSearchContent.tsx
@@ -21,6 +21,7 @@ import {animateNextChange} from '@atb/utils/animation';
 import {CheckboxWithLabel} from '@atb/components/checkbox';
 import {useAnalytics} from '@atb/analytics';
 import {useFeatureToggles} from '@atb/feature-toggles';
+import {storage} from '@atb/storage';
 
 type LocationSearchContentProps = {
   label: string;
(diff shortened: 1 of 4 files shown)


rosvik (approved):

src/stacks-hierarchy/Root_LocationSearchByTextScreen/components/LocationSearchContent.tsx
```diff
@@ -224,6 +225,24 @@ export function LocationSearchContent({
   );
 }
 
+const useOnlyStopPlacesState = (): [boolean, (b: boolean) => void] => {
```
marius-at-atb, line 228: Maybe we should have a general util for this, something like `useLocallyStoredState`?

gorandalum:
> Maybe we should have a general util for this, something like useLocallyStoredState?

I added it as a hook in src/utils now.

Some notes:
- I added tests for the hook. I think it is good practice that stuff in src/utils are throughly tested.
- To not get a dependency from a general util to the storage module the storage service is dependency injected. This also makes the hook more easily testable.
- The hook is called `usePersistedBoolState`. "Locally" is not used since the hook doesn't really know whether it is locally stored or not as it don't know how the storage service persists the data. Also it only supports booleans for now, but can be expanded if later necessary.

marius-at-atb (approved):
Lgtm. If we find that we need it, initial value for the bool and isLoading state (since it is async) may also be considered in the future.

tormoseng:
✅  Tested OK. Closes https://github.com/AtB-AS/kundevendt/issues/19643

PR #4880, opened 2024-12-12
fix: Fixed wrong arrow-icons for bus/tram tickets, and added margins

strandlie:
Follow-up to AtB-AS/kundevendt#19514. Discovered while testing.

### DIFF
--- a/src/fare-contracts/components/FareContractFromTo.tsx
+++ b/src/fare-contracts/components/FareContractFromTo.tsx
@@ -57,7 +57,13 @@ export const FareContractFromTo = (props: FareContractFromToProps) => {
     if (hasFareContract(props)) {
       const travelRight = props.fc.travelRights[0];
       if (isNormalTravelRight(travelRight)) {
-        return travelRight.direction;
+        if (!!travelRight.direction) {
+          // A travelRight between quays (e.g. for boat)
+          return travelRight.direction;
+        } else if (travelRight.tariffZoneRefs?.length ?? 0 > 1) {
+          // A travelRight between several zones (e.g. for bus)
+          return TravelRightDirection.Both;
+        }
       }
     } else if (hasRecentFareContract(props)) {
       return props.rfc.direction;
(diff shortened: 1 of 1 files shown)


gorandalum (approved):

src/fare-contracts/components/FareContractFromTo.tsx
```diff
@@ -264,11 +270,13 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
   largeContent: {
     flexDirection: 'column',
     alignItems: 'center',
+    rowGap: 4,
```
hanne-at-atb, line 273: Use `theme.spacing.xsmall` if I remember correct

src/fare-contracts/components/FareContractFromTo.tsx
```diff
@@ -57,7 +57,13 @@ export const FareContractFromTo = (props: FareContractFromToProps) => {
...
+        if (!!travelRight.direction) {
+          // A travelRight between quays (e.g. for boat)
+          return travelRight.direction;
+        } else if (travelRight.tariffZoneRefs?.length ?? 0 > 1) {
+          // A travelRight between several zones (e.g. for bus)
+          return TravelRightDirection.Both;
+        }
       }
     } else if (hasRecentFareContract(props)) {
       return props.rfc.direction;
```
gorandalum, line 69: Does recent fare contract always have direction? I would think the same issue with checking if multiple zones is present here.

Coming to think of it since zones doesn't have direction I think we can just send the `TravelRightDirection` directly to the `HarborsFromTo`. While `ZonesFromTo` is hardcoded internally to use `Both`.

strandlie: Clever thinking, but for RecentFareContracts it is only ever shown [if direction is defined](https://github.com/AtB-AS/mittatb-app/blob/master/src/recent-fare-contracts/RecentFareContract.tsx#L165) so the issue is masked there. 

But that might change, so I'll take a look at your suggestion!

strandlie: Added some further checks on `direction` for RecentFareContract now

PR #4883, opened 2024-12-16
feat: Add expand property to the button

jorelosorio:
Part of: https://github.com/AtB-AS/kundevendt/issues/17880

image


jorelosorio:
> The change it self look ok, but some more surrounding work needs to be done:
> 
> * It need to be aligned with [Figma](https://www.figma.com/design/2QTjAdekdIPuLFovQhVrY3/Components?node-id=9051-96246&t=KcKIQjG8GdXsLNYL-4), both with regards to look, icon placement and parameter name.
> * This change makes it not necessary to have the large and medium buttons, as the only difference between them are the width.

Hi, I will contact design to align the Figma, At the last meeting they agreed to do those changes, and I am going to follow up.

jorelosorio:
@ckbilstad and I had a meeting about the button and we agreed on removing the large button and keep small and large, so for both cases, the expand prop can be used, however, for the small the icon and text will be centred. I might take more effort to remove the medium button usage.

gorandalum (changes requested):
Regarding the `expand` field, should it be `expanded` to follow the convention from the other fields?

I tested this branch out locally, and a couple things stood out:
- Icons on large expanded buttons aren't placed correctly.
- Large buttons in the app are no longer expanded.

Screenshots:
<div>


</div>

jorelosorio:
I made fixes for this PR @gorandalum

jorelosorio:
The implementation of this expand button in this PR was not working properly and was moved out to a new PR https://github.com/AtB-AS/mittatb-app/pull/4907

PR #4885, opened 2024-12-16
fix: prevent serviceJourneyVehicles retries from useQuery

rosvik:
Retries are already handled by axios, so this is a quick fix to prevent retries from happening on two levels in this case. We should look into a more permanent and global solution as well.

ref. https://mittatb.slack.com/archives/C0116FMPX4Y/p1734353027280489


rosvik:
I'll cherry pick to a 1.60.1 release branch after it's merged.

gorandalum (approved):

tormoseng:
✅  Looks good!

PR #4887, opened 2024-12-17
refactor: Use Context suffix on hooks for using contexts

gorandalum:
Lots of line changes, but should be pretty safe as I've only renamed methods.

### Background
https://mittatb.slack.com/archives/C02EEG7D8EL/p1734364805443509

### Acceptance criteria
- [ ] Everything works as before. Doesn't need to be tested explicitly.

### DIFF
--- a/src/GeolocationContext.tsx
+++ b/src/GeolocationContext.tsx
@@ -334,7 +334,7 @@ const openSettingsAlert = () =>
     {cancelable: true},
   );
 
-export function useGeolocationState() {
+export function useGeolocationContext() {
   const context = useContext(GeolocationContext);
   if (context === undefined) {
     throw new Error(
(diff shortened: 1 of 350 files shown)


gorandalum:
And don't be afraid to comment if you think this is not an overall positive change. I think this is slightly better than the existing "random" naming of these hooks, but also I think it kinda violates the "program to an interface, not an implementation" principle, as with this change we now expose through the naming that the hook is implemented by using a React Context.

reidzeibel:
I think this is a positive change, since it is indeed a bit confusing with the `state` and `context`, should be useful for both new and old devs.

rosvik (approved):
I was skeptical to adding verbosity to very commonly used hooks like `useTheme`, but this PR actually has less changes than I expected, so I'm positively surprised! Either way, I'm all for consistency, and think this is a good change 👍

tormoseng:
✅  Quick walkthrough in both Android and iOS.

PR #4898, opened 2024-12-20
fix: disable fetching of vehicles when out of focus

rosvik:
We're still seeing a lot of requests towards vehicles in 1.60.1, and it seems it might be caused by requests that are active without focus. E.g. if the user navigates from trip details to departure details or switches tab, the trip query is still active and fetching updates every 20s. And if several screens with vehicles queries are open, several requests are sent every 20s.

follow up to https://github.com/AtB-AS/mittatb-app/pull/4885

### DIFF
--- a/src/travel-details-screens/use-get-service-journey-vehicles.ts
+++ b/src/travel-details-screens/use-get-service-journey-vehicles.ts
@@ -1,12 +1,14 @@
 import {getServiceJourneyVehicles} from '@atb/api/vehicles';
+import {useIsFocusedAndActive} from '@atb/utils/use-is-focused-and-active';
 import {useQuery} from '@tanstack/react-query';
 
 export const useGetServiceJourneyVehiclesQuery = (
   serviceJourneyIds: string[],
   enabled: boolean = true,
 ) => {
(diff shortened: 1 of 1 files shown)


gorandalum (approved):
Good as a quick fix. React-query should have better support for this, and I even think it shouldn't refetch in background by default, based on the [doc](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) for `refetchIntervalInBackground`.

> If set to true, queries that are set to continuously refetch with a refetchInterval will continue to refetch while their tab/window is in the background

Might be possible to solve with [`FocusManager`](https://tanstack.com/query/v4/docs/reference/focusManager) so we can hoop up the react query focus logic to the react-native active and focused state.

rosvik:
> I even think it shouldn't refetch in background by default, based on the [doc](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) for refetchIntervalInBackground

Looks like this is true for AppState, but not focus. When the app is closed/minimized, the useQuery is disabled. But if you navigate to another tab or view while still in the app, it's still active.

tormoseng:
✅  Tested OK

PR #4899, opened 2025-01-02
fix: Some date tests failing after new year

gorandalum:
The tests need to have current year provided programatically to not
need manual fixing every new year.

In addition when adding some more tests it came to light that the
formatting of the day and month differed based on whether it was
the current year or not. The full date formatting was changed to
give the same day/month format no matter if current year or not.

### Acceptance criteria
- [ ] Full dates presented correctly both for current and previous year, in english and norwegian. (Can be viewed in sent fare contract header)


reidzeibel (approved):

tormoseng:
✅

PR #4900, opened 2025-01-03
refactor: Use the new useFareContracts hook

gorandalum:
Also rename active -> available and expired -> historic.


reidzeibel (approved):
Can't find anything wrong, looks like a nice refactor to start using the new terminology!

strandlie (approved):

PR #4904, opened 2025-01-06
fix: Set BorderedInfoBox border color in same way as SectionItem separator color

strandlie:
This was previously set using opacity, which obviously only works in light mode. 

**Before:** 
<div>
    drawing
    drawing
</div>

**After:**
<div>
    drawing
    drawing
</div>

### DIFF
--- a/src/components/bordered-info-box/BorderedInfoBox.tsx
+++ b/src/components/bordered-info-box/BorderedInfoBox.tsx
@@ -3,6 +3,7 @@ import {StyleProp, View, ViewStyle} from 'react-native';
 import React, {ReactNode} from 'react';
 import {StyleSheet} from '@atb/theme';
 import {ContrastColor} from '@atb/theme/colors';
+import {addOpacity} from '@atb/utils/add-opacity';
 
 export type BorderedInfoBoxProps =
   | {
@@ -26,7 +27,7 @@ export const BorderedInfoBox = ({
   testID,
   ...props
 }: BorderedInfoBoxProps) => {
-  const styles = useStyles(type);
+  const styles = useStyles(type, backgroundColor.foreground.primary);
   return (
     <View style={[styles.container, style]}>
       {'text' in props ? (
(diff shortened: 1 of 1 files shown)


gorandalum (changes requested):
It would be nice with some before and after pictures on changes like this :)

> This was previously set using opacity, which obviously only works in light mode.

I'm not sure on that statement. I think setting opacity works both with white on black background and vice versa 🤔 But with the recent changes to the design system it might make more sense to use the `secondary` or `disabled` color in the `ContrastColor.foreground` type.

With the change made here it does no longer work for all backgrounds. Take a look in Storybook.

strandlie:
Yes I should have added pictures here. Sorry about that!

**Before:** 
<div>
    drawing
    drawing
</div>

**After:**
<div>
    drawing
    drawing
</div>
 

> > This was previously set using opacity, which obviously only works in light mode.
> 
> I'm not sure on that statement. I think setting opacity works both with white on black background and vice versa 🤔 But with the recent changes to the design system it might make more sense to use the `secondary` or `disabled` color in the `ContrastColor.foreground` type.
> 

I would also think so, but when you look at the pictures there is clearly a difference between the perceived luminance of the border, even if the contrast ratios between the border and the background are similar (1.25 for light mode, 1.20 for dark mode). The before-picture is in production today, so you can take a look. 



> With the change made here it does no longer work for all backgrounds. Take a look in Storybook.

I see that it does not work with some backgrounds. The borderColor is now set to the same color as the [Section separator](https://github.com/AtB-AS/mittatb-app/blob/master/src/components/sections/Section.tsx#L66) so that means that the separator has that issue in production today. I will check with @ckbilstad what the intention was here.

strandlie:
New discussion, and we just change opacity from 10% to 20% percent on the reasoning that `BorderedInfoBox` can be used on any background and setting its borderColor to `background[2]` means that it will disappear when used on that background, i.e. hidden knowledge.

gorandalum (approved):

src/components/bordered-info-box/BorderedInfoBox.tsx
```diff
@@ -45,10 +45,10 @@ export const BorderedInfoBox = ({
   );
 };
 
-const useStyles = (type: BorderedInfoBoxProps['type'], textColor: string) =>
-  StyleSheet.createThemeHook((theme) => ({
+const useStyles = (type: BorderedInfoBoxProps['type'], textColor: string) => {
+  return StyleSheet.createThemeHook((theme) => ({
     container: {
-      borderColor: addOpacity(textColor, 0.1),
+      borderColor: addOpacity(textColor, 0.2),
```
jorelosorio, line 51: seems like the only thing that changes was from `0.1` to `0.2` we might then revert all the other things as before, here now seems like we added a return and changed the position of the import.

hanne-at-atb:
Tested ok 👍🏼

PR #4906, opened 2025-01-06
fix: GeofencingZone onPress

marius-at-atb:
Sometimes when clicking a geofencing zone, the snackbar didn't show.
This was because of incorrect input for `turfBooleanPointInPolygon`.

Acceptance Criteria:
- [ ] Clicking on a GeofencingZone should always result in a correct Snackbar behaviour.

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -6,9 +6,8 @@ import {FOCUS_ORIGIN} from '@atb/api/geocoder';
 import {StyleSheet} from '@atb/theme';
 import {MapRoute} from '@atb/travel-details-map-screen/components/MapRoute';
 import MapboxGL, {LocationPuck, MapState} from '@rnmapbox/maps';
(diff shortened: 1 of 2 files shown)


strandlie (approved):

hanne-at-atb:
What do you mean by `correct Snackbar behaviour` in the acceptance criteria @marius-at-atb ?

marius-at-atb:
https://github.com/AtB-AS/docs-private/blob/main/geofencing-zones.md

When clicking on a zone, it shows
- black -> No entry
- red -> No parking
- yellow -> Slow zone
- very light blue -> Allowed

When a Snackbar is already visible and a zone is clicked, it should smoothly transition to show the new one.

marius-at-atb:
A click on a zone should always result in a snackbar. Previously there were some exceptions, which is the bug fixed here.

tormoseng:
Looking good. Will have to test more after the map rewrite.

PR #4907, opened 2025-01-07
feat: Add expand property to the button

jorelosorio:
Part of: https://github.com/AtB-AS/kundevendt/issues/17880

image

This PR is a replacement of: https://github.com/AtB-AS/mittatb-app/pull/4883


src/components/map/components/external-realtime-map/ExternalRealtimeMapSheet.tsx
```diff
@@ -38,6 +38,7 @@ export const ExternalRealtimeMapSheet = ({
           </GenericSectionItem>
         </Section>
         <Button
+          expand={true}
```
strandlie, line 41: We are adding `expand={true}` to a lot of buttons now. To me it would then seem like we should make `expand={true}` the default, and set `expand={false}` where it is needed in stead :)

jorelosorio: Yes, In my previous PR, I did the opposite and I found it difficult to know what should not be expanded in all the code, from the implementation perspective at least for me makes more sense that the mark was expanded and left by default not expanded, like normal button state in general.

strandlie: Seems like we may have a case for something like an `ExpandedButton` then :)

gorandalum: We might have more expanded buttons than we have non-expanded, and with regards to that it makes sense to have the default as `true`. But I'm not really a fan of boolean fields where `undefined` is evaluated as `true`, it is not intuitive for the user of the component, so I like this slightly better. If we should change something, then my personal preference would be to make the field required.

> Seems like we may have a case for something like an ExpandedButton then :)

We actually did a work going in the opposite direction of this recently, where we merged all buttons into this single `Button`. I think it has made it easier to use the code, you always use `Button` and the possible variations are right there in the typings. Also it doesn't really scale well when you have an `ExpandedButton` and `RoundedButton`, and then find out you need a button which is both expanded and rounded.

gorandalum (changes requested):
Looks mostly good!

A couple things:
- When not sending in text the icon isn't placed correctly, and the margin on one side is greater than on the other side. Here is a screenshot from the map:
Screenshot 2025-01-07 at 11 33 09

- I think the `expand` param should be named `expanded`. Then it is more inline with other parameter naming, where you can put "is" in front of the field name: *is disabled*, *is active*, *is compact* and *is expanded* .

jorelosorio:
> Looks mostly good!
> 
> A couple things:
> 
> * When not sending in text the icon isn't placed correctly, and the margin on one side is greater than on the other side. Here is a screenshot from the map:
>   Screenshot 2025-01-07 at 11 33 09
> * I think the `expand` param should be named `expanded`. Then it is more inline with other parameter naming, where you can put "is" in front of the field name: _is disabled_, _is active_, _is compact_ and _is expanded_ .

I made fixes based on your input.

gorandalum (approved):
Nice!

If you want to you can also take a look on how it works when there is no text on a large expanded button. This is from Storybook:
Screenshot 2025-01-08 at 15 49 04

It is a corner case and not likely to be used that way, but it would be beneficial if there are no combination of input which "breaks" the component.

jorelosorio:
> Nice!
> 
> If you want to you can also take a look on how it works when there is no text on a large expanded button. This is from Storybook: Screenshot 2025-01-08 at 15 49 04
> 
> It is a corner case and not likely to be used that way, but it would be beneficial if there are no combination of input which "breaks" the component.

Good catch, I made a fix for it.

simulator_screenshot_41F62FC0-3E4A-43E3-9B74-D8B4EB49F9B6

gorandalum:
> Good catch, I made a fix for it.

Nice. One last thing, if there is no text and no icon, maybe just return `null`?

jorelosorio:
> > Good catch, I made a fix for it.
> 
> Nice. One last thing, if there is no text and no icon, maybe just return `null`?

I was thinking that empty does not look good but that is quite rare case, I made an extra check and return null

hanne-at-atb:
This is looking pretty good as far as I can see when testing through the app. The only thing I notice is on full text zoom for android the icon is outside the button when the text is large @jorelosorio. See image:

PR #4909, opened 2025-01-07
feat: use secondary transport color for trip leg decorations

rosvik:
part of https://github.com/AtB-AS/kundevendt/issues/19401

In preparation for adding a VM specific transport colors, we need to use secondary transport colors for trip leg decorations in order to have enough contrast against the background in dark mode. See ad037ad90812c75553e4bb31459069632135983a for this change.

Since this made the secondary variant of transport colors more used in the app, I thought it made sense to return `TransportColor` (with both primary and secondary) from the useTransportColor hook instead of passing a `colorMode` parameter and returning ContrastColor. Because of this, the PR contains some refactorings related to usage of the hook as well.

## Acceptance criteria

- [ ] Leg decorations (the vertical line) in trip and departure details uses secondary transport colors (slightly darker than primary)
- [ ] Transport colors elsewhere in the app look like before:
	- [ ] Fare contracts
	- [ ] Live bus in map
	- [ ] Statons and scooters in the map
	- [ ] Departures
	- [ ] Trip search

### DIFF
--- a/src/components/estimated-call/LineChip.tsx
+++ b/src/components/estimated-call/LineChip.tsx
@@ -2,7 +2,7 @@ import {Statuses, StyleSheet, useThemeContext} from '@atb/theme';
 import {ThemeIcon} from '../theme-icon';
 import {ThemeText} from '../text';
 import {messageTypeToIcon} from '@atb/utils/message-type-to-icon';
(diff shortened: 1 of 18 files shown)


gorandalum (approved):

hanne-at-atb:
Tested ok ☑️

PR #4910, opened 2025-01-08
feat: Pull to refresh to fetch fare contracts from backend

gorandalum:
Can evaluate commits one-by-one to get a better overview.

### DIFF
--- a/src/ticket-history/TicketHistoryScreenComponent.tsx
+++ b/src/ticket-history/TicketHistoryScreenComponent.tsx
@@ -29,7 +29,7 @@ export const TicketHistoryScreenComponent = ({
     resubscribeFirestoreListeners,
   } = useTicketingContext();
   const {serverNow} = useTimeContext();
-  const historicFareContracts = useFareContracts(
+  const {fareContracts: historicFareContracts} = useFareContracts(
     {availability: 'historic'},
     serverNow,
   );
--- a/src/ticket-history/TicketHistoryScreenComponent.tsx
+++ b/src/ticket-history/TicketHistoryScreenComponent.tsx
@@ -26,13 +26,12 @@ export const TicketHistoryScreenComponent = ({
     isRefreshingFareContracts,
     reservations,
     rejectedReservations,
-    resubscribeFirestoreListeners,
   } = useTicketingContext();
   const {serverNow} = useTimeContext();
-  const {fareContracts: historicFareContracts} = useFareContracts(
-    {availability: 'historic'},
-    serverNow,
-  );
+  const {
+    fareContracts: historicFareContracts,
+    refetch: refetchHistoricFareContracts,
+  } = useFareContracts({availability: 'historic'}, serverNow);
 
   const {abtCustomerId: customerAccountId} = useAuthContext();
 
@@ -54,7 +53,7 @@ export const TicketHistoryScreenComponent = ({
       refreshControl={
         <RefreshControl
           refreshing={isRefreshingFareContracts}
-          onRefresh={resubscribeFirestoreListeners}
+          onRefresh={refetchHistoricFareContracts}
         />
       }
     >
--- a/src/ticketing/api.ts
+++ b/src/ticketing/api.ts
@@ -1,6 +1,10 @@
 import {APP_SCHEME} from '@env';
 import {AxiosRequestConfig} from 'axios';
-import {AddPaymentMethodResponse, ReserveOfferRequestBody} from '.';
+import {
+  AddPaymentMethodResponse,
+  type FareContract,
+  ReserveOfferRequestBody,
+} from '.';
 import {client} from '../api';
 import {
   Offer,
@@ -13,6 +17,7 @@ import {
   TicketRecipientType,
 } from './types';
 import {PreassignedFareProduct} from '@atb/configuration';
+import {convertIsoStringFieldsToDate} from '@atb/utils/date';
 
 export async function listRecentFareContracts(): Promise<
   RecentFareContractBackend[]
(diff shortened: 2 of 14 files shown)


src/ticketing/use-fare-contracts.ts
```diff
@@ -19,5 +45,28 @@ export const useFareContracts = (
...
+  const {userId} = useAuthContext();
+  return useQuery({
+    queryKey: ['FETCH_FARE_CONTRACTS', availability, userId],
+    queryFn: () => getFareContracts(mapAvailability(availability)),
+    enabled: false,
+    retry: 0,
+  });
+};
+
+const mapAvailability = (
```
marius-at-atb, line 63: A bit annoying that we needed this 😅

src/ticketing/use-fare-contracts.ts
```diff
@@ -1,14 +1,40 @@
...
+    setFareContracts(fareContractsFromFirestore);
+  }, [fareContractsFromFirestore]);
+
+  const refetch = () => {
+    getFareContractsFromBackend().then(({data, isSuccess}) => {
+      if (isSuccess) setFareContracts(data);
+    });
+  };
+
+  const filteredContracts = fareContracts.filter((fc) => {
```
marius-at-atb, line 37: Can call this `filteredFareContracts`

src/ticketing/use-fare-contracts.ts
```diff
@@ -1,14 +1,40 @@
...
+import {useEffect, useState} from 'react';
+import {getFareContracts} from '@atb/ticketing/api';
+import {useAuthContext} from '@atb/auth';
+
+type AvailabilityStatusInput = {
+  availability: Exclude<AvailabilityStatus['availability'], 'invalid'>;
+  status?: Exclude<AvailabilityStatus['status'], 'invalid'>;
+};
 
 export const useFareContracts = (
```
marius-at-atb, line 14: Is it worth adding a comment about why we have this solution?
I think it may not be obvious just from the code.

src/ticketing/api.ts
```diff
@@ -187,3 +192,14 @@ export async function getFareProducts(): Promise<PreassignedFareProduct[]> {
...
 }
+
+export async function getFareContracts(
+  availability: 'Available' | 'Historical',
+): Promise<FareContract[]> {
+  const url = `ticket/v4/list?availability=${availability}`;
+  const response = await client.get(url, {
+    authWithIdToken: true,
+  });
+
```
marius-at-atb, line 203: It would be nice to add a zod schema validation here.

src/ticket-history/TicketHistoryScreenComponent.tsx
```diff
@@ -26,13 +26,12 @@ export const TicketHistoryScreenComponent = ({
...
-    resubscribeFirestoreListeners,
   } = useTicketingContext();
   const {serverNow} = useTimeContext();
-  const {fareContracts: historicFareContracts} = useFareContracts(
-    {availability: 'historic'},
-    serverNow,
-  );
+  const {
+    fareContracts: historicFareContracts,
+    refetch: refetchHistoricFareContracts,
```
marius-at-atb, line 33: Very clear naming :+1:

marius-at-atb (approved):
Looks great!

src/utils/date.ts
```diff
@@ -602,3 +602,21 @@ export const isValidDateString = (dateString: string) => {
   const parsedDate = parseISO(dateString);
   return isValid(parsedDate);
 };
+
+/**
+ * Recursively transforms iso date strings fields to Date fields.
+ */
+export const convertIsoStringFieldsToDate = (value: any): any => {
```
jorelosorio, line 609: It seems like we need to traverse all the structure to find this and might be a bit consuming after serialization, is better if this can be done during the JSON conversion or know the structure we are working to then convert the `time format` when required.

src/ticketing/use-fare-contracts.ts
jorelosorio: maybe use a name convention to not to have to map historic as historical

src/ticketing/api.ts
gorandalum: Yeah, I think it can be an improvement for the app to add validation of api request responses. Maybe it should be its own item in the technical debt prioritisation?

As of now I won't hassle with it. It doesn't really give any benefit adding it here but not adding it on the Firestore subscription for fare contracts in `TicketingContext`, as this API endpoint only is a way to "manually" retrieve what is in Firestore.

src/ticketing/use-fare-contracts.ts
gorandalum: Nisha argumented that historical is slightly more fitting than historic, and I agree with her. I'll do renaming in our app in its own PR to not bloat this one.

ChatGPT also agrees:
> Both "historic" and "historical" can be used to describe something that is related to the past, but they have slightly different nuances.
>
> - "Historic" is typically used to refer to something that is significant or important in history. It often implies that the event or object had a notable impact or is noteworthy in some way.
> - "Historical" is more general and refers simply to something that is from or related to history, without implying significance.
>
>In the context of a ticket that was previously used, "historical ticket" would likely be the best fit, as it simply refers to a ticket that was used in the past without implying that it is historically significant.

src/utils/date.ts
gorandalum: Lets keep it in the back of our mind and think about optimising later if we find that it goes slow. Testing in simulator the mapping takes approximately 0.3ms per fare contract, and most phones are faster than the simulator. Also the mapping time is negligible when compared to the [slow rendering of long fare contract lists](https://github.com/AtB-AS/kundevendt/issues/19096).

src/ticketing/use-fare-contracts.ts
gorandalum: I added some jsdoc to the `useFareContracts`

src/utils/date.ts
jorelosorio: Great, this is nice to have in mind, might there be other places like this that all summed up might decrease performance.

PR #4914, opened 2025-01-10
feat: add bonus program button to map

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/19789
[Figma](https://www.figma.com/design/8aTgkQtYXm7l0QvGYSCdR0/Bonusprogram?node-id=962-50501&t=mibfIx0t388966M8-0)

Adds bonus program button to map when feature toggle is enabled.

image

### DIFF
--- a/src/components/map/components/bonus-program/BonusProgramMapButton.tsx
+++ b/src/components/map/components/bonus-program/BonusProgramMapButton.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import {Button} from '@atb/components/button';
+import {StyleSheet, useThemeContext} from '@atb/theme';
+import {useAnalyticsContext} from '@atb/analytics';
+import {FavoriteFill} from '@atb/assets/svg/mono-icons/places';
+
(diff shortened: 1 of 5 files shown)


src/components/map/components/bonus-program/BonusProgramMapButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type BonusProgramMapButtonProps = {
+  onPress: () => void;
+  isLoading: boolean;
+  bonusPoints: number;
+};
+export const BonusProgramMapButton = ({
+  onPress,
+  isLoading,
+  bonusPoints,
```
gorandalum, line 15: It is hard to know how the future will look, but I think plan A can be to keep as much internally in the component as possible. So if the button itself are able to retrieve the `bonusPoints` through useQuery or React.Context then the parent won't need to hassle with retrieving it.

gorandalum (approved):

strandlie (approved):

PR #4916, opened 2025-01-10
refactor: Use PurchaseSelection as route param

gorandalum:
This means that we no longer need to pass around multiple purchase
related fields when navigating between purchase screens.

I wasn't able to split this up more than this since as soon as the
route param was changed all of these places needed to be fixed.


strandlie (dismissed):

src/stacks-hierarchy/Root_PurchaseHarborSearchScreen/Root_PurchaseHarborSearchScreen.tsx
```diff
@@ -20,31 +20,35 @@ import {ScreenReaderAnnouncement} from '@atb/components/screen-reader-announceme
...
+export const Root_PurchaseHarborSearchScreen = ({
+  navigation,
+  route: {
+    params: {selection},
+  },
+}: Props) => {
   const {t} = useTranslation();
   const inputRef = useRef<InternalTextInput>(null);
   const [text, setText] = useState('');
+  const selectionBuilder = usePurchaseSelectionBuilder();
```
strandlie, line 36: I see that there is quite a lot of `selectionBuilder.fromSelection(selection)`, and the related passing of `selection` into components (along with `usePurchaseSelectionBuilder`), and I definitely see that as a good way to solve this problem. 

But just because I'm interested, did you consider pros and cons of placing this in a context instead (not necessarily part of "the arrow" but maybe placed in all Screens that use these components)? My gut feeling is that that could save a bit of code.

gorandalum:
> I see that there is quite a lot of `selectionBuilder.fromSelection(selection)`, and the related passing of `selection` into components (along with `usePurchaseSelectionBuilder`), and I definitely see that "reducer-style"-solution as a good way to solve this problem.
> 
> But just because I'm interested, did you consider pros and cons of placing this in a Context instead (not necessarily part of "the arrow" but maybe placed in all Screens that use these components)? My gut feeling is that that could save a bit of code.

Good question! I'm personally not as fond of the global state pattern, as I have been in redux codebases where there was a lot of these types of indirect dependencies between components. I rather like that the dependency is explicit when it can be, even though it gives a little more boilerplate code. But even though we had that pattern, the bulk of the boilerplate code (with `fromSelection`, `build`, etc.) would still need to be present.

tormoseng:
✅

PR #4917, opened 2025-01-10
fix: Adhere limitations in tariff zone selector

gorandalum:
When there is limitation on multiple tariff zones for a product, the
user is able to navigate to the tariff zone selection map. But in
that map all zones were actually shown, not just the one applicable
by the limitation.

### Acceptance criteria
- [ ] Only selectable tariff zones should be shown and pressable in the map
- [ ] Only selectable tariff zones should be listed in the empty search view
- [ ] Only stop places for selectable tariff zones should be returned searching

How to test:
- Either by proxy or by changing Firestore create a list of two or more (but not all) tariff zones in the `preassignedFareProduct.limitations.tariffZoneRefs`


rosvik (approved):

tormoseng:
✅  By editing the referenceData.

PR #4918, opened 2025-01-13
feat: Added GlobalMessage to TripSearchScreen

strandlie:
Closes AtB-AS/kundevendt#19786

There are no Figma sketches for this, but here is a screenshot:

Simulator Screenshot - iPhone 16 Pro - 2025-01-13 at 13 03 08

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
@@ -55,6 +55,7 @@ import {usePopOverContext} from '@atb/popover';
 import {areDefaultFiltersSelected} from './utils';
 import {useFeatureTogglesContext} from '@atb/feature-toggles';
 import {GlobalMessage, GlobalMessageContextEnum} from '@atb/global-messages';
+import {isDefined} from '@atb/utils/presence';
 
 type RootProps = DashboardScreenProps<'Dashboard_TripSearchScreen'>;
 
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
```diff
@@ -404,6 +405,18 @@ export const Dashboard_TripSearchScreen: React.FC<RootProps> = ({
               {t(TripSearchTexts.searchState.noResultReason.MissingLocation)}
             </ThemeText>
           )}
+          {!!tripPatterns.length && (
```
rosvik, line 408: Should this be shown based on loading state instead of length? We might want to show messages in cases where there are no results. Or can it be shown even if loading hasn't completed?

strandlie: Yes, good point. I thought `tripPatterns.length was a reasonable proxy for loading state, but you are right that there are cases where that is not the case. Will fix!

strandlie: For some reason, if I use `!isSearching` there is a delay after `tripPatterns` has elements which leads to a little uncomfortable experience. See the videos. The second video is using `tripPatterns.length`


https://github.com/user-attachments/assets/6cc2ccff-1276-4288-9295-a74e4b4f14a0

https://github.com/user-attachments/assets/da8d049d-3e2f-464a-85f4-ff7832da5581

gorandalum (commented):
Comments after pair-review:
- Lets just show the box regardless of searching state, as suggested by Johannes
- Use rule variables from the search results instead of search parameters, as using search parameters are a little fragile
- Suggested rule variables: transportmode/transportsubmode and authority. Should be safe showing message for AtB shuttle buses

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/utils.ts
```diff
@@ -102,7 +102,7 @@ function transportModeToEnum(
       transportMode: enumFromString(TransportMode, internal.transportMode),
       transportSubModes: internal.transportSubModes
         ?.map((submode) => enumFromString(TransportSubmode, submode))
-        .filter(Boolean) as TransportSubmode[],
+        .filter(isDefined),
```
strandlie, line 105: This is just a cast I noticed that could be replaced with something safer :)

rosvik (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/Dashboard_TripSearchScreen.tsx
```diff
@@ -404,6 +406,22 @@ export const Dashboard_TripSearchScreen: React.FC<RootProps> = ({
...
           )}
+          <View style={styles.globalMessage}>
+            <GlobalMessage
+              textColor={theme.color.background.neutral[0]}
+              globalMessageContext={GlobalMessageContextEnum.appTripResults}
+              ruleVariables={{
+                transportSubmodes: tripPatterns
+                  .flatMap((tp) => tp.legs)
+                  .map((leg) => leg.transportSubmode)
+                  .filter(isDefined),
```
gorandalum, line 417: Can also add:
- `transportModes` as a rule variable while we are at it. 
- `filter(onlyUniques)` in case we at some point want a message if more than x transportmodes or something like it.

strandlie: Done!

PR #4919, opened 2025-01-13
fix: Disable retries on response code 429

strandlie:
Closes AtB-AS/kundevendt#19693

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "@types/qrcode": "^1.5.5",
     "axios": "^1.7.3",
     "axios-better-stacktrace": "^2.1.7",
-    "axios-retry": "^4.5.0",
     "date-fns": "^3.6.0",
     "date-fns-tz": "^3.1.3",
     "emoji-datasource": "^15.1.2",
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -13,7 +13,6 @@ import {
   RequestIdHeaderName,
   Authorization,
 } from './headers';
-import axiosRetry, {isIdempotentRequestError} from 'axios-retry';
 import axiosBetterStacktrace from 'axios-better-stacktrace';
 import {Platform} from 'react-native';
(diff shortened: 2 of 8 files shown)


marius-at-atb (approved):
Lgtm.
Thorough testing recommended for this change 😃

gorandalum (commented):
Changes looks good! But I think we should at the same time disable retries in axios. I have observed that we some places retry 9 times (!) on failed requests. 3 retries from axios which are retried 3 times from use-query.

strandlie:
Added an issue for continuing the upgrade to React Query. 

As a consequence of this PR the following API-calls will not get retried untill they are upgraded to React Query:

- src/api/departures/index 
        ○ GET -> getRealtimeDepartures()
- src/api/departures/stops-nearest.ts
	○ GET -> requestNearestStops()
- src/api/geocoder
	○ GET -> autocomplete(), reverse()
- src/api/mobility -> 
	○ GET -> getVehicles(), getStations(), initViolationsReporting(), lookupVehicleByQr(), 
- src/api/profile
	○ DELETE -> deleteProfile()
- src/api/serviceJourney
	○ GET getServiceJourneyWithEstimatedCalls(), getServiceJourneyMapLegs()
- src/api/vipps-login
	○ GET authorizeUser()
- src/ticketing/api
	○ GET listRecentFareContracts()
        ○ PUT activateFareContractNow()

rosvik (approved):

gorandalum (approved):
Nice!

gorandalum:
@strandlie axios-retry also needs to be removed from yarn.lock

strandlie:
> > @strandlie axios-retry also needs to be removed from yarn.lock
> 
> 
> 
> That's fixed in https://github.com/AtB-AS/mittatb-app/pull/4941

Nice catch! 

I think we should look into master builds failing when package.json and package-lock don't match in the future. They should always be in sync on master.

PR #4922, opened 2025-01-14
fix: Device update issue, missing keychain name

jorelosorio:
It fixes issues while running update devices due to a missing parameter in the GA.

### DIFF
--- a/scripts/ios/get_ios_certs.sh
+++ b/scripts/ios/get_ios_certs.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+export KEYCHAIN_NAME=atb
 bundle exec fastlane ios get_certs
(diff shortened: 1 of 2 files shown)


scripts/ios/get_ios_certs.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+export KEYCHAIN_NAME=atb
```
reidzeibel, line 3: How does other OMS use this value @jorelosorio ?

jorelosorio: this command is in the dev machine for this case, but for the github actions, it uses CI as a name.

reidzeibel (approved):

scripts/ios/get_ios_certs.sh
gorandalum: What about when running up dev app for other transport companies? Does it still work with atb as keychain name?

PR #4926, opened 2025-01-14
fix: Use stopPositionInPattern instead of quay id

gorandalum:
Quay id doesn't uniquely identify a call in a service journey when
the service journeys visits the same quay twice. This change starts
using stopPositionInPattern to identify uniquely a call in a service
journey, as advices by Entur.

### DIFF
--- a/src/travel-aid/components/StopSignalButton.tsx
+++ b/src/travel-aid/components/StopSignalButton.tsx
@@ -20,12 +20,12 @@ import {useAnalyticsContext} from '@atb/analytics';
 
 export const StopSignalButton = ({
   serviceJourney,
-  fromQuayId,
+  fromStopPosition,
   onPress,
   status,
 }: {
   serviceJourney: ServiceJourneyWithGuaranteedCalls;
-  fromQuayId?: string;
+  fromStopPosition: number;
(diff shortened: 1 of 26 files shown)


src/travel-aid/components/StopSignalButton.tsx
```diff
@@ -38,9 +38,7 @@ export const StopSignalButton = ({
 
   if (!isTravelAidStopButtonEnabled) return null;
 
-  const selectedCall = serviceJourney.estimatedCalls.reduce((selected, cur) =>
-    cur.quay.id === fromQuayId ? cur : selected,
-  );
+  const selectedCall = serviceJourney.estimatedCalls[fromStopPosition];
```
rosvik, line 41: Is this always correct, or could we end up with a list of estimated calls that doesn't start at 0? To me, it sounds like the more robust solution would be something like

```ts
serviceJourney.estimatedCalls.find(ec => ec.stopPositionInPattern === fromStopPosition)
```

Or am I missing something?

gorandalum: I think it is always correct, but I'll just change to using `find` regardless to be safer :)

PR #4932, opened 2025-01-17
feat: added an icon for the night bus

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/19315
image
image
image


reidzeibel:
@Oleninkaa there's an issue on the code, could you check this one? https://github.com/AtB-AS/mittatb-app/actions/runs/12825257245/job/35762884591?pr=4932

Also there's a conflict on the `package.json` file, let me know if you are not sure on how to resolve the conflict :)

src/components/icon-box/utils.ts
```diff
@@ -34,7 +37,10 @@ export function getTransportModeSvg(mode?: AnyMode, subMode?: AnySubMode) {
       if (APP_ORG === 'atb' && subMode === 'shuttleBus') {
         return {svg: SvgShuttle, name: 'Shuttle'};
       }
-      return {svg: Bus, name: 'Bus'};
+
+      return subMode && TRANSPORT_SUB_MODES_BUS.includes(subMode)
```
rosvik, line 41: Don't think we need the TRANSPORT_SUB_MODES_BUS array here when it's only the one mode. A `if (submode === TransportSubmode.NightBus)` check should be good enough.

Also noticed I should have used `TransportSubmode.ShuttleBus` in the check above instead of `'shuttleBus'`, so feel free to update that if you want 😄

src/translations/screens/FareContract.ts
```diff
@@ -334,6 +337,9 @@ const FareContractTexts = {
     switch (mode) {
       case TransportMode.Bus:
       case TransportMode.Coach:
+        if (subMode === TransportSubmode.NightBus) {
```
reidzeibel, line 340: @Oleninkaa I think we can remove this, right? since we no longer use this `FareContractTexts`.

rosvik (approved):
Nice 🙌

Oleninkaa:
My first commit to the master 👶🏻🙌🏻

PR #4933, opened 2025-01-17
feat: add mapbox_sprite_url remote config variable

rosvik:
Excuse the amount of changes - I did some cleanup in remote-config.ts and sorted stuff alphabetically. The only real change is that `mapbox_sprite_url` was added (3009c14646d5ba60321b1533e592d65a37e3c8ab).

### DIFF
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -22,8 +22,8 @@ export type RemoteConfig = {
   enable_city_bikes_in_map: boolean;
   enable_extended_onboarding: boolean;
   enable_flexible_transport: boolean;
-  enable_from_travel_search_to_ticket: boolean;
   enable_from_travel_search_to_ticket_boat: boolean;
+  enable_from_travel_search_to_ticket: boolean;
   enable_geofencing_zones: boolean;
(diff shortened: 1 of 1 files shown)


src/remote-config.ts
```diff
@@ -103,32 +105,33 @@ export const defaultRemoteConfig: RemoteConfig = {
...
-  enable_shmo_deep_integration: false,
-  enable_only_stop_places_checkbox: false,
   favourite_departures_poll_interval: 30000,
   feedback_questions: '',
+  fetch_id_token_retry_count: 3,
   flex_booking_number_of_days_available: 7,
   flex_ticket_url: '',
   live_vehicle_stale_threshold: 15,
   loading_screen_delay_ms: 200,
+  mapbox_sprite_url: '',
```
rosvik, line 134: @marius-at-atb Is an empty string a good default here, or will that break things in mapbox?

marius-at-atb: It's alright. All that happens is that the icons don't show.

marius-at-atb (approved):
:+1:

PR #4936, opened 2025-01-17
feat: scooter help screen

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/19831

Adds scooter help screen, accessible from Help-button in ShmoTesting.

Uses data from firestore for scooterFaqs.

Design from [figma](https://www.figma.com/design/KFBSvkxSDhHNM8EBFELvod/Shared-Mobility?node-id=4987-40913&t=vHO00rWgUltluSNR-0)

NB! Numbered lists are not yet handled by markdown-renderer, so lists are displayed as un-numbered lists This will be solved in separate PR.


src/translations/screens/ScooterHelp.ts
```diff
@@ -0,0 +1,28 @@
...
+    'Kontakt og rapporter',
+    'Contact and report',
+    'Kontakt og rapporter',
+  ),
+  contactOperator: (operatorName: string) => {
+    return _(
+      `Kontakt ${operatorName}`,
+      `Contact ${operatorName}`,
+      `Kontakt ${operatorName}`,
+    );
```
marius-at-atb, line 15: Perhaps we should use `Operatør`/`Operator`/`Operatør` as fallback in case operatorName is an empty string?

src/stacks-hierarchy/Root_ParkingViolationsReporting/Root_ScooterHelpScreen.tsx
```diff
@@ -0,0 +1,80 @@
...
+export type ScooterHelpScreenProps =
+  RootStackScreenProps<'Root_ScooterHelpScreen'>;
+
+export const Root_ScooterHelpScreen = ({route}: ScooterHelpScreenProps) => {
+  const {vehicleId} = route.params;
+  const style = useStyles();
+  const {t, language} = useTranslation();
+  const navigation = useNavigation<RootNavigationProps>();
+  const {scooterFaqs} = useFirestoreConfigurationContext();
+  const [currentlyOpen, setCurrentlyOpen] = useState<number>();
```
marius-at-atb, line 28: `currentlyOpen` sounds like a boolean. Can you use a name that makes it clear that it is an index?

marius-at-atb (approved):
Looks really good!

PR #4937, opened 2025-01-17
feat: Frontend controlled map style

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/19557
Closes https://github.com/AtB-AS/kundevendt/issues/10821
Closes https://github.com/AtB-AS/kundevendt/issues/19451
Closes https://github.com/AtB-AS/kundevendt/issues/19259

First of all excuse the large PR, but it is kinda needed to avoid incomplete/invalid steps along the way 😅 

Feel free to ask me for closer explanations if needed!

What's happening in this PR:

- The `Map` component has been updated. That means only the map tab and choose location maps are updated for now.
- New hook: `useMapboxJsonStyle`, which uses the files inside `mapbox-styles` to get the style locally, but sprites from the url defined in remote config. This only shows geographical features, not NSR items.
- New component: `NationalStopRegistryFeatures`. This is where the NSR items are shown now. In order to access the data, new mapbox env variables have been added.
- New hooks for determining the style for icons and text for NSR and Shared Mobility items.
- No longer zooming in on a path when selecting a feature.
- Mobility components have been updated to use the new icons. These components will be removed when using the tile server, so let's not spend too much time on them..
- Dark mode support.

How the zoom logic works:
- In nsr-layers.ts, showAsDefaultAtZoomLevel is defined for each NSR layer. 3 zoom levels before, the minimized version will transition from small and invisible, to full size and visible. Then when `showAsDefaultAtZoomLevel` is reached, the icon switches to default. The label will then start to fade in as well.

Before:

After:
<section>

</section>

Acceptance Criteria:
- [ ] Light theme in light mode
- [ ] Dark theme in dark mode
- [ ] 200% zoom
- [ ] Performance - it should not be much slower than before
- [ ] When selecting a map item, it should
      - [ ] show a custom pin to indicate the selected element
      - [ ] other items should become minimized (apart from scooter clusters, that will be addressed later)
- [ ] When zooming in, the stop places should
      - [ ] first be invisible and tiny in the minimized state
      - [ ] then transition to reach full size and have no transparency in the minimized state as you zoom in
      - [ ] switch to the default icon
      - [ ] the label should then start becoming visible
      - [ ] The quays should start small, and when full size is reached, the text should start fading in
      - NOTE: different types of stop places are configured to reach these states at different zoom levels
- [ ] Vehicles and stations should still work with clustering and show correct numbers (navigate to Oslo in the map to test this)


src/components/map/Map.tsx
```diff
@@ -184,6 +191,15 @@ export const Map = (props: MapProps) => {
     async (feature: Feature) => {
       if (!isFeaturePoint(feature)) return;
 
+      // should split components instead of this, ExploreLocation should only depend on location state, not features
```
marius-at-atb, line 194: Upcoming PR..

src/components/map/Map.tsx
```diff
@@ -252,7 +269,12 @@ export const Map = (props: MapProps) => {
           onMapIdle={onMapIdle}
           onPress={onFeatureClick}
           testID="mapView"
-          {...MapViewConfig}
+          {...{
+            ...MapViewConfig,
+            // only updating Map.tsx for now.
```
marius-at-atb, line 274: Upcoming PR for the other maps

src/components/map/Map.tsx
```diff
@@ -275,12 +297,22 @@ export const Map = (props: MapProps) => {
           )}
 
           {mapLines && <MapRoute lines={mapLines} />}
+
+          <NationalStopRegistryFeatures
+            selectedFeaturePropertyId={selectedFeature?.properties?.id}
+          />
+
+          {props.selectionMode !== 'ExploreLocation' && (
+            <SelectedFeatureIcon selectedFeature={selectedFeature} />
```
marius-at-atb, line 306: Keeping SelectedFeatureIcon as a separate icon. When using the tile server (upcoming PR), this is needed to keep a visible item selected when zooming out.

src/components/map/components/mobility/Bicycles.tsx
```diff
@@ -1,91 +1,102 @@
 import React, {RefObject, useRef} from 'react';
```
marius-at-atb, line 1: All files inside `mobility/` will be deleted when switching to use the tile server, so no need to perfect these files.

src/components/map/mapbox-styles/pin-types.ts
```diff
@@ -0,0 +1,49 @@
+import {Theme, Themes} from '@atb-as/theme';
```
marius-at-atb, line 1: Note: Types and names of the icons will be iterated on in a follow-up PR.

src/components/map/Map.tsx
```diff
@@ -162,14 +166,17 @@ export const Map = (props: MapProps) => {
    * overhead.
    */
   const onMapIdle = (state: MapState) => {
+    console.log('onMapIdle');
```
jorelosorio, line 169: There are some console logs here that print some warnings.

src/components/map/Map.tsx
```diff
@@ -162,14 +166,17 @@ export const Map = (props: MapProps) => {
...
     const newMapRegion: MapRegion = {
       visibleBounds: [state.properties.bounds.ne, state.properties.bounds.sw],
       zoomLevel: state.properties.zoom,
       center: state.properties.center,
     };
-    setMapRegion((prevMapRegion) =>
-      isEqual(prevMapRegion, newMapRegion) ? prevMapRegion : newMapRegion,
-    );
+    if (!isEqual(mapRegion, newMapRegion)) {
```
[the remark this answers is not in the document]
jorelosorio: Maybe a stateless var instead. But perhaps is needed to be in that way.

src/components/map/components/SelectedFeatureIcon.tsx
```diff
@@ -0,0 +1,78 @@
...
+import {hitboxCoveringIconOnly, useMapSymbolStyles} from '@atb/components/map';
+import {Expression} from '@rnmapbox/maps/src/utils/MapboxStyles';
+import {PinType} from '../mapbox-styles/pin-types';
+
+export const SelectedFeatureIcon = ({
+  selectedFeature,
+}: {
+  selectedFeature?: Feature<Point, GeoJsonProperties>;
+}) => {
+  const pinType: PinType =
```
jorelosorio, line 13: I suggest moving this ternary to a switch can be easier to read.

src/components/map/components/SelectedFeatureIcon.tsx
```diff
@@ -0,0 +1,78 @@
...
+    selectedFeature?.properties?.entityType === 'Quay'
+      ? 'stop'
+      : selectedFeature?.properties?.vehicleTypesAvailable !== undefined // use is_virtual_station instead when using vector source
+      ? 'station'
+      : 'vehicle';
+
+  const {iconStyle, textStyle} = useMapSymbolStyles(
+    selectedFeature?.properties?.id,
+    pinType,
+    1.61, // increased text size since the icon is larger
```
jorelosorio, line 25: How does it work if font size increases?

src/components/map/components/national-stop-registry-features/nsr-utils.ts
```diff
@@ -0,0 +1,163 @@
...
+          opacityTransitionZoomRange,
+        1,
+      ],
+
+      textSize: [
+        'interpolate',
+        ['exponential', 1.5],
+        ['zoom'],
+        14,
+        textSizeFactor * 13,
```
jorelosorio, line 139: Is not very intuitive why to multiply by 13 here or 27 below, would be nice to add some comments about it.

src/components/map/components/national-stop-registry-features/use-nsr-circle-layers.ts
```diff
@@ -0,0 +1,49 @@
...
+      ...nsrLayerSourceProps,
+      id,
+      filter,
+      minZoomLevel,
+      style: {
+        circleRadius,
+        circleOpacity,
+
+        circleColor: theme.color.transport.city.primary.background,
+        circleStrokeColor: themeName === 'light' ? '#ffffff' : '#000000',
```
jorelosorio, line 43: Can this be set in the theme?

src/components/map/components/national-stop-registry-features/use-nsr-symbol-layers.ts
```diff
@@ -0,0 +1,113 @@
...
+      ? {}
+      : {
+          textSize,
+          textOpacity,
+          textField,
+          textColor,
+          textHaloColor,
+          textVariableAnchor,
+          textAnchor: 'top',
+          textFont: ['DIN Offc Pro Regular', 'Arial Unicode MS Regular'],
```
jorelosorio, line 92: Not sure if those are default fonts, we might use defaults ones.

src/components/map/hooks/use-map-symbol-styles.ts
```diff
@@ -0,0 +1,185 @@
...
+  ];
+
+  const systemId: Expression = ['get', 'system_id'];
+  const transportOperator: Expression = [
+    'case',
+    ['!', ['has', 'system_id']],
+    'generic',
+    [
+      'case',
+      ['==', ['slice', systemId, 0, 4], 'ryde'],
```
jorelosorio, line 89: can those operators be taken for a dynamic list of operators, maybe later we want to turn some of them off.

jorelosorio:
In general, the code looks good, nice work. There are some small comments here and there, in general, I see that there are a lot of values, offsets and fonts set in the map, some of them look a bit random without more context, and it might be difficult to change to another way but would be nice to have some comments, in general, the documentation of the code looks good.

src/components/map/Map.tsx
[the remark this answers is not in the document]
marius-at-atb: I like this suggestion!
It is possible to change the mapRegion state to a ref instead, and update vehicles and stations directly instead of in a useEffect:

```ts
  const mapRegionRef = useRef<MapRegion>();
  const updateVehiclesAndStations = useCallback(
    (mapRegion: MapRegion) => {
      if (!mapRegion) return;
      mapRegionRef.current = mapRegion;
      updateRegionForVehicles?.(mapRegion);
      updateRegionForStations?.(mapRegion);
    },
    [updateRegionForVehicles, updateRegionForStations],
  );
```
and then instead of setMapRegion:
```ts
    if (!isEqual(mapRegionRef.current, newMapRegion)) {
      updateVehiclesAndStations(newMapRegion);
    }
```
Also for `onDidFinishLoadingMap`:
```ts
    updateVehiclesAndStations({
      visibleBounds,
      zoomLevel,
      center,
    });
```

Seems to work well when testing, so I actually prefer this.
However it is another change, so not sure if it should be included in this PR 🤔

gorandalum (commented):
In general I think this looks good! The map code was pretty complex from before, but from what I can see there isn't too much increase in complexity from this. 

Also the map looks pretty nice, and I couldn't find any glaring bugs when trying it out. One thing I noticed was that the mobility bottom sheet was all the way up to the icon if the max height was used. The bus and mobility bottom sheets should probably have the same max height, and the icon placement should be set based on the shared max height.

When removing a feature (like the map lines) it would be nice with a reference to the decision and background/insight for it. When we did insight into the map originally the user tests showed that it was a requested feature. The reason I remember it is because it was pretty complex to implement with the correct zooming animation, and I tried to convince the designers at the time to skip it for the initial map release, with no luck 😅 

All in all a thumbs up from me! 👍 I won't approve it since my review was pretty rough, and it should still be pair-reviewed by you and Johannes.

tormoseng:
Observations (listing as I go):
1. Seems like the speed boat icon hit range is to wide in some cases. In this example it's not possible to open the train or bus station (screenshot_1)
2. When visible the bus icon should be "prioritized" on top of other transport icons (screenshot_2)

<details>
<summary>screenshot_1</summary>

IMG_5341

</details>

<details>
<summary>screenshot_2</summary>

IMG_5340

</details>

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -210,7 +210,11 @@ export const TravelDetailsMapScreenComponent = ({
           />
         )}
         {stations && (
-          <Stations stations={stations.stations} mapCameraRef={mapCameraRef} />
+          <Stations
+            selectedFeaturePropertyId="" // no active selection feature used here
```
rosvik, line 214: Should this prop be optional?

src/components/map/utils.ts
```diff
@@ -223,8 +226,8 @@ export const getVisibleRange = (visibleBounds: Position[]) => {
...
 
-export const shouldShowMapLines = (entityFeature: Feature<Point>) =>
-  isStation(entityFeature) || isStopPlace(entityFeature);
+// disable for now
+export const shouldShowMapLines = (_entityFeature: Feature<Point>) => false; //isStationFeature(entityFeature) || isStopPlaceFeature(entityFeature);
 
-export const shouldZoomToFeature = (entityFeature: Feature<Point>) =>
-  isStation(entityFeature) || isStopPlace(entityFeature);
+// disable for now
+export const shouldZoomToFeature = (_entityFeature: Feature<Point>) => false; //isStationFeature(entityFeature) || isStopPlaceFeature(entityFeature);
```
rosvik, line 233: I assume this will change soon™, but I think it's better to remove this and related logic to keep the code clean on master.

src/components/map/mapbox-styles/README.md
rosvik, line 1: Nice 🙌

src/components/map/components/SelectedFeatureIcon.tsx
marius-at-atb: This number multiplies the text size. This is needed here since the selected version of the icons is larger.

src/components/map/components/national-stop-registry-features/use-nsr-symbol-layers.ts
marius-at-atb: These are the same as in the map style currently in prod, so it will be the same as before.

src/components/map/mapbox-styles/pin-types.ts
```diff
@@ -0,0 +1,49 @@
...
+  | 'citybike'
+  | 'citybus'
+  | 'scooter'
+  | 'sharedcar'
+  | 'train'
+  | 'tram';
+
+export type StationIconCode = 'sharedcar' | 'citybike';
+
+export type PinScooterCompany = 'generic' | 'voi' | 'ryde' | 'tier'; // todo: change 'tier' to 'dott' when sprite ready
```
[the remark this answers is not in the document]
marius-at-atb: This is used in order to show correct colors per company. I suppose we could use color names instead of company names.

src/components/map/utils.ts
marius-at-atb: In case it ever becomes relevant to re-introduce something similar, I think it is nice to have an isolated PR for it.
Also this PR already has a lot going on, so I'll make an immediate follow-up PR with this :+1:

src/components/map/components/national-stop-registry-features/use-nsr-circle-layers.ts
marius-at-atb: We don't have shadow specs in design system from before. I guess we should at some point though.

src/components/map/components/SelectedFeatureIcon.tsx
marius-at-atb: https://github.com/AtB-AS/mittatb-app/pull/4937/commits/3c99ebe42a65f90397ed234e8e02335d733c678e

src/components/map/components/national-stop-registry-features/nsr-utils.ts
marius-at-atb: Comment added: https://github.com/AtB-AS/mittatb-app/pull/4937/commits/2a0c6a034d6e40b151bda09fb628c1c5b651d6cd

marius-at-atb:
@rosvik Good catch with the status bar color! Updated in https://github.com/AtB-AS/mittatb-app/pull/4937/commits/af01eebf04d948f0f650febf4e3907690571043a. Let me know if I removed too much here.

marius-at-atb:
@tormoseng 
1. Will look into this a bit.
2. Updated [here](https://github.com/AtB-AS/mittatb-app/pull/4937/commits/e1b2f6df318979bd94c7904420fceccfda310b11) so that the last listed item in nsrSymbolLayers is on top. Not 100% sure, but I think this is now the same order as in prod from before.
3. We can look into this.
4 and 5: Not a part of changes in this PR I think, but should be looked at :+1:
6: This has been addressed with onPress events directly on sources so that we don't need to wait for the querying of features where you clicked, and with useCallbacks/useMemos to minimize the number of renders of the Map component. However, it is still a bit slower than before because it supports more features. If still too slow, we probably go to the tile server directly.

src/components/map/Map.tsx
marius-at-atb: https://github.com/AtB-AS/mittatb-app/pull/5076

src/components/map/Map.tsx
marius-at-atb: https://github.com/AtB-AS/mittatb-app/pull/5077

src/components/map/Map.tsx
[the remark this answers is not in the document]
marius-at-atb: Update: MapV2 using VectorSource instead now, no longer using onMapIdle/mapRegion.

ios/Podfile.lock
```diff
@@ -3434,13 +3434,13 @@ SPEC CHECKSUMS:
   AbtCommoniOSLib: 244899edeaa2ec828566119d6d00c2cf7a93c8c1
   AbtProtobufTokenIosStubs: 76426ea8c7c8cf76a750b04bcd4be8288bafae70
   ApiLogiOSLib: bda60ece3f53fcbc6d1cfcbba4c0993057e6c2a7
-  BarcodeiOSLib: db5edae8312bf90e30673fbfe00832e2e3ea2315
+  BarcodeiOSLib: 45c0bb7f5f709c1c4147db333ee532169475e464
```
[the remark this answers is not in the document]
rosvik: Probably not. Doesn't look like the dependencies actually has changed, but rather an effect of different xcode versions.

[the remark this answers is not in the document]
marius-at-atb: Reverted

reidzeibel (approved):
I think with the feature toggle, this PR can be merged as-is, and we can fix the issues and some nitpicks on future PRs.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -33,15 +32,15 @@ import {DirectionArrow} from './components/DirectionArrow';
...
+
 import {useIsFocusedAndActive} from '@atb/utils/use-is-focused-and-active';
 import {
   MapState,
   RegionPayload,
 } from '@rnmapbox/maps/lib/typescript/src/components/MapView';
 
-const EMPTY_FILTER = {};
+// todo: re-add this
+// import {VehiclesAndStations} from '@atb/components/map/components/mobility/VehiclesAndStations';
```
rosvik, line 43:
```suggestion
```

yarn.lock
rosvik, line 1: Maybe fix the line break, so that this won't trigger uncached builds.

rosvik (approved):

tormoseng:
✅  Looking good! Will continue test the map, and note any observations in https://github.com/AtB-AS/kundevendt/issues/20260

PR #4939, opened 2025-01-17
fix: intecom update token metadata not working as intended

reidzeibel:
This fixes the issue with intercom update metadata logging, right now, the metadata for token is only updated when the user changes (login/logout), while what we want is to update it everytime the token status gets updated (renewal, error, creation, etc.).

In general:
- `updateMetadata()` is now a part of `MobileTokenContext`, with similar trigger to the ones in `GeolocationContext`.
- load native token query no longer calls the `updateMetadata()`.
- added a type for intercom token metadata (`attested | non-attested | error`).


rosvik (commented):
LGTM 👍

PR #4940, opened 2025-01-20
feat: Border radius circle on buttons

gorandalum:
### Background
Design has decided to make buttons in the app fully rounded. As this change is really quick and pretty risk free (as there are no functional changes), it is being done as a "quick issue".

This issue fixes point 1 and 2 described here:
https://github.com/AtB-AS/kundevendt/issues/19773

### Solution
Just set button border radius to circle.

Some pictures from the app after this change:
Screenshot 2025-01-20 at 11 08 04
Screenshot 2025-01-20 at 11 08 13
Screenshot 2025-01-20 at 11 08 25
Screenshot 2025-01-20 at 11 08 31
Screenshot 2025-01-20 at 11 08 42

### Acceptance criteria
- [ ] Buttons around the app are rounded

PR #4941, opened 2025-01-20
chore: remove some unused code and dependencies

rosvik:
Ran `yarn unimported` and fixed the most significant issues:

- Removed the Slider component and getRecommendedTicket, which are remnants of the ticket assistant
- Removed react-native-webview, which is no longer in use after being replaced by the in app browser
- Removed the information screen component, which isn't used after replacing it with links to atb.no

Then ran `npx yarn-deduplicate yarn.lock` to clean up yarn.lock.

### Acceptance criteria

- [ ] This PR has some general low risk changes to dependencies. If the build passes and the app doesn't crash, we should be good.


hanne-at-atb:
I clicked through quite a few screens when testing rounded buttons (https://github.com/AtB-AS/mittatb-app/pull/4940) and I did not experience any app crashes or unexpected behaviour. So I guess this is good along with letting it live for some days on master before next release.
✅

PR #4942, opened 2025-01-20
chore: Deep link pop up causes payment card to not be saved on android

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/19403


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -49,20 +56,24 @@ export const Profile_PaymentMethodsScreen = () => {
...
-  } = useAuthorizeRecurringPaymentMutation();
-
   const {
     mutateAsync: cancelRecurringPayment,
     isError: cancelRecurringPaymentError,
   } = useCancelRecurringPaymentMutation();
 
+  const [recurringPaymentId, setRecurringPaymentId] = useState<number>();
+
+  useOnRecurringPaymentReceived({
```
rosvik, line 66:
```suggestion
  // In cases where the recurring payment appears in firestore before the
  // callback is called, we can cancel the payment flow and reload the recurring
  // payment list.
  useOnRecurringPaymentReceived({
```

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +247,62 @@ const GenericError = () => {
...
+          event.addMetadata('payment', {userId});
+        });
+      },
+    );
+
+  const recurringPaymentsReceived = useMemo(() => {
+    return recurringPayments.some(
+      (rp) => rp.id === recurringPaymentId && recurringPaymentId !== undefined,
+    );
+  }, [recurringPayments, recurringPaymentId]);
```
rosvik, line 295: I assume this code is fast enough that we don't need the useMemo

rosvik (commented):
This looks like a nice solution to me! Some minor suggestions

jorelosorio:
It might be that some minimal extra changes are needed after the backed is done with the saving in the firebase.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +250,61 @@ const GenericError = () => {
...
+  ): RecurringPayment => {
+    return {
+      id: recurringPayment.id,
+      expires_at: recurringPayment.expires_at,
+      masked_pan: recurringPayment.masked_pan,
+      payment_type: recurringPayment.payment_type,
+    };
+  };
+
+  const recurringPaymentsUnsub = firestore()
```
gorandalum, line 276: Doesn't this set ut the subscription on every render? 🤔

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +250,61 @@ const GenericError = () => {
...
+      recurringPayments.some(
+        (rp) =>
+          rp.id === recurringPaymentId && recurringPaymentId !== undefined,
+      )
+    ) {
+      callback();
+    }
+
+    return recurringPaymentsUnsub;
+  }, [callback, recurringPaymentId, recurringPayments, recurringPaymentsUnsub]);
```
gorandalum, line 305: Do we need this `useEffect` and the accompanying `useState` for the recurring payments? Or could we just invoke `callback` directly in the `onSnapshot` handler?

gorandalum (changes requested):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
jorelosorio: Right! I made a fix to handle that better.

rosvik (approved):
This looks good to me 👍 Just two minor nitpicks

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +250,58 @@ const GenericError = () => {
...
+  const mapRecurringPayment = (
+    recurringPayment: FirebaseFirestoreTypes.DocumentData,
+  ): RecurringPayment => {
+    return {
+      id: recurringPayment.id,
+      expires_at: recurringPayment.expires_at,
+      masked_pan: recurringPayment.masked_pan,
+      payment_type: recurringPayment.payment_type,
+    };
+  };
```
gorandalum, line 271: Do we need to map the recurring payment? Seems a little strange to map the `expires_at`, `masked_pan`, etc., just for looking up the id.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +250,58 @@ const GenericError = () => {
...
+      .orderBy('created', 'desc')
+      .onSnapshot(
+        (snapshot) => {
+          const recurringPayments =
+            snapshot.docs.map<RecurringPayment>(mapRecurringPayment);
+          if (
+            recurringPayments.some(
+              (rp) =>
+                rp.id === recurringPaymentId &&
+                recurringPaymentId !== undefined,
```
gorandalum, line 287: When `recurringPaymentId` is undefined we don't need this subscription at all? As of now it subscribes and iterates through all elements just to always evaluate to `false` for every item and not run the `callback`.

Rather check whether `recurringPaymentId` is defined at the top of the effect and do an early return if it is not.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +250,58 @@ const GenericError = () => {
...
+    return {
+      id: recurringPayment.id,
+      expires_at: recurringPayment.expires_at,
+      masked_pan: recurringPayment.masked_pan,
+      payment_type: recurringPayment.payment_type,
+    };
+  };
+
+  useEffect(() => {
+    const recurringPaymentsUnsub = firestore()
```
gorandalum, line 274: Have you checked how often this effect is run? I suspect it can be more than we want since the `callback` function isn't a stable reference.

gorandalum (changes requested):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
jorelosorio: The only reason to have it here a useEffect is to the unsubscribe here the firebase listener, because if it is sent in a return, any better idea to do not use useeffect specially because of the callback function?

gorandalum: We need the `useEffect`. Rather it is the `callback` function that needs to be a stable reference somehow. Maybe either through `useCallback`, or use the `useRef` pattern we have in utils like `useInterval` and `useDoOnceWhen`.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -49,40 +56,48 @@ export const Profile_PaymentMethodsScreen = () => {
...
       const responseCode = queryString.parseUrl(url).query.response_code;
       const paymentId = Number(
         queryString.parseUrl(url).query.recurring_payment_id,
       );
       if (responseCode === 'OK') {
-        authorizeRecurringPayment(paymentId);
+        await refetchRecurringPayment();
       } else if (responseCode === 'Cancel') {
-        cancelRecurringPayment(paymentId);
+        await cancelRecurringPayment(paymentId);
```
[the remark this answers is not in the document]
jorelosorio: It was added because of my editor pointed out `Missing await for an async function call `

gorandalum (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
```diff
@@ -235,6 +251,55 @@ const GenericError = () => {
...
+  ): number => {
+    const recurringPayment = d.data();
+    if (!recurringPayment) {
+      throw new Error('No recurring payment data');
+    }
+
+    return recurringPayment.id;
+  };
+
+  const callbackRef = useCallback(() => callback(), [callback]);
```
gorandalum, line 274: The naming is misleading since it is not a `ref`. And also the use of `useCallback` here doesn't achieve anything.

Take a look at the [`useCallback` doc](https://react.dev/reference/react/useCallback) to understand how it works and where it should be placed to achieve what we want.

gorandalum (changes requested):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_PaymentMethodsScreen.tsx
jorelosorio: I think is best that you show here an example of how do you see it, more than I guessing what is the think you want to achieve, the useRef I used before I think works as it should.

gorandalum: From the `useCallback` [doc](https://react.dev/reference/react/useCallback):
> On next renders, React will give you the same function again if the dependencies have not changed since the last render.

The `callback` isn't a stable reference as it is recreated on every render. And from the doc quoted above since the dependencies change on every render, that means the function returned by `useCallback` will also change on every render. This means using `useCallback` here has no effect.

jorelosorio: Yes, that is clear, but I'm not quite sure what your suggestion is for the code, because the callback does not have a parameter nor uses anything/dependencies, so we have a plain object callback that might not be accurate when it is in a dependency for useEffect, useCallback, you point out those things, that is fine.

So I guess your are suggesting then something like this:

```
const callbackRef = useRef(callback);
  useEffect(() => {
    callbackRef.current = callback;
 }, [callback]);
```

I have tried different ways now. So that is why I asked if you show here the code how do you see it then we can avoid to write here too much.

gorandalum: Can make the `callback` function it self a stable reference by wrapping the creation of it with `useCallback`.

This is how the `callback` function is created now:
```ts
callback: async () => {
  closeInAppBrowser();
  await refetchRecurringPayment();
},
```

jorelosorio: Ok, I have tried that ;)

gorandalum (approved):

jorelosorio:
I confirmed with @Nisha1306 & @mjenssen about the path in Firebase, so I am going to merge it to test it properly.

PR #4943, opened 2025-01-20
feat: show toEstimatedCall notices in trip search

rosvik:
Dependent on https://github.com/AtB-AS/atb-bff/pull/371

- Refactors TripsQuery to use fragments [in the same way as the BFF does](https://github.com/AtB-AS/atb-bff/blob/main/src/service/impl/fragments/journey-gql/trips.graphql-gen.ts).
- Adds notices to toEstimatedCall in trip search (31731ad422d6d8a16704869c2e745b8dc37f4173), which closes https://github.com/AtB-AS/kundevendt/issues/19816

Before/after


gorandalum:
Hard from just changed files to see if:
- Does this also fix it for departure details screen?
- Does this also fix it for the info message icon on trip search results card?

rosvik:
> Does this also fix it for departure details screen?

No, the thought is that we'll look more into this at a later time (see proposed solution in https://github.com/AtB-AS/kundevendt/issues/19816). I'm not sure what the correct solution is, but maybe you have a suggestion. Today only show notices on departure details for the focused estimated call. See [getNoticesForServiceJourney](https://github.com/AtB-AS/mittatb-app/blob/2e785b9c710d0f78363979c21bf346c184bdb316/src/travel-details-screens/utils.ts#L67). It might instead be an idea to show it on every estimated call, like this:



But that will lead to a lot of duplicate information, which might not be very useful for many/most trips. 

With this PR as is, I believe we show notices in the same way as Entur does.

> Does this also fix it for the info message icon on trip search results card?

Yes!

gorandalum:
> I'm not sure what the correct solution is, but maybe you have a suggestion. Today only show notices on departure details for the focused estimated call.

Do the duplicate notices have the same `id`? If so we can probably use the same logic as for `situations`. Take a look at `getSituationsToShowForCall`.

rosvik:
We could probably show the message once at the top, but is that the best solution in this case? The message applies to the three middle stops in this case, but sometimes only one of the stops. See stops marked with "d" here: https://svipper.no/_f/i812f2663-b1ec-436e-be15-8db3dca2fb2b/linje-3-tromso-skjervoy.pdf

PR #4949, opened 2025-01-21
feat: Remove button compact param

gorandalum:
### Background
Design has decided to remove the compact mode for buttons. As this change is quick and pretty risk free (as there are no functional changes), it is being done as a "quick issue".

This issue fixes point 4 described here:
https://github.com/AtB-AS/kundevendt/issues/19773

### Solution
Remove the `compact` param on buttons. Buttons that previously was compact should now be small.

Some before/after pictures from the app:
<div>

</div>
<div>

</div>

### Acceptance criteria
- [ ] Buttons that previously was small are now rounded. If wanting to check them one-by-one, then use the changelog of this PR to see affected files. Or just look through the app and see that nothing seems weird or out of place.

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -60,7 +60,6 @@ export type ButtonProps = {
   leftIcon?: ButtonIconProps;
   rightIcon?: ButtonIconProps;
   active?: boolean;
-  compact?: boolean;
   expanded: boolean;
   loading?: boolean;
   style?: StyleProp<ViewStyle>;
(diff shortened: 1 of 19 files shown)


jorelosorio (approved):

hanne-at-atb:
Is this change intended in the map?
IMG_1388

gorandalum:
> Is this change intended in the map? 

It's probably fine. The bonus button I think was removed in latest sketches, and the scan feature will still be iterated upon.

hanne-at-atb:
Also, did you align with @ckbilstad that the back button in the map should change color?

rosvik:
@hanne-at-atb The back button color change was a request from @Aliaaen, done as part of the design system update. I believe it wasn't intended to be blue from before.

hanne-at-atb:
Alrighty, good to know! It will not be blue then in the next release 🔲

PR #4951, opened 2025-01-21
feat: Secondary active button styling

gorandalum:
Active buttons with type secondary had no styling representing the
active state. This was not inline with Figma were active buttons
look the same both in primary and secondary mode.


hanne-at-atb:
✅

PR #4938, opened 2025-01-17
feat: markdown renderer handles numbered lists

jorunnl:
markdown-renderer can now display numbered (ordered) lists.

Needed for https://github.com/AtB-AS/kundevendt/issues/19831
PR https://github.com/AtB-AS/mittatb-app/pull/4936 should be merged first to see it in use.

PR #4953, opened 2025-01-21
fix: width of markdown rendering + ToggleSectionItem

jorunnl:
Related to https://github.com/AtB-AS/kundevendt/issues/19831
Fixes bug: https://github.com/AtB-AS/mittatb-app/pull/4938#issuecomment-2604148912

### Fixes so markdown ThemeText fills entire width

before: 

after:

### Also removes special margins on ToggleSectionItems
More accurate with design system, subtle change
before:

after:


rosvik (approved):
Nice ✨ 

Note: Use prefix `fix: ` as commit message / pr title for bug fixes. This is described here: https://github.com/AtB-AS/mittatb-app/blob/master/CONTRIBUTING.md#types

PR #4963, opened 2025-01-23
fix: disable PaginatedDetailsHeader for non-screenreader users

rosvik:
Revert of stray change from https://github.com/AtB-AS/mittatb-app/pull/4949


gorandalum (approved):

tormoseng:
✅

PR #4893, opened 2024-12-18
refactor: Not store stop place id on favorites

gorandalum:
When thinking about it I think it is no reason to store the stop
place id on favorites, as the quay id is uniquely identifiable.

### Acceptance criteria
- [ ] Existing favorites work as expected
- [ ] Storing new favorites work as expected

PR #4965, opened 2025-01-24
fix: update widget to use new favorite type

rosvik:
With #4893 the `stopId` field was removed from the favorite departure type, which broke the iOS widget. Removing references to this field makes the widget function again.

fixes https://github.com/AtB-AS/kundevendt/issues/19834

### DIFF
--- a/ios/departureWidget/Structs.swift
+++ b/ios/departureWidget/Structs.swift
@@ -336,9 +336,8 @@ struct FavouriteDeparture: Codable {
     let quayName: String
     let quayPublicCode: String?
     let quayId: String
-    let stopId: String
 
-    init(id: String, lineId: String, lineName: String?, lineLineNumber: String, lineTransportationMode: TransportMode?, lineTransportationSubMode: TransportSubMode?, quayName: String, quayPublicCode: String, quayId: String, stopId: String) {
+    init(id: String, lineId: String, lineName: String?, lineLineNumber: String, lineTransportationMode: TransportMode?, lineTransportationSubMode: TransportSubMode?, quayName: String, quayPublicCode: String, quayId: String) {
         self.id = id
         self.lineId = lineId
         self.lineName = lineName
@@ -348,7 +347,6 @@ struct FavouriteDeparture: Codable {
         self.quayName = quayName
         self.quayPublicCode = quayPublicCode
         self.quayId = quayId
-        self.stopId = stopId
     }
 
     init(from decoder: Decoder) throws {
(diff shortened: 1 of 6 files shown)


ios/departureWidget/Structs.swift
```diff
@@ -336,9 +336,8 @@ struct FavouriteDeparture: Codable {
     let quayName: String
     let quayPublicCode: String?
     let quayId: String
-    let stopId: String
 
-    init(id: String, lineId: String, lineName: String?, lineLineNumber: String, lineTransportationMode: TransportMode?, lineTransportationSubMode: TransportSubMode?, quayName: String, quayPublicCode: String, quayId: String, stopId: String) {
+    init(id: String, lineId: String, lineName: String?, lineLineNumber: String, lineTransportationMode: TransportMode?, lineTransportationSubMode: TransportSubMode?, quayName: String, quayPublicCode: String, quayId: String) {
```
jorelosorio, line 340: Just by looking at the code, it might be that this constructor `init` can be removed and the same in `DepartureLineInfo` there is no reference to it in other places unless is a mandatory init to have due to the decoder but we can try.

rosvik: I'm getting some errors when removing them, so it might require some more changes to make that work.

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -126,18 +127,24 @@ export const RootStack = () => {
...
 
     if (path.includes('details')) {
+      const item: ServiceJourneyDeparture = {
+        serviceJourneyId: params.serviceJourneyId as string,
+        date: (params.date as string) || new Date().toISOString(),
+        serviceDate: params.serviceDate as string,
+        fromStopPosition:
+          typeof params.fromStopPosition === 'string'
+            ? parseInt(params.fromStopPosition)
+            : 0,
```
rosvik, line 137: @jorelosorio I'm having an issue where the deep link that's built by a set of `URLQueryItem`s in WidgetViewModel added quotation marks to the `fromStopPosition` param, even thought it should be a plain integer (`fromStopPosition="10"` instead of `fromStopPosition=10`).

Running `parseInt` like this works, but do you know if there's a way to fix it in Swift?

jorelosorio: I can have a look ;)

jorelosorio: I was looking into the documentation of how the URLQueryItem works but it seems like all the values must be strings, because that is what is sent to the app when it opens, so even if we send an integer while building the URL, it might at the end send a string with the `scheme://widget/......`

jorelosorio (approved):

src/stacks-hierarchy/RootStack.tsx
rosvik: Thanks for looking into it! I had another look now, and I believe `URLQueryItem` is doing it correctly, and sending `fromStopPosition=10`. What confused me was that the param type in react native is `string | number | boolean | ...`, but the data always seem to be `string` 🫠 

I added a util now to clean the code in RootStack up a bit.

PR #4966, opened 2025-01-24
fix: Consolidate radio button group

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/18735

image

### DIFF
--- a/src/components/sections/RadioGroupSection.tsx
+++ b/src/components/sections/RadioGroupSection.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import {RadioSectionItem} from './items/RadioSectionItem.tsx';
 import {Section, SectionProps} from './Section';
 import {HeaderSectionItem} from './items/HeaderSectionItem';
-import {InteractiveColor} from '@atb/theme/colors';
 
 type Props<T> = Omit<SectionProps, 'children'> & {
   items: T[];
@@ -14,7 +13,6 @@ type Props<T> = Omit<SectionProps, 'children'> & {
   hideSubtext?: boolean;
   onSelect?(item: T, index: number): void;
   headerText?: string;
-  color?: InteractiveColor;
   itemToRightAction?: (
     item: T,
   ) => React.ComponentProps<typeof RadioSectionItem>['rightAction'];
@@ -30,7 +28,6 @@ export function RadioGroupSection<T>({
   selected,
   onSelect,
   headerText,
-  color,
   accessibilityHint,
   itemToRightAction,
   ...props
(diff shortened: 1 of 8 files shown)


strandlie:
Look good! The issue says that it should be added to Storybook as well. Just let me know if you want to do that together :)

jorelosorio:
Seems  good, but I am not quite sure about the issue description in https://github.com/AtB-AS/kundevendt/issues/18735 I think this must be done in the component itself I guess to cover other places where that is used, which is why it might require to be added to it the StoryBook (This is a place to test components and check how they look)

gorandalum (changes requested):
The change looks good, but there is still some places in the app where there is no appropriate styling on the radio buttons.

I suggest making the `color` parameter on `RadioGroupSection` mandatory, and see where it breaks.

Oleninkaa:
> Then we can see which part of the apps break, and then we could go standardize everything from there. It's kind of similar to the change that @jorelosorio did some time ago about button `expand` property [in this PR](https://github.com/AtB-AS/mittatb-app/pull/4907/files)

I used this method to see were the app will break. So I need to set the initialColor in the props all the time (as Jorge did in his solution)? I thought that it would be easier to change the default color of the component, so the component looks the same everywhere where it is used. And set in props custom color if needed. I went that way because it should be updated in the storybook too.

jorelosorio:
> > Then we can see which part of the apps break, and then we could go standardize everything from there. It's kind of similar to the change that @jorelosorio did some time ago about button `expand` property [in this PR](https://github.com/AtB-AS/mittatb-app/pull/4907/files)
> 
> I used this method to see were the app will break. So I need to set the initialColor in the props all the time (as Jorge did in his solution)? I thought that it would be easier to change the default color of the component, so the component looks the same everywhere where it is used. And set in props custom color if needed. I went that way because it should be updated in the storybook too.

After that change you could run in your terminal `yarn tsc` to see if there are any compilation issues after the change.

Oleninkaa:
> After that change you could run in your terminal `yarn tsc` to see if there are any compilation issues after the change.

I ran `yarn tsc `and everything works good!

gorandalum:
The problem with having a prop like `color` as optional is that it isn't really intuitive what the default value is, unlike for example a prop like `disabled` where one intuitively understand that the default value is `false`.

In this situation since every radio section item in the app looks the same, we can do the opposite of what I suggested: Instead of making the prop required, it can just be removed all together, and just use `theme.color.interactive[2]` internally without needing a prop for it.

gorandalum (approved):
LGTM :+1:

PR #4968, opened 2025-01-27
fix: support carnets with multiple TravelRights in ValidTo

rosvik:
Carnets will sometimes come with several travel rights, where accesses are added up. E.g. a carnet with 20 uses are really two travel rights with 10 accesses each. To account for this, we should use the `flattenCarnetTravelRightAccesses` util instead of reading from the first travel right.

Related to https://github.com/AtB-AS/kundevendt/issues/4220

### DIFF
--- a/src/fare-contracts/components/ValidTo.tsx
+++ b/src/fare-contracts/components/ValidTo.tsx
@@ -1,5 +1,6 @@
 import {
   FareContract,
+  flattenCarnetTravelRightAccesses,
   getLastUsedAccess,
   isCarnetTravelRight,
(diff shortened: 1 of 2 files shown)


strandlie:
Nice catch! Should this be a util, and therefore tested and -able?

gorandalum (approved):

strandlie (approved):

rosvik:
> Should this be a util, and therefore tested and -able?

It probably should! But I think I'll revisit that after the refactor I'm working on is done 👍 Some parts of this might change a bit after that.

PR #4971, opened 2025-01-27
feat: use Zod for fare contact validation

rosvik:
_Feel free to review this one commit at a time, that should make it more managable. ✨_

In preparation for fixing https://github.com/AtB-AS/kundevendt/issues/3240, and having a shared definition of what a fare contract is in the app and webshop, this PR

- Syncs the FareContract and TravelRight type with what is defined on the backend (eventhandler): https://github.com/AtB-AS/eventhandler/tree/main/customerstore
- Uses Zod to validate the data the app recieves from Firestore. By making sure fare contract data in the app conforms to the type, I could remove a bunch of manual validation.
- Instead of using utils like `isCarnet` and `isNormalTravelRight`, we now use the same type for all fare contract and all travel right, with undefined fields to differentiate carnets from "normal" tickets.

Note: With Zod validation set up like this, fare contract that doesn't conform to the FareContract type are discarded. This might not be how we want to handle fare contracts like school tickets. I'll look into how we want to handle that.

### Acceptance criteria

- [ ] It shouldn't be possible to crash the app by putting garbage fare contact data in firestore
- [ ] Invalid fare contracts no longer show up in the app (I'm looking into a solution that displays the `orderId`)
- [ ] Otherwise, all kinds of tickets still work as before.


src/fare-contracts/FareContractInfoDetails.tsx
```diff
@@ -119,7 +115,7 @@ export const getFareContractInfoDetails = (
   userProfiles: UserProfile[],
   preassignedFareProducts: PreassignedFareProduct[],
 ): FareContractInfoDetailsProps => {
-  const firstTravelRight = fareContract.travelRights?.[0] as NormalTravelRight;
+  const firstTravelRight = fareContract.travelRights?.[0];
```
jorelosorio, line 118: This get first element have been around for a while but I wonder if there is a more reusable way to access the first element, for instance:

`fareContract.travelRight()` and avoid to have to access directly to the array? internally the function check if data is present and then be reused on many other places.

src/ticketing/types.ts
```diff
@@ -20,31 +20,61 @@ export enum TravelRightDirection {
...
+  /** Used by boat and carnet boat */
   endPointRef?: string;
+  /** Used by boat and carnet boat */
   direction?: TravelRightDirection;
+  /** Used by carnet and carnet boat */
+  maximumNumberOfAccesses?: number;
+  /** Used by carnet and carnet boat */
+  numberOfUsedAccesses?: number;
+  /** Used by carnet and carnet boat */
+  usedAccesses?: CarnetTravelRightUsedAccess[];
```
gorandalum: I have a feeling that the comments are too AtB specific to be here. For example other counties can use use zones on boat products.

src/fare-contracts/FareContractInfoDetails.tsx
gorandalum: We have some [written doc](https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md#domain-models) on strategies for these situations:
> In situations where multiple components require the same deduced value from a domain model (e.g., isValid on FareContract, derived from validFrom and validTo), it's preferable to extract this logic into a reusable utility method, such as isValid(fc: FareContract). This approach avoids passing around extended versions of the FareContract or creating unnecessary variants.

I think the pattern with reusable utility methods for retrieving the necessary data would simplify this. Those utility methods (like `getFromZone(fc)`, `getEndValidity(fc)`, `getToStopPlace(fc)`) would take in the fare contract, so components generally didn't need to bother with accessing travel rights at all. 

However, as this was a PR for adding zod validation it is a little out of scope for this PR.

gorandalum (approved):
Nice!

src/fare-contracts/FareContractInfoDetails.tsx
rosvik: Right, it probably makes sense to have a `getFirstTravelRight` util! 

For now, I noticed that we can remove the `?` here, since `travelRights` is guaranteed to exist based on the Zod type.

src/ticketing/types.ts
rosvik: Right, these we mainly to keep track myself during development, so I'll remove them. However, they are based on what I found in eventhandler, so I don't think they should be org specific.

reidzeibel (approved):
Nice!

tormoseng:
@rosvik After this merge, when I pull to refresh, no tickets are shown.

I see that the request `/ticket/v4/list?availability=Available` is called, and that the FC are present in the response. But in the app, all the tickets are gone.

rosvik:
@tormoseng I'm able to reproduce, and getting this error from Zod:

```
[ZodError: [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "undefined",
    "path": [
      "minimumSecurityLevel"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "paymentTypeGroup"
    ],
    "message": "Required"
  }
]]
```

Might be some inconsistencies in the data returned from eventhandler and ticket service. Looking into it 👍

tormoseng:
@rosvik One more observation. The tickets are now shown for pull-to-refresh, but I see that the zone information is removed after a pull-to-refresh.

<details>

https://github.com/user-attachments/assets/248353a4-1645-4365-8957-42da28bf7a09

</details>

The response from the pull-to-refresh:
```json
{
	"fareContracts": [{
		"created": "2025-02-05T12:16:35.906623Z",
		"customerAccountId": "ATB:CustomerAccount:6o5N2ETFquZ79u1OhlCJ6F95F3J2",
		"eventTimestamp": "2025-02-05T12:16:44.201255Z",
		"id": "ATB:FareContract:VHLTL39A-6o5N2ETFquZ79u1OhlCJ6F95F3J2",
		"orderId": "VHLTL39A",
		"paymentType": ["MASTERCARD"],
		"purchasedBy": "ATB:CustomerAccount:6o5N2ETFquZ79u1OhlCJ6F95F3J2",
		"qrCode": "ChYKFBISGhDTIdW3r6hE2JT8Y+TccCsAEkUKPjA8AhwElzTibfFKzRMmJ2gJPpQ6VxYF47otRUhsvd+aAhxCFxWtuFHd30fIJgiZbnWrgNVleg+E9s1VcuSMGgFOMAE=",
		"state": 2,
		"totalAmount": "43.20",
		"totalTaxAmount": "4.63",
		"travelRights": [{
			"id": "ATB:CustomerPurchasePackage:O9HKVK6U",
			"customerAccountId": "ATB:CustomerAccount:6o5N2ETFquZ79u1OhlCJ6F95F3J2",
			"startDateTime": "2025-02-05T12:16:44Z",
			"endDateTime": "2025-02-05T13:46:44Z",
			"authorityRef": "SJN:Authority:SJN",
			"fareProductRef": "ATB:PreassignedFareProduct:8808c360",
			"status": 5,
			"userProfileRef": "ATB:UserProfile:8ee842e3",
			"usageValidityPeriodRef": "",
			"type": "PreActivatedSingleTicket",
			"fareZonesRef": ["ATB:FareZone:10"],
			"tariffZoneRef": ["ATB:TariffZone:1"]
		}],
		"version": "1"
	}]
}
```

rosvik:
> The tickets are now shown for pull-to-refresh, but I see that the zone information is removed after a pull-to-refresh.

@tormoseng  This seems to be the case in prod as well 😬 The ticket endpoint returns the wrong field names. PR to fix it here: https://github.com/AtB-AS/ticket/pull/20

tormoseng:
✅  Looks good!

PR #4972, opened 2025-01-29
fix: Fix wrongly visible zones

strandlie:
From [this issue](https://mittatb.slack.com/archives/C0116FMPX4Y/p1738052534522879) we uncovered that there was previously a case where zones were hidden (when `fareProductTypeConfig?.configuration.zoneSelectionMode === 'none'`). 

This fixes that issue. 

At the same time it is a small experiment with the Component Hooks-pattern, since the `FareContractFromTo`-component was getting pretty large. Let me know you thoughts!

### Acceptance-criteria
- [ ] Zones are no longer shown for Youth-ticket in NFK (and still not for youth ticket anywhere else)
- [ ] Zones and Harbors are still shown as expected everywhere else


gorandalum (commented):
I like the pattern! As discussed I would prefer having the hook in the same file as the component, to group related code more closely together. Could also name the hook with suffix "controller" as that is kinda what it is.

strandlie:
> I like the pattern! As discussed I would prefer having the hook in the same file as the component, to group related code more closely together. Could also name the hook with suffix "controller" as that is kinda what it is.

@gorandalum Fixed now! Will also add some documentation in `docs-private`

src/fare-contracts/components/BorderedFromToBox.tsx
```diff
@@ -0,0 +1,116 @@
...
+          </ThemeText>
+          <ThemeText color={backgroundColor} typography="body__tertiary">
+            {toText}
+          </ThemeText>
+        </View>
+      </View>
+    );
+  };
+
+  const largeLayout = () => (
```
jorelosorio, line 69: It might not be a function that can be assigned, however, useCallback can be evaluated if it might improve the performance a bit.

src/fare-contracts/modules/FareContractFromTo.tsx
```diff
@@ -0,0 +1,151 @@
...
+function useFareContractFromToController({
+  preassignedFareProduct,
+  fcOrRfc,
+}: FareContractFromToControllerProps) {
+  // shouldReturnNull
+  const {fareProductTypeConfigs} = useFirestoreConfigurationContext();
+  const fareProductTypeConfig = fareProductTypeConfigs.find(
+    (c) => c.type === preassignedFareProduct?.type,
+  );
+  const shouldReturnNull =
```
jorelosorio, line 76: Why not just return null here? at the end the implementation of useFareContractFromToController returns null if `shouldReturnNull` is true, so it might not need to then continue creating all the other functions below.

src/fare-contracts/modules/FareContractFromTo.tsx
```diff
@@ -0,0 +1,151 @@
...
+export const FareContractFromTo = (props: FareContractFromToProps) => {
+  const {
+    shouldReturnNull,
+    tariffZoneRefs,
+    direction,
+    startPointRef,
+    endPointRef,
+  } = useFareContractFromToController({
+    preassignedFareProduct: props.preassignedFareProduct,
+    fcOrRfc: 'rfc' in props ? props.rfc : props.fc,
```
jorelosorio, line 35: It is possible to ask for types or instances of?

src/fare-contracts/components/BorderedFromToBox.tsx
strandlie: Maybe! I did consider it, but I concluded that these views are very lightweight and not worth adding the complexity from adding `useCallback`, in addition to that these functions are not passed as props further down. 

But I might be missing some other reasons for adding it...

jorelosorio: It needs to be evaluated depending on how many times a re-render recreates those. But if is not so much then it is ok

src/fare-contracts/modules/FareContractFromTo.tsx
strandlie: If I just return `null` here it makes the return value of this hook become this:

```
{   
    tariffZoneRefs: string[];   
    direction: TravelRightDirection | undefined;   
    startPointRef: string | undefined;   
    endPointRef: string | undefined; 
}  |  null
```

which I don't love since we then have to check if the entire object is null first. What do you think about something like this instead: 

```
if (fareProductTypeConfig?.configuration.zoneSelectionMode === 'none') {
    return {shouldReturnNull: true}
}
```

src/fare-contracts/components/BorderedFromToBox.tsx
strandlie: Every re-render recreates those functions, but it's just lightweight view-code which I think is basicly equivalent to having them in the return-statement of the component. Furthermore, I do not observe any performance issues from this. So in this case I think it's ok. 

But I appreciate the input!

src/fare-contracts/modules/FareContractFromTo.tsx
strandlie: Could we look at this together tomorrow? I tried a couple of approaches now, but it always became more complex and verbose, so I would love to explore some ideas together :)

src/fare-contracts/modules/FareContractFromTo.tsx
gorandalum: Good question! We should probably have a pattern for this across our controller hooks! I think what is done here is a good option. I like the suggested improvement with the early return of `return {shouldReturnNull: true}` as well. 

I also think returning `undefined` is a viable option. Then it would be an established pattern that when a controller hook returns `undefined` that means that the component should not render anything (return `null`).

It would make it slightly more cumbersome as the return value from the controller can't be destructured on invocation, but not to much boilerplate I guess.
```tsx
const controllerData = useMyComponentController(…)
if (!controllerData) return null;
const {field1, field2, field3} = controllerData;
```

src/fare-contracts/details/DetailsContent.tsx
```diff
@@ -156,6 +156,7 @@ export const DetailsContent: React.FC<Props> = ({
             backgroundColor={theme.color.background.neutral['0']}
             mode="large"
             fc={fc}
+            preassignedFareProduct={preassignedFareProduct}
```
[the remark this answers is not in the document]
strandlie: Ah, yes you are right. I will fix that.

src/fare-contracts/modules/FareContractFromTo.tsx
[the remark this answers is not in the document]
strandlie: Me, @rosvik and @marius-at-atb have talked about it and are introducing modules as a concept. Tomorrow we are having a meeting to decide the rules of a module. But since we haven't decided yet, I can put it into the components-folder again if it's merged before that.

src/fare-contracts/modules/FareContractFromTo.tsx
strandlie: Yes, very good question, and it leads to another important question: 

How do we then feel about "hidden rules" such as `when a controller hook returns undefined it means that the component should not render anything`? 

My take is that they are easy enough to enforce in CR and similar as long as someone remembers, but small rules like that are easy to forget and hard to document in a useful way, and arguably make the codebase harder to understand - especially on first encounter, so if there is another alternative we should choose that.

strandlie: 👆🏽 the broad question still stands. 

An important nuance in the current case of component Hooks is that if the hook has a return type of trueish types, e.g.:

```
{
   field1: string;
   field2: string;
}
```

those are problematic to deal with if we early `return {shouldReturnNull: true}`, since they then become:

```
{
   field1: string | undefined;
   field2: string | undefined
}
```

so I now agree that `return undefined` is the better solution for component hooks since it assures type safety when the component renders.

src/fare-contracts/details/DetailsContent.tsx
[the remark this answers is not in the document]
strandlie: Fixed

src/fare-contracts/modules/FareContractFromTo.tsx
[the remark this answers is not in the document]
strandlie: Moved into components now, since NFK are eager to release this

src/fare-contracts/modules/FareContractFromTo.tsx
strandlie: Have implemented the `return undefined` now, as agreed upon since NFK are eager to release this.

src/fare-contracts/components/BorderedFromToBox.tsx
```diff
@@ -0,0 +1,121 @@
...
+    justifyContent: 'space-evenly',
+    columnGap: theme.spacing.xSmall,
+  },
+  smallContentText: {
+    flexDirection: 'column',
+    paddingLeft: theme.spacing.xSmall,
+  },
+}));
+
+const icons = {
```
strandlie, line 116: This was also the logic previously, but I think this representation makes it clearer what will happen.

src/fare-contracts/modules/FareContractFromTo.tsx
gorandalum: > those are problematic to deal with if we early return {shouldReturnNull: true}, since they then become:
>
>```
>{
>   field1: string | undefined;
>   field2: string | undefined
>}
>```
>so I now agree that return undefined is the better solution for component hooks since it assures type safety when the component renders.

I know we have gone for the `undefined` return value instead of `shouldReturnNull` now, so this is a more informational comment just to point out an example of utilising typescript in a good way.

You @strandlie mention a good example of what could be a problem with the `shouldReturnNull: true` early return option if we had inefficient typing. A return type like this wouldn't be good:
```ts
type ExampleReturnType = {
  shouldReturnNull: boolean;
  field1: string | undefined;
  field2: string | undefined
}
```
But with a typing like this the type safety of the fields would still be preserved:
```ts
type ExampleReturnType =
  | { shouldReturnNull: true }
  | { shouldReturnNull: false; field1: string;, field2: string };
```

As then it could look like this in component:
```ts
const controllerData = useMyComponentController(…)
if (controllerData.shouldReturnNull) return null;
const {field1, field2, field3} = controllerData; // Now the compiler knows that field1, field2 are defined
```

PR #4973, opened 2025-01-29
fix: token issue fix and many more logging

reidzeibel:
Fix issue with mobile token where the token gets recreated whenever issues happened during fetching the local token.
Also adds many more loggings.

So, what happened on the issue in 1.61.4 is that I added a function to detect if a user needs to start fresh, there was an issue for old users when their data was truncated from backoffice and the token is already deleted, where their local device still has the old token. It means that the user should just remove that local token, and then init a new token, problem solved.

The problem was this code: 
```
export const wipeToken = async (tokensIds: string[], traceId: string) => {
  for (const id of tokensIds) {
    logToBugsnag('Wiping token with id ' + id);
    try {
      await tokenService.removeToken(id, traceId);
    } catch (err: any) {
      /**
       * There are cases where remove token fails due to the backoffice token
       * already removed, in this case, just wipe the local token.
       */
      if (isTokenDeletedError(err)) {
        await mobileTokenClient.clear();
      }
    }
  }
  await mobileTokenClient.clear();
};
```

In that `wipeToken` function, I added a catch block, and check if the token is deleted in backoffice, just clear the token. 

Unfortunately, I didn't realize that you need to actually **_throw_** the error for the error to propagate and stop the `clear()` function at the bottom from being called. Therefore, when an issue happens during fetching, the `wipeToken` gets called, then the local token is cleared, prompting an `/init` before calling `/remove`, causing the remote token to still be there, and a new token gets recreated everytime.

In this PR I tried fixing the issue by calling **_throw_** in the `catch` block, and added many more loggings.

Please point out if there's anything weird, or ask if you have any questions!

### DIFF
--- a/src/mobile-token/utils.ts
+++ b/src/mobile-token/utils.ts
@@ -54,7 +54,7 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
             description: `Token already deleted in backoffice ${id}, start new`,
           },
         });
-        await mobileTokenClient.clear();
+        break; // No need to do anything, proceed to clear the local token
       } else {
         notifyBugsnag(err, {
           errorGroupHash: 'token',
--- a/src/mobile-token/utils.ts
+++ b/src/mobile-token/utils.ts
@@ -60,7 +60,7 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
           errorGroupHash: 'token',
           metadata: {
             traceId,
-            description: `Token already deleted in backoffice ${id}, start new`,
+            description: `Other error during wipe token, throw the error`,
           },
         });
         // if it is not entity deleted error
--- a/src/mobile-token/utils.ts
+++ b/src/mobile-token/utils.ts
@@ -15,8 +15,7 @@ import Bugsnag from '@bugsnag/react-native';
 import {getAxiosErrorType} from '@atb/api/utils';
 import {tokenService} from './tokenService';
 import {mobileTokenClient} from './mobileTokenClient';
-import {logToBugsnag} from '@atb/utils/bugsnag-utils';
-import {isAxiosError} from 'axios';
+import {logToBugsnag, notifyBugsnag} from '@atb/utils/bugsnag-utils';
 
 export const MOBILE_TOKEN_QUERY_KEY = 'mobileToken';
 
@@ -48,7 +47,24 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
        * already removed, in this case, just wipe the local token.
        */
       if (isEntityDeletedError(err)) {
+        notifyBugsnag(Error(`Token removed from backoffice ${id}`), {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Token already deleted in backoffice ${id}, start new`,
+          },
+        });
         await mobileTokenClient.clear();
+      } else {
+        notifyBugsnag(err, {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Token already deleted in backoffice ${id}, start new`,
+          },
+        });
+        // if it is not entity deleted error
+        throw err;
       }
     }
   }
@@ -78,17 +94,20 @@ export const getMobileTokenErrorHandlingStrategy = (
     errorResolution = err.resolution;
   } else if (err instanceof RemoteTokenStateError) {
     errorResolution = mapTokenErrorResolution(err);
-  } else if (isEntityDeletedError(err)) {
-    /**
-     * This handles the situation where local token is still stored, while
-     * the remote token is already deleted, should only happens with a very
-     * old account that got their account history truncated in backoffice.
-     */
-    return 'reset';
   } else {
     return 'unspecified';
   }
 
+  logToBugsnag(
+    `mobile token handle error ${err} with resolution ${errorResolution}`,
+  );
+
+  notifyBugsnag(err, {
+    errorGroupHash: 'token',
+    metadata: {
+      description: `Handling error ${err} with resolution of ${errorResolution}`,
+    },
+  });
   // handle the error resolution
   switch (errorResolution) {
     case TokenErrorResolution.RESET:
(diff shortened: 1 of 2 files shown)


src/mobile-token/utils.ts
```diff
@@ -48,7 +47,24 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
...
+            description: `Token already deleted in backoffice ${id}, start new`,
+          },
+        });
+        break; // No need to do anything, proceed to clear the local token
+      } else {
+        notifyBugsnag(err, {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Token already deleted in backoffice ${id}, start new`,
```
tormoseng, line 63: Should be equal as in the if()? Just checking that it's not an copy-paste error

reidzeibel: Yes that's a copy-paste error, fixing

reidzeibel: fixed

src/mobile-token/utils.ts
```diff
@@ -48,7 +47,24 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
...
+      } else {
+        notifyBugsnag(err, {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Other error during wipe token, throw the error`,
+          },
+        });
+        // if it is not entity deleted error
+        throw err;
```
strandlie, line 67: What happens when the error is thrown? What catches the error?

src/mobile-token/utils.ts
```diff
@@ -78,17 +94,20 @@ export const getMobileTokenErrorHandlingStrategy = (
...
+  logToBugsnag(
+    `mobile token handle error ${err} with resolution ${errorResolution}`,
+  );
+
+  notifyBugsnag(err, {
+    errorGroupHash: 'token',
+    metadata: {
+      description: `Handling error ${err} with resolution of ${errorResolution}`,
+    },
+  });
```
gorandalum, line 110: Do we need to do both log and notify right after each other?

src/mobile-token/utils.ts
```diff
@@ -105,17 +124,12 @@ export const getMobileTokenErrorHandlingStrategy = (
...
+ * This usually happens when the token is removed from backoffice
+ * and they return error code 500.
  */
-const isEntityDeletedError = (err: any): boolean => {
-  return (
-    isAxiosError(err) && err.code === '500' && err.message.includes('NOT_FOUND')
-  );
-};
+const isEntityDeletedError = (err: any): boolean =>
+  'statusCode' in err && err.statusCode === 500;
```
gorandalum, line 132: When I see this code it seems a little iffy to handle every 500 as not found error 🤔 Do we have control over this status code and what message it can return?

src/mobile-token/utils.ts
reidzeibel: This one is to preserve the existing behavior before adding the try-catch block, it was like this originally:

```
export const wipeToken = async (tokensIds: string[], traceId: string) => {
  for (const id of tokensIds) {
    logToBugsnag('Wiping token with id ' + id);
    await tokenService.removeToken(id, traceId);
  }
  await mobileTokenClient.clear();
};
```

So there's no handling of errors whatsoever, throw any errors to the ones that calls this function.

src/mobile-token/utils.ts
```diff
@@ -48,7 +47,24 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
...
       if (isEntityDeletedError(err)) {
-        await mobileTokenClient.clear();
+        notifyBugsnag(Error(`Token removed from backoffice ${id}`), {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Token already deleted in backoffice ${id}, start new`,
+          },
+        });
+        break; // No need to do anything, proceed to clear the local token
```
gorandalum, line 57: This part here with the `break` isn't really intuitive. Why don't we need to iterate over the other token ids if one of them is already deleted? 🤔

src/mobile-token/utils.ts
```diff
@@ -48,7 +47,24 @@ export const wipeToken = async (tokensIds: string[], traceId: string) => {
...
+      } else {
+        notifyBugsnag(err, {
+          errorGroupHash: 'token',
+          metadata: {
+            traceId,
+            description: `Other error during wipe token, throw the error`,
+          },
+        });
```
gorandalum, line 65: Should this just be log to Bugsnag instead? I would think the thrown err gets notified to Bugsnag when it is catcher higher up.

src/mobile-token/utils.ts
reidzeibel: Unfortunately we don't have any controls, the error that gets thrown is: 

```
{"error": "Internal Server Error", "message": "An internal server error occurred", "statusCode": 500}
```

I tried using `isAxiosError` function but it returned false, and since the check only called on the `/remove` endpoint, it should be safe, looking at the grafana logs with error 500, all of them is token deleted errors.

src/mobile-token/utils.ts
reidzeibel: Right, this one should be left at do nothing and not `break`, to allow continuing removal of other tokens.

src/mobile-token/utils.ts
reidzeibel: Yes, we can do that, thanks for the reminder!

src/mobile-token/utils.ts
gorandalum: But we don't need a breadcrumb when we right after generate an error for the same thing?

src/mobile-token/utils.ts
gorandalum: How do we see in Grafana that they are token deleted errors if they are just generic 500 errors? 🤔

reidzeibel: It gets the message from Grafana, see below, not sure why it doesn't show the message on the log or error object though...

image

src/mobile-token/utils.ts
reidzeibel: In my understanding, if we generate error first, the breadcrumb that gets called next won't get sent. I added this to at least get more logs in Bugsnag, since we need to identify the exact point of errors on the token, and the current bugsnag logs doesn't seem to give us a clear image of what actually happens.

src/mobile-token/utils.ts
strandlie: I agree that handling all 500-errors will probably bite us at a later time. If we have to do it like this now, we should look into changing the tokenService such that it will a sensible error asap.

reidzeibel: @strandlie It will be handled on the server side now, thanks to @gorandalum suggestion!

reidzeibel: This code has been removed

gorandalum (approved):

PR #4974, opened 2025-01-29
fix: upgrade `phone` to support 9 digit Bosnia and Herzegovina numbers

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/19919

After https://github.com/AfterShip/phone/pull/420, the issue with 9 digit phone numbers from Bosnia and Herzegovina is resolved by upgrading the phone package.


gorandalum (approved):
Nice work! 🎉

PR #4975, opened 2025-01-29
feat: updated vehiclecard

Dahly96:
closes: https://github.com/AtB-AS/kundevendt/issues/19914

| Before | After |
|--------|-------|
|  |  |

updated vehiclecard and added new icons to the design system which made some changes to the path of some already in use icons.

### DIFF
--- a/src/mobility/components/VehicleCard.tsx
+++ b/src/mobility/components/VehicleCard.tsx
@@ -1,8 +1,6 @@
-import {VehicleId} from '@atb/api/types/generated/fragments/vehicles';
 import React from 'react';
 import {useTranslation} from '@atb/translations';
 import {StyleSheet} from '@atb/theme';
-import {useVehicle} from '@atb/mobility/use-vehicle';
 import {GenericSectionItem, Section} from '@atb/components/sections';
 import {OperatorNameAndLogo} from './OperatorNameAndLogo';
 import {View} from 'react-native';
@@ -12,7 +10,6 @@ import {VehicleCardStat} from './VehicleCardStat';
 import {ScooterTexts} from '@atb/translations/screens/subscreens/MobilityTexts';
 import {
   formatNumberLang,
-  formatPrice,
   formatPricePerUnit,
   formatRange,
   getBatteryLevelIcon,
--- a/src/mobility/components/VehicleCard.tsx
+++ b/src/mobility/components/VehicleCard.tsx
@@ -13,18 +13,31 @@ import {ScooterTexts} from '@atb/translations/screens/subscreens/MobilityTexts';
 import {
   formatNumberLang,
   formatPrice,
+  formatPricePerUnit,
   formatRange,
   getBatteryLevelIcon,
 } from '../utils';
+import {PricingPlanFragment} from '@atb/api/types/generated/fragments/mobility-shared';
 
 type Props = {
-  vehicleId: VehicleId;
+  pricingPlan: PricingPlanFragment;
+  currentFuelPercent: number | undefined;
+  currentRangeMeters: number;
+  operatorName: string;
+  brandLogoUrl: string | undefined;
 };
 
-export const VehicleCard = ({vehicleId: id}: Props) => {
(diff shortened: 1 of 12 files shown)


src/mobility/components/VehicleCardStat.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {ThemeText} from '@atb/components/text';
+import {SvgProps} from 'react-native-svg';
+
+type Props = {
+  icon: (SvgProps: SvgProps) => JSX.Element;
+  stat: string;
+  description: string;
+};
+
+export const VehicleCardStat = ({stat, description, icon}: Props) => {
```
jorelosorio, line 14: We might have already a component with an icon, title and subtitle...

src/mobility/components/VehicleCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+} from '../utils';
+
+type Props = {
+  vehicleId: VehicleId;
+};
+
+export const VehicleCard = ({vehicleId: id}: Props) => {
+  const {t, language} = useTranslation();
+  const styles = useSheetStyle();
+  const {vehicle, operatorName, brandLogoUrl} = useVehicle(id);
```
marius-at-atb, line 27: During a trip, this data will come from the Shared Mobility booking data instead of GBFS. (During a trip, the vehicle does not exist in GBFS.)
So this should be a "dumb" component that doesn't use any APIs, but instead accepts the exact props it needs to render.

src/mobility/components/VehicleCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+              ) : null}
+            </>
+          ) : null}
+        </View>
+      </GenericSectionItem>
+    </Section>
+  );
+};
+
+const useSheetStyle = StyleSheet.createThemeHook((theme) => {
```
marius-at-atb, line 79: This should always be called `useStyles`.

marius-at-atb:
Remember to link the issue, and when changing an existing component, also add a before and after screenshot in the description ;)

src/mobility/components/VehicleCardStat.tsx
Dahly96: i dont think we have actualy

src/mobility/utils.ts
```diff
@@ -163,6 +171,44 @@ export const formatRange = (rangeInMeters: number, language: Language) => {
   return `${rangeInKm} km`;
 };
 
+const getPercentageBattery = (batteryPercentage: number) => {
+  let newBatteryPercentage = batteryPercentage;
+  if (batteryPercentage % 1 !== 0) {
```
marius-at-atb, line 176: If the percentage is 20.5, this will show 2050%, so also check if it is between 0 and 1. Should probably round the number otherwise.

src/mobility/components/VehicleCardStat.tsx
```diff
@@ -0,0 +1,38 @@
...
+    </View>
+  );
+};
+
+const useSheetStyle = StyleSheet.createThemeHook(() => {
+  return {
+    contentItem: {
+      flexDirection: 'column',
+      alignItems: 'center',
+      gap: 5,
```
marius-at-atb, line 35: This should use a standard size. Probably `theme.spacing.small`. If Figma has `5`, talk to the designer.

src/mobility/utils.ts
```diff
@@ -163,6 +171,44 @@ export const formatRange = (rangeInMeters: number, language: Language) => {
...
+  } else if (newBatteryPercentage >= 16) {
+    return BatteryMedium;
+  } else if (newBatteryPercentage >= 3) {
+    return BatteryLow;
+  } else {
+    return BatteryEmpty;
+  }
+};
+
+export const formatPrice = (
```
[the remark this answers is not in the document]
Dahly96: yes, ill create a more general function and use that for everything

marius-at-atb (approved):
Nice!

PR #4976, opened 2025-01-29
fix: changes in purchase overview

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/19653

Also requested changes in [firestore](https://github.com/AtB-AS/firestore-configuration/pull/721) for the `Velg varighet ` text.

image


src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Summary.tsx
```diff
@@ -35,12 +35,8 @@ export function Summary({
...
   const {theme} = useThemeContext();
 
-  const formattedPrice = formatDecimalNumber(price, language, 2);
-  const formattedOriginalPrice = formatDecimalNumber(
-    originalPrice,
-    language,
-    2,
-  );
+  const formattedPrice = formatDecimalNumber(price, language);
+  const formattedOriginalPrice = formatDecimalNumber(originalPrice, language);
```
rosvik, line 39: I don't think this will give the wanted behavior in all cases. I assume we still have to show decimals if the price is not a whole number. I expect we want something like:

- For price `42,00 kr` -> `42 kr`
- For price `42,99 kr` -> `42,99 kr`
- For price `42,90 kr` -> `42,90 kr`
- For price `42,99001 kr` -> `42,99 kr`

Could you look into making a new util for this? I also noticed an old comment from the `formatDecimalNumber`:

```
When RN 0.65 is released, this can be replaced by the 'toLocaleString' method,
which as of now is not working on Android.
```

We're at react native 0.73 now, so that might be worth looking into as well 😄 It has a `maximumFractionDigits` that might help you.

rosvik: @ckbilstad The more I think about this, the more I think it might not be a good idea. What's the thought behind the change?

Should the same logic apply to the purchase summary screen, and everywhere else in the app? (Probably not things like VAT calculations that usually has decimals) Could the user be confused by it, and assume we're rounding the price?

Oleninkaa: > Should the same logic apply to the purchase summary screen, and everywhere else in the app? (Probably not things like VAT calculations that usually has decimals) Could the user be confused by it, and assume we're rounding the price?

For the whole price (49.00) kroner it would look like this. I think it can be confusing for the user that it has no zeros after. I think it's a common thing to display total price with zeros even if it's a whole price. 

image

ckbilstad:
It was a proposed adjustment with the intention to reduce cognitive load and the amount of information in the purchase flow, particularly handy for those with dyscalculia eg. It originated in a codesign-session after a discussion and look at how other comparable products handle similar cases (Entur, Ruter, Flytoget, Norgestaxi). I will address it in a design-meeting later today to see if someone still feel the need and have the opportunity to dive deeper into how to handle all the cases. Ill let you know.

rosvik:
@Oleninkaa Maybe you could remove the `formatDecimalNumber` changes from this PR, then make a new one once we know what's wanted from design? The text changes look good 👍

Mariblaas:
We have been looking into it, notify me when we can have a chat! @rosvik

Mariblaas:
If there are decimals, show 2 (42,90 kr)

Always show 2 decimals: 
- mva
- receipt sent on email

Oleninkaa:
So I've created separate util for Prices. The old one is being used in mobility for distances. We leave two zeros only for VAT values.

Summary

Price_summary

ticket_details

Skjermbilde 2025-01-31 kl  16 03 12


I also fixed price thingy for scooters `(10.9 => 10.90)`. But it can be wrong solution for that 🥲


Before:
screen_before

After:
Simulator Screenshot - iPhone 16 Pro Max - 2025-01-31 at 14 34 08

src/utils/numbers.ts
```diff
@@ -1,13 +1,19 @@
...
+  fractionDigits?: number,
+  isVat?: boolean /**Always show 2 decimals in VAT value*/,
+) => {
+  const minFractionDigits = isVat ? 2 : str % 1 == 0 ? 0 : 2;
+  return str.toLocaleString(language, {
+    minimumFractionDigits: minFractionDigits,
+    maximumFractionDigits: fractionDigits || 2,
+    useGrouping: false,
+  });
+};
```
gorandalum, line 15: The implementation looks good, but the naming could be improved. It is slightly nitpicking, but I think it is good learning in the thought behind it :)

As this is a generic utility function it shouldn't use domain specific terms like 'fare' or 'vat'. Instead it should be named according to what it does. So the field `isVat` should be instead `alwaysShowTwoDecimals` or something like it. Also in the name of the function the 'fare' part should be changed to a wording which reflects what is special about the price formatting for fares as opposed to other prices.

gorandalum (approved):
Good work 👍

PR #4977, opened 2025-01-30
fix: Fixed verticalPadding on PhoneInputSectionItem

strandlie:
Bug found while testing AtB-AS/kundevendt#4220

Makes `PhoneInputSectionItem` equal to `TextInputSectionItem`. Also added `PhoneInputSectionItem` to the Design System screen, because it was missing from there.


reidzeibel (approved):
LGTM!

PR #4978, opened 2025-01-30
refactor: remove some unused FareContract/TraveRight fields

rosvik:
Some fields that are always present in Firestore, are not returned by the ticket service. After adding stricter Zod validation of fields, this caused tickets from the endpoint to fail validation. Since these fields were not used (in a significant way) in the app, I opted to treat them as deprecated and remove them from the app entirely.

At the same time, it shows that we should use the ticket service as source of truth for fare contract and travel right types in the app.

fixes comment https://github.com/AtB-AS/mittatb-app/pull/4971#issuecomment-2623942274

### DIFF
--- a/src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
+++ b/src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
@@ -27,7 +27,6 @@ import {
   isOfFareProductRef,
   useFirestoreConfigurationContext,
 } from '@atb/configuration';
-import {onlyUniquesBasedOnField} from '@atb/utils/only-uniques';
 import {useTimeContext} from '@atb/time';
 import {getDeviceNameWithUnitInfo} from './utils';
 import {TokenToggleInfo} from '@atb/token-toggle-info';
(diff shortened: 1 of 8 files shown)


src/select-travel-token-screen/SelectTravelTokenScreenComponent.tsx
```diff
@@ -72,9 +71,8 @@ export const SelectTravelTokenScreenComponent = ({onAfterSave}: Props) => {
   );
 
   // Filter for unique travel rights config types
-  const availableFareContractsTypes = availableTravelRights
-    .filter(onlyUniquesBasedOnField('type'))
```
rosvik, line 76: This is the only place we use the `travelRight.type` field, and seems to be a (small) performance optimization. Removing it should have no functional change. We're also talking about removing the `requiresTokenOnMobile` option, which should remove the need for all of this code.

marius-at-atb (approved):
Lgtm

gorandalum (approved):

mjenssen (approved):
LGTM 🚀

PR #4979, opened 2025-01-31
feat: contact scooter operator confirmation screen

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/19943

Adds ConfirmationScreen component.
* This is now used for contact scooter operator confirmation screen.
* It is also used for the report parking violations confirmation screen.

This PR also renames some folders and files to clean up

This screen is also updated, using the same screen component:


src/stacks-hierarchy/Root_ScooterHelp/components/ConfirmationScreenComponent.tsx
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  const icon = themeName === 'dark' ? CheckDark : CheckLight;
+
+  const iconColorProps = {
+    color: theme.color.status.valid.secondary.foreground.primary,
+  };
+
+  return (
+    <ScreenContainer rightHeaderButton={{type: 'close', onPress: close}}>
```
marius-at-atb, line 38: Is it a bit strange to have to close buttons on the same screen? Has this been discussed with design?

marius-at-atb (approved):
Lgtm! Just some nitpicks in the comments.

src/stacks-hierarchy/Root_ScooterHelp/components/ConfirmationScreenComponent.tsx
jorunnl: Yes, double checked with design. This is the wanted look / functionality

src/translations/dictionary.ts
```diff
@@ -49,6 +49,16 @@ const dictionary = {
...
     },
   },
+  appNavigation: {
+    close: {
+      text: _('Lukk', 'Close', 'Lukk'),
+      a11yHint: _(
+        'Aktivér for å lukke',
+        'Activate to close',
+        'Aktivér for å lukke',
+      ),
```
gorandalum, line 59: Just a FYI, not something that needs to be changed now. We use accessibility hints in the situation where it is not intuitive for the vision-impaired user what the effect of the button is. When hints are not necessary we try to not use them as less texts and translations in the app are beneficial.

Here the label is "close", and together with the accessibility role "button", it is intuitive for the user what the effect of the button is.

jorunnl: Aha, thanks for the FYI, I'll remember that

PR #4983, opened 2025-02-03
fix: Added back some Android-specific padding

strandlie:
Follow up to AtB-AS/kundevendt#4220

Added back some Android-specific padding, since it turned out it was needed to get it _just_ right.

### DIFF
--- a/src/components/sections/items/PhoneInputSectionItem.tsx
+++ b/src/components/sections/items/PhoneInputSectionItem.tsx
@@ -3,6 +3,7 @@ import {
   AccessibilityInfo,
   Keyboard,
   NativeSyntheticEvent,
+  Platform,
   TextInput as InternalTextInput,
   TextInputFocusEventData,
   TextInputProps as InternalTextInputProps,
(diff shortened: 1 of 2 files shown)


src/components/sections/items/PhoneInputSectionItem.tsx
```diff
@@ -113,12 +114,20 @@ export const PhoneInputSectionItem = forwardRef<InternalTextInput, Props>(
       }
     };
 
-    // Remove padding from topContainerStyle
-    const {padding: _dropThis, ...topContainerStyle} = topContainer;
```
strandlie, line 117: This is not needed anymore, since `padding` is replaced with `paddingVertical` and `paddingHorizontal` in ´useSectionItem`

src/components/sections/items/TextInputSectionItem.tsx
```diff
@@ -183,15 +197,13 @@ const useInputStyle = StyleSheet.createTheme((theme) => ({
...
     rowGap: theme.spacing.small,
   },
+  internalInputContainer: {
+    justifyContent: 'center',
+  },
   label: {
     minWidth: 60 - theme.spacing.medium,
     paddingRight: theme.spacing.xSmall,
   },
-  inputContainer: {
```
strandlie, line 190: Removed because it is not used.

gorandalum (approved):

PR #4985, opened 2025-02-04
refactor: Slight refactor for TravelTokenBox usage

gorandalum:
Moved the TravelTokenBox out of the FareContractAndReservationsList
component. This is to adhere to the single responsibility principle,
the component for listing fare contracts should not be responsible
for also showing token related info.

In addition removed the color parameter for the TravelTokenBox
since it was always themed to interactive color 2 and could just as
well be specified internally in the component.


reidzeibel (approved):
Looks tidier in my opinion 👍🏼

PR #4986, opened 2025-02-04
refactor: Cleaner configuration of empty state in fare contract list

gorandalum:
This refactor started when I wanted to make the 'mode' parameter
non-optional, since it wasn't intuitive what the default value was.
But then making it required made it obvious that the possible
modes (historic and sent) was not enough as in the my tickets tab
neither historic or sent was fitting.

All in all I decided to remove the mode parameter and make all of the
empty state config parametrised from the outside. This also removed
the circular dependency where ticket history screen used fare contract
list, and the fare contract list used the TicketHistoryMode type from
the ticket history screen.

Also cleaned up some props that either could be made required or
removed.

### DIFF
--- a/src/components/empty-state/index.tsx
+++ b/src/components/empty-state/index.tsx
@@ -1,1 +1,1 @@
-export {EmptyState} from './EmptyState';
+export {EmptyState, type EmptyStateProps} from './EmptyState';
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):
Cool refactor!

PR #4989, opened 2025-02-05
feat: Improve ticket history performance

gorandalum:
Improve the ticket history performance by using FlatList instead of
ScrollView. Some notes on the solution:
- Using a default FlatList without special tuning or extra memoing
seems to improve performance adequately.
- A vertical scrolling FlatList can't be used inside a vertical
scroll ScrollView. This means that we couldn't use the
FullScreenView with the parallax scrolling effect as that requires
an animated ScrollView.
- Didn't change the active tickets tab as that would be a slightly
larger refactor since we would need to do some more changes to get
rid of the ScrollView there. Shouldn't be a problem as active
tickets tab doesn't render many fare contracts.

### DIFF
--- a/src/ticket-history/TicketHistoryScreenComponent.tsx
+++ b/src/ticket-history/TicketHistoryScreenComponent.tsx
@@ -1,6 +1,5 @@
 import {StyleSheet} from '@atb/theme';
 import {
-  type FareContract,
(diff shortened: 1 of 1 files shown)


rosvik (approved):
Sweet!

PR #4994, opened 2025-02-05
refactor: use `@atb-as/utils` package for fare contracts

rosvik:
In order to create a shared definition about what a fare contract is between the app and webshop, I've created the `@atb-as/utils` package: https://github.com/AtB-AS/utils/

Please take a look at the structure of that repo while doing review of this PR. The code has been copied over as it was in the app.

closes https://github.com/AtB-AS/kundevendt/issues/19981

### DIFF
--- a/src/fare-contracts/FareContractAndReservationsList.tsx
+++ b/src/fare-contracts/FareContractAndReservationsList.tsx
@@ -3,12 +3,13 @@ import {FareContractOrReservation} from '@atb/fare-contracts/FareContractOrReser
 import {Reservation} from '@atb/ticketing';
 import {useAnalyticsContext} from '@atb/analytics';
 import {EmptyState} from '@atb/components/empty-state';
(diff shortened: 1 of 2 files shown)


gorandalum (approved):
Nice!

Some notes on the utils package, a couple suggestion that could be done while we are at it:
- Add the `Type` suffix the types to adhere to the [guidelines](https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md#types) on the subject.
- Could `flattenTravelRightAccesses` take in fareContract instead of travelRights? Also don't need to mention "flatten" in the method name since that is what technically is done, but a user of the utility method don't need to be exposed to the technicality of the implementation. Just `getAccesses(fc)` would suffice. 

Even though they are just copied 1-to-1 I think it would be nice to do these slight improvements now before we start using the in webshop and travel-planner as well.

PR #5004, opened 2025-02-10
fix: prevent markdown overflow in global messages

rosvik:
Fixes an issue with markdown overflowing introduced by https://github.com/AtB-AS/mittatb-app/pull/4953

fixes https://mittatb.slack.com/archives/CR75QKX6G/p1739112362530579?thread_ts=1739111540.478569&cid=CR75QKX6G

### DIFF
--- a/src/stacks-hierarchy/Root_TicketInformationScreen.tsx
+++ b/src/stacks-hierarchy/Root_TicketInformationScreen.tsx
@@ -61,7 +61,7 @@ export const Root_TicketInformationScreen = (props: Props) => {
             />
             <Section>
               <GenericSectionItem>
-                <View style={{flex: 1}}>
+                <View style={styles.descriptionContainer}>
                   {fareProductTypeConfig && (
                     <View style={styles.descriptionHeading}>
                       <TransportationIconBoxList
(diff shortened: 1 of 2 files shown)


reidzeibel (approved):
Aha, found the culprit!

tormoseng:
Remember to also merge to the `1.63` release branch 😄

src/stacks-hierarchy/Root_TicketInformationScreen.tsx
```diff
@@ -61,23 +61,28 @@ export const Root_TicketInformationScreen = (props: Props) => {
...
-                    <ThemeText typography="body__primary--bold">
-                      {getTextForLanguage(fareProductTypeConfig.name, language)}
-                    </ThemeText>
-                  </View>
-                )}
-                <ThemeText typography="body__secondary" isMarkdown={true}>
-                  {getTextForLanguage(
-                    preassignedFareProduct.productDescription,
-                    language,
+                <View style={{flex: 1}}>
```
strandlie, line 64: I prefer having even simple styling like this in a `styles`-object, because it gives some context as to why it exists through its name. When its directly applied like this you kind of have to guess.

strandlie (approved):

PR #5007, opened 2025-02-11
feat: In app review

jorelosorio:
Close https://github.com/AtB-AS/kundevendt/issues/3644

simulator_screenshot_6F81415F-53C9-4729-A258-01589445717B

### DIFF
--- a/src/analytics/types.ts
+++ b/src/analytics/types.ts
@@ -16,4 +16,5 @@ export type AnalyticsEventContext =
   | 'Ticket assistant'
   | 'Ticketing'
   | 'Trip details'
-  | 'Trip search';
+  | 'Trip search'
+  | 'InAppReview';
--- a/src/analytics/types.ts
+++ b/src/analytics/types.ts
@@ -17,4 +17,4 @@ export type AnalyticsEventContext =
   | 'Ticketing'
   | 'Trip details'
   | 'Trip search'
-  | 'InAppReview';
+  | 'In App Review';
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -344,8 +344,9 @@ export function getConfig(): RemoteConfig {
   const vehicles_poll_interval =
     values['vehicles_poll_interval']?.asNumber() ??
     defaultRemoteConfig.vehicles_poll_interval;
-  const enable_in_app_review = values['enable_in_app_review']?.asBoolean() ??
-  defaultRemoteConfig.enable_in_app_review;
+  const enable_in_app_review =
+    values['enable_in_app_review']?.asBoolean() ??
+    defaultRemoteConfig.enable_in_app_review;
 
   return {
     customer_feedback_url,
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -73,6 +73,7 @@ export type RemoteConfig = {
   use_flexible_on_egressMode: boolean;
   use_trygg_overgang_qr_code: boolean;
   vehicles_poll_interval: number;
+  enable_in_app_review: boolean;
 };
 
 export const defaultRemoteConfig: RemoteConfig = {
@@ -146,6 +147,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   use_flexible_on_egressMode: true,
   use_trygg_overgang_qr_code: false,
   vehicles_poll_interval: 20000,
+  enable_in_app_review: false,
 };
 
 export type RemoteConfigKeys = keyof RemoteConfig;
@@ -342,6 +344,8 @@ export function getConfig(): RemoteConfig {
   const vehicles_poll_interval =
     values['vehicles_poll_interval']?.asNumber() ??
     defaultRemoteConfig.vehicles_poll_interval;
+  const enable_in_app_review = values['enable_in_app_review']?.asBoolean() ??
+  defaultRemoteConfig.enable_in_app_review;
 
   return {
     customer_feedback_url,
@@ -410,6 +414,7 @@ export function getConfig(): RemoteConfig {
     use_flexible_on_egressMode,
     use_trygg_overgang_qr_code,
     vehicles_poll_interval,
+    enable_in_app_review,
   };
 }
 
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -51,6 +51,7 @@ export type RemoteConfig = {
   enable_vehicle_operator_logo: boolean;
   enable_vehicles_in_map: boolean;
   enable_vipps_login: boolean;
+  enable_in_app_review: boolean;
   favourite_departures_poll_interval: number;
   feedback_questions: string;
   fetch_id_token_retry_count: number;
@@ -73,7 +74,6 @@ export type RemoteConfig = {
   use_flexible_on_egressMode: boolean;
   use_trygg_overgang_qr_code: boolean;
   vehicles_poll_interval: number;
-  enable_in_app_review: boolean;
 };
 
 export const defaultRemoteConfig: RemoteConfig = {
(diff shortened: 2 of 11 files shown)


src/utils/use-in-app-review.ts
```diff
@@ -0,0 +1,59 @@
...
+        analytics.logEvent(
+          'InAppReview',
+          'In-app review failed to be presented',
+        );
+      }
+    } catch (error) {
+      // Log and report any errors encountered during the in-app review request.
+      // For detailed error codes, refer to:
+      // https://github.com/MinaSamir11/react-native-in-app-review#errors-and-google-play-app-store-error-codes
+      Bugsnag.notify(error as any);
```
reidzeibel, line 54:
```suggestion
      notifyBugsnag(error as any);
```

have to add import: `import {notifyBugsnag} from '../utils/bugsnag-utils';`

reidzeibel: Also, should we `throw` anything here? It doesn't have to be, just thinking out loud. 🤔

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -216,7 +222,12 @@ export const TravelDetailsMapScreenComponent = ({
       <View style={controlStyles.backArrowContainer}>
         <BackArrow
           accessibilityLabel={t(MapTexts.exitButton.a11yLabel)}
-          onBack={onPressBack}
+          onBack={() => {
```
reidzeibel, line 225: What if the user navigates back through some form of navigation? Will it still show the in-app review prompt?

(e.g. in some android devices you have the on-screen navigation button, or you can do it through gesture)

src/travel-details-screens/components/Trip.tsx
```diff
@@ -185,6 +187,9 @@ export const Trip: React.FC<TripProps> = ({
                           vehicleWithPosition: legVehiclePosition,
                           mode: leg.mode,
                           subMode: leg.transportSubmode,
+                          onScreenClose: async () => {
+                            await requestReview();
+                          },
```
rosvik, line 192: Could we remove the await here, and make it fire-and-forget? (Same for departure details)

```suggestion
                          onScreenClose: () => {
                            requestReview();
                          },
```

src/analytics/types.ts
```diff
@@ -16,4 +16,5 @@ export type AnalyticsEventContext =
   | 'Ticket assistant'
   | 'Ticketing'
   | 'Trip details'
-  | 'Trip search';
+  | 'Trip search'
+  | 'InAppReview';
```
rosvik, line 20: I guess this should be `'In app review'` to follow the naming convention used elsewhere 🤓

src/remote-config.ts
```diff
@@ -73,6 +73,7 @@ export type RemoteConfig = {
   use_flexible_on_egressMode: boolean;
   use_trygg_overgang_qr_code: boolean;
   vehicles_poll_interval: number;
+  enable_in_app_review: boolean;
```
rosvik, line 76: We should try to keep these flags sorted alphabetically, so they're easier to find in the debug screen.

rosvik (commented):
This looks good! Some nitpicks on the code, and also you should add a check on whether the live bus feature was active. Shouldn't show the prompt without of bus in map.

Otherwise, seems to be working well locally 💪

src/utils/use-in-app-review.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    // Changed to an async function
+    // Check if in-app review functionality is available on this device.
+    if (!InAppReview.isAvailable()) {
+      analytics.logEvent(
+        'InAppReview',
+        'Device does not support in-app review',
+      );
+      return;
+    }
```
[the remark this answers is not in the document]
jorelosorio: Maybe, or because the system does not meet the in-app review requirements. But we support versions for that.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
jorelosorio: True, I added extra code to handle this, good catch.

src/utils/use-in-app-review.ts
jorelosorio: Like an exception? the app rating is fired after an action I think is best to fail in silence and report it.

reidzeibel: Yes, then no worries with the throw!

reidzeibel (approved):

src/utils/use-in-app-review.ts
```diff
@@ -0,0 +1,57 @@
...
+export function useInAppReviewFlow() {
+  const analytics = useAnalyticsContext();
+  const {isInAppReviewEnabled} = useFeatureTogglesContext();
+
+  const requestReview = useCallback(async () => {
+    if (!isInAppReviewEnabled) {
+      return;
+    }
+
+    // Changed to an async function
```
rosvik, line 16:
```suggestion
```

rosvik (approved):
Couple of more nits, otherwise good stuff 👍

src/utils/use-in-app-review.ts
```diff
@@ -0,0 +1,57 @@
...
+import {useAnalyticsContext} from '@atb/analytics';
+import {useFeatureTogglesContext} from '@atb/feature-toggles';
+import {useCallback} from 'react';
+import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
+
+export function useInAppReviewFlow() {
+  const analytics = useAnalyticsContext();
+  const {isInAppReviewEnabled} = useFeatureTogglesContext();
+
+  const requestReview = useCallback(async () => {
```
gorandalum, line 11: If we want the `requestReview` to just be fire and forget, it should return `void` instead of `Promise<void>`, so IDEs doesn't complain about unhandled promise.

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -51,6 +51,11 @@ export type TravelDetailsMapScreenParams = {
   mode?: AnyMode;
   subMode?: AnySubMode;
   mapFilter?: MapFilterType;
+  // An event to notify that the screen close button was pressed and can take some actions before it is closed,
+  // it differs from `onPressBack` which is called/overridden by the parent call-in Dashboard_TravelDetailsMapScreen.tsx
+  // By doing this, it prevents replacing `onPressBack` with a different action that prevents calling the navigation
+  // at the top view or avoids overflow in the stack.
+  onScreenClose?: () => void;
```
gorandalum, line 58: Shouldn't have functions in screen params types, since functions are not serialisable. 
https://reactnavigation.org/docs/5.x/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state

jorelosorio: I removed it, and instead, I used a different approach.

src/utils/use-in-app-review.ts
```diff
@@ -0,0 +1,98 @@
...
+              'In App Review',
+              'Review prompt launched successfully',
+            );
+          } else {
+            analytics.logEvent(
+              'In App Review',
+              'In-app review failed to be presented',
+            );
+          }
+        }
```
rosvik, line 87: Shouldn't check `if (hasFlowFinishedSuccessfully)` twice here. This will never reach the `else` -case.

rosvik (commented):
Noticed one issue with the hook, but otherwise, it looks like this should work 👍 

I spent some time thinking about the log messages here, since it's kind of important that we know what these log messages mean for when we use it to make decisions about the next iteration. To me, it looks like these are the situations that can appear, and the log messages I'd expect:

- Device doesn't support in-app review (e.g. OS is too old)
	- `Device does not support in-app review`
- User has disabled the prompts system-wide. (System Preferences > App Store > In-App Ratings & Reviews)
	- `Device does not support in-app review`
- We decide not to ask because it has been shown before
	- No logs
- We decide not to ask because we've tried the last 5 days
	- No logs
- We decide to ask, but the OS won't allow us to ask for review
	- `In-app review failed to be presented`
- We decide to ask, and it was successful
	- `Review prompt launched successfully`

Is this how we expect this to work in practice? Such a list might be good input for testers.

PR #5010, opened 2025-02-11
fix: phone input overflow

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/20017

Solution: 
* Fixes so that phone number input inside PhoneInputSectionItem has correct width
  * Now functions the same as all other inputs when content is too long (size scales down for placeholder, ... at the end for placeholder and input)
* _number-pad_ is now default keyboard type for PhoneInputSectionItem

 Screenshots:


rosvik (approved):
💪

PR #5016, opened 2025-02-12
feat: added tripcard

Dahly96:
closes: https://github.com/AtB-AS/kundevendt/issues/19915

Skjermbilde 2025-02-13 kl  08 47 02


src/mobility/components/TripCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  bookingId: string;
+};
+
+export const TripCard = ({bookingId}: Props) => {
+  const styles = useStyles();
+  const {serverNow} = useTimeContext();
+  const {theme} = useThemeContext();
+  const lineColor = theme.color.background.neutral[0].background;
+  const backgroundColor = useTransportColor('scooter', 'escooter');
+  const {data: booking} = useShmoBookingQuery(bookingId, 30000);
```
marius-at-atb, line 20: Won't hurt to set 15000ms here I think

marius-at-atb:
Oh and one more thing - In Figma there is no separation line right beneath the LineWithVerticalBars.

src/mobility/components/TripCard.tsx
marius-at-atb, line 20: In the sketches there is also a disabled mode.
Perhaps TripCard can take in a parameter "isDisabled", and if it is true, use
- `Foreground/Dynamic/Disabled` as background color
- `animate=false` for `LineWithVerticalBars`

gorandalum (commented):
Just a note from my side:
I think it would be beneficial to find another term than "trip" to use in the context of mobility. Trip is a word strongly associated with the trip search part of the app.

It is beneficial to have a [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html) in the team, and using "trip" both for trip search and mobility rentals introduces some ambiguity in our terminology.

src/mobility/queries/use-shmo-booking-query.tsx
```diff
@@ -9,7 +9,10 @@ export const getShmoBookingQueryKey = (
   acceptLanguage: string,
 ) => ['GET_SHMO_BOOKING_QUERY_KEY', bookingId, acceptLanguage];
 
-export const useShmoBookingQuery = (bookingId?: ShmoBooking['bookingId']) => {
+export const useShmoBookingQuery = (
+  bookingId?: ShmoBooking['bookingId'],
+  refetchInterval: number | false = false,
```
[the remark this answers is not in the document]
Dahly96: In the docs it does default to false, so im just keeping the defaultvalue as is if its not used.

marius-at-atb (approved):
🙌

rosvik (approved):

PR #5017, opened 2025-02-12
fix: use frontText instead of lineName in departures widget

rosvik:
fixes issue described in https://mittatb.slack.com/archives/C0116FMPX4Y/p1739351827212559

Note: This adds destination display with `frontText` and `via` to the widget, but I haven't added logic to actually display the `via` data, only `frontText`.

### DIFF
--- a/ios/AtbAppIntent/IntentHandler.swift
+++ b/ios/AtbAppIntent/IntentHandler.swift
@@ -27,7 +27,7 @@ class IntentHandler: INExtension, UseLocationIntentHandling {
 
                     let favoriteDeparture = FavoriteDeparture(
                         identifier: departure.id,
-                        display: "\(departure.lineLineNumber) \(departure.lineName ?? NSLocalizedString("all_variations", comment: ""))"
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):
Looks ok from me, maybe @jorelosorio have some comments?

gorandalum (approved):
Looks good! But should probably add `via` as soon as possible. I know at least Fram uses the `via` field in their trip data.

jorelosorio (approved):
Nice work, I miss Swift hehe

rosvik:
> But should probably add via as soon as possible.

@jorelosorio I guess that's a Swift task up for grabs ✌️

marius-at-atb:
The relevant function:
https://github.com/AtB-AS/mittatb-app/blob/master/src/travel-details-screens/utils.ts#L122-L143

tormoseng:
✅  Tested OK! 

Issue for handling `via`: https://github.com/AtB-AS/kundevendt/issues/20038

PR #5019, opened 2025-02-13
refactor: Remove transitionPreset not serialisable warning

gorandalum:
### Background

We have the possibility to specify a transition when navigating, but
using this functionality gave this warning:
"Non-serializable values were found in the navigation state"

Instead of passing the actual TransitionPreset object as a param, now
the param instead is a string which will be mapped to a
TransitionPreset object before usage.

In addition renamed the param to transitionOverride to signal that
the param is ment to be used only for overriding the transition
specified in the stack navigator.

Screenshot 2025-02-13 at 12 39 46

### Acceptance criteria
- [ ] Navigation animations in the app are unchanged. Do some random sampling.


rosvik (approved):
Nice!

tormoseng:
✅

PR #5021, opened 2025-02-13
refactor: Make backgroundColor mandatory on secondary/tertiary buttons

gorandalum:
Secondary and tertiary buttons always need to know which background
they are placed on to use the correct text and border color. We should
not rely on just setting a default value in the button and hope it is
correct. It might even be that the default value works in the theme of
one transport company, but not for another transport company.

### Acceptance criteria
- [ ] Secondary and tertiary buttons around the app looks as before. Take a look in DesignSystem screen and in Storybook, as well as some random sampling around the app.


rosvik (approved):

tormoseng:
✅

PR #5028, opened 2025-02-17
made the tripDetailsection more accessible

Dahly96:
https://github.com/AtB-AS/kundevendt/issues/19889


gorandalum (approved):

src/mobility/components/ShmoTripDetailsSectionItem.tsx
```diff
@@ -30,22 +30,32 @@ export const ShmoTripDetailsSectionItem = ({
   const timeUsed = getTimeBetweenFormatted(startDateTime, endDateTime);
   return (
     <View style={[topContainer, styles.container]}>
-      {withHeader && (
-        <View style={styles.heading}>
-          <ThemeText typography="body__secondary" color="secondary">
+      <View style={styles.leftSection} accessible={true}>
```
rosvik, line 33: If I remember correctly, there should always be a `accessibilityLabel` for views with `accessible={true}` to make it work properly on android:

> When a view is marked as accessible, it is a good practice to set an accessibilityLabel on the view, so that people who use VoiceOver or TalkBack know what element they have selected.

https://reactnative.dev/docs/accessibility#accessibilitylabel

marius-at-atb: Perhaps we could add en eslint rule for this?

marius-at-atb (approved):

rosvik (approved):

PR #5031, opened 2025-02-19
feat: implement token status endpoint

reidzeibel:
closes: https://github.com/AtB-AS/kundevendt/issues/19963

Changes: 
- Implement endpoint for sending token status to /token/v1/status.
- Send token status along with intercom metadata update.
- Add condition to send the token status when the query status is not `loading`, so that only the final error/success state will be sent. Without the query status, it will send the status update when the query is still on initialization/empty state.

Please let me know if checking for `query status !== 'loading'` is not the correct solution here, I've looked into `isLoading` and `isFetching` and it seems that checking on `isLoading` is the way to go.

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -9,7 +9,6 @@ import React, {
 import {useAuthContext} from '@atb/auth';
 import {useRemoteConfigContext} from '@atb/RemoteConfigContext';
 import {
-  IntercomTokenStatus,
   MobileTokenStatus,
   RemoteToken,
   Token,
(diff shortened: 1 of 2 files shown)


src/mobile-token/MobileTokenContext.tsx
```diff
@@ -114,14 +114,28 @@ export const MobileTokenContextProvider: React.FC = ({children}) => {
...
+        'AtB-Mobile-Token-Error-Correlation-Id': tokenErrorCorrelationId,
+      });
+
+      // token status endpoint
+      tokenService.postTokenStatus(
+        tokenId,
+        tokenStatus,
+        tokenErrorCorrelationId,
+      );
+    }
```
gorandalum, line 137: Can test if this would be more easily readable with splitting the `nativeTokenStatus !== 'loading'` into different ifs for success/error. Pseudocode:
```ts
if (nativeTokenStatus === 'success') {
  updateMetadata({
    'AtB-Mobile-Token-Id': tokenId,
    'AtB-Mobile-Token-Status': nativeToken.attested ? 'attested' : 'non_attested',
    'AtB-Mobile-Token-Error-Correlation-Id': undefined,
  });
} else if (nativeTokenStatus === 'error') {
  updateMetadata({
    'AtB-Mobile-Token-Id': undefined,
    'AtB-Mobile-Token-Status': 'error',
    'AtB-Mobile-Token-Error-Correlation-Id': traceId.current,
  });
}
```

gorandalum (approved):

strandlie (approved):

src/mobile-token/MobileTokenContext.tsx
reidzeibel: Seems like a good idea, easier to distinguish as well

PR #5033, opened 2025-02-19
feat: pay with bonus point checkbox

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/19988

### DIFF
--- a/src/bonus/BonusPriceTag.tsx
+++ b/src/bonus/BonusPriceTag.tsx
@@ -27,6 +27,7 @@ export const BonusPriceTag = ({price, ...props}: Props) => {
 
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   container: {
+    backgroundColor: theme.color.background.neutral[0].background,
     borderColor: theme.color.foreground.inverse.secondary,
     borderWidth: theme.border.width.slim,
     borderRadius: theme.border.radius.regular,
--- a/src/mobility/components/PayWithBonusPointsCheckbox.tsx
+++ b/src/mobility/components/PayWithBonusPointsCheckbox.tsx
@@ -1,0 +1,105 @@
+import React, {ComponentProps} from 'react';
+import {BonusPriceTag} from '@atb/bonus';
+import {Checkbox} from '@atb/components/checkbox';
+import {PressableOpacity} from '@atb/components/pressable-opacity';
+import {GenericSectionItem} from '@atb/components/sections';
+import {ThemeText} from '@atb/components/text';
+import {ThemeIcon} from '@atb/components/theme-icon';
+import {BonusProductType} from '@atb/configuration/types';
+import {StyleSheet, useThemeContext} from '@atb/theme';
+import {StarFill} from '@atb/assets/svg/mono-icons/bonus';
+import {
+  BonusProgramTexts,
+  getTextForLanguage,
+  useTranslation,
+} from '@atb/translations';
+import {View} from 'react-native';
+
+type Props = ComponentProps<typeof View> & {
+  userBonusPoints: number;
+  bonusProduct: BonusProductType;
+  isChecked: boolean;
+  onPress: () => void;
+};
+
+export const PayWithBonusPointsCheckbox = ({
+  userBonusPoints,
+  bonusProduct,
+  isChecked,
+  onPress,
+  ...props
+}: Props) => {
+  const styles = useStyles();
+  const {t, language} = useTranslation();
+  const {theme} = useThemeContext();
+  const background = theme.color.background.neutral[0].background;
+  const selectedBackground = theme.color.interactive[2].active.background;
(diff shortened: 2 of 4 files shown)


src/mobility/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+      ]}
+    >
+      <PressableOpacity
+        onPress={onPress}
+        accessibilityRole="button"
+        accessibilityLabel={t(BonusProgramTexts.checkbox.a11yLabel)}
+        accessibilityHint={t(BonusProgramTexts.checkbox.a11yHint)}
+        disabled={disabled}
+        style={[styles.container, disabled && {opacity: 0.5}]}
+      >
```
gorandalum, line 59: There exists a `GenericClickableSectionItem` that is ment to be used for this pattern. Take a look if that can be used instead.

jorunnl: I struggeled with making the 'checked' styling work as intended when using the `GenericClickableSectionItem` because `Property 'style' does not exist on type 'IntrinsicAttributes & { onPress?(): void; } & AccessibilityProps & BaseSectionItemProps & { children?: ReactNode; } & RefAttributes<...>'`

image

So I used the GenericSectionItem to not make changes to the GenericClickableSectionItem. 
Do you think of a better solution? Should I make changes to `GenericClickableSectionItem` to make it work?

strandlie (approved):

src/mobility/components/PayWithBonusPointsCheckbox.tsx
jorunnl: A better solution in progress in:
- https://github.com/AtB-AS/mittatb-app/pull/5039

strandlie (approved):

src/mobility/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {StarFill} from '@atb/assets/svg/mono-icons/bonus';
+import {
+  BonusProgramTexts,
+  getTextForLanguage,
+  useTranslation,
+} from '@atb/translations';
+import {View} from 'react-native';
+
+type Props = SectionProps & {
+  userBonusPoints: number;
```
marius-at-atb, line 22: `userBonusPoints` can be extracted with a hook inside the component instead of being a parameter.

src/translations/screens/subscreens/BonusProgram.ts
```diff
@@ -1,36 +1,57 @@
...
-      title: _(
-        'Hvordan funker Bonus?',
-        'How does the Bonus work?',
-        'Korleis fungerer bonus?',
-      ),
-      description: _(
-        'Tjen poeng hver gang du kjøper en billett for buss eller trikk i sone A.',
-        'Earn points every time you buy a bus or tram ticket in zone A.',
-        'Ten poeng kvar gong du kjøper ein billett for buss eller trikk i sone A.',
+  costA11yLabel: (price: number) =>
```
marius-at-atb, line 4: For `costA11yLabel` and `yourBonusPointsA11yLabel` as well, let's standardise the use of price vs points vs amount.

marius-at-atb (approved):
Nice work!

src/components/sections/items/GenericClickableSectionItem.tsx
```diff
@@ -2,15 +2,20 @@ import React, {PropsWithChildren, forwardRef} from 'react';
...
+  PressableOpacityProps,
+} from '@atb/components/pressable-opacity';
 
 type Props = PropsWithChildren<
   SectionItemProps<
     {
       onPress?(): void;
-    } & AccessibilityProps
+    } & AccessibilityProps &
+      PressableOpacityProps
```
gorandalum, line 15: Maybe just add `disabled` instead of the full `PressableOpacityProps`? We want to expose parameters on a need-to-have basis.

In addition, could the opacity when disabled be handled inside `GenericClickableSectionItem`? So parent components that specify `disabled={true}` doesn't need to do the styling themselves.

src/bonus/BonusPriceTag.tsx
```diff
@@ -7,26 +7,32 @@ import {BonusProgramTexts, useTranslation} from '@atb/translations';
...
 
   return (
     <View {...props} style={[styles.container, props.style]}>
-      <ThemeText>{price}</ThemeText>
+      <ThemeText
+        accessible={true}
+        accessibilityLabel={t(BonusProgramTexts.costA11yLabel(amount))}
+      >
+        {amount}
+      </ThemeText>
```
gorandalum, line 23: There is something iffy about the accessibility stuff here. I think we should either do it correctly, or not at all (if it is only used in map). We can have a talk about it.

PR #5035, opened 2025-02-20
feat: Turn on Posthog auto capture

gorandalum:
- Turns on autocomplete for Posthog. Enabled for screen navigations and app open/close, but not all touches as that generated a ton of events.
- Needed to move the `AnalyticsProvider` inside of `NavigationContainer` for the autocapture of navigation events to work.
- Removed the custom app status listener.
- Slightly simplified creation of Posthog client with a `useMemo` instead of `useState` and `useEffect`.
- Removed the open ShareTravelHabits screen event as it is now being autocaptured.
- Removed the `client.reset` since all user id changes generates a new `client.identify`. The reset function is there for going back to the unidentified state, which our users never should be in.


rosvik (approved):
LGTM 👍

PR #5036, opened 2025-02-20
feat: add self service refund of fare contracts

rosvik:
https://github.com/user-attachments/assets/761c3f76-fa63-41cd-a0f3-5903c6c015a0

closes https://github.com/AtB-AS/kundevendt/issues/20058

### DIFF
--- a/src/modules/fare-contracts/components/RefundBottomSheet.tsx
+++ b/src/modules/fare-contracts/components/RefundBottomSheet.tsx
@@ -1,0 +1,92 @@
+import {getAxiosErrorMetadata} from '@atb/api/utils';
+import {Confirm} from '@atb/assets/svg/mono-icons/actions';
+import {
+  BottomSheetContainer,
+  useBottomSheetContext,
+} from '@atb/components/bottom-sheet';
+import {Button} from '@atb/components/button';
+import {MessageInfoBox} from '@atb/components/message-info-box';
+import {GenericSectionItem, Section} from '@atb/components/sections';
+import {ThemeText} from '@atb/components/text';
+import {StyleSheet} from '@atb/theme';
+import {refundFareContract} from '@atb/ticketing';
+import {FareContractTexts, useTranslation} from '@atb/translations';
+import Bugsnag from '@bugsnag/react-native';
+import React, {useState} from 'react';
+import {ScrollView} from 'react-native-gesture-handler';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+type Props = {
+  fareContractId: string;
+};
+
+export const RefundBottomSheet = ({fareContractId}: Props) => {
+  const styles = useStyles();
+  const {t} = useTranslation();
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<boolean>(false);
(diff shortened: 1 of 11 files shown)


src/modules/fare-contracts/components/RefundBottomSheet.tsx
```diff
@@ -0,0 +1,111 @@
...
+            type="error"
+          />
+        )}
+        <Section>
+          <GenericSectionItem type="spacious">
+            <ThemeText>
+              {t(FareContractTexts.refund.bottomSheetDescription)}
+            </ThemeText>
+          </GenericSectionItem>
+        </Section>
```
marius-at-atb, line 62: UX question - should this be hidden when `refundOptions?.is_refundable === false`?

src/modules/fare-contracts/components/RefundBottomSheet.tsx
```diff
@@ -0,0 +1,111 @@
...
+          loading={
+            refundStatus === 'loading' || refundOptionsStatus === 'loading'
+          }
+          interactiveColor={theme.color.interactive.destructive}
+        />
+        <Button
+          backgroundColor={theme.color.background.neutral[1]}
+          onPress={() => {
+            close();
+          }}
```
marius-at-atb, line 89: Can be just `onPress={close}`

src/ticketing/types.ts
```diff
@@ -193,3 +193,11 @@ export type AddPaymentMethodResponse = {
   recurring_payment_id: number;
   terminal_url: string;
 };
+
+/**
+ * Defined by RefundOptionsResponse in
+ * https://github.com/AtB-AS/sales/blob/main/sales-service/src/handlers/sales/refund.rs#L31
+ */
```
marius-at-atb, line 200: Good to have comments like this :+1:

marius-at-atb (approved):
Very nice!
I guess you can remove "wip" from the PR title now? 😄

src/modules/fare-contracts/components/RefundBottomSheet.tsx
rosvik:
```suggestion
          onPress={close}
```

src/ticketing/types.ts
rosvik: I guess the line number will change over time though, so removing that for now ✌️

```suggestion
/**
 * Defined by RefundOptionsResponse in
 * https://github.com/AtB-AS/sales/blob/main/sales-service/src/handlers/sales/refund.rs
 */
```

src/modules/fare-contracts/components/RefundBottomSheet.tsx
rosvik: Had a talk with @hannahATB, and no, we want to show the info even when the ticket is not available for refund.

rosvik:
Slight change of plans here, and we decided to only show the refund option if the ticket is refundable. Updated screen recording. the major change since last review is here: f05c30f26eaa046b2ec851e38fd40b3cef2e9070

reidzeibel (approved):
new changes looks good!

marius-at-atb (approved):

PR #5038, opened 2025-02-21
feat:login scooter sheet

Dahly96:
closes https://github.com/AtB-AS/kundevendt/issues/20078

Added functionality for the login button in the scooter sheet


reidzeibel:
Could you update the PR title to follow the convention here: https://github.com/AtB-AS/mittatb-app/blob/master/CONTRIBUTING.md

Also don't forget to add the acceptance criteria on the main issue.

src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
```diff
@@ -49,6 +51,9 @@ export const useUpdateBottomSheetWhenSelectedEntityChanges = (
     useBottomSheetContext();
   const analytics = useMapSelectionAnalytics();
   const navigation = useNavigation<RootNavigationProps>();
+  const hasReservationOrAvailableFareContract =
+    useHasReservationOrAvailableFareContract();
+  const {enable_vipps_login} = useRemoteConfigContext();
```
reidzeibel, line 56: Maybe should also add this on the dependency array for the hook, we are currently suppressing the warning using eslint, but since we use it in this hook `src/components/map/hooks/use-auto-select-map-item.tsx`, seems like it should also be added here.

reidzeibel:
Also there's a console.log in this file : `src/mobility/components/ShmoActionButtons.tsx`

src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
Dahly96: yes, done

strandlie (approved):
Very clean and readable code! I like it!

src/components/map/hooks/use-auto-select-map-item.tsx
```diff
@@ -92,6 +97,22 @@ export const useAutoSelectMapItem = (
...
+                    );
+                  } else if (enable_vipps_login) {
+                    navigation.navigate('Root_LoginOptionsScreen', {
+                      showGoBack: true,
+                      transitionOverride: 'slide-from-bottom',
+                    });
+                  } else {
+                    navigation.navigate('Root_LoginPhoneInputScreen', {});
+                  }
+                }}
```
marius-at-atb, line 115: Since the loginCallback is identical for both places, it should only be defined once:
- Should this function be in its own hook?
- Perhaps it also doesn't need to be a prop for ScooterSheet -> maybe just use it directly inside `ShmoActionButton`?

Dahly96: My understanding is that we should limit the amount of components directly implementing navigation. So i let it be at the level where navigation is used. And that is why im keeping the function as is

marius-at-atb: Right, yes that can be its own issue

PR #5039, opened 2025-02-21
feat: add active and interActiveColor to section item

jorunnl:
Adds active and interActiveColor as props for all section items

Uses active for radio button section items instead of separate styling

Part of https://github.com/AtB-AS/kundevendt/issues/19988

### DIFF
--- a/src/components/sections/use-section-item.ts
+++ b/src/components/sections/use-section-item.ts
@@ -14,6 +14,7 @@ export type BaseSectionItemProps = {
   radiusSize?: keyof Theme['border']['radius'];
   testID?: string;
   active?: boolean;
+  interactiveColor?: InteractiveColor;
 };
 
 export type SectionReturnType = {
@@ -30,9 +31,10 @@ export function useSectionItem({
   radius,
   radiusSize,
   active,
+  interactiveColor: customInteractiveColor,
 }: BaseSectionItemProps): SectionReturnType {
   const {theme} = useThemeContext();
-  const interactiveColor = getInteractiveColor(theme);
+  const interactiveColor = customInteractiveColor ?? getInteractiveColor(theme);
 
   const topContainer: ViewStyle = {
     ...mapToPadding(theme, type),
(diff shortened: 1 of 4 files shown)


src/components/sections/use-section-item.ts
```diff
@@ -8,27 +13,37 @@ export type BaseSectionItemProps = {
...
+  const interactiveColor = customInteractiveColor ?? getInteractiveColor(theme);
 
   const topContainer: ViewStyle = {
     ...mapToPadding(theme, type),
     ...mapToBorderRadius(theme, radiusSize, radius),
     backgroundColor: transparent
       ? undefined
-      : theme.color.background.neutral[0].background,
+      : interactiveColor[active ? 'active' : 'default'].background,
+    borderColor: interactiveColor[active ? 'outline' : 'default'].background,
```
gorandalum, line 45: Do the section item now end up with a border even if `transparant: true`?

gorandalum (approved):
Looks good! Just take a look at the comment regarding border color.

src/components/sections/use-section-item.ts
```diff
@@ -8,27 +13,37 @@ export type BaseSectionItemProps = {
...
+  interactiveColor?: InteractiveColor;
 };
 
 export type SectionReturnType = {
   topContainer: ViewStyle;
   contentContainer: ViewStyle;
+  interactiveColor: InteractiveColor<ContrastColor>;
 };
 
+const getInteractiveColor = (theme: Theme) => theme.color.interactive[2];
```
strandlie, line 26: I'm not sure I see the value of this being a function, when it is this short

gorandalum: It is a pattern we have around the app to set a color for a whole screen/component. It is ment to simulate the pattern where the color is set as a const at the top of the file, but it needs to be a function as it requires access to the `theme`.

Screenshot 2025-02-24 at 15 41 16

src/components/sections/use-section-item.ts
jorunnl: Yes. I think that makes sense since otherwise `transparent non-active` and `transparent active` will look the same

jorunnl: Good catch, fixed now so that transparent only have border color if it is active:
image

PR #5040, opened 2025-02-24
fix: renew token correctly in v3 mobile token

reidzeibel:
addresses https://github.com/AtB-AS/kundevendt/issues/19973#issuecomment-2671012285

- Moved the error handling to outer function on the `validate` service.
- Removed error logging on token renewal handling, it is not an error.
- Changed `Mobiletoken sdk remote logger error` to `Mobiletoken sdk remote logger caught an error` for clarity, that the error is caught by the remote logger, and **NOT** that the remote logger is error.
- Improved renewal handling by returning `unspecified` handling to prevent a `reset` call derived from the sdk suggestion.

Acceptance Criteria:
Same as https://github.com/AtB-AS/kundevendt/issues/19973

### DIFF
--- a/src/mobile-token/abtClientLogger.ts
+++ b/src/mobile-token/abtClientLogger.ts
@@ -33,7 +33,7 @@ export const remoteLogger: RemoteLogger = {
     notifyBugsnag(err, {
       errorGroupHash: 'token',
       metadata: {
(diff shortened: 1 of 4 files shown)


gorandalum (approved):

tormoseng:
✅

PR #5042, opened 2025-02-25
fix: handle reattestation error during validate

reidzeibel:
part of https://github.com/AtB-AS/kundevendt/issues/19973

Currently reattestation of mobile token leads to a `reset` instead, in this PR I added a handling so that reattestation issues are handled appropriately.

The problem is that inside the `validate` call, we have 2 function. The outer part is provided from the SDK to add a reattestation header to wrap the actual validate function, while the inner part is calling the validate endpoint. (see the code)

- If we put `catch(handleError)` on the inner part: Reattestation works, Renewal doesn't work.
- If we put `catch(handleError)` on the outer part: Reattestation doesn't work, Renewal works.

So the solution is to differentiate the error handling for both part. Reattestation is handled on the inner part, while Renewal and others are handled on the outer part.

**Acceptance Criteria:**
- [ ] Reattestation works
- [ ] Renewal from expiration works
- [ ] Upgrading from v2 to v3 works

### DIFF
--- a/src/mobile-token/tokenService.ts
+++ b/src/mobile-token/tokenService.ts
@@ -23,6 +23,10 @@ import {
 import {getDeviceName} from 'react-native-device-info';
 import {isRemoteTokenStateError, parseBffCallErrors} from './utils';
 import {abtClient} from './mobileTokenClient';
+import {
+  TokenReattestationRemoteTokenStateError,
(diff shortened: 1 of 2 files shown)


gorandalum (approved):

tormoseng:
✅

PR #5043, opened 2025-02-26
feat: onboarding comp

Dahly96:
closes https://github.com/AtB-AS/kundevendt/issues/20075

Added the fundation for the onboarding components.But i decided to split it up into 1 PR pr. component. So in this PR the location component is finished. But payment and terms is not done.

image

image


strandlie (approved):

reidzeibel:
Don't forget to ensure that the merge commit uses the convention, so the release notes generator can include the commit in the list (e.g. feat, fix)

src/stacks-hierarchy/Root_ShmoOnboardingScreen.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {ShmoRequirementEnum} from '@atb/mobility/types';
+import {useShmoRequirements} from '@atb/mobility/use-shmo-requirements';
+import {useNavigation} from '@react-navigation/native';
+
+export type ShmoOnboardingProps = {};
+
+export const Root_ShmoOnboardingScreen = ({}: ShmoOnboardingProps) => {
+  const {requirements, hasBlockers} = useShmoRequirements();
+
+  const navigation = useNavigation();
```
gorandalum, line 14: Should not use `useNavigation`. Get it from the `NavigationProp` instead.

src/mobility/components/onboarding/LocationBlocker.tsx
```diff
@@ -0,0 +1,37 @@
...
+export type LocationBlockerProps = {};
+
+export const LocationBlocker = ({}: LocationBlockerProps) => {
+  const {t} = useTranslation();
+
+  const {requestLocationPermission} = useGeolocationContext();
+
+  const setCurrentLocationOrRequest = useCallback(async () => {
+    await requestLocationPermission();
+  }, [requestLocationPermission]);
```
gorandalum, line 18: Note that this might often be blocked, since the user has not accepted it in the normal onboarding. Should probably help the user navigate to the settings.

src/mobility/components/onboarding/LocationBlocker.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <OnboardingScreenComponent
+        illustration={<MyLocation height={220} />}
+        title={t(MobilityTexts.shmoRequirements.location.locationTitle)}
+        description={t(
+          MobilityTexts.shmoRequirements.location.locationDescription,
+        )}
+        buttonText={t(MobilityTexts.shmoRequirements.location.locationButton)}
+        buttonOnPress={setCurrentLocationOrRequest}
+      />
+    </FullScreenView>
```
gorandalum, line 35: Some architectural comments:

**Screen components are generally full screen**
Here the `OnboardingScreenComponent` needs to be wrapped in a `FullScreenView`, which means that it doesn't really follow the pattern we have for screen components.

**Screen component usage pattern**
The normal pattern for using screen components are:
 ```
 SomeScreen -> SomeScreenComponent -> OrdinaryComponents
 ```
 
 Here the pattern is:
 ```
Root_ShmoOnboardingScreen -> LocationBlocker -> OnboardingScreenCompoent
```

Which makes it seem like the pattern is this:
 ```
 SomeScreen -> OrdinaryComponent -> SomeScreenComponent
 ```

**Nesting components** 
Do we really need the middle layer with those `Blocker`-components? Deep nesting of components bring complexity. Could it be just done with something like this?
 ```ts
 const getBlockerTexts = (requirements) => {
   if (requirements.includes...) {
     return {title, description, image, buttonText}
   } else if (requirements.includes) {
     …
   }
   …
 }
 ```
 
And then use the `OnboardingScreenComponent` directly in the `Root_ShmoOnboardingScreen` with these texts?

File: src/mobility/components/onboarding/LocationBlocker.tsx (hunk 2 above)
Comment (Dahly96) [reply src/mobility/components/onboarding/LocationBlocker.tsx:18]: Not sure what you are refering to. This will open the settings dialog if the user already has declined it in the past

File: src/mobility/components/onboarding/LocationBlocker.tsx (hunk 3 above)
Comment (Dahly96) [reply src/mobility/components/onboarding/LocationBlocker.tsx:35]: We could use the OnboardingScreenComp directly in root. I was trying to separate the logic from location, payment and terms. Then everything needs to be inn the root file. But im going to see if i can create reusable hooks for the payment flow, location function and the terms. So it might not be as bad as i think. But for this PR im only creating the logic for location

File: src/mobility/components/onboarding/LocationBlocker.tsx (hunk 2 above)
Comment (gorandalum) [reply src/mobility/components/onboarding/LocationBlocker.tsx:18]: Great! Didn't remember that.

File: src/translations/screens/subscreens/MobilityTexts.ts [hunk 4]
```diff
@@ -24,26 +24,54 @@ export const MobilityTexts = {
...
+    payment: {
+      paymentTitle: _(
+        'Vi trenger et betalingskort',
+        'We need a payment card',
+        'Vi treng eit betalingskort',
+      ),
+      paymentDescription: _(
+        'Du må legge til et Visa, Mastercard eller American express for å bruke en el-sparkesykkel',
+        'You need to add a Visa, Mastercard or American express to use an electric scooter',
+        'Du må legge til et Visa, Mastercard eller American express for å bruke ein el-sparkesykkel',
```
marius-at-atb, line 62: American Express should be written with a capital `E`.

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -24,26 +24,54 @@ export const MobilityTexts = {
...
     ),
+    shmoBlockers: _('Fullfør steg', 'Complete steps', 'Fullfør steg'),
+    shmoBlockersInfoMessage: (numberOfBlockers: number) =>
+      _(
+        `Du må igjennom ${numberOfBlockers} steg før du kan starte turen`,
+        `You need to go through ${numberOfBlockers} steps before you can start the trip`,
+        `Du må gå igjennom ${numberOfBlockers} steg før du kan starte turen`,
+      ),
+    location: {
+      locationTitle: _(
```
marius-at-atb, line 41: My preference is that we don't repeat the parent object name every time.
I.e. for this case it can be just `location.title` instead of `location.locationTitle`

src/mobility/components/onboarding/LocationBlocker.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {useTranslation} from '@atb/translations';
+
+export type LocationBlockerProps = {};
+
+export const LocationBlocker = ({}: LocationBlockerProps) => {
+  const {t} = useTranslation();
+
+  const {requestLocationPermission} = useGeolocationContext();
+
+  const setCurrentLocationOrRequest = useCallback(async () => {
```
marius-at-atb, line 15: `requestLocationPermission` is already wrapped in a `useCallback` in `GeolocationContextProvider`, so this wrapper is unnecessary. Can just use `requestLocationPermission` directly.

marius-at-atb (commented):
Very nice! Just some small fixes in the comments

src/translations/screens/subscreens/MobilityTexts.ts
gorandalum: What about just "betalingskort"? What payment methods that currently are enabled is configurable [through Firestore](https://github.com/AtB-AS/firestore-configuration/blob/main/documents/atb-mobility-platform/paymentTypes.yaml).

src/mobility/components/onboarding/LocationBlocker.tsx
gorandalum, line 1: I think these files should be named `XXXScreenComponent` as well, since when they are just a small wrapping layer for a screen component they are in practice also screen components.

gorandalum (approved):

src/mobility/components/onboarding/LocationBlocker.tsx
Dahly96: yes, ill change it

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -14,7 +14,7 @@ export const MobilityTexts = {
   formFactor: (formFactor: FormFactor) => {
     switch (formFactor) {
       case FormFactor.Scooter:
-        return _('Sparkesykkel', 'Electric scooter', 'Elektrisk sparkesykkel');
+        return _('El-sparkesykkel', 'El-scooter', 'El-sparkesykkel');
```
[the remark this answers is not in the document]
Dahly96: yes, ill change

src/translations/screens/subscreens/MobilityTexts.ts
Dahly96: went with betalingskort after a talk with Øyvind

src/mobility/components/onboarding/LocationBlocker.tsx
Dahly96: ill change it

marius-at-atb (approved):
Nice

PR #5047, opened 2025-02-27
feat: use PayWithBonusPointsCheckbox for Trondheim bysykkel

jorunnl:
Adds use of PayWithBonusPointsCheckbox in BikeStation Bottom sheet.
Will display checkbox if an active BonusProduct for the given operator and form factor exists AND isBonusProgramEnabled is true

### DIFF
--- a/src/mobility/components/BikeStationBottomSheet.tsx
+++ b/src/mobility/components/BikeStationBottomSheet.tsx
@@ -146,7 +146,9 @@ export const BikeStationBottomSheet = ({
                 <PayWithBonusPointsCheckbox
                   bonusProduct={bonusProduct}
                   isChecked={useBonusPoints}
-                  onPress={() => setUseBonusPoints(!useBonusPoints)}
+                  onPress={() =>
+                    setUseBonusPoints((useBonusPoints) => !useBonusPoints)
+                  }
                   style={styles.useBonusPointsSection}
                 />
               )}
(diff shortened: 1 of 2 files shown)


src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -130,6 +142,14 @@ export const BikeStationBottomSheet = ({
                   </View>
                 </GenericSectionItem>
               </Section>
+              {isBonusProgramEnabled && bonusProduct && (
+                <PayWithBonusPointsCheckbox
+                  bonusProduct={bonusProduct}
+                  isChecked={useBonusPoints}
+                  onPress={() => setUseBonusPoints(!useBonusPoints)}
```
marius-at-atb, line 149: Best practice when updating a value like this is to use `setUseBonusPoints(useBonusPoints => !useBonusPoints)` in order to be 100% sure it is the up-to-date value.

marius-at-atb (approved):
Nice!

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -55,7 +59,15 @@ export const BikeStationBottomSheet = ({
...
   const {operatorBenefit} = useOperatorBenefit(operatorId);
+  const {isBonusProgramEnabled} = useFeatureTogglesContext();
+  const {bonusProducts} = useFirestoreConfigurationContext();
+  const bonusProduct = findRelevantBonusProduct(
+    bonusProducts,
+    operatorId,
+    FormFactor.Bicycle,
+  );
 
+  const [useBonusPoints, setUseBonusPoints] = useState(false);
```
gorandalum, line 70: Note for later: I would avoid using `use` as prefix on any variables other than hooks, as it is such an establish connotation in React projects.

PR #5049, opened 2025-02-28
feat: add debug menu to specify endpoint for mobile token server

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20084

Pretty much self-explanatory, adds debug menu for specifying the mobile token server endpoint.
If there's a better alternative than using the `storage` function I would love to know more about it 😄 

[mobile token debug menu](https://github.com/user-attachments/assets/7388199e-0781-4877-9f66-8d5e9a62ff08)

### DIFF
--- a/src/mobile-token/DebugTokenServerAddress.tsx
+++ b/src/mobile-token/DebugTokenServerAddress.tsx
@@ -10,8 +10,7 @@ export const DebugTokenServerAddress = () => {
 
   const saveDebugIpAddress = async (ipAddress: string) => {
     if (ipAddress.length > 0) {
-      const fullAddress = `http://${ipAddress}:8080`;
-      await storage.set('@ATB_debug_token_server_ip_address', fullAddress);
+      await storage.set('@ATB_debug_token_server_ip_address', ipAddress);
     } else {
       await storage.remove('@ATB_debug_token_server_ip_address');
     }
@@ -34,8 +33,8 @@ export const DebugTokenServerAddress = () => {
       <TextInput
         onChangeText={(text) => setIpAddress(text)}
         value={ipAddress}
-        inputMode="numeric"
-        placeholder="e.g. 10.100.1.89 (Leave blank to use default)"
+        inputMode="text"
+        placeholder="e.g. http://10.100.1.89:8080 (Leave blank to use default)"
         clearButtonMode="always"
       />
       <Button
--- a/src/mobile-token/DebugTokenServerAddress.tsx
+++ b/src/mobile-token/DebugTokenServerAddress.tsx
@@ -1,0 +1,58 @@
+import {Button} from '@atb/components/button';
+import {ThemeText} from '@atb/components/text';
+import {storage} from '@atb/storage';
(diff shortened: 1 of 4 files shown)


src/mobile-token/DebugTokenServerAddress.tsx
```diff
@@ -0,0 +1,58 @@
...
+import {StyleSheet} from '@atb/theme';
+import {useEffect, useState} from 'react';
+import {TextInput, View} from 'react-native';
+
+export const DebugTokenServerAddress = () => {
+  const [ipAddress, setIpAddress] = useState<string>('');
+
+  const saveDebugIpAddress = async (ipAddress: string) => {
+    if (ipAddress.length > 0) {
+      const fullAddress = `http://${ipAddress}:8080`;
```
rosvik, line 13: Not quite sure what's useful to test in practice, but should we have the input be the whole address instead of just the IP? If it's useful to e.g. for point a prod device to staging servers, where we don't have access to a server running on a local IP.

jorelosorio: We usually save that into the .env file and read from it

reidzeibel: We can change it to a whole IP (probably with a `http://` prefix added) 👍🏼 

Also, pointing a prod device to staging server won't work, because the tokens on Prod are different from tokens from Staging.

This is just to make it easier for testing on staging, either for reproducing a bug, or testing/developing new functionality. For example: we want to replicate a bug with creating tokens, we can just point the token server to local, instead of changing some parameters from the code to allow debug. These changes might be minor, but the added overhead of having to modify the server is something that we could have avoided with this menu.

src/mobile-token/DebugTokenServerAddress.tsx
```diff
@@ -0,0 +1,58 @@
+import {Button} from '@atb/components/button';
+import {ThemeText} from '@atb/components/text';
+import {storage} from '@atb/storage';
```
marius-at-atb, line 3: Regarding storage, we recently created `usePersistedBoolState`.
We might want to either create one for string as well, or make the existing one general purpose. @gorandalum

gorandalum: Yeah, it is possible to refactor it into a more generic `usePersistedState`.

reidzeibel: I will make the refactor on another PR 👍🏼

src/mobile-token/DebugTokenServerAddress.tsx
reidzeibel: Refactored to allow full address

reidzeibel:
nudging @rosvik 😄

rosvik (approved):
Apologies for the slow response 🫣 Looks good 👍

reidzeibel:
@rosvik no worries! lots of stuff happened last week, thank you for the review!

PR #5050, opened 2025-02-28
refactor: Use state for purchase selection

gorandalum:
Use state for the purchase selection object instead of relying on `setParams`. This is to reduce PostHog screen events, which are created when using `setParams`.


marius-at-atb (approved):
Nice!

PR #5055, opened 2025-03-04
fix: accessibility pay with bonus points checkbox

jorunnl:
Fixes accessibility role + state + label on PayWithBonusPointsCheckbox

Also fixes the name of a useState variable to not start with 'use'.

### DIFF
--- a/src/mobility/components/BikeStationBottomSheet.tsx
+++ b/src/mobility/components/BikeStationBottomSheet.tsx
@@ -67,7 +67,7 @@ export const BikeStationBottomSheet = ({
     FormFactor.Bicycle,
   );
 
-  const [useBonusPoints, setUseBonusPoints] = useState(false);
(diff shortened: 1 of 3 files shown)


src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -145,9 +145,9 @@ export const BikeStationBottomSheet = ({
               {isBonusProgramEnabled && bonusProduct && (
                 <PayWithBonusPointsCheckbox
                   bonusProduct={bonusProduct}
-                  isChecked={useBonusPoints}
+                  isChecked={payWithBonusPoints}
                   onPress={() =>
-                    setUseBonusPoints((useBonusPoints) => !useBonusPoints)
+                    setPayWithBonusPoints((useBonusPoints) => !useBonusPoints)
```
marius-at-atb, line 150: useBonusPoints -> payWithBonusPoints here as well

PR #5057, opened 2025-03-04
fix: Ensure correct button notification color

gorandalum:
Instead of letting the parent of the button specify the background
color for the notification, it is instead injected from the button,
making it always correct.

Before:
Screenshot 2025-03-04 at 13 15 33
Screenshot 2025-03-04 at 13 15 49

After:
Screenshot 2025-03-04 at 13 11 47
Screenshot 2025-03-04 at 13 12 35


rosvik (approved):

rosvik (approved):
Even better!

PR #5060, opened 2025-03-04
fix: disabled accessibility pay with bonus check box

jorunnl:
fix: disabled accessibility pay with bonus check box


src/mobility/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -51,7 +51,7 @@ export const PayWithBonusPointsCheckbox = ({
       <GenericClickableSectionItem
         active={isChecked}
         onPress={onPress}
-        onAccessibilityTap={onPress}
+        onAccessibilityTap={() => !disabled && onPress()}
```
gorandalum, line 54: I don't think we need `onAccessibilityTap` on something that already is pressable. We haven't used it anywhere else in the app until now.

gorandalum (approved):

PR #5064, opened 2025-03-05
feat: new style for payment option in app/webshop

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/18186

### If user is **logged in**:

Profile saved cards           |  Preselected last saved payment option
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 08 15|  Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 08 55

User clicks on `Edit`         |  If selects `New Card`
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 09 05| Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 09 21

### If user **is anonymous**:

Two options to select    |  If selects `New Card`
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 10 14 |  Simulator Screenshot - iPhone 16 Pro Max - 2025-03-05 at 14 10 19

I have tried to create a `PaymentSelectionCard` component for both **delete** and **edit**  card functionalities, but I have encountered some issues with types. For **delete**, we should use `RecurringPayment`, but for **edit**, we should use `PaymentMethod`.

The main goal was to split one payment selection component into three: `SinglePaymentMethod` for 'Vipps', `MultiplePaymentMethodsRadioSection` for the group of payment methods, and `PaymentSelectionCard` for actions such as delete and edit.

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
@@ -10,7 +10,6 @@ import {BottomSheetContainer} from '@atb/components/bottom-sheet';
 import {FullScreenFooter} from '@atb/components/screen-footer';
 import {PaymentBrand} from './PaymentBrand';
 import {useFirestoreConfigurationContext} from '@atb/configuration/FirestoreConfigurationContext';
-import {getExpireDate} from '../../utils';
 import {Checkbox} from '@atb/components/checkbox';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
 import {
@@ -19,7 +18,9 @@ import {
 } from '@atb/stacks-hierarchy/types';
 import {useAuthContext} from '@atb/auth';
 import {PaymentType, humanizePaymentType} from '@atb/ticketing';
-import {getRadioA11y} from '@atb/components/radio';
+import {RadioIcon, getRadioA11y} from '@atb/components/radio';
+import {MessageInfoText} from '@atb/components/message-info-text';
+import AnonymousPurchases from '@atb/translations/screens/subscreens/AnonymousPurchases';
 
 type Props = {
   onSelect: (
@@ -44,6 +45,7 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
   const [shouldSave, setShouldSave] = useState(
     currentOptions?.shouldSavePaymentMethod ?? false,
   );
+  const {authenticationType} = useAuthContext();
 
   const {paymentTypes} = useFirestoreConfigurationContext();
   const defaultPaymentMethods: PaymentMethod[] = paymentTypes.map(
@@ -51,6 +53,16 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
       paymentType,
       savedType: SavedPaymentMethodType.Normal,
     }),
+  );
+  const singlePaymentMethod = defaultPaymentMethods.filter(
+    (method) =>
+      method.paymentType !== PaymentType.Amex &&
+      method.paymentType !== PaymentType.Mastercard &&
+      method.paymentType !== PaymentType.Visa,
+  );
+
+  const multiplePaymentMethod = defaultPaymentMethods.filter(
+    (method) => method.paymentType !== PaymentType.Vipps,
   );
   const [selectedMethod, setSelectedMethod] = useState(
     currentOptions?.paymentMethod,
@@ -60,13 +72,18 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
     <BottomSheetContainer
       title={t(SelectPaymentMethodTexts.header.text)}
       fullHeight
+      maxHeightValue={
+        authenticationType !== 'phone'
+          ? 0.33
+          : 0.33 + (recurringPaymentMethods?.length ?? 0) * 0.13
+      }
     >
       <View style={{flex: 1}}>
         <View style={{flexShrink: 1, flexGrow: 1}}>
           <ScrollView style={styles.paymentMethods}>
-            {defaultPaymentMethods.map((method, index) => {
+            {singlePaymentMethod.map((method, index) => {
               return (
-                <PaymentMethodView
+                <SinglePaymentMethod
                   key={method.paymentType}
                   paymentMethod={method}
                   shouldSave={shouldSave}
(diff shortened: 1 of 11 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -128,6 +173,107 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
 };
 
+type MultiplePaymentMethodsProps = {
+  paymentMethod?: PaymentMethod;
+  selected: boolean;
+  onSelect: (value: PaymentMethod[]) => void;
+  shouldSave: boolean;
+  onSetShouldSave: (val: boolean) => void;
+  paymentGroup: PaymentMethod[];
+  index: number;
```
rosvik, line 183: Could we replace the `index` param with `testID`?

rosvik (commented):
Nice! Noticed a few things reading through this, but it looks and works great when testing locally.

Also, there's a bit of commented out code here. Better to remove that, the we can re-implement it later if needed.

src/components/bottom-sheet/BottomSheetContainer.tsx
```diff
@@ -31,7 +32,7 @@ export function BottomSheetContainer({
         title={title}
         focusTitleOnLoad={focusTitleOnLoad}
       />
-      {children}
+      <ScrollView>{children}</ScrollView>
```
jorunnl, line 35: I think this should not be done for the entire BottomSheetContainer component. See if it can only be done where you need it.

jorunnl (commented):
Looks and works very well 👍 Some feedback in comments. Let me know if we should look at them together 🤗

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -128,6 +165,107 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
+  paymentMethod?: PaymentMethod;
+  selected: boolean;
+  onSelect: (value: PaymentMethod[]) => void;
+  shouldSave: boolean;
+  onSetShouldSave: (val: boolean) => void;
+  paymentGroup: PaymentMethod[];
+  testID?: string;
+};
+
+const MultiplePaymentMethodsRadioSection: React.FC<
```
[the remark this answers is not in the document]
rosvik: We had a talk about this a while back, and decided to try splitting them up, to avoid having too much complexity in one component. There's some differences between them in functionality, even though they look similar.

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -52,59 +56,92 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
+            shouldSave={shouldSave}
+            onSetShouldSave={setShouldSave}
+            selected={
+              !selectedMethod?.recurringCard &&
+              selectedMethod?.paymentType ===
+                multiplePaymentMethod[0]?.paymentType
+            }
+            onSelect={() => {
+              setShouldSave(true);
+              setSelectedMethod(multiplePaymentMethod[0]);
```
rosvik, line 111: I'm a bit unsure if saving the first payment method here is the best way to do it. Could we use a different state to more clearly communicate what is selected? Something like this maybe:

```ts
type PaymentSelection = {
	kind: 'new_recurring' | 'saved_recurring' | 'non_recurring',
	paymentMethod: PaymentMethod | undefined,
}
```

I haven't looked into the details here, so you might have to change it a bit, but do you think something like it could work? If it instead makes it harder to work with, feel free to ignore 😄

Oleninkaa:
I forgot to mention that there are also some changes on this screen. @rosvik @jorunnl 

simulator_screenshot_3E8CE45F-0D36-4126-B0A3-C69E0BC1A826

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -418,6 +403,30 @@ const SinglePaymentMethod: React.FC<PaymentMethodProps> = ({
   );
 };
 
+function defineSelection(
```
reidzeibel, line 406: `defineSelection` seems to be too generic, I would pick something more specific, for example `getPaymentSelection`

Oleninkaa:
When a user pays with a new card and selects to save the card, the next time they visit the `Root_PurchaseConfirmationScreen`, they will see the previously saved payment method. @rosvik Johannes suggested this simple solution. We just added a **delay** to make it work for a newly added card.

I also removed unnecessary elements for the single payment method that I hadn't noticed before. The `PaymentType.PaymentCard` is used when a user selects `MultiplePaymentMethods` and adds a new card in NETS. After the payment flow, we determine the actual type of the card. This is because NETS doesn't really care what card is typed in.

src/stacks-hierarchy/saved-payment-utils.ts
```diff
@@ -131,6 +131,7 @@ export const saveLastUsedRecurringPaymentOrType = async (
     });
   } else {
     try {
+      await new Promise((resolve) => setTimeout(resolve, 3000));
```
reidzeibel, line 134: @Oleninkaa why do we need this timeout? 🤔

reidzeibel (approved):
Looks good, but need to wait for code owners for merging!

src/stacks-hierarchy/utils.ts
```diff
@@ -15,7 +15,7 @@ export function getExpireDate(iso: string): string {
   // and the date that shows on the card is the month before the card expires
   // Example: The card expires the moment the date is 02.2021, but the date on the card is 01.2021
   date.setDate(date.getDate() - 1);
-  return format(date, 'MM/yy');
+  return format(date, 'dd.MM.yy');
```
marius-at-atb, line 18: I can't find where `getExpireDate` is used. Or is it just me?
We can delete it if unused. Otherwise we should probably add some tests for it, to confirm that it works as intended.

marius-at-atb (approved):
Great work!
All the comments are syntactic sugar/best practice suggestions/clean-ups, not blocking :)

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -92,62 +127,170 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
...
+const MultiplePaymentMethodsRadioSection: React.FC<
+  MultiplePaymentMethodsProps
+> = ({
   paymentMethod,
   selected,
   onSelect,
   shouldSave,
   onSetShouldSave,
-  index,
+  paymentGroup,
```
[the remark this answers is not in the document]
Oleninkaa: I don't think it's possible, since we need this property later in the function `getPaymentSelection(paymentMethod, shouldSave)` 😮

src/stacks-hierarchy/utils.ts
Oleninkaa: We will use it for the saved card/NETS expiry messages when we implement this feature

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
```diff
@@ -234,8 +229,11 @@ const MultiplePaymentMethodsRadioSection: React.FC<
                 </ThemeText>
               </View>
               <View style={styles.paymentIconsGroup}>
-                {paymentGroup.map((payment) => (
-                  <PaymentBrand paymentType={payment.paymentType} />
+                {paymentMethodsInGroup.map((payment) => (
```
marius-at-atb, line 232: Typically the variable names in map functions should match the array items.
In this case, `paymentMethodInGroup` or just `paymentMethod` would be more precise than just `payment`.

marius-at-atb (approved):
Very nice!

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/components/SelectPaymentMethodSheet.tsx
reidzeibel: @Oleninkaa If you fix this, since it's very simple, no need to request for another re-review, you can safely merge :)

PR #5069, opened 2025-03-07
refactor: Update certs

jorelosorio:
This PR refactors the action that nukes the certs and forces updating the provisioning profiles.

image

This PR also includes updates to fastlane.


Gemfile.lock
reidzeibel, line 1: @jorelosorio do you need to update this file or is it a stray file added to the commit?

jorelosorio: It needs to be modified since the fastlane was updated as well.

reidzeibel (approved):

tormoseng:
✅  Tested OK + changed the Slack alerts to 2026-02-10 (1 month before the next expiry date)

PR #5075, opened 2025-03-14
feat: intercept NFC reading when app is in foreground

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20247

This PR will make our Android App to intercept all NFC reading when it is in foreground, no new libraries included, only added some native code.


marius-at-atb (approved):
Lgtm!

PR #5077, opened 2025-03-14
DRAFT: Map v2 for remaining maps

marius-at-atb:
This uses the new map for all remaining maps in the app, supporting everything as before + dark mode as well.

| Before | After |
|--------|-------|
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

Resolves https://github.com/AtB-AS/kundevendt/issues/20249

PR #5078, opened 2025-03-17
Separate explore location map

marius-at-atb:
This cleans up the code by separating the explore location mode in the map to its own component, and fixes the bug where an item on the map was selected when clicked.

| Before (icons minimized when they shouldn't be) | After |
|--------|-------|
|  |  |

Part of https://github.com/AtB-AS/kundevendt/issues/19216

### DIFF
--- a/src/components/map/ExploreLocationMap.tsx
+++ b/src/components/map/ExploreLocationMap.tsx
@@ -1,0 +1,133 @@
+import {
+  getCurrentCoordinatesGlobal,
+  useGeolocationContext,
+} from '@atb/GeolocationContext';
+import {FOCUS_ORIGIN} from '@atb/api/geocoder';
+import {StyleSheet} from '@atb/theme';
+import MapboxGL, {LocationPuck} from '@rnmapbox/maps';
+import {Feature} from 'geojson';
+import React, {useCallback, useMemo, useRef, useState} from 'react';
+import {View} from 'react-native';
+import {MapCameraConfig, MapViewConfig} from './MapConfig';
+import {SelectionPin} from './components/SelectionPin';
+import {LocationBar} from './components/LocationBar';
+import {PositionArrow} from './components/PositionArrow';
+import {useControlPositionsStyle} from './hooks/use-control-styles';
+import {SelectionLocationCallback} from './types';
+
+import {isFeaturePoint} from './utils';
+import {Location} from '@atb/favorites';
+import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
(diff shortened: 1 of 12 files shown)


src/stacks-hierarchy/Root_LocationSearchByMapScreen.tsx
```diff
@@ -36,19 +33,10 @@ export const Root_LocationSearchByMapScreen: React.FC<Props> = ({
...
-          selectionMode="ExploreLocation"
-        />
-      ) : (
-        <Map
-          onLocationSelect={onLocationSelect}
-          initialLocation={initialLocation}
-          selectionMode="ExploreLocation"
-        />
-      )}
+      <ExploreLocationMap
```
rosvik, line 36: Are we not adding the feature toggle to this map? I guess it's less risky than the main map.

src/components/map/ExploreLocationMap.tsx
```diff
@@ -0,0 +1,133 @@
...
+            flex: 1,
+          }}
+          pitchEnabled={false}
+          onPress={onFeatureClick}
+          testID="exploreLocationMapView"
+          {...{
+            ...MapViewConfig,
+            // only updating Map.tsx for now.
+            ...customMapViewConfigProps,
+          }}
```
rosvik, line 94: I found this a bit confusing 🤔 Can we write it like so? (Also not sure if the comment still applies)

```suggestion
          {...MapViewConfig}
          {...customMapViewConfigProps}
```

rosvik (approved):
Nice! Looks good and works well locally.

src/stacks-hierarchy/Root_LocationSearchByMapScreen.tsx
marius-at-atb: Its coming in the next PR: https://github.com/AtB-AS/mittatb-app/pull/5077/files/e1503d5bfeec43704e5fd8df115ac96b05fdc75d..6763b12e3e565033b9c7d0c61c68145294abfbe0#diff-5a86c4dada4a01870eb25bab183849195acfbc143f858c731ceef81d9d102bbd

src/components/map/ExploreLocationMap.tsx
marius-at-atb: This one is also updated in the [next PR](https://github.com/AtB-AS/mittatb-app/pull/5077/files/e1503d5bfeec43704e5fd8df115ac96b05fdc75d..6763b12e3e565033b9c7d0c61c68145294abfbe0#diff-5a86c4dada4a01870eb25bab183849195acfbc143f858c731ceef81d9d102bbd), so merging this as is first.

PR #5079, opened 2025-03-17
feat: add signed token as header to list token

reidzeibel:
part of https://github.com/AtB-AS/kundevendt/issues/20216

some reference that may be useful: https://mittatb.slack.com/archives/C079LPXV1NF/p1742217909662349

Related to https://github.com/AtB-AS/abt-token-server/pull/95, which will utilize the stuff that we added in this PR.

The PR will add `signedToken` as header to the `/list` endpoint request, which in turn will allow the `list` function to `add inspection` rights if there are no tokens with that right on the list. 

Also now the list token will only fetched by the app when the native token (on-device token) is successfully fetched/generated, instead of being called regardless of the native token status. This is to accommodate the change to the `/list` function above, since the token inspection right is added when we call this endpoint.

### Acceptance criteria

- [ ] `Atb-Signed-Token` header is sent when app calls `/list` endpoint
- [ ] Try to prevent the token from being created (intercept `/init` or `/activate`), then the `/list` endpoint should not be called without any token. (can also use token Sabotage I think, to prevent token creation and trigger fallback)

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -135,7 +135,10 @@ export const MobileTokenContextProvider = ({children}: Props) => {
     data: remoteTokens,
     status: remoteTokensStatus,
     error: remoteTokenError,
-  } = useListRemoteTokensQuery(enabled, nativeToken);
+  } = useListRemoteTokensQuery(
+    enabled && nativeToken !== undefined,
+    nativeToken,
+  );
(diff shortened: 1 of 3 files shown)


rosvik (approved):
LGTM 👍

strandlie (approved):

tormoseng:
✅

tormoseng:
✅

PR #4896, opened 2024-12-19
feat: Improve FareContract ticket inspection

strandlie:
Closes AtB-AS/kundevendt#4220

I have tried to follow conventions in the code base as I understand them, but please feel free to comment and make suggestions :) 

[Figma](https://www.figma.com/design/2QTjAdekdIPuLFovQhVrY3/Components?node-id=9314-86777&node-type=canvas&t=7drnAkdCRM29OiBg-0)

I would add some screenshots, but there are a lot of cases so I would encourage checking out the branch and looking at it on your machine.

PR #5085, opened 2025-03-19
fix: make spacing even on spacious sections

rosvik:
Reverts a change made in https://github.com/AtB-AS/mittatb-app/pull/4896. Seems the padding change to spacious sections is no longer needed for fare contracts. This makes padding even for sections 

- in the following bottom sheets:
	- External realtime map sheet
	- Activate now
	- Consume carnet
	- Situations
- the mobile token toggle radio buttons (which is due to change anyway)
- the "multiple traveller selection" component in the purchase flow was spacious, but it was overridden by internal styles. I cleaned this up a bit.

<div>

</div>

### Acceptance criteria

- [ ] The sections mentioned above look good, and have even spacing above and below the content


hanne-at-atb:
Tested ok ✅

PR #5087, opened 2025-03-19
fix: Remove Stations From TravelDetailsMapScreenComponent

marius-at-atb:
Stations were included in this map in https://github.com/AtB-AS/mittatb-app/pull/5084, but wasn't there from before, so shouldn't be there now at least.

### DIFF
--- a/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
+++ b/src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
@@ -74,7 +74,7 @@ export const TravelDetailsMapScreenComponent = ({
   const isFocusedAndActive = useIsFocusedAndActive();
 
   const {isMapV2Enabled} = useFeatureTogglesContext();
-  const mapViewConfig = useMapViewConfig(true);
+  const mapViewConfig = useMapViewConfig(false);
 
   const features = useMemo(() => createMapLines(legs), [legs]);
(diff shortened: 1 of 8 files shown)


src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
```diff
@@ -75,7 +74,7 @@ export const TravelDetailsMapScreenComponent = ({
   const isFocusedAndActive = useIsFocusedAndActive();
 
   const {isMapV2Enabled} = useFeatureTogglesContext();
-  const mapViewConfig = useMapViewConfig(true);
+  const mapViewConfig = useMapViewConfig(false);
```
rosvik, line 77: Just a note: I think it's nicer if the options to a function like this is an object, in this case `{useMapViewConfig: boolean, useDarkModeForV1?: boolean}`. Makes it easier to understand what the parameters actually do.

rosvik (approved):

rosvik (approved):
💯

src/travel-details-map-screen/TravelDetailsMapScreenComponent.tsx
marius-at-atb: Agreed, https://github.com/AtB-AS/mittatb-app/pull/5087/commits/3b10f494e556e3b05c73e7a65848f639cb766f03 :+1:

PR #5088, opened 2025-03-19
fix: use correct text color for button loading indicator

rosvik:
There was an issue in the app where the loading indicator on buttons don't get the correct color, and would sometimes be almost invisible:

Screenshot 2025-03-19 at 16 10 11

This fixes this by setting it to the same color as the text color:

<div>

</div>

### Acceptance criteria

- [ ] Loading indicators on buttons are always visible, and have the same color as the text
	- You can use storybook to verify this :)


marius-at-atb (approved):
Lgtm!

hanne-at-atb:
Tested ok ✅

PR #5089, opened 2025-03-20
feat: first iteration of start/stop scooter in the app

Dahly96:

Created the first iteration of start/stop scooter. There is some things that is commented in the code that will be changed but it will be done in its own PR.

the step of taking image it not implemented yet. So the final finishing call is now made on the "close/lukk" button on the last summary page. If you navigate back and forth from the map the bottomsheet will autoopen becuase it still has a active trip until finished event is called from now being the close/lukk button.

there is also no change paymentmethod component in the version. That will come later


src/mobility/components/ShmoActionButton.tsx
```diff
@@ -1,22 +1,71 @@
...
+      const lastRecurringPaymentMethod =
+        recurringPaymentMethods[recurringPaymentMethods.length - 1];
+
+      const recurringPaymentId =
+        lastRecurringPaymentMethod.savedType === 'recurring'
+          ? lastRecurringPaymentMethod?.recurringCard?.id
+          : 0;
+      const initReqBody: InitShmoOneStopBookingRequestBody = {
+        recurringPaymentId: recurringPaymentId ?? 0,
+        coordinates: {latitude: 0, longitude: 0},
```
marius-at-atb, line 62: coordinates should be the user location.

src/mobility/utils.ts
```diff
@@ -333,6 +334,16 @@ export const getNewFilterState = (
...
 
+export const formatErrorMessage = (
+  errorResponse: any,
+  t: TFunc<typeof Language>,
+) => {
+  return (errorResponse && errorResponse.response.status === 500) ||
+    errorResponse.response.status === 404
+    ? t(dictionary.genericErrorMsg)
+    : errorResponse.response.data;
+};
```
marius-at-atb, line 345: Do we have an issue to come back and update this? If not let's create it.

src/mobility/components/sheets/ScooterSheet.tsx
```diff
@@ -70,16 +73,16 @@ export const ScooterSheet = ({
   return (
     <BottomSheetContainer
       title={t(MobilityTexts.formFactor(FormFactor.Scooter))}
-      maxHeightValue={0.5}
+      maxHeightValue={0.6}
```
marius-at-atb, line 76: We may want to increase `SLIGHTLY_RAISED_MAP_PADDING` if increasing `maxHeightValue` here, so that there is space between the top of the bottom sheet and where the map item is centered.

src/components/map/hooks/use-active-shmo-booking.tsx
```diff
@@ -0,0 +1,119 @@
...
+  // NOTE: This ref is not used for anything since the map doesn't support
+  // screen readers, but a ref is required when opening bottom sheets.
+  const onCloseFocusRef = useRef<RefObject<any>>(null);
+
+  const closeBottomSheet = useCallback(() => {
+    close();
+    setBottomSheetCurrentlyAutoSelected(undefined);
+  }, [close, setBottomSheetCurrentlyAutoSelected]);
+
+  const flyToMapItemLocation = useCallback(async () => {
```
marius-at-atb, line 41: In this case it is actually `flyToUserLocation`.
Also it can be good to experiment with whether it is best to wait for onVehicleReceived (current solution), or call it immediately.

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+type Props = {
+  onClose: () => void;
+  onVehicleReceived?: () => void;
+  navigateSupportCallback: () => void;
+};
+
+export const ActiveScooterSheet = ({
+  onClose,
+  onVehicleReceived,
```
marius-at-atb, line 35: In this case it should be called `onActiveBookingReceived`. (Unless it can just be removed, depending on the [conclusion here](https://github.com/AtB-AS/mittatb-app/pull/5089/files#))

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -0,0 +1,170 @@
...
+        bookingId: activeBooking.bookingId,
+        shmoBookingEvent: startFinishingEvent,
+      });
+    }
+  }, [activeBooking, sendShmoBookingEvent]);
+
+  return (
+    <BottomSheetContainer
+      title={t(MobilityTexts.formFactor(FormFactor.Scooter))}
+      maxHeightValue={0.7}
```
marius-at-atb, line 76: Here as well, we may want to consider this in relation to `SLIGHTLY_RAISED_MAP_PADDING`

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const {t} = useTranslation();
+  const {theme} = useThemeContext();
+  const styles = useStyles();
+
+  const pricePlan: PricingPlanFragment = {
+    price: activeBooking?.pricingPlan.price ?? 0,
+    perMinPricing: activeBooking?.pricingPlan
+      .perMinPricing as unknown as Array<PricingSegmentFragment>,
+    perKmPricing: [],
+  };
```
marius-at-atb, line 48: Hmm, perhaps `VehicleCard` should support an inferred type from `ShmoPricingPlanSchema` instead, so that `activeBooking?.pricingPlan` can be used directly 🤔

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -0,0 +1,170 @@
...
+          )}
+          {!isLoading && !isError && activeBooking && (
+            <>
+              <ScrollView style={styles.container}>
+                <ShmoTripCard bookingId={activeBooking.bookingId} />
+                <VehicleCard
+                  pricingPlan={pricePlan}
+                  currentFuelPercent={activeBooking.asset.stateOfCharge ?? 0}
+                  currentRangeMeters={
+                    (activeBooking.asset.currentRangeKm as number) * 1000 // Convert km to meters
```
marius-at-atb, line 94: We usually try to avoid the `as` keyword, and in this case it covers up that currentRangeKm can be undefined? I think `(activeBooking.asset?.currentRangeKm || 0)*1000` will work?

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -0,0 +1,170 @@
...
+                  }
+                  operatorName={activeBooking.operator.name}
+                  brandLogoUrl=""
```
marius-at-atb, line 97: `brandLogoUrl` should either be optional as prop to `VehicleCard`, or added here. To include it, see the `findOperatorBrandImageUrl` function.

marius-at-atb (changes requested):
Nice work!
Some comments that should be looked at.

src/mobility/components/ShmoActionButton.tsx
```diff
@@ -1,22 +1,71 @@
...
+  } = useInitShmoOneStopBookingMutation();
+
+  const {recurringPaymentMethods} = usePreviousPaymentMethods();
+
+  const initShmoBooking = useCallback(() => {
+    if (
+      vehicleId &&
+      recurringPaymentMethods &&
+      recurringPaymentMethods.length > 0
+    ) {
```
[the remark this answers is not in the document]
Dahly96: i removed the if, becuase there is a message for both bad paymentid and assetid, so ille let the errorresponse handle the message

src/mobility/components/ShmoActionButton.tsx
Dahly96: yes adding

src/mobility/utils.ts
Dahly96: created now: https://github.com/AtB-AS/kundevendt/issues/20287

src/mobility/components/sheets/ScooterSheet.tsx
Dahly96: this is to make space for when if a errormessage is shown. So i don think this will be an issue, since it wont be that high most of the time, only when displaying an errormessage

Dahly96: i might look into it when adding the paymentcard component, then it will be higher

src/components/map/hooks/use-active-shmo-booking.tsx
Dahly96: i do think that waiting like it is now might be the best. But we can always change it if we feel its not the best later aswell

Dahly96: i changed the name

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: yes updated name

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: changed to checking currentRangeKm first

`currentRangeMeters={
                    activeBooking.asset?.currentRangeKm
                      ? activeBooking.asset.currentRangeKm * 1000
                      : 0
                  }`

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: made it optional

src/mobility/components/sheets/ScooterSheet.tsx
marius-at-atb: :+1:

src/mobility/components/sheets/ActiveScooterSheet.tsx
marius-at-atb: That works. However, from [the drawings](https://github.com/AtB-AS/kundevendt/issues/20230) it looks like the logo should be there?

src/mobility/components/sheets/ScooterSheet.tsx
```diff
@@ -169,6 +175,7 @@ const useStyles = StyleSheet.createThemeHook((theme) => {
     footer: {
       marginBottom: Math.max(bottom, theme.spacing.medium),
       marginHorizontal: theme.spacing.medium,
+      gap: theme.spacing.medium,
```
marius-at-atb, line 178: This line causes extra padding also in the App Switch case:
| Current | How it should be |
|--------|-------|
|  |  |

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: i believe we had a discussion where we created a draft without logo and just the operator name? Since we dont have the logo, we decided that it was okey in the first version? so i guess it can be optional until we have logos we can use

src/mobility/components/sheets/ScooterSheet.tsx
Dahly96: fixed

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: yes ill add this

PR #5095, opened 2025-03-21
fix: use bonus folder for bonus code

jorunnl:
Moves bonus-related code from the mobility folder to the bonus folder


rosvik (commented):
I think `src/modules/bonus/` would be a good fit for this. Could we move it there instead?

(And then also update `.github/CODEOWNERS`)

reidzeibel (approved):

rosvik (approved):

PR #5096, opened 2025-03-21
fix: edit storybook radio button

Oleninkaa:
### Closes https://github.com/AtB-AS/kundevendt/issues/20015
From implementation of https://github.com/AtB-AS/kundevendt/issues/18735.

I hope it added some visible functionality for `RadioGroupSection` and `RadioSectionItem` components.

| Radio Section | Listed Section Items |
|--------|--------|
| simulator_screenshot_D9A60063-8B29-4DB8-97EC-23A0D553B6A7 | simulator_screenshot_205D63DA-4359-45A3-BF4A-C4C44BF8393E |

### DIFF
--- a/src/storybook/stories/Section.stories.tsx
+++ b/src/storybook/stories/Section.stories.tsx
@@ -79,6 +79,23 @@ export const ListedSectionItems: Meta<SectionMetaProps> = {
                   onPress={() => {}}
                   selected={false}
                 />
+                <RadioSectionItem
+                  text="RadioSectionItem selected"
+                  onPress={() => {}}
+                  selected={true}
+                />
+                <RadioSectionItem
+                  text="RadioSectionItem transparent"
+                  onPress={() => {}}
+                  selected={false}
+                  transparent
+                />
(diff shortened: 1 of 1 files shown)


src/storybook/stories/Section.stories.tsx
```diff
@@ -191,12 +208,25 @@ export const RadioSection: Meta<SectionMetaProps> = {
...
+
+              <RadioGroupSection
+                items={[
+                  'Spacious radio group option 1',
+                  'Spacious radio group option 2',
+                ]}
+                keyExtractor={(s) => s}
+                itemToText={(t) => t}
+                selected="Spacious radio group option 1"
+                type="spacious"
```
rosvik, line 227: It's important to return `RadioGroupSection` as the root node here, not the `View`, since `args` are applied to it. `args` is where the user selection from storybook comes from, including the `type` param. 

That is, I think we can keep this like it was, and have `type` as an storybook option, not a hard coded variant.

reidzeibel (approved):

PR #5097, opened 2025-03-24
revert: remove world cup shuttle bus design

rosvik:
This reverts commit ea93fc90b26a706fdc52228461b9833e6dee416b and part of d37e2c7b3d90e4c4ab6ecdf7aceb443acfd37f83

part of https://github.com/AtB-AS/kundevendt/issues/20191


src/utils/use-transport-color.ts
```diff
@@ -27,7 +27,6 @@ const getTransportColorKey = (
     case 'coach':
       if (subMode === 'localBus') return 'city';
       if (subMode === 'airportLinkBus') return 'airportExpress';
```
marius-at-atb, line 29: Just something I noticed while looking at this - we should probably use the `TransportSubmode` enum in this file.

marius-at-atb (approved):
Lgtm

PR #5098, opened 2025-03-24
fix: prevent refund options requests when orderId is empty string

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/20058#issuecomment-2748168960


reidzeibel (approved):

src/modules/fare-contracts/components/RefundBottomSheet.tsx
```diff
@@ -25,9 +26,10 @@ import {useSafeAreaInsets} from 'react-native-safe-area-context';
 type Props = {
   orderId: string;
   fareProductType: string | undefined;
+  state: FareContractState;
```
marius-at-atb, line 29: Can we call it `fareContractState` instead?
Since any state variable in the whole app can be called state..

PR #5103, opened 2025-03-25
feat: add token renewal logic based on token and device details

reidzeibel:
fixes https://github.com/AtB-AS/kundevendt/issues/20216#issuecomment-2747889695

This PR will add logic for token renewal based on remote token details and device details. The remote token details came from this PR https://github.com/AtB-AS/abt-token-server/pull/96. 

Token renewal logic:

- Get local token
- Get remote tokens
- Check if local token id exists on remote tokens
- Get the remote token that has the same id as local
- Find the details of that remote token, and compare to the current device situation (app version, OS level, device ID, etc.).
- If any of the details are different between device situation and remote token details, renew the token.

I made a custom hook for this PR and moved all renewal/reset logic into the hook `use-validate-token`.

### Acceptance criteria
- [ ] Updating from older version of app should renew the token

### DIFF
--- a/src/mobile-token/MobileTokenContext.tsx
+++ b/src/mobile-token/MobileTokenContext.tsx
@@ -22,13 +22,7 @@ import {
   ActivatedToken,
   AttestationSabotage,
 } from '@entur-private/abt-mobile-client-sdk';
-import {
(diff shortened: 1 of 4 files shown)


rosvik (approved):
LGTM 👍

PR #5104, opened 2025-03-26
fix: add missing query dependency to token toggle mutation

reidzeibel:
addresses https://github.com/AtB-AS/kundevendt/issues/20216#issuecomment-2747955834

Only added `nativeToken` to the query dependency so it can update the data properly.

### DIFF
--- a/src/mobile-token/hooks/use-toggle-token-mutation.tsx
+++ b/src/mobile-token/hooks/use-toggle-token-mutation.tsx
@@ -25,7 +25,7 @@ export const useToggleTokenMutation = () => {
           LIST_REMOTE_TOKENS_QUERY_KEY,
           userId,
           nativeToken?.tokenId,
-          secureContainer
+          secureContainer,
         ],
         tokens,
       );
--- a/src/mobile-token/hooks/use-toggle-token-mutation.tsx
+++ b/src/mobile-token/hooks/use-toggle-token-mutation.tsx
@@ -14,7 +14,7 @@ type Args = {
 export const useToggleTokenMutation = () => {
   const queryClient = useQueryClient();
   const {userId} = useAuthContext();
-  const {nativeToken} = useMobileTokenContext();
+  const {nativeToken, secureContainer} = useMobileTokenContext();
   return useMutation({
     mutationFn: ({tokenId, bypassRestrictions}: Args) =>
       tokenService.toggle(tokenId, uuid(), bypassRestrictions),
@@ -24,7 +24,8 @@ export const useToggleTokenMutation = () => {
           MOBILE_TOKEN_QUERY_KEY,
           LIST_REMOTE_TOKENS_QUERY_KEY,
           userId,
-          nativeToken,
+          nativeToken?.tokenId,
+          secureContainer
         ],
         tokens,
       );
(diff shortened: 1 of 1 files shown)


src/mobile-token/hooks/use-toggle-token-mutation.tsx
```diff
@@ -13,12 +14,18 @@ type Args = {
...
     mutationFn: ({tokenId, bypassRestrictions}: Args) =>
       tokenService.toggle(tokenId, uuid(), bypassRestrictions),
     onSuccess: (tokens) => {
       queryClient.setQueryData(
-        [MOBILE_TOKEN_QUERY_KEY, LIST_REMOTE_TOKENS_QUERY_KEY, userId],
+        [
+          MOBILE_TOKEN_QUERY_KEY,
+          LIST_REMOTE_TOKENS_QUERY_KEY,
+          userId,
+          nativeToken,
```
rosvik, line 27: I believe query keys should be serializable. Could we use an ID instead here?

reidzeibel: so, use `nativeToken.id` instead of the whole `nativeToken`? Can do that :D

reidzeibel: Looks like `nativeToken` is serializable, this is a `JSON.stringify` version that I logged using `console.log` 
```
native token serializable {"tokenId":"7f8ed08b-6135-4b8c-954a-fc40e0c03c63","validityStart":1742984527645,"validityEnd":1745576527645,"attested":true,"attestRequired":false,"invalidated":false}
```

reidzeibel: Should only be when the token ID is changed, but if we change that, the List token query also needs to be updated, I can make it on another PR before everything is released, what do you think?

It's going to look like this, will only take in strings so it is always serializable: 

```
export const useListRemoteTokensQuery = (
  enabled: boolean,
  tokenId?: string,
  secureContainer?: string,
) => {
  const {userId} = useAuthContext();
  return useQuery({
    queryKey: [
      MOBILE_TOKEN_QUERY_KEY,
      LIST_REMOTE_TOKENS_QUERY_KEY,
      userId,
      secureContainer,
    ],
    queryFn: () => {
      logToBugsnag('Listing tokens for user ' + userId);
      return tokenService.listTokens(secureContainer, uuid());
    },
    enabled,
    select: (tokens) =>
      tokens.map(
        (t): Token => ({
          ...t,
          isThisDevice: t.id === tokenId,
          isInspectable: isInspectable(t),
          type: isTravelCardToken(t) ? 'travel-card' : 'mobile',
          travelCardId: getTravelCardId(t),
        }),
      ),
  });
};
```

rosvik: I think that would be nice 👍

reidzeibel: Refactor PR https://github.com/AtB-AS/mittatb-app/pull/5106

I'll merge that first and then push the change to this PR

tormoseng:
✅

PR #5114, opened 2025-03-28
feat: support for active and inactive trip

Dahly96:

You can now send support request to an operator before a trip is started and when in an active trip. Before a trip is started you can get info about an operator from the GBFS feed, but when the trip is active its removed from the feed, and we need to use the booking request instead to retrieve the same info about the operator and the scooter.


src/mobility/components/sheets/ScooterSheet.tsx
```diff
@@ -35,6 +35,7 @@ type Props = {
   navigateSupportCallback: () => void;
   loginCallback: () => void;
   startOnboardingCallback: () => void;
+  navigation: any;
```
jorunnl, line 38: Can you specify type here?
```suggestion
  navigation: RootNavigationProps;
```

jorunnl (commented):
Looks good, tiny feedback from me:)

src/mobility/components/sheets/ScooterSheet.tsx
Dahly96: yes! updated

src/stacks-hierarchy/navigation-types.ts
```diff
@@ -65,12 +65,15 @@ type Root_ParkingViolationsConfirmationParams = {
 };
 
 type Root_ScooterHelpScreenParams = {
-  vehicleId: string;
+  operatorName: string;
+  operatorId: string;
+  vehicleId: string | null;
```
marius-at-atb, line 70: We can then fetch `operatorName` from
```ts
  const operators = useOperators();
  const operatorName = operators.byId(operatorId)?.name;
```
inside `Root_ContactScooterOperatorScreen`, so we don't need it as param here.
Generally it is desirable to keep the required input params at a minimum.

src/components/map/hooks/use-active-shmo-booking.tsx
```diff
@@ -70,7 +70,9 @@ export const useShmoActiveBottomSheet = (
                   navigateSupportCallback={() => {
```
marius-at-atb, line 70: `navigateSupportCallback` can be moved to a shared function here.
Also, I just noticed - the `navigateSupportCallback` prop should be renamed since it is the whole handler, not a callback.

Dahly96: i think we wait with this until the other PR about camera. Since im splitting up active trip and finishing

src/stacks-hierarchy/navigation-types.ts
Dahly96: changed

marius-at-atb: Nice, and remember to also remove `operatorName` as from both `Root_ScooterHelpScreenParams` and `Root_ContactScooterOperatorScreenParams`.

src/stacks-hierarchy/Root_ScooterHelp/Root_ContactScooterOperatorScreen.tsx
```diff
@@ -31,22 +30,24 @@ import {useProfileQuery} from '@atb/queries';
...
 export type Root_ContactScooterOperatorScreenProps =
   RootStackScreenProps<'Root_ContactScooterOperatorScreen'>;
 
 export const Root_ContactScooterOperatorScreen = ({
   route,
 }: Root_ContactScooterOperatorScreenProps) => {
-  const {vehicleId, operatorId} = route.params;
+  const {operatorId, vehicleId} = route.params;
+  const operators = useOperators();
+  const operatorName = operators.byId(operatorId)?.name;
```
marius-at-atb, line 43: Al the  ?? '' can be just on this line instead.

marius-at-atb (approved):
Looks good!

PR #5116, opened 2025-04-03
fix: Upgrade mapbox version

marius-at-atb:
Context: https://mittatb.slack.com/archives/C07A7TH7ATH/p1743578359152499

Follow up after https://github.com/AtB-AS/mittatb-app/pull/5102

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1018,7 +1018,7 @@ PODS:
     - FirebaseCoreInternal (~> 10.0)
     - GoogleUtilities/Environment (~> 7.12)
     - GoogleUtilities/Logger (~> 7.12)
-  - FirebaseCoreExtension (10.15.0):
+  - FirebaseCoreExtension (10.29.0):
     - FirebaseCore (~> 10.0)
   - FirebaseCoreInternal (10.29.0):
     - "GoogleUtilities/NSData+zlib (~> 7.8)"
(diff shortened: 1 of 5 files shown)


ios/Podfile.lock
```diff
@@ -1018,7 +1018,7 @@ PODS:
     - FirebaseCoreInternal (~> 10.0)
     - GoogleUtilities/Environment (~> 7.12)
     - GoogleUtilities/Logger (~> 7.12)
-  - FirebaseCoreExtension (10.15.0):
+  - FirebaseCoreExtension (10.29.0):
```
jorelosorio, line 1021: Does this one belong to the map?

jorelosorio:
Looks good, documentation says that is supported the version 11 of maps.

```
Using V11
@rnmapbox 10.1 supports both 10.16.* and 11.0.* versions, but defaults to 10.16.*. To use 11.0.* please configure according to your platform:
```

https://rnmapbox.github.io/docs/install?v11-insructions=android#using-v11

But we need to do the same for Android:

```
buildscript {
    ext {
// highlight-start
        RNMapboxMapsVersion = '11.0.0'
// highlight-end
    }
}
```

marius-at-atb:
@jorelosorio Good thinking! I'll look into upgrading Android as well here.

marius-at-atb:
@jorelosorio Android also [updated](https://github.com/AtB-AS/mittatb-app/pull/5116/commits/e6630e51d384de35eee09ab781aa20bb67f866e1) now. Didn't crash in the emulator (although as usual very slow), so let's see on real devices.

ios/Podfile.lock
marius-at-atb: Update: to avoid checksum collision, we leave it in.

jorelosorio (approved):

tormoseng:
✅  Seems to be fine. No Bugsnag issues during the testing. Will be tested and followed up until next release - and after release.

PR #5122, opened 2025-04-07
fix: give vipps button on reservations some width to prevent overflow

rosvik:
There was a bug with flex layout in reservations, where the "Go to vipps for payment" would get 0 width, and a lot of height on android. The theory is that this could lead to double purchases because users couldn't find their active ticket, because it's hidden below a reservation.

closes https://github.com/AtB-AS/kundevendt/issues/20293

### Before/after

<div>

</div>


reidzeibel (approved):
🚀

HanneLH:
Tested ok on Android and iOS ✅

PR #5123, opened 2025-04-07
feat:parking photo scooter

Dahly96:

Added taking picture after stopping a scooter trip

### DIFF
--- a/src/components/map/hooks/use-active-shmo-booking.tsx
+++ b/src/components/map/hooks/use-active-shmo-booking.tsx
@@ -72,8 +72,8 @@ export const useShmoActiveBottomSheet = (
                     });
                   }}
                   photoNavigation={() => {
-                    closeBottomSheet();
                     mapSelectionCloseCallback();
+                    closeBottomSheet();
                     navigation.navigate('Root_ParkingPhotoScreen', {
                       bookingId: activeBooking.bookingId,
                     });
--- a/src/components/map/hooks/use-auto-select-map-item.tsx
+++ b/src/components/map/hooks/use-auto-select-map-item.tsx
@@ -24,6 +24,7 @@ import {useRemoteConfigContext} from '@atb/RemoteConfigContext';
 import {InteractionManager} from 'react-native';
 import {ShmoBookingState} from '@atb/api/types/mobility';
 import {FinishedScooterSheet} from '@atb/mobility/components/sheets/FinishedScooterSheet';
+import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
 
 export type AutoSelectableMapItem =
   | VehicleExtendedFragment
@@ -50,6 +51,7 @@ export const useAutoSelectMapItem = (
   const hasReservationOrAvailableFareContract =
     useHasReservationOrAvailableFareContract();
   const {enable_vipps_login} = useRemoteConfigContext();
+  const {isShmoDeepIntegrationEnabled} = useFeatureTogglesContext();
 
   // NOTE: This ref is not used for anything since the map doesn't support
   // screen readers, but a ref is required when opening bottom sheets.
@@ -106,19 +108,21 @@ export const useAutoSelectMapItem = (
               bottomSheetToAutoSelect?.shmoBookingState ===
               ShmoBookingState.FINISHED
             ) {
-              BottomSheetComponent = (
-                <FinishedScooterSheet
-                  bookingId={bottomSheetToAutoSelect.id}
-                  onClose={closeBottomSheet}
-                  navigateSupportCallback={(operatorId, bookingId) => {
-                    closeBottomSheet();
-                    navigation.navigate('Root_ScooterHelpScreen', {
-                      operatorId,
-                      bookingId,
-                    });
-                  }}
-                />
-              );
+              if (isShmoDeepIntegrationEnabled) {
+                BottomSheetComponent = (
+                  <FinishedScooterSheet
+                    bookingId={bottomSheetToAutoSelect.id}
+                    onClose={closeBottomSheet}
+                    navigateSupportCallback={(operatorId, bookingId) => {
+                      closeBottomSheet();
+                      navigation.navigate('Root_ScooterHelpScreen', {
+                        operatorId,
+                        bookingId,
+                      });
+                    }}
+                  />
+                );
+              }
             } else {
               BottomSheetComponent = (
                 <ScooterSheet
@@ -207,5 +211,6 @@ export const useAutoSelectMapItem = (
     enable_vipps_login,
     hasReservationOrAvailableFareContract,
     close,
+    isShmoDeepIntegrationEnabled,
   ]);
 };


src/components/map/hooks/use-active-shmo-booking.tsx
```diff
@@ -45,7 +41,7 @@ export const useShmoActiveBottomSheet = (
       flyToLocation({
         coordinates: {
           latitude: coordiantes?.latitude ?? 0,
-          longitude: coordiantes?.latitude ?? 0,
+          longitude: coordiantes?.longitude ?? 0,
```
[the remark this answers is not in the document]
jorelosorio: Good catch copilot hehe

src/MapContext.tsx
```diff
@@ -32,6 +32,7 @@ export enum AutoSelectableBottomSheetType {
 export type AutoSelectableBottomSheet = {
   type: AutoSelectableBottomSheetType;
   id: string;
+  state?: ShmoBookingState;
```
marius-at-atb, line 35: Maybe `shmoBookingState` is a better name here, just to make it clear what kind of state it is?

src/mobility/components/sheets/ActiveScooterSheet.tsx
```diff
@@ -52,18 +58,42 @@ export const ActiveScooterSheet = ({
       const startFinishingEvent: ShmoBookingEvent = {
         event: ShmoBookingEventType.START_FINISHING,
       };
-      await sendShmoBookingEvent({
+      const res = await sendShmoBookingEvent({
```
marius-at-atb, line 61: This is an interesting case which might even be discussed with product owner(s).
Is it better to wait until we are sure the startFinishingEvent succeeded, or is it better with optimistic UI – assuming it goes well, and handle problems later if they occur instead.
Totally fine to start like this though :+1:

src/mobility/components/sheets/FinishedScooterSheet.tsx
```diff
@@ -12,58 +12,28 @@ import {useSafeAreaInsets} from 'react-native-safe-area-context';
...
   onClose,
   navigateSupportCallback,
+  bookingId,
 }: Props) => {
-  const {data: activeBooking, isLoading, isError} = useActiveShmoBookingQuery();
   const {t} = useTranslation();
   const {theme} = useThemeContext();
   const styles = useStyles();
-
+  const {data: booking, isLoading, isError} = useShmoBookingQuery(bookingId);
```
marius-at-atb, line 34: Since there is an upcoming initiative regarding booking for boat, maybe we should foresee that it is good practice to always describe the word `booking`.
So in this case, `booking` -> `shmoBooking`.
In other places it could then be something like `boatBooking`, `placeBooking` or `seatBooking`.
@Dahly96 @rosvik @strandlie Thoughts on this?

src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const compressed = await Image.compress(photo.path, {
+      maxHeight: 1024,
+      maxWidth: 1024,
+      quality: 0.7,
+    });
+
+    // Convert to Base64
+    const image = await fetch(compressed);
+    const imageBlob = await image.blob();
+    const base64Image = (await blobToBase64(imageBlob)) as string;
```
marius-at-atb, line 63: In order to get rid of the `as string` both here and in `Root_ParkingViolationsQrScreen`, we can update the util function. I think this will work:
```ts
export const blobToBase64 = (blob: Blob): Promise<string> => {
  const reader = new FileReader();
  reader.readAsDataURL(blob);

  return new Promise((resolve, reject) => {
    reader.onloadend = () => {
      const {result} = reader;
      if (typeof result !== 'string') {
        reject(new Error('Invalid blob type'));
        return;
      }
      resolve(result);
    };

    reader.onerror = () => {
      reject(new Error('FileReader failed'));
    };
  });
};

```

src/stacks-hierarchy/Root_ScooterHelp/Root_ContactScooterOperatorScreen.tsx
```diff
@@ -265,13 +270,14 @@ const useStyles = StyleSheet.createThemeHook((theme) => ({
...
-  vehicleId: string | null,
+  vehicleId: string | undefined,
+  bookingId: string | undefined,
   onSuccess: () => void,
 ) => {
   const {phoneNumber: authPhoneNumberWithPrefix} = useAuthContext();
   const {prefix: authPhonePrefix, phoneNumber: authPhoneNumber} =
     getParsedPrefixAndPhoneNumber(authPhoneNumberWithPrefix);
-  const {data: activeShmoBooking} = useActiveShmoBookingQuery();
+  const {data: booking} = useShmoBookingQuery(bookingId);
```
marius-at-atb, line 280: Here you should remove this line, and just pass in `bookingId` directly below.

src/stacks-hierarchy/navigation-types.ts
```diff
@@ -64,16 +64,26 @@ type Root_ParkingViolationsConfirmationParams = {
...
+    | {vehicleId: string; bookingId?: string}
+    | {vehicleId?: string; bookingId: string}
+  );
+
+type Root_ContactScooterOperatorScreenParams = BaseScooterHelpParams &
+  (
+    | {vehicleId: string; bookingId?: string}
+    | {vehicleId?: string; bookingId: string}
+  );
+
```
marius-at-atb, line 86: Can be simplified I think:
```ts
type ScooterHelpParams = {operatorId: string} & (
   { vehicleId: string } | { bookingId: string }
)
type Root_ScooterHelpScreenParams = ScooterHelpParams;
type Root_ContactScooterOperatorScreenParams = ScooterHelpParams;
```

src/mobility/components/sheets/ActiveScooterSheet.tsx
Dahly96: yes, we will see when testing!

src/components/map/hooks/use-auto-select-map-item.tsx
```diff
@@ -94,36 +102,53 @@ export const useAutoSelectMapItem = (
...
+            if (bottomSheetToAutoSelect?.state === ShmoBookingState.FINISHED) {
+              BottomSheetComponent = (
+                <FinishedScooterSheet
+                  bookingId={bottomSheetToAutoSelect.id}
+                  onClose={closeBottomSheet}
+                  navigateSupportCallback={(operatorId, bookingId) => {
+                    closeBottomSheet();
+                    navigation.navigate('Root_ScooterHelpScreen', {
+                      operatorId: operatorId,
+                      bookingId: bookingId,
```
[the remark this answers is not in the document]
Dahly96: updating

src/stacks-hierarchy/navigation-types.ts
Dahly96: updating

Dahly96: i think im going to keep it as is, because with this typescript cant be sure whats there and will complain when trying to extract it from params

src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
Dahly96: okey, updating

src/stacks-hierarchy/navigation-types.ts
marius-at-atb: The oneOf params can be extracted like so:
```ts
  const vehicleId = 'vehicleId' in route.params ? route.params?.vehicleId : undefined;
  const bookingId = 'bookingId' in route.params ? route.params?.bookingId : undefined;
```

Since we technically want to require one of them, and also don't know what to do with both of them, it is preferable this way.

marius-at-atb (approved):
🙌

src/stacks-hierarchy/navigation-types.ts
Dahly96: yes this works and i was looking at this exact implementation, it just not as readable. But i can make it this way

src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
```diff
@@ -0,0 +1,107 @@
...
+      };
+      return await sendShmoBookingEvent({
+        bookingId: bookingId,
+        shmoBookingEvent: finishEvent,
+      });
+    }
+  };
+
+  const onConfirmImage = async (photo: PhotoFile) => {
+    //const base64Data = await RNFS.readFile(photo.path, 'base64');
```
marius-at-atb, line 53:
```suggestion
```

Dahly96: removed

PR #5124, opened 2025-04-09
fix: color and null value bonus balance

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/19986

fixes to the bonus balance component:
- null now shows as --
- Color for large is now primary, while small is secondary

### DIFF
--- a/src/modules/bonus/components/UserBonusBalance.tsx
+++ b/src/modules/bonus/components/UserBonusBalance.tsx
@@ -16,9 +16,7 @@ const TYPOGRAPHY_BONUS_BALANCE: Record<Props['size'], TextNames> = {
   large: 'body__primary--jumbo--bold',
 };
 
-
 export const UserBonusBalance = ({color, size}: Props) => {
-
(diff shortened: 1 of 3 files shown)


src/modules/bonus/components/UserBonusBalance.tsx
```diff
@@ -15,31 +15,36 @@ const TYPOGRAPHY_BONUS_BALANCE: Record<Props['size'], TextNames> = {
   large: 'body__primary--jumbo--bold',
 };
 
+const getColor = (theme: Theme, size: Props['size']) => {
+  const colorMap: Record<Props['size'], string> = {
+    small: theme.color.foreground.dynamic.secondary,
+    large: theme.color.foreground.dynamic.primary,
+  };
+  return colorMap[size];
+};
```
marius-at-atb, line 24: Deciding color based on size isn't standard, so it could be considered to accept color as a prop to `UserBonusBalance` instead.

marius-at-atb (approved):

PR #5126, opened 2025-04-10
feat: move bottomsheet above tabbar

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20077

* moved bottomsheet above tabbar on demand
* made bottomsheet close when navigating away from the maptab

you could review commit for commit

### DIFF
--- a/src/components/bottom-sheet/AnimatedBottomSheet.tsx
+++ b/src/components/bottom-sheet/AnimatedBottomSheet.tsx
@@ -14,16 +14,15 @@ export function AnimatedBottomSheet({
   animatedOffset,
   children,
   onLayout,
-  tabBarHeight,
+  bottomOffset = 0,
 }: {
   animatedOffset: Animated.Value;
   children: ReactNode;
-  tabBarHeight?: number;
+  bottomOffset?: number;
   onLayout: (ev: LayoutChangeEvent) => void;
 }) {
   const styles = useStyles();
   const {height: windowHeight} = useWindowDimensions();
-  const bottomOffset = tabBarHeight ?? 0;
   const translateY = useMemo(
     () =>
       animatedOffset.interpolate({
@@ -37,7 +36,7 @@ export function AnimatedBottomSheet({
     <View
       style={{
         position: 'absolute',
-        bottom: tabBarHeight,
+        bottom: bottomOffset,
         left: 0,
         right: 0,
         height: windowHeight,
--- a/src/components/bottom-sheet/AnimatedBottomSheet.tsx
+++ b/src/components/bottom-sheet/AnimatedBottomSheet.tsx
@@ -6,6 +6,7 @@ import {
 } from 'react-native';
 import React, {ReactNode, useMemo} from 'react';
 import {StyleSheet, Theme} from '@atb/theme';
+import {shadows} from '../map';
 
 const getThemeColor = (theme: Theme) => theme.color.background.neutral[1];
 
@@ -49,6 +50,7 @@ export function AnimatedBottomSheet({
           ...styles.bottomSheet,
           transform: [{translateY}],
           maxHeight: windowHeight,
+          ...shadows,
         }}
         onLayout={onLayout}
       >
--- a/src/components/bottom-sheet/AnimatedBottomSheet.tsx
+++ b/src/components/bottom-sheet/AnimatedBottomSheet.tsx
@@ -1,4 +1,9 @@
-import {Animated, LayoutChangeEvent, useWindowDimensions} from 'react-native';
+import {
+  Animated,
+  LayoutChangeEvent,
+  useWindowDimensions,
+  View,
+} from 'react-native';
 import React, {ReactNode, useMemo} from 'react';
 import {StyleSheet, Theme} from '@atb/theme';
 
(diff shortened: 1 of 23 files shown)


src/components/bottom-sheet/BottomSheetContext.tsx
```diff
@@ -151,7 +165,7 @@ const BottomSheetOnBackDrop = ({
         easing: Easing.out(Easing.exp),
         useNativeDriver: true,
       }).start(),
-    [animatedOffset, isOpen],
```
marius-at-atb, line 154: I guess `tabBarHeight` isn't needed in this dependency array since the variable isn't used in the effect?

src/components/map/MapV2.tsx
```diff
@@ -216,7 +223,12 @@ export const MapV2 = (props: MapProps) => {
           coordinates: mapPositionToCoordinates(
             featureToSelect.geometry.coordinates,
           ),
-          padding: SLIGHTLY_RAISED_MAP_PADDING,
+          padding: {
+            ...SLIGHTLY_RAISED_MAP_PADDING,
+            paddingBottom:
+              SLIGHTLY_RAISED_MAP_PADDING.paddingBottom + tabBarHeight,
```
marius-at-atb, line 229: Since this code is used in several places, it's probably a good idea to make a custom function for it.

src/components/map/hooks/use-map-selection-change-effect.tsx
```diff
@@ -47,14 +48,15 @@ export const useMapSelectionChangeEffect = (
...
-  useTriggerCameraMoveEffect(cameraFocusMode, mapCameraRef);
+  useTriggerCameraMoveEffect(cameraFocusMode, mapCameraRef, tabBarHeight);
   const {selectedFeature, onReportParkingViolation} =
     useUpdateBottomSheetWhenSelectedEntityChanges(
       mapProps,
       distance,
       mapSelectionAction,
       mapViewRef,
       closeCallback,
+      tabBarHeight,
```
marius-at-atb, line 59: Rather than prop drilling `tabBarHeight`, it can be extracted with `useBottomTabBarHeight` in the leaf node.

src/components/bottom-sheet/BottomSheetHeader.tsx
```diff
@@ -91,7 +96,15 @@ const useStyle = StyleSheet.createThemeHook((theme) => {
       padding: theme.spacing.small,
       borderRadius: 100,
     },
-
+    headerBar: {
+      width: 75,
```
marius-at-atb, line 100: How come this width is hard coded?
Maybe `flex: 1` somewhere could do the trick instead?

src/components/map/hooks/use-map-selection-change-effect.tsx
Dahly96: useTabBarHeight needs to be called in a screen used in the tabbar. So the hook will be undefined if its used directly in a component. Thats why its passed down from map

src/components/bottom-sheet/BottomSheetContext.tsx
Dahly96: yes, forgot to remove this one. There was a point i was trying to see if it would rerender when changing text size while having the sheet open

src/components/bottom-sheet/BottomSheetHeader.tsx
Dahly96: this is the little "bottomsheet" bar thats at the top of the component. The width was set in figma and i went with that. Dont think we have a standard for it, since its new.

image

src/components/map/MapV2.tsx
Dahly96: fixed and made a function

src/components/map/hooks/use-map-selection-change-effect.tsx
marius-at-atb: Right. Still possible to store it in a context to avoid prop drilling I guess, but that's not perfect either so ok..

src/components/map/utils.ts
```diff
@@ -175,6 +176,17 @@ export function flyToLocation({
     });
 }
 
+export function getTabBarPaddingBottom(tabBarHeight: number | undefined) {
```
marius-at-atb, line 179: Should be called something like `getMapPadding` instead.

src/components/bottom-sheet/BottomSheetContext.tsx
```diff
@@ -52,12 +57,16 @@ export const BottomSheetContextProvider = ({children}: Props) => {
   const [contentFunction, setContentFunction] = useState<() => ReactNode>(
     () => () => null,
   );
+  const [tabBarHeight, setTabBarHeight] = useState<number | undefined>(
+    undefined,
+  );
```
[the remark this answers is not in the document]
Dahly96: changed to this

src/components/bottom-sheet/AnimatedBottomSheet.tsx
```diff
@@ -1,43 +1,62 @@
...
 
 export function AnimatedBottomSheet({
   animatedOffset,
   children,
   onLayout,
+  tabBarHeight,
 }: {
   animatedOffset: Animated.Value;
   children: ReactNode;
+  tabBarHeight?: number;
```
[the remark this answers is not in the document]
Dahly96: renamed

tormoseng:
Ref X. The app is crashing when trying to choose a place from the map in the location search.
```
Error · Couldn't find the bottom tab bar height. Are you inside a screen in Bottom Tab Navigator?
```

https://app.bugsnag.com/atb-as/mittatb/errors/680b39e2f0d0ce3f5d25e3f1?filters[event.since]=30d&filters[app.release_stage]=staging&filters[release.seen_in]=1.67%20(Any)

tormoseng:
✅  Error mentioned above is retested OK

The rest is _TBD_

tormoseng:
@Dahly96 As mentioned in the office. If I:
- scan the test e-scooter
- navigate away
- re-enter the map
- open the chosen e-scooter
- close the bottom sheet

Then it seems like the map is in a "chosen e-scooter" state with only minimized e-scooter icons. See screen recording.

<details>

https://github.com/user-attachments/assets/111ac890-4e63-475a-9aa9-21da7fed1cb1

</details>

PR #5127, opened 2025-04-10
feat: update illustration

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20046

Pretty straightforward PR, major version bump to the design system, added light & dark variant for many assets, and maintained compatibility with the OMS apps (should use old assets

### DIFF
--- a/src/error-boundary/FullScreenErrorView.tsx
+++ b/src/error-boundary/FullScreenErrorView.tsx
@@ -1,11 +1,10 @@
 import {Button} from '@atb/components/button';
 import {ThemeText} from '@atb/components/text';
-import {StyleSheet, useThemeContext} from '@atb/theme';
+import {StyleSheet, themes} from '@atb/theme';
 import {ThemedCrashSmall} from '@atb/theme/ThemedAssets';
 import React from 'react';
 import {Linking, View} from 'react-native';
 import {SafeAreaView} from 'react-native-safe-area-context';
-import {Theme} from '@atb/theme/colors';
 import {CUSTOMER_SERVICE_URL} from '@env';
 
 type ErrorProps = {
@@ -15,10 +14,6 @@ type ErrorProps = {
 
 export function FullScreenErrorView({onRestartApp}: ErrorProps) {
   const styles = useStyles();
-  const {theme} = useThemeContext();
-
-  const getBackgroundColor = (theme: Theme) =>
-    theme.color.background.neutral[2];
 
   return (
     <SafeAreaView style={styles.safearea}>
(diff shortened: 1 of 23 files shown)


src/error-boundary/FullScreenErrorView.tsx
```diff
@@ -1,55 +1,51 @@
...
 
 type ErrorProps = {
   onRestartApp: () => void;
   errorCode?: string;
 };
 
-export function FullScreenErrorView({onRestartApp, errorCode}: ErrorProps) {
+export function FullScreenErrorView({onRestartApp}: ErrorProps) {
   const styles = useStyles();
-  const buildNumber = getBuildNumber();
```
marius-at-atb, line 18: Hmm, probably not safe to use `useThemeContext` in this component, since it is a fallback when something else has gone very wrong.
For this particular screen I think we should rather hard code a color.

src/error-boundary/FullScreenErrorView.tsx
```diff
@@ -58,25 +54,36 @@ export function FullScreenErrorView({onRestartApp, errorCode}: ErrorProps) {
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   safearea: {
     flex: 1,
+    justifyContent: 'space-between',
     backgroundColor: theme.color.background.neutral[2].background,
   },
+  svg: {
+    marginTop: 200,
```
jorelosorio, line 61: It looks like there is a lot of spacing; maybe centring is better? Or does it look on small screens?

src/mobility/components/BikeStationBottomSheet.tsx
```diff
@@ -139,7 +139,7 @@ export const BikeStationBottomSheet = ({
                     />
                     <BrandingImage
                       logoUrl={brandLogoUrl}
-                      fallback={<ThemedCityBike />}
+                      fallback={<ThemedCityBike height={50} width={50} />}
```
jorelosorio, line 142: It would be nice to have that 50 somewhere in the theme if we use it frequently

src/error-boundary/FullScreenErrorView.tsx
reidzeibel: removed the margin, it looks ok

src/error-boundary/FullScreenErrorView.tsx
reidzeibel: updated

marius-at-atb (approved):
Lgtm!
Also +1 for the thorough screenshot overview in the PR description!

PR #5131, opened 2025-04-23
docs: update problem fixing command

adriansberg:
`bundle exec pod install` should be used instead of the `npx pod-install
ios` due to how dependencies are managed in the commands.


rosvik (approved):
🚀

PR #5134, opened 2025-04-24
feat: let the user cancel ongoing payments

rosvik:
_Might make sense to review one commit at a time_

This PR allows the user to cancel ongoing payments, and prevents several purchase flows going at once. 

Also sets a `isUser` query parameter for all cancel calls, indicating whether or not it was manually triggered by the user. This should allow us to monitor the amount of people who use the button.

https://github.com/user-attachments/assets/b850216c-88f6-4316-85fe-f4a4057dc629

closes https://github.com/AtB-AS/kundevendt/issues/20293

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -119,13 +119,12 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
     recipient,
     shouldSavePaymentMethod,
   });
-  const cancelPaymentMutation = useCancelPaymentMutation();
-  useEffect(() => {
-    if (cancelPaymentMutation.isSuccess) {
+  const cancelPaymentMutation = useCancelPaymentMutation({
+    onSuccess: () => {
       cancelPaymentMutation.reset();
       reserveMutation.reset();
-    }
-  }, [cancelPaymentMutation, reserveMutation]);
+    },
+  });
 
   useOpenVippsAfterReservation(
     reserveMutation.data?.url,
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -213,7 +213,10 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
             reserveMutation.reset();
             setVippsNotInstalledError(false);
             if (reserveMutation.isSuccess) {
-              cancelPaymentMutation.mutate(reserveMutation.data);
+              cancelPaymentMutation.mutate({
+                reservation: reserveMutation.data,
+                isUser: false,
+              });
               analytics.logEvent('Ticketing', 'Payment cancelled');
             }
             setSelectedPaymentMethod(paymentMethod);
@@ -237,7 +240,10 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
           type: 'back',
           onPress: () => {
             if (reserveMutation.isSuccess) {
-              cancelPaymentMutation.mutate(reserveMutation.data);
+              cancelPaymentMutation.mutate({
+                reservation: reserveMutation.data,
+                isUser: false,
+              });
               analytics.logEvent('Ticketing', 'Payment cancelled');
             }
             navigation.pop();
@@ -324,7 +330,10 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
           onGoToPayment={goToPayment}
           onCancelPayment={() => {
             if (reserveMutation.isSuccess) {
-              cancelPaymentMutation.mutate(reserveMutation.data);
+              cancelPaymentMutation.mutate({
+                reservation: reserveMutation.data,
+                isUser: true,
+              });
             }
           }}
         />
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -10,7 +10,7 @@ import {
 } from '@atb/modules/global-messages';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy/navigation-types';
 import {StyleSheet, useThemeContext} from '@atb/theme';
-import {PaymentType, ReserveOffer} from '@atb/ticketing';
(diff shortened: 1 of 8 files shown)


src/ticketing/api.ts
```diff
@@ -170,11 +170,12 @@ export async function reserveOffers({
...
-  payment_id: number,
-  transaction_id: number,
+  paymentId: number,
+  transactionId: number,
+  isUser: boolean,
 ): Promise<void> {
-  const url = `ticket/v3/payments/${payment_id}/transactions/${transaction_id}/cancel`;
-  await client.put(url, {}, {authWithIdToken: true});
+  const url = `ticket/v3/payments/${paymentId}/transactions/${transactionId}/cancel?isUser=${isUser}`;
+  await client.put(url, undefined, {authWithIdToken: true});
```
rosvik, line 178: ☝️ FYI @Nisha1306 and @adriansberg

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -297,61 +295,101 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
+          onSelectPaymentMethod();
+        }}
+        testID="choosePaymentMethodButton"
+        ref={onCloseFocusRef}
+      />
+    );
+
+  return (
+    <Button
+      expanded={true}
```
adriansberg, line 443:
```suggestion
      expanded
```

adriansberg:
Based on the recording in the description, even though the user cancels payment the ticket is still processed and active?

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -112,6 +120,12 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
     shouldSavePaymentMethod,
   });
   const cancelPaymentMutation = useCancelPaymentMutation();
+  useEffect(() => {
+    if (cancelPaymentMutation.isSuccess) {
+      cancelPaymentMutation.reset();
+      reserveMutation.reset();
+    }
```
marius-at-atb, line 127: To avoid the useEffect, this function can be passed in to `onSuccess` in `useCancelPaymentMutation` instead. (Same as in e.g. `useSendSupportRequestMutation`)

marius-at-atb:
@adriansberg Good question - is it possible to send a cancel callback or listener to the pop up? When does a user not want to cancel the payment, having already pressed cancel once?

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
rosvik: Ah, thanks! That's much cleaner.

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
rosvik: The app code mostly has `={true}` set for props like this. I'm open to changing that convention as bigger refactor (maybe an ESLint rule for one or the other), but for now I'll stick with this.

adriansberg: Ah, TIL! Let's stick to that convention then.

rosvik:
> Based on the recording in the description, even though the user cancels payment the ticket is still processed and active?

The in app browser we use for NETS doesn't seem to support a `onClose` callback, which means I haven't found a way to cancel the payment automatically. Please let me know if I've missed something 😅 https://github.com/proyecto26/react-native-inappbrowser

This flow makes more sense for Vipps, where the app is waiting for payments in the background while Vipps is open. If you then return to the app before completing payment, it makes more sense that you have to manually cancel.

adriansberg:
> This flow makes more sense for Vipps, where the app is waiting for payments in the background while Vipps is open. If you then return to the app before completing payment, it makes more sense that you have to manually cancel.

That sounds like it makes a lot more sense, yes. Been experiencing that confusion myself when the Vipps payment isn't completed.

adriansberg (approved):

marius-at-atb:
https://github.com/proyecto26/react-native-inappbrowser/issues/415#issuecomment-1647860408
Haven't checked, but this might work for onClose handling?

The Vipps case makes sense, so I guess we want to include the solution from this PR anyway :+1:

rosvik:
> Haven't checked, but this might work for onClose handling?

Aaahh, I've been looking in the entirely wrong place 🤦 Thanks!

rosvik:
We try again! Added a little refactor in 28621b453b2d9ed47a18bdb4d8cfa8d0db6ed82d, and added a `onCancel` callback in 70b34d1f3ef2867d46016837cb92f2eeb60571a0 ✨ 

https://github.com/user-attachments/assets/35b04444-f600-4903-8e8c-9f6025b3c5e2

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -297,64 +302,159 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
   );
 };
 
+type PaymentButtonProps = {
+  isSearchingOffer: boolean;
+  isOfferError: boolean;
+  totalPrice: number;
+  mode: 'TravelSearch' | 'Ticket' | undefined;
+  onCloseFocusRef: RefObject<any>;
+  reserve: OfferReservation | undefined;
```
marius-at-atb, line 342: Could we find a name/solution that makes the difference between `OfferReservation` and `Reservation` more clear here?
Atm. `reserveMutation.data` is used both as `reservation` for `cancelPaymentMutation.mutate`, and `reserve` here, which I didn't intuitively understand 🤔

marius-at-atb (approved):
Nice work!

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
rosvik: Good point! Looks like `ReserveOffer` is used in the new sales service, so we could start using that for naming. https://github.com/AtB-AS/sales/blob/596055de91dc22b05b5b0c243993b75b467cb0d2/sales-service/src/handlers/sales/reserve.rs#L85

rosvik:
@adriansberg I may have made you some more conflicts with this 🙈 7d99ea5d8183c36001577ab27595de053b585838

PR #5141, opened 2025-04-28
feat: use 'cancel' instead of 'undo' for ticket refunds

rosvik:
We got some feedback that we should use "cancel" for refunds on English in the app. Still using "angre" in Norwegian.


marius-at-atb (approved):

Sebstorvik (approved):
LGTM🚀

PR #5142, opened 2025-04-29
refactor: create native-bridges module

rosvik:
As part of https://github.com/AtB-AS/kundevendt/issues/20391, I propose that we collect all code that uses `NativeModules` in `modules/native-bridges`. Should make it easier to keep code on the native side in sync.

Cleaned up a few tests as well, since some were failing.


src/travel-details-screens/utils.ts
```diff
@@ -506,7 +505,7 @@ export function getLineAndTimeA11yLabel(
       : undefined,
   ]
     .filter(isDefined)
-    .join(screenReaderPause);
```
rosvik, line 509: Shouldn't think this would cause a test to fail, but due to imports dependent on theme in `@atb/components/text`, and the "ChangeNativeBridge module is not linked" check, it did 🫠

marius-at-atb (approved):

PR #5149, opened 2025-05-02
feat: Add paymentflow in scootersheet

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20075

Added paymentcard component in the scootersheet to be able to switch between paymentcards, added navigation to the profile page for adding a new paymentcard
Also finished the onboarding component to add a paymentcard before being able to start a scootertrip

might be able to do commit for commit from 30.april


src/modules/payment/SelectPaymentMethodSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+  PaymentMethod,
+  PaymentSelection,
+  SavedPaymentMethodType,
+  VippsPaymentMethod,
+} from './types';
+import {SinglePaymentMethod} from './SinglePaymentMethod';
+import {MultiplePaymentMethodsRadioSection} from './MultiplePaymentMethodsRadioSection';
+
+type Props = {
+  onSelect: (
```
strandlie, line 28: Is `onSelect` and `currentOptions` a kind of `[state, setState]` from above? It would seem so since both has `paymentMethod` and `shouldSavePaymentMethod` as properties. In that case could we name it something like `setCurrentOptions` and `currentOptions`?

src/modules/payment/SelectPaymentMethodSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+  return (
+    <BottomSheetContainer
+      title={t(SelectPaymentMethodTexts.header.text)}
+      onClose={onClose}
+    >
+      <ScrollView>
+        <View style={{flex: 1}}>
+          <View style={styles.paymentMethods}>
+            <View>
+              {recurringPaymentMethods &&
```
strandlie, line 87: This `recurringPaymentMethods` is redundant when you use `recurringPaymentMethods?.length`after

src/modules/payment/SelectPaymentMethodSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+                  onSelect={(val: PaymentSelection) => {
+                    setSelectedMethod(val);
+                  }}
+                  index={index}
+                />
+              ))}
+            </View>
+
+            <View>
+              {recurringPaymentMethods &&
```
strandlie, line 112: Redundant as well

strandlie (approved):
Looks good! Suggested some small improvements, if you want

src/modules/payment/SelectPaymentMethodSheet.tsx
Dahly96: iv not really done anything to this file other then moving it, so i dont think i should be changing anything in this PR, but it might be right what you mention here

src/components/map/hooks/use-update-bottom-sheet-when-selected-entity-changes.tsx
```diff
@@ -92,6 +92,26 @@ export const useUpdateBottomSheetWhenSelectedEntityChanges = (
...
+
+                  setTimeout(() => {
+                    navigation.navigate('Root_TabNavigatorStack', {
+                      screen: 'TabNav_ProfileStack',
+                      params: {
+                        screen: 'Profile_PaymentMethodsScreen',
+                      },
+                    });
+                  }, 100);
+                }}
```
marius-at-atb, line 114: Rather than navigating twice with a timeout, can you look into achieving this with just one call?
`navigation.reset` might be helpful, which you can see an example of in the `useEnterApp` hook

Dahly96: I created a custom hook to solve this better. It now keeps the hierarchy when navigating to a screen inside another tab. The only drawback is that the state in the other tabs is reset. So after adding a card, if you go back to the map its like you opened the map for the first time. But i think for this case its the intended usage, you are supposed to be taken out and away if you want to add a paymentcard

marius-at-atb (changes requested):
Very nice, I like the new payment module.
See comments :)

src/components/map/hooks/use-enter-payment-methods.tsx
```diff
@@ -0,0 +1,41 @@
...
+                  ],
+                  index: 1,
+                },
+              },
+            ],
+            index: 0,
+          },
+        },
+      ],
+    });
```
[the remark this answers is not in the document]
Dahly96: updated with a fix

src/components/map/hooks/use-auto-select-map-item.tsx
```diff
@@ -93,6 +94,97 @@ export const useAutoSelectMapItem = (
...
+    (vehicleId: string) => {
+      async function selectPaymentMethod(vehicleId: string) {
+        openBottomSheet(
+          () => {
+            return (
+              <SelectShmoPaymentMethodSheet
+                onSelect={() => {
+                  openScooterSheet(vehicleId);
+                }}
+                onClose={closeBottomSheet}
```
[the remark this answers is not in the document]
Dahly96: updated with a fix, added tabbarheight to the paymentbottomsheet aswell, since its used in the map

src/components/map/utils.ts
```diff
@@ -187,6 +189,12 @@ export function getMapPadding(tabBarHeight: number | undefined) {
   }
 }
 
+export function isCardPaymentMethod(
```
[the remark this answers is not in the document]
Dahly96: agreed, moving it

src/MapContext.tsx
```diff
@@ -43,6 +56,38 @@ export const MapContextProvider = ({children}: Props) => {
...
+      }
+    }
+
+    //use fallback
+    if (recurringPaymentMethods?.length) {
+      const fallbackPaymentMethod =
+        recurringPaymentMethods[recurringPaymentMethods.length - 1];
+      setSelectedPaymentMethod(fallbackPaymentMethod);
+    }
+  }, [previousPaymentMethod, recurringPaymentMethods]);
```
marius-at-atb, line 89: It is a bit unintuitive that payment info is stored in MapContext.
To make it more obvious why this is, I suggest we extract the logic into a hook, and then use it like this here:
```ts
const [selectedShmoPaymentMethod, setSelectedShmoPaymentMethod] = useSelectedShmoPaymentMethod()
```

marius-at-atb (approved):

PR #5154, opened 2025-05-05
feat: Prompt in app review prompt after a news / information was read

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/20328

image


src/utils/use-in-app-review.ts
```diff
@@ -13,84 +13,93 @@ export function useInAppReviewFlow() {
...
+          }
+        } catch (error) {
+          // Log and report any errors encountered during the in-app review request.
+          // For detailed error codes, refer to:
+          // https://github.com/MinaSamir11/react-native-in-app-review#errors-and-google-play-app-store-error-codes
+          notifyBugsnag(
+            `Error: ${(error as Error).message}, Stack: ${
+              (error as Error).stack
+            }, Context: ${context}`,
           );
```
rosvik, line 97: I think you can pass `error` as the first param like before, then context as metadata (see the notifyBugsnag definition).

rosvik:
Ok, but can we still make them more consistent? Maybe 

'Announcements: Bottom sheet dismissed'
'Trip details: Bus in map dismissed'
'Departure details: Bus in map dismissed'

reidzeibel (approved):

PR #5156, opened 2025-05-05
fix: Use boat icon instead of ferry icon for boat

marius-at-atb:
Fixes a bug where the ferry icon was used for boat as well in the map.

| Before | After |
|--------|-------|
|  |  |


rosvik (approved):

tormoseng:
✅

PR #5160, opened 2025-05-07
feat:fixed locationarrow onclick

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20585


marius-at-atb (approved):

rosvik:
@Dahly96 FYI I think not having a space after `feat:` breaks the release note generator, so best to include that going forward ✌️

PR #5162, opened 2025-05-08
feat: added reset consent in debug menu

Dahly96:

added reset consent in debug menu and some small changes to css in scootersheet. Also changed flyTo on the locationarrow in map to moveTo, so it does not animate in the map over longer distances, reducing api calls to the tileserver when moving quickly over the map.

iv change flyTo function for all usage. So it should probably be tested where its used. From click in item in map, location arrow inn map, starting scooter trip, follow bus, scanning qr and if there is other cases of using flyTo in map


src/mobility/components/sheets/ScooterSheet.tsx
```diff
@@ -123,8 +124,14 @@ export const ScooterSheet = ({
               {defaultPaymentMethod &&
                 isShmoDeepIntegrationEnabled &&
                 isMapV2Enabled &&
+                !hasBlockers &&
                 operatorIsIntegrationEnabled && (
-                  <Section style={{paddingHorizontal: theme.spacing.medium}}>
+                  <Section
+                    style={{
+                      paddingHorizontal: theme.spacing.medium,
```
reidzeibel, line 131: I think we could extract these as well to match the existing code style

reidzeibel (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_DebugInfoScreen.tsx
```diff
@@ -256,6 +256,10 @@ export const Profile_DebugInfoScreen = () => {
             text="Reset one time popovers"
             onPress={() => storage.remove(StorageModelKeysEnum.OneTimePopOver)}
           />
+          <LinkSectionItem
+            text="Reset scooter consent"
+            onPress={() => storage.remove('@ATB_scooter_consent')}
```
marius-at-atb, line 261: We should probably extract the `@ATB_scooter_consent` into an enum and use it both here and in `useShmoRequirements`.

src/components/map/MapV2.tsx
```diff
@@ -308,7 +308,7 @@ export const MapV2 = (props: MapProps) => {
                   padding: getMapPadding(tabBarHeight),
                   mapCameraRef,
                   zoomLevel: 15,
-                  animationDuration: 200,
+                  animationMode: 'moveTo',
```
marius-at-atb, line 311: Here I was thinking to only use this option if the distance is e.g. > 10km.
By making the adjustment inside the `flyToLocation` function, it will also work for the flying after scanning.

marius-at-atb (approved):
Lgtm!

PR #5137, opened 2025-04-25
refactor: use sales service

adriansberg:
This pull request focuses on transitioning from use ticket-legacy to using the new sales service for endpoints related to ticketing.

We've discovered some discrepancies along the way while transitioning the endpoints, related pull requests in the sales repository are listed in the end of this description.

In addition to transitioning to using the new endpoints we focused on making the switch to using camelCase for all of the responses from the backend, so some of the backend PRs are related to that as well. 

closes https://github.com/AtB-AS/kundevendt/issues/20453

---

related backend PRs: 
- https://github.com/AtB-AS/sales/pull/43
- https://github.com/AtB-AS/sales/pull/40
- https://github.com/AtB-AS/sales/pull/39
- https://github.com/AtB-AS/sales/pull/34
- https://github.com/AtB-AS/sales/pull/33
- 
- https://github.com/AtB-AS/sales/pull/31
- https://github.com/AtB-AS/sales/pull/35
- https://github.com/AtB-AS/sales/pull/47

PR #5164, opened 2025-05-08
fix: migrate saved payment methods to new format

rosvik:
This fixes a crash that was caused by some small breaking changes when migrating to the new sales service (https://github.com/AtB-AS/mittatb-app/pull/5137). Since this change forces us to migrate to a new format in local storage, it also gave us the opportunity to change the storage format a bit. I tried to simplify things, and be more consistent with naming:

- Removes `SavedPaymentMethodType`, since this is redundant information that isn't really used for anything.
- Renaming `StoredPaymentMethods` to `PreviousPaymentMethods`.
- Renames `recurringCard` to `recurringPayment`.


reidzeibel (approved):
Looks clean to me!

src/modules/payment/previous-payment-utils.ts
```diff
@@ -3,12 +3,14 @@ import {storage} from '@atb/storage';
...
 import {useAuthContext} from '@atb/auth';
 import {useListRecurringPaymentsQuery} from '@atb/ticketing/use-list-recurring-payments-query';
-import {PaymentMethod, SavedPaymentMethodType} from './types';
+import {PaymentMethod} from './types';
 import {useFirestoreConfigurationContext} from '@atb/configuration';
 import {parseISO} from 'date-fns';
 import {PaymentType, listRecurringPayments} from '@atb/ticketing';
 import {onlyUniques} from '@atb/utils/only-uniques';
 
+const PREVIOUS_PAYMENT_METHODS_STORAGE_KEY = '@ATB_previous_payment_methods';
```
marius-at-atb, line 12: This should probably be defined in `StorageModel.tsx`?
Also I think that users will have to re-enter their info due to the renaming of the string 🤔

src/modules/payment/SelectPaymentMethodSheet.tsx
```diff
@@ -51,10 +50,7 @@ export const SelectPaymentMethodSheet: React.FC<Props> = ({
 
   const {paymentTypes} = useFirestoreConfigurationContext();
   const defaultPaymentMethods: PaymentMethod[] = paymentTypes.map(
-    (paymentType) => ({
-      paymentType,
-      savedType: SavedPaymentMethodType.Normal,
-    }),
+    (paymentType) => ({paymentType}),
```
strandlie, line 53: Is this map really necessary now? Can't it just be deleted?

src/modules/payment/previous-payment-utils.ts
rosvik: > This should probably be defined in StorageModel.tsx?

Right, I used the enum now 👍 Reminds me we really should clean this up soon.

rosvik:
Some guy just moved all the files on master, so I had a bunch of conflicts here 😡 Should be good to go now!

PR #5167, opened 2025-05-09
fix: don't read zones if there are no zones on recent ticket results.

reidzeibel:
Fixes a crash that happens when you have "Recent Tickets" purchased in around 2024, I think the newly updated recent tickets endpoint caused this, so we need to guard this by checking if the zones are defined or not in the code.

To test this, use test account 45645645 and open the ticket purchase screen.

Backend team will make a fix as well on their side.

Acceptance Criteria: 
- [ ] App does not crash if the recent ticket doesn't have zones, try account 45645645

### DIFF
--- a/src/recent-fare-contracts/use-recent-fare-contracts.ts
+++ b/src/recent-fare-contracts/use-recent-fare-contracts.ts
@@ -121,11 +121,11 @@ const mapBackendRecentFareContracts = (
 
   const fromTariffZone = findReferenceDataById(
     tariffZones,
-    recentFareContract.zones[0],
+    recentFareContract.zones?.[0],
   );
   const toTariffZone = findReferenceDataById(
     tariffZones,
-    recentFareContract.zones.slice(-1)[0],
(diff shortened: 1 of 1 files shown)


tormoseng:
@reidzeibel For my test user, I see the recent ticket, but the zones are not visible even though it exists in /recent. This is the recent response:
```json
[{
	"products": ["ATB:PreassignedFareProduct:8808c360"],
	"zones": ["ATB:FareZone:10"],
	"users": {
		"ATB:UserProfile:8ee842e3": 1
	},
	"paymentMethod": "MASTERCARD",
	"totalAmount": "45.00",
	"createdAt": "2025-05-09T09:11:03Z"
}]
```

tormoseng:
But will this be fixed in your task going from TariffZones > FareZones in the app?

reidzeibel:
@tormoseng yes it will be fixed after my task is ready, this one is just to prevent crash when the zones are not there.

tormoseng:
✅  Tested OK!

PR #5168, opened 2025-05-09
fix: save recurring payments as previous payment method

rosvik:
There was a bug related to the onSuccess callback on the reserveMutation in PurchaseConfirmationScreen that broke saving of recurring payment.

It seems the root of the issue was a "stale closure" problem, and works as follows:

1. A closure was provided as useReserveOfferMutation's onSuccess, that had a callback to onPaymentCompleted.
2. onPaymentCompleted is a useCallback, with a dependency array updating the function every time recurringPaymentId changes.
3. The assumption was that once recurringPaymentId changes, the onSuccess closure would update as well. But this is not the case, since onSuccess itself doesn't have a dependency array, and becomes stale if any state other that the argument (reserveOfferResponse) changes.

I believe this is why the onSuccess, onError and onSettled callbacks are deprecated and removed in TanStack Query v5.
https://tanstack.com/query/v4/docs/framework/react/reference/useQuery

closes https://github.com/AtB-AS/kundevendt/issues/20437

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
@@ -53,6 +53,7 @@ import SvgClose from '@atb/assets/svg/mono-icons/actions/Close';
 import {ThemeText} from '@atb/components/text';
 import {MutationStatus} from '@tanstack/react-query';
 import {PaymentSelectionSectionItem} from '@atb/modules/payment';
+import {useDoOnceWhen} from '@atb/utils/use-do-once-when';
 
 type Props = RootStackScreenProps<'Root_PurchaseConfirmationScreen'>;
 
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -113,22 +114,27 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
-              isUser: false,
-            }),
-        );
-      }
-    },
   });
+  useDoOnceWhen(
+    () => {
+      if (reserveMutation.status !== 'success') return;
+      if (paymentMethod?.paymentType === PaymentType.Vipps) return;
```
marius-at-atb, line 121: I suppose these lines should be a part of the condition instead. (The first one already is actually)

marius-at-atb (approved):

Oleninkaa (approved):
Good work! Now we save and receive correct data about the previous payment method. I tested it on the bug about showing the previous payment card, and now it works correctly! 👍🏻

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
marius-at-atb: Right, keep as is then 👍

PR #5176, opened 2025-05-12
refactor: create mobility and map modules

rosvik:
- Moved `src/mobility` to `src/modules/mobility`
- Moved `src/components/map` to `src/modules/map`
- Moved `src/MapContext.tsc` into `src/modules/map`


marius-at-atb (approved):
Lgtm

PR #5179, opened 2025-05-13
refactor: migrate TariffZones to FareZones

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20567

Renamed almost everything `TariffZone`to `FareZone`, the exception is the stuff from Entur GraphQL (generated from AtB BFF), these are not changed at all. The `TariffZone` returned by Entur GraphQL queries also contains the `FareZone` value, so it is safe to leave it as is.


src/fare-zones-selector/FareZoneResults.tsx
```diff
@@ -3,69 +3,69 @@ import {Pin} from '@atb/assets/svg/mono-icons/map/';
...
-                accessibilityHint={t(TariffZoneSearchTexts.zones.item.a11yHint)}
+                accessibilityHint={t(FareZoneSearchTexts.zones.item.a11yHint)}
                 accessibilityRole="button"
                 hitSlop={insets.symmetric(8, 1)}
-                onPress={() => onSelect(tariffZone)}
-                style={styles.tariffZoneButton}
-                testID={'tariffZone' + tariffZone.name.value + 'Button'}
+                onPress={() => onSelect(fareZone)}
+                style={styles.fareZoneButton}
+                testID={'fareZone' + fareZone.name.value + 'Button'}
```
rosvik, line 53: Probably need to update tests in `e2e/` when you change testIDs.

src/screen-components/travel-details-screens/utils.ts
```diff
@@ -246,14 +246,14 @@ export function getFilteredLegsByWalkOrWaitTime(tripPattern: TripPattern) {
 
 export function canSellCollabTicket(tripPattern: TripPattern) {
   const someLegsByTrain = someLegsAreByTrain(tripPattern);
-  const tariffZonesHaveZoneA = (tariffZones?: TariffZone[]) =>
-    tariffZones?.some((a) => a.id === 'ATB:TariffZone:1');
+  const fareZonesHaveZoneA = (fareZones?: FareZone[]) =>
+    fareZones?.some((a) => a.id === 'ATB:FareZone:1');
```
rosvik, line 250: Zone A is `ATB:FareZone:10` now

```suggestion
    fareZones?.some((a) => a.id === 'ATB:FareZone:10');
```

src/modules/favorites/types.ts
```diff
@@ -22,7 +22,7 @@ export type SearchLocation = {
   category: FeatureCategory[];
   label?: string;
   postalcode?: string;
-  tariff_zones?: string[];
+  fare_zones?: string[];
```
rosvik, line 25: I don't think we can or should change this field. It's connected to the type returned by the geocoder, which I guess haven't changed?

(Also, we store data on this format in local storage, so when changing it, we need to consider backwards compatibility)

rosvik (commented):
Good work 💯 Just a few minor things

src/fare-zones-selector/FareZoneResults.tsx
reidzeibel: I've updated stuff on the e2e purchase overview part 👍🏼

rosvik (approved):

PR #5182, opened 2025-05-13
fix: support tinted variant of widget

rosvik:
This fixes an issue where the widget would have unreadable information if tinted/accented mode is enabled on iOS 18.

## Acceptance criteria

- [ ] Widget has readable information when the home screen is set to Tinted (edit home screen -> Edit -> Customize -> Tinted)
- [ ] The widget still works with older iOS versions (Older than iOS 18)

### DIFF
--- a/ios/departureWidget/Views/ChipView.swift
+++ b/ios/departureWidget/Views/ChipView.swift
@@ -1,6 +1,8 @@
 import SwiftUI
 
 struct ChipView: View {
+    @Environment(\.widgetRenderingMode) var widgetRenderingMode
+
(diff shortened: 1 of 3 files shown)


strandlie (approved):

strandlie:
Fun to approve a Swift-PR!

reidzeibel (approved):

rosvik:
Seems to be crashing the build due to some iOS 16 features being used without a version check. Proposed fix here: https://github.com/AtB-AS/mittatb-app/pull/5184

tormoseng:
✅

PR #5183, opened 2025-05-14
fix: prevent crash on Android when selecting favorite departure

reidzeibel:
Fixes issue `Cannot read property 'refreshWidgets' of null` from [BugSnag](https://app.bugsnag.com/atb-as/mittatb/errors?filters[event.since]=all&filters[app.release_stage][]=store&filters[app.release_stage][]=production&filters[search][]=refresh&filters[search][]=widgets&filters[app.type]=android)

### Steps to reproduce
1. Go to departures/avganger
2. Select any bus stops
3. Select any bus departures
4. Mark as favorite
5. Select any options that are shown on the bottom sheet
6. App crashes

### Acceptance Criteria
- [ ] Mark as favorite does not crash the app on Android
- [ ] Mark as favorite works as intended on iOS and Android

### DIFF
--- a/src/modules/native-bridges/widget-updater.ts
+++ b/src/modules/native-bridges/widget-updater.ts
@@ -1,4 +1,4 @@
-import {NativeModules} from 'react-native';
+import {NativeModules, Platform} from 'react-native';
 
 // Found in /ios/BridgeModules/WidgetUpdaterBridge.m
 interface WidgetUpdaterBridge {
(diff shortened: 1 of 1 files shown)


reidzeibel:
@tormoseng this should be on 1.67 because it will fix quite a lot of issues 

image

tormoseng (approved):

rosvik (approved):
😬 Good catch!

I'm having doubts if throwing these errors are a good idea at all. I thought they would be easier to spot, and meant we would spot bugs earlier, but there might be better ways to do that.

reidzeibel:
Merging now, should test on Staging both iOS and Android

tormoseng:
✅  Android
❓ iOS shortly

PR #5184, opened 2025-05-14
fix: set minimum iOS version for widget to 16.6

rosvik:
Some of the features used in c0d3da03 requires iOS 16, and makes the build fail, since it targets iOS 15 and above. It's easier to bump the minimum version than to add version checks, and will affect a negligible amount of users, so I propose we do that for the widget and intent. (not the app itself)


rosvik:
Build seems to be working, but will take a while to complete. Build ID will be 1747228047 if you want to test on a device with iOS <16 @tormoseng

reidzeibel (approved):

tormoseng:
✅  Seems to work as before on iOS 15 - but without the widget.

PR #5186, opened 2025-05-15
chore: remove validate function from tokenService

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20600


marius-at-atb (approved):
Lgtm

PR #5188, opened 2025-05-16
feat: change to new token server endpoint

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20603

The changing of endpoint itself is trivial, but to make this work, it requires the PR below from Platform Team (all merged and ready)
- [Fixing cleanup function](https://github.com/AtB-AS/token/pull/29), before the PR, the newly activated token is immediately removed, causing the token to always turn to fallback state.
- [Expose device details](https://github.com/AtB-AS/token/pull/28), without this PR, tokens will always be renewed once it is activated, because the logic check returns undefined.
- [Change to v1 instead of v4](https://github.com/AtB-AS/token/pull/27), fixes routing issue
- [Add inspection](https://github.com/AtB-AS/token/pull/25), fixes missing inspection issue.

### DIFF
--- a/src/modules/mobile-token/hooks/use-validate-token.tsx
+++ b/src/modules/mobile-token/hooks/use-validate-token.tsx
@@ -178,7 +178,7 @@ export const useValidateToken = (
 
       if (token && token.type === 'mobile') {
         checkShouldRenew(token).then((shouldRenew) => {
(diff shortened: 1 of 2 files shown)


rosvik (approved):
Clean 👌

PR #5190, opened 2025-05-16
refactor: Refactor use-trips-query to use React Query

strandlie:
The intention is to keep the functionality the same, which it seems to do. 

A little unsure whether I should also remove Axios from the base `tripsSearch`-fetch function. Please give your input.

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
@@ -1,22 +1,20 @@
-import {CancelToken, isCancel} from '@atb/api';
 import {tripsSearch} from '@atb/api/trips';
-import {Modes} from '@atb/api/types/generated/journey_planner_v3_types';
 import {TripPattern} from '@atb/api/types/trips';
 import {ErrorType, getAxiosErrorType} from '@atb/api/utils';
 import {Location} from '@atb/modules/favorites';
-import {useRemoteConfigContext} from '@atb/modules/remote-config';
 import {useSearchHistoryContext} from '@atb/modules/search-history';
 import type {SearchStateType, TripSearchTime} from '../types';
 
 import {isValidTripLocations} from '@atb/utils/location';
 import Bugsnag from '@bugsnag/react-native';
-import {CancelTokenSource} from 'axios';
-import {useCallback, useEffect, useRef, useState} from 'react';
+import {useCallback, useEffect, useMemo, useState} from 'react';
 import {useJourneyModes} from './hooks';
 import {useAnalyticsContext} from '@atb/modules/analytics';
 import {TravelSearchFiltersSelectionType} from '@atb/modules/travel-search-filters';
 import {TripPatternWithKey} from '@atb/screen-components/travel-details-screens';
 import {createQuery, sanitizeSearchTime, SearchInput} from './utils';
+import {useInfiniteQuery} from '@tanstack/react-query';
+import {useRemoteConfigContext} from '@atb/modules/remote-config';
 
 export function useTripsQuery(
   fromLocation: Location | undefined,
@@ -29,8 +27,7 @@ export function useTripsQuery(
 ): {
   tripPatterns: TripPatternWithKey[];
   timeOfLastSearch: string;
-  loadMore: (() => {}) | undefined;
-  clear: () => void;
+  loadMore: (() => void) | undefined;
   searchState: SearchStateType;
   error?: ErrorType;
 } {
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [fromLocation, toLocation, searchTime, filtersSelection],
+  const {data, error, fetchNextPage, hasNextPage, isFetching, status} =
+    useInfiniteQuery(
+      ['trips', fromLocation, toLocation, searchTime, filtersSelection],
+      fetchTrips,
+      {
+        getNextPageParam: (lastPage) => lastPage.nextPageCursor,
+        retry: 3,
```
rosvik, line 108: Were there any retries on error before this change? I guess it makes sense to have on things like network errors, but trip search is quite expensive compared to other journey planner requests (and has a lower rate limit), so we should be careful about making a change here.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
-  }, [search, pageCursor, tripPatterns]);
+  useEffect(() => {
+    if (status === 'success' && tripPatterns.length > 0) {
+      analytics.logEvent('Trip search', 'Search performed', {
+        searchTime,
+        filtersSelection: toLoggableFiltersSelection(filtersSelection),
+        numberOfHits: tripPatterns.length,
+      });
+    }
+  }, [status, tripPatterns, analytics, searchTime, filtersSelection]);
```
rosvik, line 169: Could maybe use a `useDoOnceWhen` instead of useEffect here, or at least consider it. I used it for a similar case in https://github.com/AtB-AS/mittatb-app/pull/5168, and I think it makes sense for things that are basically `onSuccess` callbacks.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
strandlie: You are right. I was thinking of [this task](https://github.com/AtB-AS/kundevendt/issues/19835) when I added the `retry`, but I see now that according to that documentation it never had `retry` to begin with. 

But I agree that network errors _could_ be allowed to retry, but that should not be a part of this change. I can set it to `retry: false` now.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
-                searchInput,
-                cancelTokenSource,
-                filtersSelection,
-                journeySearchModes,
-              );
-
-              const tripPatternsWithKeys = decorateTripPatternWithKey(
-                results.trip.tripPatterns,
-              );
+  const fetchTrips = async ({pageParam}: {pageParam?: string}) => {
```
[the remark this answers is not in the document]
strandlie: Yes. And definitely in the future, since that is why I was making this change in the first place :)

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
+    try {
+      // Fire and forget add journey search entry if both locations are not from geo
+      if (
+        fromLocation.resultType !== 'geolocation' &&
+        toLocation.resultType !== 'geolocation'
+      ) {
+        (async () =>
+          await addJourneySearchEntryStable([fromLocation, toLocation]))();
+      }
+    } catch (e) {}
```
jorelosorio, line 158: This try catch should have any action or track the issue, or handle the issue properly.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
+  }, [status, tripPatterns, analytics, searchTime, filtersSelection]);
 
   return {
     tripPatterns,
     timeOfLastSearch: timeOfSearch,
-    loadMore: !!pageCursor ? loadMore : undefined,
-    clear: clearTrips,
-    searchState: searchState,
-    error: errorType,
+    loadMore: hasNextPage ? fetchNextPage : undefined,
```
jorelosorio, line 174: load more could be undefined at some point, I suggest keeping this function always available and internally call fetchNextPage if hasNextPage, otherwise return an empty array or no more results found.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
+    const query = createQuery(
+      fromLocation,
+      toLocation,
+      searchInput,
+      arriveBy,
+      filtersSelection,
+      journeySearchModes,
+    );
+
+    Bugsnag.leaveBreadcrumb('searching', {
```
jorelosorio, line 75: Maybe we could send a query directly here...

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-query.ts
```diff
@@ -29,199 +27,160 @@ export function useTripsQuery(
...
-              const tripPatternsWithKeys = decorateTripPatternWithKey(
-                results.trip.tripPatterns,
-              );
+  const fetchTrips = async ({pageParam}: {pageParam?: string}) => {
+    if (
+      !fromLocation ||
+      !toLocation ||
+      !isValidTripLocations(fromLocation, toLocation)
+    ) {
+      throw new Error('Invalid trip locations');
```
jorelosorio, line 53: Perhaps proper handle this issue

strandlie:
@jorelosorio @rosvik I guess we don't actually need this refactor any more when we get the new BFF-endpoint? I wanted this change to be able to used the cached result in the TripSelectionScreen, but we can cache the query to the new BFF endpoint instead now

PR #5192, opened 2025-05-19
feat: sort ticket history page by creation

rosvik:
Had a talk with @Mariblaas, and we agreed that it doesn't make sense to show reservations at the top of ticket history. We should rather always sort by creation date, so that the user doesn't have to scroll past lots of old cancelled payments before reaching their previous purchase.

## Acceptance criteria

- [ ] In ticket history, tickets and reservations are always sorted by date.
- [ ] Active tickets should be unaffected.


reidzeibel (approved):
Let's Get This Merged!

tormoseng:
✅

PR #5194, opened 2025-05-20
fix: add a11y message for expired/expiring cards in payment

adriansberg:
Accessibility message for expired or expiring cards were missing in the payment bottom sheet. 

The recurring payment's a11y label should now have the expiry message embedded in it.

fixes https://github.com/AtB-AS/kundevendt/issues/20096


HanneLH (approved):
🌸

reidzeibel (approved):

PR #5195, opened 2025-05-20
feat: use stream service to reload fare contracts

rosvik:
This implements the first iteration of using the [stream service](https://github.com/AtB-AS/stream/) to rely on websocket events to fetch fare contracts from the backend, outside of firestore. It's feature toggled, which means we'll be able to roll out gradually.

It works like this: 

1. On app launch, a websocket connection is created to `/stream/v1`. The thought is to keep this connection open as long as the app is running.
2. When the app receives a message on the format `{ event: "FARE_CONTRACT" }`, we tell the react query client to invalidate queries with queryKey `"FETCH_FARE_CONTRACTS"`, which will make fare contract data stale globally, and cause a reload wherever it's used.
4. In the ticketing context, we'll ignore fare contract snapshots from Firestore (as long as the feature toggle is enabled), and instead update it though `useGetFareContractsQuery`. The data will then propagate throughout the app, just like the firestore data did.

closes https://github.com/AtB-AS/kundevendt/issues/20460

---

This should make it possible to start testing the integration, but there are still some parts missing:

- Storage of fare contract data on device, so that if the user is without internet connection for a while, we should store the previous result.
- Some indication that the websocket connection is lost. Maybe time to revisit https://github.com/AtB-AS/kundevendt/issues/9427

### DIFF
--- a/src/modules/event-stream/use-setup-event-stream.ts
+++ b/src/modules/event-stream/use-setup-event-stream.ts
@@ -1,0 +1,48 @@
+import {useSubscription} from '@atb/api/use-subscription';
+import {useQueryClient} from '@tanstack/react-query';
+import {useCallback} from 'react';
+import {WS_API_BASE_URL} from '@env';
+import {getIdTokenGlobal, useAuthContext} from '../auth';
+import Bugsnag from '@bugsnag/react-native';
+import {handleStreamEvent} from './handle-stream-event';
+import {StreamEvent} from './types';
+import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
+
+export const useSetupEventStream = () => {
+  const queryClient = useQueryClient();
+  const {abtCustomerId} = useAuthContext();
+  const {isEventStreamEnabled} = useFeatureTogglesContext();
+
+  const url = `${WS_API_BASE_URL}stream/v1`;
+
+  const onMessage = useCallback(
+    (event: WebSocketMessageEvent) => {
+      const message = StreamEvent.safeParse(JSON.parse(event.data));
+      if (!message.success) {
+        Bugsnag.leaveBreadcrumb('Received unknown message from stream', {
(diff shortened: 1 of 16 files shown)


src/modules/event-stream/use-setup-event-stream.ts
```diff
@@ -0,0 +1,48 @@
...
+export const useSetupEventStream = () => {
+  const queryClient = useQueryClient();
+  const {abtCustomerId} = useAuthContext();
+  const {isEventStreamEnabled} = useFeatureTogglesContext();
+
+  const url = `${WS_API_BASE_URL}stream/v1`;
+
+  const onMessage = useCallback(
+    (event: WebSocketMessageEvent) => {
+      const message = StreamEvent.safeParse(JSON.parse(event.data));
```
jorelosorio, line 20: What if the JSON fails? We could have a try catch here

marius-at-atb (approved):
Looks great!

mjenssen (approved):
LGTM! 🚀

src/modules/event-stream/use-setup-event-stream.ts
rosvik: Good point! Added a util for this kind of thing here: 42fcba11c3a1174654f38f7666f32a0fd11321e2

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -89,6 +91,8 @@ export const RootStack = () => {
 
   useBeaconsContext();
   useTestIds();
+  useSetupReactQueryWindowFocus();
+  useSetupEventStream();
```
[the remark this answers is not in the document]
rosvik: Yes, that probably makes sense once we need to communicate between the app and stream in other ways than invalidating queries. For another time ✌️

PR #5196, opened 2025-05-21
feat: added forceEnd when there is no active booking

Dahly96:
Closes http://github.com/AtB-AS/kundevendt/issues/20598

We need to handle cases where entur force end the trip without a user action. So if entur ends the trip, they will set the booking to FINISHED so we skip the photo stage and just closes the scootersheet and cleans up the state


marius-at-atb (approved):
Nice!

PR #5197, opened 2025-05-22
feat: simplified the previous payment logic

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20604

Ended up reworking how we keep track of paymentcards in the shmo payment bottomsheet. Now we only rely on the last used card, and when we select cards in the bottomsheet we save the card to last used instead of setting it to state

### DIFF
--- a/src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
+++ b/src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import {View} from 'react-native';
 import {StyleSheet, useThemeContext} from '@atb/theme';
 import {Button} from '@atb/components/button';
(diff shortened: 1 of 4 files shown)


src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
```diff
@@ -79,12 +91,23 @@ export const SelectShmoPaymentMethodSheet = ({
...
+                  selectedPaymentMethod &&
+                  selectedPaymentMethod.recurringPayment &&
+                  userId
+                ) {
+                  setIsLoading(true);
+                  await savePreviousPaymentMethodByUser(userId, {
+                    paymentType: selectedPaymentMethod?.paymentType,
+                    recurringPayment: selectedPaymentMethod.recurringPayment,
+                  });
+                  setIsLoading(false);
```
marius-at-atb, line 105: Just a note that react-query is suitable not just for api calls, but any async logic you want to keep state for.
While this is ok as is, I wouldn't mind using that here as well :)

marius-at-atb (approved):
Looks good!

strandlie (approved):

PR #5201, opened 2025-05-23
fix: fixed issue with live map

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20788

fixed issue where the map is all the way zoomed out sometimes when opening live bus


marius-at-atb (approved):
Lgtm!

PR #5202, opened 2025-05-23
feat: Add Slot Layers

marius-at-atb:
Closes https://github.com/AtB-AS/kundevendt/issues/20702

There is a problem in mapbox that `layerIndex` doesn't work.
However, using `aboveLayerId` works well.
In this PR, new placeholder layers are added so that we have full control over the layer order.
Current order, from bottom to top:
```ts
const slotLayerIds: MapSlotLayerId[] = [
  MapSlotLayerId.GeofencingZones, // bottom
  MapSlotLayerId.Vehicles,
  MapSlotLayerId.Stations,
  MapSlotLayerId.NSRItems,
  MapSlotLayerId.SelectedFeature, // top
];
```
How to place a `SymbolLayer`:
```ts
aboveLayerID={MapSlotLayerId.Stations}
```

### DIFF
--- a/src/modules/map/hooks/use-mapbox-json-style.tsx
+++ b/src/modules/map/hooks/use-mapbox-json-style.tsx
@@ -6,7 +6,7 @@ import {useFirestoreConfigurationContext} from '@atb/modules/configuration';
 import {getTextForLanguage, useTranslation} from '@atb/translations';
 import {useVehiclesAndStationsVectorSource} from '../components/mobility/VehiclesAndStations';
 
-// since zIndex doesn't work in mapbox, but aboveLayerId does, add some slot layer ids to use
+// since layerIndex doesn't work in mapbox, but aboveLayerId does, add some slot layer ids to use
 export enum MapSlotLayerId {
   GeofencingZones = 'geofencingZones',
   Vehicles = 'vehicles',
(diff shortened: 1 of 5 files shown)


src/modules/map/hooks/use-mapbox-json-style.tsx
```diff
@@ -6,6 +6,35 @@ import {useFirestoreConfigurationContext} from '@atb/modules/configuration';
...
+  Vehicles = 'vehicles',
+  Stations = 'stations',
+  NSRItems = 'nsrItems',
+  SelectedFeature = 'selectedFeature',
+}
+
+const slotSourceKey = 'slotSource';
+// This source only exists for slots layers, no data is fetched.
+const slotSource: StyleJsonVectorSourcesObj = {
+  [slotSourceKey]: {type: 'vector'}, // type is required, but otherwise doesn't matter here.
```
marius-at-atb, line 21:
```suggestion
  [slotSourceKey]: {type: 'shape'}, // type is required, but otherwise doesn't matter here.
```
@tormoseng after merging this, let's make sure there are no network calls to "nowhere" because of this. If so, we should make this change.

Dahly96 (approved):

PR #5210, opened 2025-05-26
fix: Revert "Update @react-native-async-storage+async-storage+1.19.8" and apply patch properly

jorelosorio:
There was some extra information that had to be changed as well, like the index IDs and more. This version reverts my previous change and makes a fix.


marius-at-atb (approved):
Works for me after `yarn clean:install` :+1:

PR #5211, opened 2025-05-26
fix: fixed posthog in bottomsheets

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20568

So i created a common client to use in both the analytics context and the bottomsheet context, but only initiated it one place, to make sure we dont do redundant calls and also keep history clean for users.


rosvik (approved):
sweet

PR #5212, opened 2025-05-27
feat: follow user location when in active trip

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/20800

When a user is in a active trip the map will follow the user. If a user interacts with the map, it will be disabled and will be actived again if the user presses the location arrow while in an active trip.

If the user restarts the app while in an active trip, the user will have to reactive the follow user feature with the location arrow again

### DIFF
--- a/src/modules/map/MapV2.tsx
+++ b/src/modules/map/MapV2.tsx
@@ -3,7 +3,13 @@ import {
   useGeolocationContext,
 } from '@atb/modules/geolocation';
 import {FOCUS_ORIGIN} from '@atb/api/geocoder';
-import MapboxGL, {LocationPuck} from '@rnmapbox/maps';
+import {
+  LocationPuck,
+  UserTrackingMode,
+  Viewport,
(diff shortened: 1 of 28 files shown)


marius-at-atb:
Very nice!
I wonder if it is possible to simplify it even further by using `activeBooking` in `MapV2`, and not needing `followUser, setFollowUser` in the context 🤔

src/modules/map/MapV2.tsx
```diff
@@ -42,7 +48,7 @@ import {
   isVehiclesClusteredFeature,
 } from '@atb/modules/mobility';
 
-import {Snackbar, useSnackbar} from '@atb/components/snackbar';
+import {Snackbar, useSnackbar, useStableValue} from '@atb/components/snackbar';
```
marius-at-atb, line 51: Btw I just noticed - we should move `useStableValue` out of the `snackbar` folder to be a general util.

PR #5215, opened 2025-05-28
chore: Add Håkons phone as test device in NFK and Troms

strandlie:


reidzeibel:
Don't forget to run the script!

PR #5218, opened 2025-05-28
fix: Always refetch shmo items on mount

marius-at-atb:
E.g. when scanning a vehicle very shortly after someone else has parked it, the GBFS data may not be updated yet.
If so, the user will get an error, and because of caching, this would continue to show as a missing vehicle even after the data has been updated.
To fix this, vehicles and stations are always refetched with this PR.

Acceptance Criteria:
Every time a bottom sheet is opened, a new request should be sent to get the data
   - [ ] Scooters
      - [ ] From clicking on map
      - [ ] From scanning (both starting from selected, and non-selected bottom sheet state)
   - [ ] Bicycle Stations
      - [ ] From clicking on map
   - [ ] Car Sharing Stations
      - [ ] From clicking on map
- [ ] Requests should not be fired tooooo many times/at other times

### DIFF
--- a/src/modules/mobility/queries/use-bike-station-query.tsx
+++ b/src/modules/mobility/queries/use-bike-station-query.tsx
@@ -8,4 +8,5 @@ export const useBikeStationQuery = (id: string) =>
     queryFn: ({signal}) => getBikeStation(id, {signal}),
     staleTime: ONE_MINUTE_MS,
     cacheTime: ONE_MINUTE_MS,
+    refetchOnMount: 'always',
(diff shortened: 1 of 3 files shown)


strandlie (approved):

marius-at-atb:
@strandlie It will just return an empty list of vehicles instead of error, but I don't think it hurts to add retries. E.g. if the internet connection is unstable, maybe it will retry more quickly.

tormoseng:
✅

PR #5220, opened 2025-05-30
feat: Bump utils and use TicketOffer from utils

strandlie:
Removed local Offer-type, and used the common TicketOffer from utils

### Acceptance Criteria
- [ ] Everything involving offers works as before

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@atb-as/config-specs": "5.16.0",
     "@atb-as/generate-assets": "17.2.0",
     "@atb-as/theme": "^13.1.5",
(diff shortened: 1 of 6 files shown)


rosvik (approved):
Nice ✨ Not sure if it even makes sense to add zod parsing in this case, but that would be a natural next step.

HanneLH:
@strandlie 
Ikke sikker om dette er knyttet til endringene i denne oppgaven, men vist velger en annen sone enn A for 24 timers billetten så ser det ut som man kan reise mellom to soner nå! 

IMG_2708801FFB66-1
IMG_2708801FFB66-2

PR #5222, opened 2025-06-02
Feat: add car sharing bonus checkbox

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/20668

adds bonus checkbox to car sharing bottom sheet

only visible if car sharing bonus product exists in firestore


strandlie (approved):

strandlie:
I find it a little weird that the checkbox is not centered in the Pay with bonus points-box, but that might be a personal preference :)

PR #5229, opened 2025-06-03
feat: new mobile token endpoint error handling

reidzeibel:
closes https://github.com/AtB-AS/kundevendt/issues/20605

This adds implementation of error handling for the new token server endpoint.

Also in this PR:
- More inline documentation for various mobile-token related functions & hooks
- Removes redundant error checking
- use `nullish` instead of `optional` for the `ErrorResponse`


rosvik (approved):
LGTM 👍

PR #5232, opened 2025-06-04
fix: allow bus in map to overlap location puck

rosvik:
ref. https://github.com/AtB-AS/kundevendt/issues/20799
> Also relevant: look into avoiding that the icon gets hidden when the live icon and user location marker get too close.

### Acceptance criteria

- [x] Bus in map icon isn't hidden when it gets close to the user puck


tormoseng:
✅

PR #5233, opened 2025-06-04
feat: add warnings from the app and geofencing zones in bottomsheet

Dahly96:


src/modules/map/hooks/use-shmo-warnings.tsx
```diff
@@ -0,0 +1,84 @@
...
+  useEffect(() => {
+    if (vehicle?.isDisabled) {
+      setWarningMessage(t(ShmoWarnings.scooterDisabled));
+    } else if (warningMessage !== null) {
+      setWarningMessage(null);
+    }
+  }, [locationIsAvailable, t, vehicle?.isDisabled, warningMessage]);
+
+  const throttledCallback = useRef(
+    throttle(async (loc: {latitude: number; longitude: number}) => {
```
marius-at-atb, line 42: For consistency, let's use `coordinates: Coordinates` here

src/modules/map/utils.ts
```diff
@@ -128,27 +136,43 @@ export const mapPositionToCoordinates = (p: Position): Coordinates => ({
   latitude: p[1],
 });
 
-export const getFeaturesAtClick = async (
-  clickedFeature: Feature<Point>,
+export const getFeaturesAtPoint = async (
+  point: [number, number],
```
marius-at-atb, line 140: Not great by mapbox, they have a function called `getPointInView` which accepts `coordinate: Position` as point input 😅 
So naming won't be perfect here, but let's at least use the same type. (`Position` rather than `[number, number]`)

src/modules/map/utils.ts
```diff
@@ -128,27 +136,43 @@ export const mapPositionToCoordinates = (p: Position): Coordinates => ({
...
+ * Gets features at a clicked location
+ */
+export const getFeaturesAtClick = async (
+  clickedFeature: Feature<Point>,
+  mapViewRef: RefObject<MapboxGL.MapView | null>,
+  filter?: Expression,
+  layerIds?: string[],
+) => {
+  const coords = mapPositionToCoordinates(clickedFeature.geometry.coordinates);
+  return getFeaturesAtPoint(
```
marius-at-atb, line 168: Nice!

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -324,19 +324,19 @@ export const ParkAndRideTexts = {
...
   },
   slow: {
     title: _('Saktesone', 'Slow Zone', 'Saktesone'),
     description: _(
-      'Farten blir redusert i dette området',
-      'The speed will be reduced in this area',
-      'Farta vert redusert i dette området',
+      'Farten blir redusert her',
+      'The speed will be reduced here',
+      'Farten blir redusert her',
```
marius-at-atb, line 339:
```suggestion
      'The speed will be reduced here',
      'Farta vert redusert her',
```

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -324,19 +324,19 @@ export const ParkAndRideTexts = {
...
   allowed: {
-    title: _('Tillatt område', 'Allowed Area', 'Tillatt område'),
+    title: _('Parkeringssone', 'Parking Zone', 'Parkeringssone'),
     description: _(
-      'Her kan du både kjøre og parkere',
-      'Here you can both drive and park',
-      'Her kan du både køyre og parkere',
+      'Du kan parker her',
+      'You can park here',
+      'Du kan parkere her',
```
marius-at-atb, line 331: Hmm I don't think we can change it like this. Let's discuss it on Tuesday.

src/modules/map/hooks/use-shmo-warnings.tsx
```diff
@@ -0,0 +1,84 @@
...
+        setGeofencingZoneMessage(null);
+      }
+    }, 1000),
+  ).current;
+
+  useEffect(() => {
+    if (coordinates) {
+      throttledCallback(coordinates);
+    }
+  }, [coordinates, throttledCallback]);
```
marius-at-atb, line 81: I just realised, the GeofencingZones depend on `vehicleId`, so it should also be a dependency that triggers a re-run.

src/modules/map/hooks/use-shmo-warnings.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const {locationIsAvailable} = useGeolocationContext();
+  const {vehicle} = useVehicle(vehicleId);
+
+  useEffect(() => {
+    if (vehicle?.isDisabled) {
+      setWarningMessage(t(ShmoWarnings.scooterDisabled));
+    } else if (warningMessage !== null) {
+      setWarningMessage(null);
+    }
+  }, [locationIsAvailable, t, vehicle?.isDisabled, warningMessage]);
```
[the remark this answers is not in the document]
Dahly96: yes, at first it was a prioritized order of multiple messages. But in the first version we decided on just the one. So ill change it

src/modules/map/hooks/use-shmo-warnings.tsx
```diff
@@ -0,0 +1,75 @@
...
+    : null;
+
+  const throttledCallback = useRef(
+    throttle(async (coordinates: Coordinates) => {
+      if (!mapViewRef?.current) return null;
+
+      const featuresAtLocation = await getFeaturesAtPoint(
+        [coordinates.longitude, coordinates.latitude],
+        mapViewRef,
+      );
```
marius-at-atb, line 39: Hmm, if the location is the same as where a point feature is, such as e.g. stopPlaces, getFeatureToSelect will select that one. We should add a filter to the features first.

marius-at-atb: Btw while filters with mapbox expressions are supported by `getFeaturesAtPoint`, I think in this case it is easier to just reuse the existing `isFeatureGeofencingZone` js function
```ts
const geofenzingZoneFeatures = featuresAtLocation.filter(f => isFeatureGeofencingZone(f)) 
```

src/modules/map/hooks/use-shmo-warnings.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+      if (!featuresAtLocation || featuresAtLocation.length === 0) return;
+
+      const featureToSelect = getFeatureToSelect(featuresAtLocation, [
+        coordinates.longitude,
+        coordinates.latitude,
+      ]);
+
+      if (isFeatureGeofencingZone(featureToSelect)) {
```
marius-at-atb, line 48: This check should not be necessary after https://github.com/AtB-AS/mittatb-app/pull/5233/files#r2137814983

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -324,7 +324,7 @@ export const ParkAndRideTexts = {
 
 export const GeofencingZoneExplanations: GeofencingZoneExplanationsType = {
   allowed: {
-    title: _('Tillatt område', 'Allowed Area', 'Tillatt område'),
+    title: _('Parkeringssone', 'Parking Zone', 'Parkeringssone'),
```
marius-at-atb, line 327: Let's revert this as well for now

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -364,19 +364,15 @@ export const GeofencingZoneExplanations: GeofencingZoneExplanationsType = {
...
+    title: _('Parkering forbudt', 'No Parking Zone', 'Parkering forboden'),
     description: _(
       'Du kan ikke parkere i dette området',
       'You cannot park in this area',
       'Du kan ikkje parkere i dette området',
     ),
   },
   noEntry: {
-    title: _('Kjøring forbudt-sone', 'No Riding Zone', 'Kjøring forboden-sone'),
+    title: _('Kjøring forbudt', 'No Riding', 'Kjøring forboden'),
```
marius-at-atb, line 375:
```suggestion
    title: _('Kjøring forbudt', 'No Riding', 'Køyring forboden'),
```

PR #5235, opened 2025-06-04
Feat: bonus points earned visible on ticket

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/19985

### Adds Bonus points earned visible on bought tickets.
* It is visible on active tickets and under details on all tickets.
* The query will retry 3 times with 500 ms delay if bonus service says 0 points earned for a ticket, to handle situation where eventhandler is delayed to call bonus service.

NB! illustrations are not added yet, part of a separate issue:
- https://github.com/AtB-AS/kundevendt/issues/20669

### DIFF
--- a/src/components/sections/items/ExpandableSectionItem.tsx
+++ b/src/components/sections/items/ExpandableSectionItem.tsx
@@ -1,6 +1,6 @@
 import React, {useEffect, useState} from 'react';
 import {AccessibilityProps, View} from 'react-native';
-import {StyleSheet, Theme} from '@atb/theme';
+import {StyleSheet, Theme, useThemeContext} from '@atb/theme';
 import {SectionTexts, useTranslation} from '@atb/translations';
 import {ThemeText} from '@atb/components/text';
 import {NavigationIcon} from '@atb/components/theme-icon';
@@ -10,8 +10,10 @@ import {useSectionStyle} from '../use-section-style';
 import {animateNextChange} from '@atb/utils/animation';
 import {TextNames} from '@atb/theme/colors';
 import {PressableOpacity} from '@atb/components/pressable-opacity';
+import {LabelType} from '@atb-as/config-specs';
 import {Tag} from '@atb/components/tag';
 import {TagInfoTexts} from '@atb/translations/components/TagInfo';
+import {insets} from '@atb/utils/insets';
 
 type Props = SectionItemProps<
   {
@@ -19,6 +21,7 @@ type Props = SectionItemProps<
     prefixNode?: React.ReactNode;
     suffixNode?: React.ReactNode;
     textType?: TextNames;
+    label?: LabelType;
     showIconText?: boolean;
     testID?: string;
     accessibility?: AccessibilityProps;
@@ -46,12 +49,14 @@ export function ExpandableSectionItem({
   suffixNode,
   textType,
   showIconText = false,
+  label,
   accessibility,
   testID,
   ...props
 }: Props) {
   const {contentContainer, topContainer} = useSectionItem(props);
   const sectionStyle = useSectionStyle();
+  const {theme} = useThemeContext();
   const styles = useStyles();
   const {t} = useTranslation();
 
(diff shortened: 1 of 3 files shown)


src/modules/bonus/queries/use-bonus-amount-earned-query.tsx
```diff
@@ -0,0 +1,51 @@
...
+    ),
+    queryFn: () =>
+      getBonusAmountEarned(
+        isLoggedIn,
+        isBonusProgramEnabled,
+        fareContractId,
+        disabled,
+      ),
+    retry: 3,
+    retryDelay: 500,
```
strandlie, line 48: Why use a fixed retry, and not the exponential backoff that is default to give the service a bit more time to get into the right state?

jorunnl: Good point, thanks.
We should set a limited number of retries since 0 is a valid response (but can be wrong), but there is no need to have a specific retryDelay here, will remove

src/modules/bonus/api/api.ts
```diff
@@ -27,3 +27,32 @@ export const buyValueCodeWithBonusPoints = (
...
+  return client
+    .get(`/bonus/v1/fare-contract/${fareContractId}/amount`, {
+      authWithIdToken: true,
+    })
+    .then((response) => {
+      const value =
+        response.data.amount === null ? 0 : Number(response.data.amount);
+      if (value == 0) {
+        throw new Error();
+      }
```
marius-at-atb, line 55: 0 points doesn't necessarily mean an error. E.g. escooter trip fareContracts don't yield bonus points.

src/modules/fare-contracts/details/DetailsContent.tsx
```diff
@@ -126,6 +130,8 @@ export const DetailsContent: React.FC<Props> = ({
     accesses &&
     accesses.maximumNumberOfAccesses <= MAX_ACCESSES_FOR_CARNET_FOOTER;
 
+  const {data: earnedBonusPoints} = useBonusAmountEarnedQuery(fc.id, false);
```
marius-at-atb, line 133: It might be nice to make `disabled` optional with default `false` :)

src/modules/bonus/api/api.ts
```diff
@@ -27,3 +27,32 @@ export const buyValueCodeWithBonusPoints = (
...
+  disabled: boolean,
+): Promise<number> => {
+  if (
+    !isLoggedIn ||
+    !isBonusProgramEnabled ||
+    fareContractId === undefined ||
+    disabled
+  ) {
+    return Promise.resolve(0);
+  }
```
marius-at-atb, line 44: The entries in the query key should be linked to the dependencies in the API call, in this case only `fareContractId`.
So all these requirements belong as part of the react query `enabled` field instead.

marius-at-atb (changes requested):
Nice work!
Just need to have another look at the query implementation.

src/modules/bonus/api/api.ts
marius-at-atb: Then you would simply invalidate the query based on the fareContract id given in the event

marius-at-atb: @mjenssen @rosvik @strandlie
We could make a new event with simply EventKind=BONUS_AMOUNT_EARNED. However it may also be a case for expanding StreamEvent to include an extra field, which could contain the specific id, so that we don't have to invalidate all the queries.

marius-at-atb: Great! We can use that field then. @rosvik I guess we should update the `StreamEvent` type in the app.

marius-at-atb (approved):
Approving as is, with the "query hack" to be fixed in a new issue :)

src/modules/bonus/api/api.ts
jorunnl: Added separate task to use stream svc: https://github.com/AtB-AS/kundevendt/issues/20930

PR #5236, opened 2025-06-05
fix: use tabs, not spaces, in ios-devices.txt

rosvik:
follow up to https://github.com/AtB-AS/mittatb-app/pull/5231

Used spaces instead of tabs, so now the action fails 🙈 https://github.com/AtB-AS/mittatb-app/actions/runs/15460539805/job/43520769475


reidzeibel (approved):
👾

PR #5238, opened 2025-06-05
fix: handle type checking more thoroughly

reidzeibel:
ref slack discussion https://mittatb.slack.com/archives/C0116FMPX4Y/p1749116516725149

This does the checking more gracefully and thoroughly, should no longer crash on wrong type.


rosvik:
I think we should try to tackle the issue at the source, just after we receive the snapshot from firestore. If the issue is that fare contracts doesn't have any travel rights, there are 31 places in the app we use `travelRights[0]`, so I suspect fixing isOfFareProductRef will just make something else fail.

PR #5239, opened 2025-06-05
fix: Workaround for Fastlane

jorelosorio:
This is a workaround fix for the issue that occurs during device updates.

Related to this entry: https://github.com/AtB-AS/mittatb-app/actions/runs/15465542508/job/43536658822

It uses: https://github.com/fastlane/fastlane/pull/29580/files

And builds seems to work:

https://github.com/AtB-AS/mittatb-app/actions/runs/15466164121


rosvik:
Good work finding a fix 💯 Is there any downsides to waiting for an fixed fastlane version instead of relying on the fork? If only the ios-devices.txt update is affected, I guess it's not a big problem to wait a bit.

jorelosorio:
> Good work finding a fix 💯 Is there any downsides to waiting for an fixed fastlane version instead of relying on the fork? If only the ios-devices.txt update is affected, I guess it's not a big problem to wait a bit.

Agree, I do not like that fix but at least the action was ran, so no need to merge this and wait for a proper fix or when that PR is merged into fastlane.

rosvik:
@jorelosorio New fastlane release with the fix 🚀 https://github.com/fastlane/fastlane/releases/tag/2.228.0

jorelosorio:
Updated now

strandlie (approved):

PR #5242, opened 2025-06-05
feat: use Tag component instead of the ListItem

Oleninkaa:
Closes https://github.com/AtB-AS/kundevendt/issues/20454 

Now we are using the `Tag` Component in the app. I also replaced `ListItem` since it is being used in a few places. and has the same purpose as Tag.  Additionally, I added it to the design system to display all of the states.

Skjermbilde 2025-06-05 kl  16 07 19


src/components/tag/Tag.tsx
```diff
@@ -0,0 +1,115 @@
...
+      case 'primary':
+        return theme.color.status.info.primary.background;
+      case 'secondary':
+        return theme.color.background.neutral[0].background;
+      default:
+        return theme.color.status[tagType].secondary.background;
+    }
+  };
+
+  const iconComponent = useMemo(() => {
```
strandlie, line 46: I think we can be a little clever with the typing here, in a way that means we don't need this memoized function. 

What do you think about something like this? This means the component only accepts an icon if the type is `secondary`. 

```
type BaseTagProps = {
  label: string[];
  size?: 'small' | 'regular';
  customStyle?: StyleProp<ViewStyle>;
};

type SecondaryTagProps = BaseTagProps & {
  tagType: 'secondary';
  icon: (props: SvgProps) => JSX.Element;
};

type OtherTagProps = BaseTagProps & {
  tagType: Exclude<TagStatuses, 'secondary'>;
  icon?: never;
};

export type TagProps = SecondaryTagProps | OtherTagProps;
```

If we do this, it means we can do 

`<ThemeIcon svg={icon ?? statusTypeToIcon(tagType, true, themeName)} size={iconSize} />`

strandlie (approved):

PR #5243, opened 2025-06-06
fix: bonus product design and checkbox text and functionality

jorunnl:
closes: https://github.com/AtB-AS/kundevendt/issues/20886

Updates design and fixes related to bonus. Specifically:

On the profile page:
1. Put bonus products in same section (instead of separate sections)

For the checkbox on Bonus products in bottom sheets:
1. fix so that the checkbox is always unchecked when it is disabled
2. Change from 'Du har 42 ★' to 'Du har 42 poeng'

### DIFF
--- a/src/modules/bonus/components/PayWithBonusPointsCheckbox.tsx
+++ b/src/modules/bonus/components/PayWithBonusPointsCheckbox.tsx
@@ -17,17 +17,18 @@ import {View} from 'react-native';
 import {MessageInfoBox} from '@atb/components/message-info-box';
 import {UserBonusBalance} from './UserBonusBalance';
 import {isDefined} from '@atb/utils/presence';
+import {useEffect} from 'react';
 
 type Props = SectionProps & {
   bonusProduct: BonusProductType;
   isChecked: boolean;
-  onPress: () => void;
+  setIsChecked: (checked: boolean) => void;
 };
 
 export const PayWithBonusPointsCheckbox = ({
   bonusProduct,
   isChecked,
-  onPress,
+  setIsChecked,
   ...props
 }: Props) => {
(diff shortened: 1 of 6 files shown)


src/modules/bonus/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -44,6 +45,12 @@ export const PayWithBonusPointsCheckbox = ({
 
   const isDisabled = isError || userBonusBalance < bonusProduct.price.amount;
 
+  useEffect(() => {
```
strandlie, line 48: I'm not sure I understand this, and when I see an `useEffect` I don't understand my alarm bells go off...

Is this really an effect? I.e. is this state that changes based on some external system, like an async API or file system? If this is only internal state, I think you could solve this without `useEffect` by calculating `isChecked` on every render.

jorunnl: This is an actual useEffect triggerd by API-call to get userBonusBalance when a user does a purchase, so that the checkbox can not be checked when user does not have sufficient bonus points to buy the bonus product

src/modules/bonus/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -62,7 +69,7 @@ export const PayWithBonusPointsCheckbox = ({
       <Section {...props}>
         <GenericClickableSectionItem
           active={isChecked}
-          onPress={onPress}
+          onPress={() => setIsChecked(!isChecked)}
```
marius-at-atb, line 72: Remember best practice is
```ts
setIsChecked(isChecked => !isChecked)
```

PR #5250, opened 2025-06-10
feat: persist fare contract query in AsyncStorage

rosvik:
Uses a [`PersistQueryClientProvider`](https://tanstack.com/query/latest/docs/framework/react/plugins/persistQueryClient) in ReactQueryProvider, which supports storing of query data on the device, so that it's available without an internet connection on app launch. It's made opt-in, since there are some gotchas with non-serializable data that makes it break for cases like the mobile token `ActivatedToken` class.

I've enabled it just for fare contracts for now, which needed some improvements to how we handle Date objects (See https://github.com/AtB-AS/utils/pull/23)

closes https://github.com/AtB-AS/kundevendt/issues/20780


marius-at-atb (approved):
Looks great!

PR #5261, opened 2025-06-17
fix: update Clock Skew error handling

reidzeibel:
resolves https://mittatb.slack.com/archives/C03DTC05VRQ/p1749126042005239 

Fix the TokenClockSkewError so that the SDK properly recalibrates and can continue with the token process. 🎉 

Also updated the SDK version to 3.3.14-hotfix1, to not get the other updates before Entur tested it.

### Acceptance Criteria:
- [x] Clock Skew error doesn't happen anymore, try to move back the device time by 1 hour, expire the token in backoffice, and launch the app.


reidzeibel:
@jorelosorio @rosvik is updating the podfile look ok here?

reidzeibel:
Tested on iOS Simulator and Android Emulator, both works fine using version 3.3.14-hotfix1

tormoseng:
Seems good, but will re-test after https://github.com/AtB-AS/mittatb-app/pull/5271

tormoseng:
✅  I think we can say it works. A small clock skew is handled OK. But with an hour clock skew it seems that no /list request is sent. A test of 1.68 shows that this isn't a new problem.

PR #5267, opened 2025-06-18
refactor: Moved booking into its own module

strandlie:
Related to AtB-AS/kundevendt#20761

This PR does two things: 
- Refactor the booking logic into its own module, since it is used several places
- Separate what time window is queried and what trip patterns are displayed. This is different, since the last departure might have departed today, but we still don't want users to buy a regular ticket.


rosvik (approved):

HanneLH:
Tested ok 🎉

PR #5273, opened 2025-06-19
feat: add breadcrumbs for logging ID token on API error

reidzeibel:
Based on our discussion about the ID token logging

This PR will add the breadcrumb for logging ID token on API error, and add metadata with the same content for easier viewing on some errors.

- Added getter for global ID token expiration time, similar to how we get the global ID token customer number
- Changed the function signature for `secondsToTokenExpiry`, now taking `string` instead of `IdTokenResult`, and updates existing usage.
- Added function to convert that expiration time to string: 
  - `valid` means that the ID token expiration time is **more than** 300 seconds (5 minutes)
  - `expiring` means that the ID token expiration time is **between** 0 and 300 seconds
  - `expired` means that the ID token expiration time is **less than** 0 seconds
- Add breadcrumbs and metadata on API error handling inside `client.ts` file, with the `ID token` and its status (can be `valid`, `expiring`, or `expired) using the above function.

Bugsnag error will most likely look like this:
image

### Acceptance Criteria
- [x] API errors will have breadcrumbs and/or metadata containing ID Token

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -16,9 +16,12 @@ import {
 } from './headers';
 import axiosBetterStacktrace from 'axios-better-stacktrace';
 import {Platform} from 'react-native';
-import {getCurrentUserIdGlobal, getIdTokenGlobal} from '@atb/modules/auth';
-import {getIdTokenValidityStatus} from '@atb/modules/auth/utils';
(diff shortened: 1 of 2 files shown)


rosvik (approved):
LGTM 👍

tormoseng:
✅ 

Screenshot 2025-06-23 at 10 29 13

PR #5276, opened 2025-06-20
feat: add publicCode to global message contexts

rosvik:
Adds `publicCode` / `publicCodes` as strings to departure details, trip details and trip results, so that we can add global messages for specific line numbers.

Also adds `publicCodeAsNumber` to departure details, so line numbers in a range can be targeted, which can be useful for cases like 100-999 for regional buses. This is harder to do with the rule engine when there is more than one public code.

### Acceptance criteria

- [ ] Able to use `publicCodes` as a rule variable on trip results and trip details. (with e.g. `contains`)
- [ ] Able to use `publicCode` as a rule variable on departure details. (with e.g. `equals`)


rosvik:
@Sebstorvik I don't think TPW has support for rules at all 😅

reidzeibel (approved):

reidzeibel:
Maybe update the Global message documentation as well on https://github.com/AtB-AS/docs-private/blob/main/global-messages.md

Sebstorvik:
Aaah, right. Mixed it with context and scheduler. I'll create an issue for TPW rules that we can look into later if we see the need for it 😊

PR #5277, opened 2025-06-20
fix: Improve display of rejected payments

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/20510

Screenshot_1750410985


rosvik (commented):
Could we have a util for this similar to how `sortFcOrReservationByValidityAndCreation` was before? And update/replace the tests. `sortFcOrReservationByValidityAndCreation` isn't used for anything else, so it can be replaced.

strandlie (approved):

PR #5287, opened 2025-06-24
feat(spar): add edit vehicle registration screen

adriansberg:
Introduce a new screen for editing vehicle registrations, along with the necessary API mutation and query hooks. Update navigation to include the new edit screen and modify existing components to support the editing functionality.

fixes https://github.com/AtB-AS/kundevendt/issues/21079


rosvik (approved):
Nice 🙌

src/modules/smart-park-and-ride/types.ts
```diff
@@ -1,3 +1,4 @@
 export type VehicleRegistration = {
+  id: string;
   licensePlate: string;
 };
```
marius-at-atb, line 4: Remember to follow the same practice as for `SendSupportRequestBodySchema` - using zod schema and validating form inputs with that.

PR #5292, opened 2025-06-27
fix: always use backoff for websocket reconnects

rosvik:
Fixes an issue where the event stream connection was aggressively trying to reconnect when the connection was closed with code 1001. See https://mittatb.slack.com/archives/C02DL21RK7C/p1751010497949739. Now we always use `retryWithCappedBackoff`, regardless of error code.

## Acceptance criteria

- [x] When a websocket is closed with event code 1001 (or any other code), the app retries with exponential backoff, capped at 10s. With this, I expect we'll see a _lot_ less websocket errors in prod.


marius-at-atb (approved):

tormoseng:
✅

PR #5301, opened 2025-07-03
fix:scroll with keyboard on 2x view enrollment screen

jorunnl:
closes: https://github.com/AtB-AS/kundevendt/issues/21088#issuecomment-3018587810

Adds KeyboardAvoidingView to Enrollment screen so user can see input field with keyboard even with 2x screen.

Before: with 2x zoom user can not scroll down to see input field when keyboard is present
After: user can scroll all the way down with keyboard present (also I'm listening to Åge Aleksandersen in this screenshot, but ignore that difference 🙈)


marius-at-atb (approved):
🎶

PR #5302, opened 2025-07-04
chore: Rewrite AppDelegate to Swift

jorelosorio:
Closes https://github.com/AtB-AS/kundevendt/issues/21156


rosvik:
Seems to work well after the fix 💯 Started some staging builds to see if that still works 

https://github.com/AtB-AS/mittatb-app/actions/runs/16069680661
https://github.com/AtB-AS/mittatb-app/actions/runs/16069683707

PR #5307, opened 2025-07-07
feat(spar): update layout with nickname

adriansberg:
Introduce an optional nickname field for vehicle registrations, enhancing user experience by allowing users to assign names to their vehicles. Update related components and hooks to accommodate this new feature.

fixes https://github.com/AtB-AS/kundevendt/issues/21081


src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
```diff
@@ -38,32 +38,50 @@ export const Root_SmartParkAndRideAddScreen = ({navigation}: Props) => {
...
+            )}
+            onChangeText={setNickname}
+            value={nickname}
+            inlineLabel={false}
+            autoFocus={true}
+          />
+          <TextInputSectionItem
+            label={t(SmartParkAndRideTexts.add.inputs.licensePlate.label)}
+            placeholder={t(
+              SmartParkAndRideTexts.add.inputs.licensePlate.placeholder,
```
marius-at-atb, line 79: These should have a maxLength just for health and safety

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideEditScreen.tsx
```diff
@@ -58,6 +62,16 @@ export const Root_SmartParkAndRideEditScreen = ({
       }
     >
       <Section style={styles.content}>
+        <TextInputSectionItem
```
marius-at-atb, line 65: Let's add a `maxLength` here as well

marius-at-atb (approved):
Lgtm

src/modules/smart-park-and-ride/components/LicensePlateInputSection.tsx
```diff
@@ -12,24 +12,19 @@ import {ThemeText} from '@atb/components/text';
 import {StyleSheet} from '@atb/theme';
 
 type LicensePlateInputSectionItemProps = {
```
marius-at-atb, line 14: Should be renamed to `LicensePlateInputSectionProps`, and probably also be structured as
```ts
{
   textInputSectionItemProps: {label?: string} & Omit<React.ComponentProps<typeof TextInputSectionItem>, 'label'>
}
```

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
```diff
@@ -22,6 +22,9 @@ export const Root_SmartParkAndRideAddScreen = ({navigation}: Props) => {
   const [nickname, setNickname] = useState('');
   const [licensePlate, setLicensePlate] = useState('');
   const {theme} = useThemeContext();
+  const insets = useSafeAreaInsets();
+
+  const keyboardVerticalOffset = insets.top + 44; // 44 is the height of the header
```
marius-at-atb, line 27: Rather than a hard coded value, is it possible to calculate in e.g. `useHeaderLayouts`?
Otherwise maybe just use
`const headerHeight = 44;`
with no need for a comment..

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
```diff
@@ -46,14 +49,17 @@ export const Root_SmartParkAndRideAddScreen = ({navigation}: Props) => {
...
-        </FullScreenFooter>
+        </View>
       }
     >
-      <View style={styles.container}>
+      <KeyboardAvoidingView
+        contentContainerStyle={styles.container}
+        behavior="position"
+        keyboardVerticalOffset={keyboardVerticalOffset}
+      >
```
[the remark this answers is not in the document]
adriansberg: The inputs doesn't go down that far on that screen, but can add it there as well.

src/modules/smart-park-and-ride/components/LicensePlateInputSection.tsx
adriansberg: Alternatively we can change the component to be isolated as a section item, and wrap Section outside the component, that way it will be more flexible and the namings etc. will fit better.

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
adriansberg: Not sure how we should access that value if it is calculated in `useHeaderLayouts`?

Could go with something like passing an e.g. `onLayout` to the `headerProps` which is passed on to `ScreenHeader`, and that sets a state. But not too keen on doing that, sounds like much work for little gain.

C.f. Slack thread: https://mittatb.slack.com/archives/C02EEG7D8EL/p1752052176541499?thread_ts=1752051010.697549&cid=C02EEG7D8EL

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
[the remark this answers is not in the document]
adriansberg: I understand what you are thinking, I can look into it.

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
marius-at-atb: Hmm yea not looking for a heavy solution here. I suppose we could let screen headers update state in a Context as well, but maybe not worth it.
However I don't like that it is hardcoded somewhere else than where the header logic is defined, so perhaps define the variable in `ScreenHeader`, and export it.

adriansberg: The thing is that this is pure guesstimation. The header does not set an explicit height, so IMO it's weird to export a variable that says something about the height of the header if that variable is never used to decide anything about the height. I'm fine with having a local variable instead of a comment here though.

marius-at-atb: 44 is exactly right :+1:

src/stacks-hierarchy/Root_SmartParkAndRide/Root_SmartParkAndRideAddScreen.tsx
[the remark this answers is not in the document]
adriansberg: Implemented a proposed solution now.

PR #5309, opened 2025-07-08
refactor: add getThemedAsset util to ThemedAssets

rosvik:
Refactor of the ThemedAssets file to avoid duplication, and make it easier to add new illustrations in the future ✨


rosvik:
> Nice! Btw would also be great if we no longer need empty fallback illustrations such as [these](https://github.com/AtB-AS/design-system/pull/315/commits/bc4baad3146c7ddb230153b572b58168ec88db48).

I agree! Not sure how we're going to solve it though.

PR #5311, opened 2025-07-08
feat: update user info design on the Profile screen

rosvik:
Updates the section at the top of the Profile page ✨ 

https://github.com/user-attachments/assets/5cfc9eca-79b9-4589-8fb6-84a90327f194

closes https://github.com/AtB-AS/kundevendt/issues/20064

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
@@ -65,7 +65,6 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
           style={style.contentContainer}
         >
           <View style={style.mediumGap}>
-            <ContentHeading text={t(ProfileTexts.sections.account.heading)} />
             <UserInfo
               navigateToEditProfileScreen={() =>
                 navigation.navigate('Profile_EditProfileScreen')
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
@@ -1,26 +1,34 @@
+import {LogIn, LogOut} from '@atb/assets/svg/mono-icons/profile';
 import {ActivityIndicatorOverlay} from '@atb/components/activity-indicator-overlay';
 import {ScreenReaderAnnouncement} from '@atb/components/screen-reader-announcement';
 import {LinkSectionItem, Section} from '@atb/components/sections';
 import {ThemeText} from '@atb/components/text';
 import {useAuthContext} from '@atb/modules/auth';
+import {useMobileTokenContext} from '@atb/modules/mobile-token';
 import {useRemoteConfigContext} from '@atb/modules/remote-config';
 import {StyleSheet, Theme} from '@atb/theme';
 import {ProfileTexts, useTranslation} from '@atb/translations';
 import {useIsLoading} from '@atb/utils/use-is-loading';
 import {useLocalConfig} from '@atb/utils/use-local-config';
+import Bugsnag from '@bugsnag/react-native';
 import {APP_ORG_NUMBER, IS_QA_ENV} from '@env';
 import React from 'react';
 import {View} from 'react-native';
 import {getBuildNumber, getVersion} from 'react-native-device-info';
 import {ProfileScreenProps} from './navigation-types';
-
+import {destructiveAlert} from './utils';
 import {ContentHeading} from '@atb/components/heading';
 import {FullScreenView} from '@atb/components/screen-view';
+import {useAnalyticsContext} from '@atb/modules/analytics';
 import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
 import {useStorybookContext} from '@atb/modules/storybook';
-import {useTicketingContext} from '@atb/modules/ticketing';
+import {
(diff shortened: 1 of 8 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -84,131 +65,37 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
...
-                          setIsLoading(false);
-                        }
-                      },
-                    })
+            <UserInfo
+              navigateToEditProfileScreen={() =>
+                navigation.navigate('Profile_EditProfileScreen')
+              }
+            />
+            {authenticationType === 'anonymous' && (
```
marius-at-atb, line 73: I see that e.g. `isLoggedIn` is used in many places, and redeclared every time. Perhaps it should be a util such as
```ts
const {isLoggedIn} = useLoginStatus()
```
And then just `!isLoggedIn` instead of `authenticationType === 'anonymous'` here?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -84,131 +65,37 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
...
+                      'Root_LoginAvailableFareContractWarningScreen',
+                      {},
+                    );
+                  } else if (enable_vipps_login) {
+                    navigation.navigate('Root_LoginOptionsScreen', {
+                      showGoBack: true,
+                      transitionOverride: 'slide-from-bottom',
+                    });
+                  } else {
+                    navigation.navigate('Root_LoginPhoneInputScreen', {});
```
marius-at-atb, line 92: Not a part of your changes, but this is another function that is duplicated in several places, that would be nice to extract and reuse instead.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -328,6 +215,56 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
               </Section>
             </>
           )}
+          {authenticationType === 'phone' && (
```
marius-at-atb, line 218: Ref the other comment, would be nice to just use `isLoggedIn` here.
It think this is a good prep for email login as well.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/components/UserInfo.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const {authenticationType, retryAuth, authStatus} = useAuthContext();
+  const {t} = useTranslation();
+  const analytics = useAnalyticsContext();
+
+  return (
+    <Section>
+      {authenticationType === 'phone' && (
+        <LoggedInInfoSectionItem onPress={navigateToEditProfileScreen} />
+      )}
+      {authenticationType === 'anonymous' && <LoggedOutInfoSectionItem />}
```
marius-at-atb, line 34: Same `isLoggedIn` suggestion here

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
strandlie: Are you thinking of `navigation.navigate`? Isn't that a function that is already extrcted and reused?

src/utils/phone-number-utils.ts
```diff
@@ -4,9 +4,12 @@ import parsePhoneNumber from 'libphonenumber-js';
  * Formats the phone number to the international standard. If the given phone
  * number could not be parsed or formatted, then the input phone number will
  * be returned unmodified.
+ *
+ * Uses a non-breaking space character (U+00A0) to prevent line breaks in the middle of
+ * phone numbers.
  */
 export const formatPhoneNumber = (raw: string): string =>
-  parsePhoneNumber(raw)?.formatInternational() || raw;
+  parsePhoneNumber(raw)?.formatInternational().replaceAll(' ', ' ') || raw;
```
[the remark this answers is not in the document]
strandlie: Very nice!

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
rosvik: I agree that it's not the best solution we have today, but I've thought we'll have to revisit when we add email. It's sometimes important to distinguish between things like `authenticationType !== 'anonymous'` and `authStatus === 'authenticated'`, so we need to think of something that is both precise and understandable.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
rosvik: Summary of our discussion: We need to be careful when creating reusable utils that depend on the navigation, which I guess is why it hasn't been done before. Might work in this case since it's all Root screens, but the stack your navigating in can be different, and miss some of the screens. Won't change anything for now, but maybe easier with the TS improvements in react navigation 7? 🤞 https://github.com/AtB-AS/kundevendt/issues/21199

PR #5312, opened 2025-07-08
feat: add posthog to bonus

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/21044
Adds posthog logging events to bonus checkbox, bonus product clicks and feedback button on bonus page


src/modules/bonus/components/PayWithBonusPointsCheckbox.tsx
```diff
@@ -57,12 +59,20 @@ export const PayWithBonusPointsCheckbox = ({
       ),
     );
 
+  const onPressWithLogging = () => {
+    onPress();
+    logEvent('Bonus', 'bonus points checkbox toggled', {
+      bonusProductId: bonusProduct.id,
+      newState: isChecked,
+    });
```
reidzeibel, line 67: Would it make more sense to put this directly inside the `onPress`? the other change in this commit has the functions directly inside it. Just thinking out loud here 🤔

PR #5315, opened 2025-07-09
test: Test using proxy to capture requests

tormoseng:
This is an test to see if I can capture requests during E2E tests. The motivation is to see that the token requests behave - and possibly add a verification script of the requests after an auth scenario.


tormoseng:
Closing this PR as I'm not able to make it work on GitHub action (works locally) due to:
- no access to install system certs on GitHub emulators with `adb`
- have to manually add user certs (or make a flaky test script doing it in UI)
- without certs all https requests are stopped in the TLS handshake, making only the domain (and not the path) visible in the log

PR #5318, opened 2025-07-10
refactor: zod type safety for Announcements

marius-at-atb:
Rather than using manual mapping, now using zod safeParse to ensure valid data for announcements.

Note to self after summer vacation:
- Should do some more testing before merge, and data in firestore should be updated
- A natural follow-up to this PR, is to do the same for GlobalMessages

Resolves https://github.com/AtB-AS/kundevendt/issues/21179


marius-at-atb:
@rosvik Updates in firestore: make sure that the data fulfils the required props and has valid formats (see logged warnings from zod on this branch). Since we only load Announcements from the V2 collection, I guess we don't have to consider V1 anymore.
Also we should check in with the OMS partners in case they are using Announcements as well.

src/modules/announcements/types.ts
```diff
@@ -1,53 +1,67 @@
...
+  summaryTitle: LanguageAndTextTypeArray.optional(),
+  summary: LanguageAndTextTypeArray,
+  summaryImage: z.string().base64().optional(),
+  fullTitle: LanguageAndTextTypeArray,
+  body: LanguageAndTextTypeArray,
+  mainImage: z.string().base64().optional(),
+  isDismissable: z.boolean().optional(),
+  appPlatforms: z.array(AppPlatformSchema),
+  appVersionMin: z.string(),
+  appVersionMax: z.string(),
```
marius-at-atb, line 50: We might consider making these optional, since they seem to have been treated as such from before, and appliesToAppVersion also supports it

marius-at-atb:
@rosvik I did a little performance test with a 430kb image.
Validating the whole image string as base64 takes about 10x longer than just the ends (~14ms vs ~1.4ms).
So I think it's a good tradeoff we've landed on here.

rosvik:
Ah, I see! Sweet 🙌

PR #5321, opened 2025-07-10
Better e scooter icon transitions

marius-at-atb:
DRAFT, not ready for review

Stations not yet supported.

Overlapping e scooter icons for better transitions between zoom levels


marius-at-atb:
Feedback from designers: the changes in this PR are positive, clearly better on iOS. However Android seems slower, and on a really old and slow Android device, the app crashed in the map, likely because of the extra memory required.
Currently looking into whether we should increase the minimum Android target version.

marius-at-atb:
Further Android testing - we probably need a faster tile server for the UX to be as good there.
Holding this back for now.

marius-at-atb:
Replaced by https://github.com/AtB-AS/mittatb-app/pull/5992

PR #5325, opened 2025-07-11
fix: add a11y button role to UserInfo

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/20064#issuecomment-3061248151


HanneLH (approved):
🌸

PR #5327, opened 2025-07-11
fix: use expanded true for buttons in onboarding screens

rosvik:
Fixes observations from https://mittatb.slack.com/archives/C0116FMPX4Y/p1751958946526719


HanneLH (approved):
🌸

HanneLH:
Tested 🎉

PR #5328, opened 2025-07-11
test: e2e for auth and ticket purchase and mobile token

tormoseng:
NOTE! Added related test-ids to the tests below

Add E2E app tests for:
- login
- purchase a ticket
- verifying the purchased ticket (separate file to avoid session-mess with in-app browser during purchase)
- verify mobile token + barcode in the purchased ticket (verify in toggle + debug + ticket barcode)

Fix E2E app tests for:
- ticket page as anonymous
- ticket page as auth
- remove service disruption link test

Added also some performance measure metrics to measure the time on certain test commands / app operations. Considers only timing and not resource usage.


tormoseng:
Sorry for the rather large PR. Most important code-wise is to take a look on the commits with test-ids. :-)

rosvik (approved):
LGTM 👍

PR #5330, opened 2025-07-14
feat: add voice over to label/Tag

jorunnl:
Addresses comment: https://github.com/AtB-AS/kundevendt/issues/19985#issuecomment-2962270032

Voice over now says 'new functionality' or 'beta functionality' to be more descriptive for voice over users.

### Acceptance Criteria

- [ ] 'New' label reads 'new functionality'
- [ ] 'Beta' label reads 'beta functionality'

### Test input

- [ ] Screenreader

### DIFF
--- a/src/components/sections/items/ExpandableSectionItem.tsx
+++ b/src/components/sections/items/ExpandableSectionItem.tsx
@@ -105,7 +105,8 @@ export function ExpandableSectionItem({
         {suffixNode}
         {label && (
           <Tag
-            labels={[t(TagInfoTexts.labels[label])]}
+            labels={[t(TagInfoTexts.labels[label].text)]}
+            a11yLabel={t(TagInfoTexts.labels[label].a11y)}
             tagType="primary"
             customStyle={styles.labelContainer}
           />
--- a/src/components/sections/items/LinkSectionItem.tsx
+++ b/src/components/sections/items/LinkSectionItem.tsx
@@ -73,7 +73,7 @@ export const LinkSectionItem = forwardRef<any, Props>(
         disabled={disabled}
         accessibilityLabel={
           label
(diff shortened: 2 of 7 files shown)


rosvik (approved):
Sweet ✨

hanne-at-atb:
I notice that under my profile now, it reads out the "new" tag twice on android. First when entering the button for for instance bonus, it reads "bonus, new functionality" then when I swipe to next element that is the "new" tag and it reads out "new, button". So it seems like there is a button with button on top of it. I guess the tag should not be a button in this case?

strandlie:
> I notice that under my profile now, it reads out the "new" tag twice on android. First when entering the button for for instance bonus, it reads "bonus, new functionality" then when I swipe to next element that is the "new" tag and it reads out "new, button". So it seems like there is a button with button on top of it. I guess the tag should not be a button in this case? Can you help me evaluate this @strandlie ?

Looking!

strandlie:
I think the problem here is that all the tags (`PrimaryTag`, `SecondaryTag` and ´SemanticTag`) have gotten [accesibilityRole="button"](https://github.com/AtB-AS/mittatb-app/pull/5330/files#diff-db77068c96dcd3198dcedf04f0e042339ee0539f423a15d42584095ca1d7368aR96). 

That must be an error? I have not seen Tags used as standalone buttons anywhere, they are always subcomponents of other larger components that themselves can or cannot be clickable. 

As you found out @hanne-at-atb the issue is masked on iOS, since it ignores accessibility labels inside containers which themselves have accessibility labels (such as the `LinkSectionItem` into Bonus in the Profile page), but removing the `accessibilityRole="button" from the tags fixes the problem. I can make a PR if this is not intentional?

PR #5333, opened 2025-07-15
feat: add icons to LinkSectionItems in Profile

rosvik:
Adds `rightIcon` and `leftIcon` props to LinkSectionItem, using the same patterns as the Button component, as well as adds icons to the Profile and Settings page.

closes https://github.com/AtB-AS/kundevendt/issues/20067


adriansberg (approved):
Looks good to me! 🤩 

Should also look into preventing the `LinkSectionItem` from being taller when there's a tag, IMO.

rosvik:
> Should also look into preventing the LinkSectionItem from being taller when there's a tag, IMO.

True. Have talked to Mari about it, and it's an issue in figma as well. We could probably make a hack in the app, but not sure what's best.

PR #5335, opened 2025-07-15
fix: Fixed crooked bars and a faulty border on FareContract

strandlie:
It now looks like this:
simulator_screenshot_549EBA11-BC92-4E6C-8AE7-1706D7D57FF4


hanne-at-atb:
Now everything looks really well! ✅

PR #5340, opened 2025-07-15
fix: sort fare contracts by validity

rosvik:
After https://github.com/AtB-AS/mittatb-app/pull/5277, we no longer sort fare contracts by validity, which means upcoming fare contracts or inactive carnets can show above other active fare contracts. This makes the rules we have on sorting more explicit, and improves the tests to reflect the rules.

### Acceptance criteria

- [ ] Only the most recent "failed" reservation is shown
- [ ] Processing reservations are all shown as before
- [ ] Inactive carnets show always below active tickets, even though they were bought after an active ticket
- [ ] Tickets that start in the future always show below active tickets, even though they were bought after an active ticket


strandlie (approved):

hanne-at-atb:
This works as expected based on the acceptance criteria as far as I can see ✅

PR #5342, opened 2025-07-16
feat: fetch server time from identity

rosvik:
Uses the [new time endpoint](https://github.com/AtB-AS/identity/pull/61) to fetch server time instead of relying on the mobile token SDK.

### Acceptance criteria

- [ ] Reloads every 60s, on network reconnects and app reopen
- [ ] If the request fails, we fall back to device time.
- [ ] Server time is used for ticket validity info. The trip planner part of the app still uses device time.

### DIFF
--- a/src/api/identity.ts
+++ b/src/api/identity.ts
@@ -12,6 +12,13 @@ import {v4 as uuid} from 'uuid';
 
 export const VIPPS_CALLBACK_URL = `${APP_SCHEME}://auth/vipps`;
 
+export const getServerTime = async () => {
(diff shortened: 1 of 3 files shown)


hanne-at-atb:
When I change the device time, it overrides the recently fetched server time. If you have a valid ticket and change the time for it to become invalid, it hows as invalid based on the device time.
Even less ideal if you have an invalid ticket, ie. it expired yesterday, you change the time to when it was valid yesterday. It looks valid in the app, and this can be misused in visual inspection.

PR #5343, opened 2025-07-16
feat: add feedback section item in Profile

rosvik:
- Adds a "Gi tilbakemeldinger om appen" section item to Profile
- Adds a notification indicator to the Profile icon in the tab bar that shows when the user has received a intercom message.
- Changes intercom notifications to be red/error, instead of green/valid.
- Some changes to notification indicators in general, done together with @oyvindmikkelsenatb
	- Offsets the notification indicator slightly away from the icon (globally).
	- Changes the travel search filter notification indicator to be blue/info instead of green/valid.

closes https://github.com/AtB-AS/kundevendt/issues/20068

### DIFF
--- a/src/components/theme-icon/NotificationIndicator.tsx
+++ b/src/components/theme-icon/NotificationIndicator.tsx
@@ -66,8 +66,8 @@ const getIndicatorSize = (
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   indicator: {
     position: 'absolute',
-    right: 0,
-    top: 0,
+    right: -4,
+    top: -2,
     borderRadius: theme.border.radius.large,
     zIndex: 10,
     overflow: 'hidden',
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
@@ -74,7 +74,7 @@ export const Root_TabNavigatorStack = () => {
     }
     if (unreadCount) {
       return {
-        color: theme.color.status.valid.primary,
+        color: theme.color.status.error.primary,
         backgroundColor: interactiveColor.default,
       };
     }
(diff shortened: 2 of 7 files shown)


src/components/theme-icon/NotificationIndicator.tsx
```diff
@@ -66,8 +66,8 @@ const getIndicatorSize = (
 const useStyles = StyleSheet.createThemeHook((theme) => ({
   indicator: {
     position: 'absolute',
-    right: 0,
-    top: 0,
+    right: -4,
```
adriansberg, line 69: Is this half the width (and height below here) of the notification indicator?

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -63,6 +65,21 @@ export const Root_TabNavigatorStack = () => {
     nextOnboardingScreen?.name && goToScreen(false, nextOnboardingScreen);
   }, [nextOnboardingSection?.initialScreen, goToScreen, navigation]);
 
+  const getProfileNotification = (): ThemeIconProps['notification'] => {
+    if (customerNumber === undefined) {
```
adriansberg, line 69: Could this just be `!customerNumber`, or is it a good reason for explicitly checking against `undefined`?

adriansberg (approved):
Looks good :shipit:

src/components/theme-icon/NotificationIndicator.tsx
rosvik: This was decided mainly on feel, and what looked best 😅 The notification indicator already lives a bit of a life of it's own outside of the design system.

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -63,6 +65,21 @@ export const Root_TabNavigatorStack = () => {
...
+      return {
+        color: theme.color.status.error.primary,
+        backgroundColor: interactiveColor.default,
+      };
+    }
+    if (unreadCount) {
```
marius-at-atb, line 75: Any reason why this case isn't just included as `|| unreadCount` in the if above instead?

PR #5344, opened 2025-07-16
fix: update time context every second

rosvik:
Some logic that to update the time context's `now` was overwritten in https://github.com/AtB-AS/mittatb-app/pull/5342, so adding it back here. Also improves on some names and comments, and add display of server time offset in the debug menu 🕺


mjenssen (approved):
LGTM 🚀

PR #5347, opened 2025-07-17
feat: remove chat and service disruption icons

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/20069


src/screen-components/nearby-stop-places/NearbyStopPlacesScreenComponent.tsx
```diff
@@ -156,61 +158,61 @@ export const NearbyStopPlacesScreenComponent = ({
...
-        />
-      ) : (
-        <EmptyState
-          title={t(NearbyTexts.stateAnnouncements.noAccessToLocationTitle)}
-          details={t(NearbyTexts.stateAnnouncements.noAccessToLocation)}
-          illustrationComponent={
-            <ThemedOnBehalfOf
-              height={90}
-              style={styles.emptyStopPlacesIllustration}
+    <>
```
adriansberg, line 161: Why did this need to change? Do you have a screenshot from before vs after (though probably identical)?

rosvik: Right, I should have explained that in the PR 😅 The problem was the parallax scroll header, which hides the title until you scroll for a bit.  This change disables the parallax scroll on departures, making both the header and search fields always visible.

adriansberg (approved):
LGTM 🎉

PR #5348, opened 2025-07-18
fix: correctly set duration consts and preemptiveRefreshPercent

rosvik:
`ONE_DAY_MS` is not in use, but the `HALF_DAY_MS` const (which is 12 minutes instead of 12 hours), is used by

- shouldPreemptiveRenew in mobileTokenClient
- cacheTime and staleTime of useFetchOnBehalfOfAccountsQuery
- cacheTime and staleTime of useGeofencingZonesQuery

In the case of shouldPreemtiveRenew, this means that mobile tokens have been renewed when there was 12 minutes OR it has lived for 10% "of the tokens total time to live". The latter condition might have saved us from issues here, but it seems to be set by mistake.

From mobileTokenClient.ts:

```ts
  /*
  Check if token should be renewed. Will return true if either:
   - Token is expired
   - It is less than 12 hours until token expires
   - Less than 10 % of the token validity time is left
   */
  shouldRenew: (token: ActivatedToken) =>
    abtClient.shouldPreemptiveRenew(token, HALF_DAY_MS, 10),
```

> Less than 10 % of the token validity time is left

Sounds like this has it backwards. From the SDK:

```ts
/**
 * Returns true if the token has lived longer than this percent of the tokens total time to live.
 */
preemptiveRefreshPercent: number,
```

---

So, this changes the mobile token to use 12 hours remaining or 90% elapsed, instead of 12 minutes remaining or 10% elapsed 🙈

### DIFF
--- a/src/modules/mobile-token/mobileTokenClient.ts
+++ b/src/modules/mobile-token/mobileTokenClient.ts
@@ -82,10 +82,10 @@ export const mobileTokenClient = {
   Check if token should be renewed. Will return true if either:
    - Token is expired
    - It is less than 12 hours until token expires
-   - Less than 10 % of the token validity time is left
+   - It has lived longer than 90% percent of its total time to live
    */
   shouldRenew: (token: ActivatedToken) =>
-    abtClient.shouldPreemptiveRenew(token, HALF_DAY_MS, 10),
+    abtClient.shouldPreemptiveRenew(token, HALF_DAY_MS, 90),
   currentTimeMillis: () => abtClient.getCurrentTimeMillis(),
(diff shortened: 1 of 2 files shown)


adriansberg (approved):
This looks like stuff we need to test thoroughly 😅

hanne-at-atb:
Not entirely able to test this is staging with backoffice modifications as of now. We have [a thread with Entur](https://mittatb.slack.com/archives/C03DTC05VRQ/p1754040428055119) to look into why token in debug menu is out of sync with Backoffice.
Together with @mjenssen and @strandlie we consider releasing this in the next version as being of low risk after testing that the caching of the two endpoints are not done after 12 minutes now. 
Moving this along to approved ✅

PR #5352, opened 2025-07-18
feat: use products api instead of firestore

rosvik:
- Uses `useGetFareProductsQuery` instead of `useFirestoreConfigurationContext` for all uses of `preassignedFareProducts`
- Persists fare products in local storage, which is cached for one week, the same as fare contracts.
- Makes pull to refresh on the active tickets and purchase tab reload preassigned fare products.
- Removes the reliance of firestore data entirely.

Note: If we want to keep the fallback to firestore data for a bit longer, 143ea65e4b6e0d2a737367ef6d4b253dd5a5ea21 can be reverted.

closes https://github.com/AtB-AS/kundevendt/issues/17568


rosvik:
I'm leaving for vacation in just a few minutes, and this is quite a critical change, so this might be something that can wait until I, or more developers are back 🤷 Could also keep the firestore fallback as mentioned in the description if we want to reduce risk a bit.

strandlie:
I think its smart to let this sit until Johannes is back.

strandlie (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_TicketingStack/Ticketing_TicketTabNavStack/TicketTabNav_PurchaseTabScreen/TicketTabNav_PurchaseTabScreen.tsx
```diff
@@ -25,8 +25,17 @@ export const TicketTabNav_PurchaseTabScreen = ({navigation}: Props) => {
...
   const {theme} = useThemeContext();
-  const {recentFareContracts, loading} = useRecentFareContracts();
-  const {data: fareProducts} = useGetFareProductsQuery();
+  const {
+    recentFareContracts,
+    isLoading: isLoadingRecentFareContracts,
+    refresh: refetchRecentFareContracts,
+  } = useRecentFareContracts();
+  const {
+    data: fareProducts,
```
marius-at-atb, line 34: Everywhere else this is called `preassignedFareProducts`

PR #5355, opened 2025-07-29
fix: Prevented booking flow to proceed without selecting fromTo

strandlie:
It was possible to proceed in the booking flow without selecting from and to-places. This is prevented now. 

Also uncovered that we passed `selection` to the `Summary`-component, which was not necessary since the same logic was covered outside of the component. This also leads to better separation of concerns and less coupling of components. Finally, this lead to a renaming of the `isError` prop of `summary` to `isDisabled` since that is functionally what it does now.


hanne-at-atb:
This works well now! ✅

PR #5365, opened 2025-08-05
feat: fixed map jumping with followUserLocation state

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21269

When starting a trip, and the map goes in followUserLocation mode, that map location state does not update. So when ending the trip the map snaps back to the previous location before starting the trip. Same if the user start manually moving the map while in an active trip, it will then snap back to previous position before followUserLocation where activated. So with this PR i update the camera state based on the user location while the follow mode is on

### DIFF
--- a/src/modules/map/MapV2.tsx
+++ b/src/modules/map/MapV2.tsx
@@ -9,8 +9,9 @@ import {
   Viewport,
   Camera,
   MapView,
+  UserLocation,
 } from '@rnmapbox/maps';
-import {Feature} from 'geojson';
+import {Feature, Position} from 'geojson';
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {View} from 'react-native';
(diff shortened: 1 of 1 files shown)


src/modules/map/MapV2.tsx
```diff
@@ -319,11 +327,20 @@ export const MapV2 = (props: MapProps) => {
...
-              startingCoordinates.longitude,
-              startingCoordinates.latitude,
-            ]}
+            zoomLevel={followCameraState?.zoomLevel ?? DEFAULT_ZOOM_LEVEL}
+            centerCoordinate={
+              followCameraState
+                ? [
+                    followCameraState.centerCoordinate[0],
+                    followCameraState.centerCoordinate[1],
+                  ]
```
marius-at-atb, line 336: This is same as just `followCameraState.centerCoordinate`

PR #4154, opened 2023-12-21
feat: update RNMapBox dependency and update usage of UserLocation

reidzeibel:
Replaces PR : https://github.com/AtB-AS/mittatb-app/pull/4141
Closes : https://github.com/AtB-AS/kundevendt/issues/16093

This PR updates the RNMapBox dependency to fix the issue with using `UserLocation` with `onMapIdle` on Android.

Also replaces the deprecated `UserLocation` with the new `LocationPuck` as suggested by RNMapBox (`NativeUserLocation` is also deprecated).

~~iOS users might have to run `pod update MapboxMaps` before `pod install` for it to be able to run. Please let me know if I need to update something regarding this issue.~~ just pod install should work now, I have updated the podfile.lock.

Screenshots: 

Android: 
image

iOS: 
image

PR #5373, opened 2025-08-12
fix: fixed bug on android with crash and optimized followUser functionallity

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21269

### DIFF
--- a/src/modules/map/MapV2.tsx
+++ b/src/modules/map/MapV2.tsx
@@ -11,7 +11,7 @@ import {
   MapView,
   MapState,
 } from '@rnmapbox/maps';
-import {Feature, Position} from 'geojson';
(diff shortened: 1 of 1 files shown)


src/modules/map/MapV2.tsx
```diff
@@ -373,27 +372,6 @@ export const MapV2 = (props: MapProps) => {
           )}
 
           <LocationPuck puckBearing="heading" puckBearingEnabled={true} />
-          <UserLocation
```
marius-at-atb, line 376: Just for reference: https://github.com/AtB-AS/mittatb-app/pull/4154

marius-at-atb (approved):
Lgtm!

PR #5369, opened 2025-08-08
refactor: Refactored component into TripPatternLegsSummary

strandlie:
A step towards https://github.com/AtB-AS/kundevendt/issues/20769

PR #5374, opened 2025-08-13
feat: add info from enriched fare contract

strandlie:
Closes https://github.com/AtB-AS/kundevendt/issues/20769

https://github.com/AtB-AS/mittatb-app/pull/5369 must be merged first. 

Adds a small section to the FareContract with info about where it is valid between, if it has information about DatedServiceJourney.

<details>

</details>


src/components/journey-legs-summary/JourneyLegsSummary.tsx
```diff
@@ -9,60 +8,68 @@ import React from 'react';
 import type {SalesTripPatternLeg} from '@atb/api/types/sales';
 import {StyleSheet} from '@atb/theme';
 
-export function TripPatternLegsSummary({legs}: {legs?: SalesTripPatternLeg[]}) {
+export function JourneyLegsSummary({
```
marius-at-atb, line 11: `Journey` on its own is not an existing entity in NeTEX.
However `SalesTripPatternLeg` isn't either. I'm not familiar with it from before, but is it a subset of `Leg`?
Would it be a good idea to keep the existing name, or rename this to e.g. `SalesTripPatternLegsSummary` or just `LegsSummary`?

src/modules/fare-contracts/use-fare-contract-legs.ts
```diff
@@ -0,0 +1,71 @@
...
+export function useFareContractLegs(
+  dsjRefs: DatedServiceJourneyRefsType | undefined,
+): FareContractLegsType {
+  const {data} = useQuery({
+    queryKey: ['datedServiceJourneys', dsjRefs?.datedServiceJourneyRef],
+    queryFn: async () => {
+      if (!!dsjRefs?.datedServiceJourneyRef) {
+        return getDatedServiceJourney(dsjRefs?.datedServiceJourneyRef);
+      }
+    },
```
marius-at-atb, line 27: The if check should be handled inside `getDatedServiceJourney`, so you only need this here:
`queryFn: () => getDatedServiceJourney(dsjRefs?.datedServiceJourneyRef),`

src/modules/fare-contracts/use-fare-contract-legs.ts
```diff
@@ -0,0 +1,71 @@
+import {useQuery} from '@tanstack/react-query';
+import {ONE_MINUTE_MS, ONE_SECOND_MS} from '@atb/utils/durations';
+import {getDatedServiceJourney} from '@atb/api/bff/servicejourney';
+import type {DatedServiceJourneyRefsType} from '@atb-as/utils';
+import type {SalesTripPatternLeg} from '@atb/api/types/sales';
+import type {DatedServiceJourney} from '@atb/api/types/generated/journey_planner_v3_types';
+
+type FareContractLegsType = {
+  legs: SalesTripPatternLeg[];
+};
```
marius-at-atb, line 10: Hmm, I'm getting a bit confused with the different names/entities.
Perhaps this can just return `SalesTripPatternLeg[]` directly so that FareContractLegsType can be removed?

src/components/journey-legs-summary/JourneyLegsSummary.tsx
marius-at-atb: Normally we stick to the same domain model as from Entur, and then use leaf node utils to convert: https://github.com/AtB-AS/docs-private/blob/main/kundevendt-development-guidelines.md#domain-models

marius-at-atb (commented):
Nice work!
See some comments regarding naming/domain model/types.

src/modules/fare-contracts/use-fare-contract-legs.ts
strandlie: Yeah, I just made it return an object out of habit since "all hooks return object". I know they don't, so I can change it :)

src/components/journey-legs-summary/JourneyLegsSummary.tsx
marius-at-atb: Btw while looking into this I noticed that `OfferFromLegsResponse` should be defined with zod to parse the response. Could you make a separate PR for that?

src/modules/fare-contracts/use-fare-contract-legs.ts
```diff
@@ -33,39 +30,96 @@ export function useFareContractLegs(
...
+    ),
+    serviceJourneyEstimatedCalls: estimatedCalls,
+    fromEstimatedCall: fromCall,
+    toEstimatedCall: toCall,
+    line: datedServiceJourney.serviceJourney.line,
   };
 }
+
+// Utility function to map TransportMode to Mode
+export function mapTransportModeToMode(mode?: TransportMode): Mode | undefined {
```
marius-at-atb, line 98: Could consider a separate util file for this one in `src/utils`

marius-at-atb (approved):
Great!

PR #5375, opened 2025-08-13
feat: Added handling for FINISHING state in active booking

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21342
Closes https://github.com/AtB-AS/kundevendt/issues/21405

If a user finished the scooter trip but did not take picture, they should be prompet to take a picture when reentering the mapscreen, before they can do anything else there. They need to finish up their active booking.

simulator_screenshot_52A8AA5B-B902-4B5F-8F5F-825CB7948854

### DIFF
--- a/src/modules/map/hooks/use-active-shmo-booking.tsx
+++ b/src/modules/map/hooks/use-active-shmo-booking.tsx
@@ -62,6 +62,14 @@ export const useShmoActiveBottomSheet = (
         activeBooking &&
         activeBooking.asset.formFactor === FormFactor.Scooter
       ) {
+        const photoNavigation = () => {
+          mapSelectionCloseCallback();
+          closeBottomSheet();
+          navigation.navigate('Root_ParkingPhotoScreen', {
+            bookingId: activeBooking.bookingId,
+          });
+        };
+
         switch (activeBooking.state) {
           case ShmoBookingState.IN_USE:
             openBottomSheet(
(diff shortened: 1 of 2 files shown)


src/modules/map/hooks/use-active-shmo-booking.tsx
```diff
@@ -88,6 +91,25 @@ export const useShmoActiveBottomSheet = (
...
+              () => (
+                <FinishingScooterSheet
+                  onForceClose={mapSelectionCloseCallback}
+                  photoNavigation={() => {
+                    mapSelectionCloseCallback();
+                    closeBottomSheet();
+                    navigation.navigate('Root_ParkingPhotoScreen', {
+                      bookingId: activeBooking.bookingId,
+                    });
+                  }}
```
marius-at-atb, line 105: Can you extract this into a function and re-use both here and in ActiveScooterSheet above?

src/modules/mobility/components/sheets/FinishingScooterSheet.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }, [
+    activeBooking?.asset.operator.id,
+    activeBooking?.bookingId,
+    logEvent,
+    photoNavigation,
+  ]);
+
+  return (
+    <BottomSheetContainer maxHeightValue={0.7} disableHeader={true}>
+      {isShmoDeepIntegrationEnabled && (
```
marius-at-atb, line 66: Hmm, you should only end up in this screen if you have an unfinished trip, and need a way to close it, so I think we should remove the `isShmoDeepIntegrationEnabled` check here.

src/modules/mobility/components/sheets/FinishingScooterSheet.tsx
```diff
@@ -0,0 +1,132 @@
...
+            </View>
+          )}
+          {!isLoading && !isError && activeBooking && (
+            <View style={styles.container}>
+              <View style={styles.contentWrapper}>
+                <ThemeText typography="heading--big">
+                  {t(MobilityTexts.finishing.header)}
+                </ThemeText>
+                <View style={styles.illustrationWrapper}>
+                  <ThemedBeacons height={102} />
```
marius-at-atb, line 80: @strandlie ref your comment on semantic naming etc. Here's an example where the Beacons illustration is used in non-beacons context.

marius-at-atb (commented):
Looks good!

PR #5378, opened 2025-08-18
fix: gracefully handle invalid fare contract data

rosvik:
We had an issue in prod where accounts with old school tickets would crash while we did the initial mapping of firestore timestamps. The app could crash if e.g. `travelRights` were set to `[ null ]`. This change tries to handle all those issues by ignoring any errors thrown in the mapping functions, and discarding the fare contracts instead. 

Since the goal is to no longer rely on firestore for fare contract data (and remove all of this code in the process), I'm not aiming for super clean reliable code here, rather something simple that fixes the issues we're having in prod.

## Acceptance criteria

- [x] Any invalid data in fare contracts or travel rights should no longer crash the app, but instead be discarded. This includes `null` values, empty travel rights array, invalid timestamps, missing fields etc.

## Test input

Made an ios staging build with build number `1755516993`

### DIFF
--- a/src/modules/ticketing/firestore.ts
+++ b/src/modules/ticketing/firestore.ts
@@ -5,6 +5,7 @@ import firestore, {
 import {addHours} from 'date-fns';
 import {CustomerProfile, Reservation} from './types';
 import {FareContractType} from '@atb-as/utils';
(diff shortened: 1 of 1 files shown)


tormoseng:
✅

PR #5380, opened 2025-08-19
fix: enrollment text update and remove code on enrollment

jorunnl:
As discussed with @hannahATB 
image

## 📋 Acceptance criteria

- [x] Info text on 'Nye konsepter' screen is updated and understandable
- [x] Text to tell user they joined a group is updated and understandable

## 💣 Test input

- [x] When user enters correct code, code is cleared on 'join'-button press
- [x] When user enters incorrect code, code is not cleared on 'join'-button press
- [x] 200% Zoom
- [x] Screen reader
- [x] Light and dark mode
- [x] Language
- [ ]


jorunnl:
@hanne-at-atb I will add the d in selected
I agree that we test with users not on users. 

Any objections on the wording from you @hannahATB ?

PR #5383, opened 2025-08-20
feat: use stream service to invalidate bonus amount query

jorunnl:
Updates StreamEventSchema to reflect backend

Invalidate Bonus amount query on FARE_CONTRACT event


rosvik (approved):
LGTM 💯

PR #5387, opened 2025-08-20
fix(spar): use correct secondary button color

adriansberg:
Update the secondary button's background color to the correct accent color for better visual consistency.


tormoseng:
✅

PR #5388, opened 2025-08-20
feat: Support isDismissable

marius-at-atb:
isDismissable was included in the Announcement type, but not respected.
Now the close x button is hidden if isDismissable is true.

| isDismissable=true or not set, x shows | isDismissable=false, x hidden |
|--------|-------|
|  |  |

Resolves https://github.com/AtB-AS/kundevendt/issues/21179


marius-at-atb:
@rosvik Closing this since as you mention it may be better to not support it :+1:
Remove from the type instead: https://github.com/AtB-AS/mittatb-app/pull/5389

PR #5396, opened 2025-08-22
fix: bonus checkbox posthog value logged is correct

jorunnl:
newState logged in posthog for checkbox toggle is now actual newState value


tormoseng:
✅

PR #5397, opened 2025-08-22
fix: fetch server time even when traveling back in time

rosvik:
This re-applies changes from https://github.com/AtB-AS/mittatb-app/pull/5342, and applies a fix so that we always fetch server time after the user manually sets device time in settings. I'll try to summarize my findings.

The issue with the previous PR seems to be when the device time is set back in time, e.g. from 12:00 to 10:00. In this case, the useServerTimeQuery will fetch server time only when the clock gets back to 12:01. Finding the root cause of this was a bit of a journey, but it boils down to two issues with the react query setup:

## 1. `refetchOnWindowFocus: true`

This should cause the app to reload server time whenever the app gets focus. So if the user exits the app to change their device time, it should refresh at once when they return to the app. However, from the [docs](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery):

> refetchOnWindowFocus
> - If set to `true`, the query will refetch on window focus if the data is stale.

and 

> staleTime
> - Defaults to `0`

When the device time is set back in time the data isn't stale yet, since "two hours in the past" is less than `0` 🤷 Setting `refetchOnWindowFocus: "always"` should fix this, and refetch server time even when not stale. 8c0372ce1b3e23d826f3a0dfcccb5b66d425be93

## 2. `refetchInterval: ONE_MINUTE_MS`

However, `refetchOnWindowFocus` wasn't the only mechanism that should have caused updates 🙈 There should have been a refetch every minute, regardless of device time.

Initially, I thought this was a bug in react query. It's not! [Internally react query uses a plain old `setInterval`](https://github.com/TanStack/query/blob/a1b1279be3befc37a9d4fbb2fa79826bfb0955a4/packages/query-core/src/queryObserver.ts#L397). This is not _supposed_ to be affected by device time, and in a browser it's not. In React Native, however, it is.

**This means the issue we're seeing with server time applies to every `setInterval` in the app.** So if we have other functionality that rely on it, it might also break when device time is set back in time. Might be something to look into, but our `useInterval` hook uses `setTimeout`, which might work better.

---

Anyway, thanks for reading my rant 💛

TL;DR: I changed `true` to `"always"`. That's it.

### DIFF
--- a/src/modules/time/use-server-time-query.ts
+++ b/src/modules/time/use-server-time-query.ts
@@ -8,6 +8,6 @@ export const useServerTimeQuery = (enabled: boolean) =>
     queryFn: () => getServerTime(),
     refetchInterval: ONE_MINUTE_MS,
     refetchOnReconnect: true,
-    refetchOnWindowFocus: true,
+    refetchOnWindowFocus: 'always',
     enabled,
   });
--- a/src/modules/time/use-server-time-query.ts
+++ b/src/modules/time/use-server-time-query.ts
@@ -1,0 +1,13 @@
+import {useQuery} from '@tanstack/react-query';
+import {ONE_MINUTE_MS} from '@atb/utils/durations';
+import {getServerTime} from '@atb/api/identity';
+
+export const useServerTimeQuery = (enabled: boolean) =>
+  useQuery({
+    queryKey: ['getServerTime'],
+    queryFn: () => getServerTime(),
+    refetchInterval: ONE_MINUTE_MS,
+    refetchOnReconnect: true,
+    refetchOnWindowFocus: true,
+    enabled,
+  });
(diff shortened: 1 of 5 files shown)


src/modules/time/use-server-time-query.ts
```diff
@@ -0,0 +1,13 @@
+import {useQuery} from '@tanstack/react-query';
+import {ONE_MINUTE_MS} from '@atb/utils/durations';
+import {getServerTime} from '@atb/api/identity';
+
+export const useServerTimeQuery = (enabled: boolean) =>
```
marius-at-atb, line 5: Perhaps move
```ts
const {isServerTimeEnabled} = useFeatureTogglesContext();
```
inside this hook, and don't accept enabled as a param?

marius-at-atb (approved):
Well done tracking this down!

hanne-at-atb:
What a journey you've been on! Happy to hear you tracked down the true that should be always.

src/modules/time/use-server-time-query.ts
rosvik: Makes sense! -> e58406ae7852af1b3d7ee7e0e2a4615e45d7ef8e

PR #5401, opened 2025-08-22
fix: Earning 0 points is not visble on bought ticket

jorunnl:
| Before | After |
|----------|----------|
| image    | image   |


hanne-at-atb:
Tested ok ✅  Take it away, release!

PR #5405, opened 2025-08-25
feat: Add hasValidFareContract variable for Announcements

cbrevik:
Closes https://github.com/AtB-AS/kundevendt/issues/21182

Have not talked to @Sebstorvik if we should consider adding hasValidFareContract to globalRuleVariables in GlobalMessages or not. But I understood from discussion with @marius-at-atb and @rosvik that it's probably not necessary?


marius-at-atb (approved):
Lgtm!

PR #5409, opened 2025-08-28
feat: updated posthog analytics for map

Dahly96:
Updated to use new checks t registrer clicks on items in the map. And also added one that registreres when a scooter is scanned


src/stacks-hierarchy/Root_ScanQrCodeScreen.tsx
```diff
@@ -91,14 +96,22 @@ export const Root_ScanQrCodeScreen: React.FC<Props> = ({navigation}) => {
 
       if (!!type && !!id) {
         setBottomSheetToAutoSelect({type, id});
+        analytics.logEvent('Map', 'Scooter selected', {
```
marius-at-atb, line 99: Worth adding "through scan" here?

marius-at-atb (approved):

PR #5411, opened 2025-08-28
draft: WIP on new bottomsheet v2

Dahly96:


marius-at-atb (changes requested):
Great work 🎉 
So much cleaner and simpler now.

A few comments as one would expect for such a big refactor, but only easy fixes I think.

src/components/bottom-sheet-map/BottomSheetMap.tsx
```diff
@@ -0,0 +1,196 @@
...
+                      color={theme.color.foreground.dynamic.secondary}
+                    >
+                      {subText}
+                    </ThemeText>
+                  )}
+                </View>
+              </View>
+            )}
+
+            {rightIconText && rightIcon && (
```
strandlie, line 137: What do you think about collection `rightIconText` and `rightIcon` into some common structure? I think that would make sense, since they are both required to show this? 

Something like: 
```
type RightButton = {
   text: string;
   icon: Icon;
}
```

Dahly96: Good idea! i like the type approach. I do think then the designers need to agree on some common patterns to link to the different types. But i could make an issue for it, but maybe not for this PR. And there is also not a strict rule, atleast now for it to be both icon and text. So im changing it to "or", so one of them has to be there for the button to be defined,  and conditional rendering for the items inside, like the icon and the text

marius-at-atb (changes requested):
Well done figuring out the animation!
Some more cleanup/details uncovered which should be looked at before merging.

marius-at-atb (approved):
Lgtm 🚀

PR #5412, opened 2025-08-29
feat: Update map dark style

marius-at-atb:
Color adjustments to the dark mode map style.

| Before | After |
|--------|-------|
|  |  |

(Note: icons also being changed simultaneously, but that is not a part of this PR.)

Resolves https://github.com/AtB-AS/kundevendt/issues/21184


rosvik (approved):
Looks nice ✨

PR #5414, opened 2025-09-03
feat(aztec): remote configurable aztec code dimensions

mathiazom:
For the FRAM Aztec rollout, we'd like to be able to tweak dimensions via remote config.

Defined `aztec_code_max_height` and `aztec_code_padding` for this. The current values have been preserved as defaults. 

~~The default setup using `themes['light'].spacing.large` doesn't seem ideal, so open to other approaches!~~

Initial proposal (first commit):
> Since the padding default is retrieved from `theme`, we can't give a proper default from the start (`RemoteConfigContextProvider` is not wrapped by `ThemeContextProvider`). Also, remote config only supports `string | number | boolean` defaults. I chose to use `Number.MIN_SAFE_INTEGER` as a `undefined`/`null` alternative. `NaN` would make more sense, but causes errors in React Native. Also, `Number.MIN_SAFE_INTEGER` will have the safe effect as `0`, so it's safe even if you forget to check the value. Open to other solutions 🤗

### DIFF
--- a/src/modules/fare-contracts/details/Barcode.tsx
+++ b/src/modules/fare-contracts/details/Barcode.tsx
@@ -86,6 +86,7 @@ function useScreenBrightnessIncrease() {
  */
 const MobileTokenAztec = ({fc}: {fc: FareContractType}) => {
   const styles = useStyles();
+  const {aztec_code_max_height, aztec_code_padding} = useRemoteConfigContext();
   const {t} = useTranslation();
   const {data: signedToken} = useGetSignedTokenQuery();
   const [aztecCodeError, setAztecCodeError] = useState(false);
@@ -108,7 +109,10 @@ const MobileTokenAztec = ({fc}: {fc: FareContractType}) => {
 
   return (
     <View
-      style={styles.aztecCode}
+      style={[
+        styles.aztecCode,
+        {padding: aztec_code_padding, maxHeight: aztec_code_max_height},
+      ]}
       accessible={true}
       accessibilityLabel={t(FareContractTexts.details.barcodeA11yLabel)}
       testID="mobileTokenBarcode"
@@ -180,6 +184,7 @@ const LoadingBarcode = () => {
 
 const StaticAztec = ({fc}: {fc: FareContractType}) => {
   const styles = useStyles();
+  const {aztec_code_max_height, aztec_code_padding} = useRemoteConfigContext();
   const {t} = useTranslation();
   const [aztecXml, setAztecXml] = useState<string>();
   const onCloseFocusRef = useRef<RefObject<any>>(null);
@@ -197,7 +202,12 @@ const StaticAztec = ({fc}: {fc: FareContractType}) => {
   if (!aztecXml) return null;
 
   return (
-    <View style={styles.aztecCode}>
+    <View
+      style={[
+        styles.aztecCode,
+        {padding: aztec_code_padding, maxHeight: aztec_code_max_height},
+      ]}
+    >
       <PressableOpacity
         onPress={onOpenBarcodePress}
         accessibilityRole="button"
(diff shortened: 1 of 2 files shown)


src/modules/fare-contracts/details/Barcode.tsx
```diff
@@ -249,28 +250,31 @@ const StaticQrCode = ({fc}: {fc: FareContractType}) => {
...
-    marginLeft: 'auto',
-    marginRight: 'auto',
-    maxWidth: 250,
-  },
-  staticQrCodeSmall: {
-    maxWidth: 125,
-  },
-}));
+const useStyles = () => {
+  const {aztec_code_max_height, aztec_code_padding} = useRemoteConfigContext();
```
mikaelbr, line 254: Wonder if this warrants a comment actually. 

> "why in all that is holy is this configurable" 
– someone in 2 years

I don't think the need here is self explanatory, so describing why this is needed I think is nice. "Some barcode readers are sensitive on size, etc etc blah blah"

src/modules/remote-config/remote-config.ts
```diff
@@ -82,6 +85,8 @@ export type RemoteConfig = {
 };
 
 export const defaultRemoteConfig: RemoteConfig = {
+  aztec_code_max_height: 275,
+  aztec_code_padding: themes['light'].spacing.large,
```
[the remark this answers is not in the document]
mathiazom: That makes sense actually, design changes might deviate from code reading requirements. Update to use current value for "large spacing" (20)

src/modules/fare-contracts/details/Barcode.tsx
rosvik: I was just about to ask why (in all that is holy) the padding is configurable. Is the height not enough?

mikaelbr: Might want to add a note in the comment on darkmode/lightmode. There is some concern that the surrounding white border field in darkmode has implications to how well the barcode is read. Meaning, if it is too small it is harder to read. This isn't really confirmed (in my mind), but still being able to experiment to be a bit more empirical is nice.

Note: Testing all this locally just wont do and we've used several hours trying to check read stability/speed, but out in the real world a lot of things happen (different devices, different readers, sun, movement etc etc)

mikaelbr: Could be clever with maths etc, but I think it is just as simple and versatile to just treat the values separately

src/modules/fare-contracts/details/Barcode.tsx
```diff
@@ -249,28 +250,32 @@ const StaticQrCode = ({fc}: {fc: FareContractType}) => {
...
+const useStyles = () => {
+  // Some code readers are sensitive to size and padding. Configurable parameters allow quick response to reading issues.
+  const {aztec_code_max_height, aztec_code_padding} = useRemoteConfigContext();
+  return StyleSheet.createThemeHook(() => ({
+    aztecCode: {
+      width: '100%',
+      aspectRatio: 1,
+      padding: aztec_code_padding,
+      backgroundColor: '#FFFFFF',
+      maxHeight: aztec_code_max_height,
```
rosvik, line 262: Nit: The most common pattern for dynamic styling values in the app is to set the style prop to something like `style={[styles.aztecCode, {padding: aztec_code_padding, maxHeight: aztec_code_max_height}]}`, and avoid logic like this in the useStyles hook.

rosvik (approved):
LGTM 👍

src/modules/fare-contracts/details/Barcode.tsx
mathiazom: Yea, was not sure about this. I guess I wanted to avoid reading remote config each time `styles.aztecCode` is used, currently four times. But nice to follow the same pattern as elsewhere :+1:

mathiazom: Updated ✅

rosvik (approved):

PR #5415, opened 2025-09-03
fix: prevent cached response when traveling back in time

rosvik:
Part II of https://github.com/AtB-AS/mittatb-app/pull/5397! While the previous edition of our time travel adventures sure were exciting, the journey continues ✨

I've tried to explain what I did in comments, and won't write an essay on it this time. In short, these changes should make fare contracts follow server time when setting the device clock back in time, fixing https://github.com/AtB-AS/kundevendt/issues/21276#issuecomment-3223672232.

Let's not talk about ba211bbbaf1153b659cc3330a9d22e99878d1566.


adriansberg (approved):
Tested in simulator, working as expected 🤩

PR #5419, opened 2025-09-04
feat: Upgrade to React Native 0.81

cbrevik:
Partly fixes https://github.com/AtB-AS/kundevendt/issues/21229

**Important**: This PR does not enable new architecture. Because of [several visual bugs](https://github.com/AtB-AS/kundevendt/issues/21229#issuecomment-3269589722) this has been postponed. I do however upgrade to the latest version of React Native, and bump several dependencies.

## Highlights
* We are now using React `19` and React Native `0.81.1`.
* Rewrote AppDelegate to Swift (again).
* Updates the app to support `react-navigation` `7`+. It also refactors PostHog-tracking to accommodate changes in navigation event handling, since PostHog does not support autocapture with `react-navigation` `7`+.
* Bump Mapbox to `11.13.4`.
* Bumps required Node-version to `22`+, since this is the minimum version for React Native `0.81`.
* Build tools and SDK versions for Android in development is now `36`.
* Added [document detailing how to upgrade to new architecture](https://github.com/AtB-AS/mittatb-app/pull/5419/files?short_path=28ea518#diff-28ea5186f27be40f3de85769525de8f5462d7b9e4dc8ce05dcc541d0f23d4089) when the time comes.

## Other notable changes
* Added the `navigationInChildEnabled` prop to `NavigationContainer` to support nested navigation with React Navigation v7. This is [only temporarily supported](https://reactnavigation.org/docs/upgrading-from-6.x/#the-navigate-method-no-longer-navigates-to-screen-in-a-nested-child-navigator), and should be addressed in a separate PR. I've not done a lot of research on how substantial of a change this entails. Just that I noticed some parts of the app navigation broke, since they depend on being to able to navigate to nested child navigators implicitly.
* Removed `src/utils/transition-spec.ts` since it's unused and I couldn't import the type needed there anymore with upgrading react-navigation.
* Upgraded `posthog-react-native` to v4 and disabled autocapture, and manually send screen events to PostHog in `trackNavigation` instead. This is because [PostHog does not support autocapture](https://posthog.com/docs/libraries/react-native#capturing-screen-views) with `react-navigation` 7+. 
* Changed the bootsplash generation script to explicitly create assets for both `beacons` and `app` flavors instead of using the `$APP_FLAVOR` variable. Just a QOL thing so you can build either app flavor locally.
* Add negative margins to ThemeText in tab bar button to prevent early text breaking in tab bar label
* Lots of views have minor changes like `JSX.Element` -> `React.JSX.Element`, this is because of React `19`.
* Lots of minor lint and formatting-related changes, because of more strict checks with updated dependencies.
* Migrated from `@testing-library/react-hooks` (deprecated) to `@testing-library/react-native`
* Removed linting of js/jsx, this either way did not work with current setup and just threw an error now
* Had to force resolution of `zod` to a specific version, since our other `@atb` dependencies do not agree, and with newer versions `zod` will derp if there are conflicting versions installed

## Future follow up from this PR:
* Update app navigation to be specific about navigation container so we can at some point remove `navigationInChildEnabled`. This will make our navigation better statically typed.
* Investigate visual bugs, one of the biggest issues was the use of `overflow: visible` in `ScrollChildrenIOS` (part of `ParallaxScroll` being used everywhere): https://github.com/AtB-AS/mittatb-app/blob/master/src/components/parallax-scroll/ParallaxScroll.tsx#L152 - only an issue if new architecture is enabled
* Create reproducible sample projects for the visual bugs we are experiencing, and have them fixed, before we can upgrade to new architecture


src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
```diff
@@ -197,7 +201,14 @@ function tabSettings(
...
       <ThemeText
         typography="body__secondary"
-        style={{color, textAlign: 'center', lineHeight}}
+        style={{
+          color,
+          textAlign: 'center',
+          lineHeight,
+          // react-navigation v7 adds 5px padding to the tab bar label, and breaks text too early
+          // this workaround is to use negative margins to extend beyond parent padding
+          marginHorizontal: -5,
```
strandlie, line 210: There is no config i react-navigation v7 to turn off this padding?

strandlie (approved):
Too big to really know what I'm approving, but at least it LGTM!

src/stacks-hierarchy/Root_TabNavigatorStack/Root_TabNavigatorStack.tsx
cbrevik: No 😔 The wrapped button has different variants, but you cannot override its style like that: https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/views/BottomTabItem.tsx#L355-L371

You CAN override the button itself, but then I'd have to make a new component based on their elements package, just to override `marginHorizontal` on it. Which I thought would be more likely to have regressions later than just a simple style-hack here.

src/modules/diagnostics/trackNavigation.ts
```diff
@@ -1,12 +1,19 @@
...
 
 export function trackNavigation(currentRouteName: string) {
   if (previousRouteName !== currentRouteName) {
     Bugsnag.leaveBreadcrumb('navigate', {route: currentRouteName});
     analytics().logScreenView({screen_name: currentRouteName});
+
+    // Send screen view to PostHog manually since we disabled autocapture
+    const postHogClient = getPosthogClientGlobal();
+    if (postHogClient) {
+      postHogClient.screen(currentRouteName);
```
rosvik, line 15: Could add this to test input, so we verify that Posthog works before the release.

android/build.gradle
```diff
@@ -2,18 +2,18 @@
...
+        RNMapboxMapsVersion = '11.13.4'
 
         // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
-        ndkVersion = "26.1.10909125"
-        kotlin_version = "1.9.25"
+        ndkVersion = "27.1.12297006"
+        kotlinVersion = "2.1.20"
         // Required for react-native-camera
         // https://github.com/teslamotors/react-native-camera-kit/blob/master/example/android/build.gradle
         kotlinVersion = "1.9.24"
```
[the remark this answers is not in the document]
cbrevik: Presumably fixed by [a8b3d73](https://github.com/AtB-AS/mittatb-app/pull/5419/commits/a8b3d73d7c6dafd02dfaa13385efa38ca9afa3e6)

rosvik (approved):
Nice work 💯

PR #5423, opened 2025-09-05
test: e2e performance measure improvements

tormoseng:
NOTE! Added a test-ids in a separate commit.

Updated the e2e tests used for measuring performance with `flashlight`. Placed the tests in separate files to get measures per file/context. Also updated and improved the map actions.


rosvik:
Screenshot 2025-09-25 at 12 27 45

I wanted to see the test results from the [test run](https://github.com/AtB-AS/mittatb-app/actions/runs/17944761863), but the artifacts show as expired, do you know why? 🤔

rosvik (approved):

PR #5424, opened 2025-09-05
refactor: rewrite axios client error handling

adriansberg:
Update the error handling mechanism to utilize the new `ErrorResponse` type instead of the previous `AxiosError`. This change improves consistency and clarity in error management across the application.

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -94,10 +94,7 @@ async function requestIdTokenHandler(config: InternalAxiosRequestConfig) {
 }
 
 function responseErrorHandler(error: AxiosError) {
-  const errorJson: any = error.toJSON();
-  // TODO: `error.status` returns undefined, but calling `toJSON` and indexind
-  // `status` works.
-  const status = errorJson['status'];
+  const status = error.response?.status ?? error.status;
 
   const errorResponse: ErrorResponse = getErrorResponse(error) ?? {
     http: {
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -1,7 +1,12 @@
 import axios, {AxiosError, InternalAxiosRequestConfig} from 'axios';
 import {v4 as uuid} from 'uuid';
 import {API_BASE_URL, APP_VERSION, IOS_BUNDLE_IDENTIFIER} from '@env';
-import {getAxiosErrorMetadata, getAxiosErrorType} from './utils';
+import {
+  ErrorResponse,
+  getAxiosErrorMetadata,
+  getAxiosErrorType,
+  getErrorResponse,
+} from './utils';
 import Bugsnag from '@bugsnag/react-native';
 import {
   AppIdentifierHeaderName,
@@ -23,12 +28,6 @@ import {
   getIdTokenValidityStatus,
 } from '@atb/modules/auth';
 
-type InternalUpstreamServerError = {
-  errorCode: 602;
-  shortNorwegian: string;
-  shortEnglish: string;
-};
-
 export const client = createClient(API_BASE_URL);
 
 const DEFAULT_TIMEOUT = 15000;
@@ -37,8 +36,6 @@ declare module 'axios' {
   export interface AxiosRequestConfig {
     // Use id token as bearer token in authorization header
     authWithIdToken?: boolean;
-    // Force refresh id token from firebase before request
-    forceRefreshIdToken?: boolean;
     // Whether the error logging to Bugsnag should be skipped for a given error
     skipErrorLogging?: (error: AxiosError) => boolean;
   }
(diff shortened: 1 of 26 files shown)


src/api/utils.ts
```diff
@@ -18,12 +18,14 @@ export const ErrorResponse = z.object({
...
 
-export type ErrorType =
-  | 'unknown'
-  | 'default'
-  | 'network-error'
-  | 'timeout'
-  | 'cancel';
+type ErrorType = 'unknown' | 'default' | 'network-error' | 'timeout' | 'cancel';
+
+export type AxiosErrorKind =
```
adriansberg, line 23: There were quite a few places that accessed the axios error type directly. Since we now intercept the error and make it an `ErrorResponse` it won't be possible any more. We do set the `kind` to a matching error type, so we can do something like this. 🤷

src/api/client.ts
```diff
@@ -97,61 +93,16 @@ async function requestIdTokenHandler(config: InternalAxiosRequestConfig) {
...
+function responseErrorHandler(
+  error: AxiosError,
+): Promise<ErrorResponse | HttpErrorResponse> {
+  const errorResponse = parseErrorResponse(error);
 
-function responseErrorHandler(error: AxiosError) {
-  if (shouldSkipLogging(error)) {
-    return Promise.reject(error);
+  if (!shouldSkipLogging(error)) {
+    notifyError(error);
```
rosvik, line 102: With this change does `notifyError` get called more often than before? Previously, the `'timeout'` and `'networkError'` case didn't call it.

src/api/utils.ts
```diff
@@ -9,21 +9,26 @@ const HttpError = z.object({
...
-  http: HttpError,
   kind: z.string(),
   message: z.string().nullish(),
   details: z.array(z.unknown()).nullish(),
 });
 export type ErrorResponse = z.infer<typeof ErrorResponse>;
+/** https://github.com/AtB-AS/amp-rs/blob/main/amp-http/src/lib.rs */
+export const HttpErrorResponse = ErrorResponse.extend({
+  http: HttpError,
+});
```
rosvik, line 21: I still think it would be nice if we could keep one of these types a 1:1 with the backend type. 

One solution could be to keep `ErrorResponse` on the same format as before, then maybe the name `RequestError` would make sense, which has the `http` field overwritten as undefined.

Not a blocker though. I'm ok with the current solution as well, and open for other suggestions 🤷

src/modules/auth/auth-utils.ts
```diff
@@ -41,14 +41,14 @@ export const authSignInWithPhoneNumber = async (
     await authenticateWithSms(phoneNumberWithPrefix, language);
     dispatch({type: 'SIGN_IN_INITIATED', phoneNumber: phoneNumberWithPrefix});
   } catch (e: any) {
-    const error = getAxiosErrorMetadata(e);
-    if (error.responseStatus === 400) {
-      const errorData = error.responseData && JSON.parse(error.responseData);
+    const error = e as HttpErrorResponse;
```
rosvik, line 44: We can't cast to `HttpErrorResponse`, right? Since the `http` field might be undefined.

src/modules/auth/auth-utils.ts
```diff
@@ -85,15 +85,15 @@ export const authConfirmCode = async (
     const token = await verifySms(phoneNumber, code);
     return await authSignInWithCustomToken(token);
   } catch (e) {
-    const error = getAxiosErrorMetadata(e as any);
-    const status = error.responseStatus;
+    const error = e as HttpErrorResponse;
```
rosvik, line 88: Same as above

src/modules/geocoder/use-geocoder.ts
```diff
@@ -37,9 +38,10 @@ export function useGeocoder(
             locations: response?.data?.map(mapFeatureToLocation),
           });
         } catch (err) {
-          if (!isCancel(err)) {
+          const error = err as ErrorResponse;
+          if (!(error.kind === 'AXIOS_CANCEL')) {
             console.warn(err);
-            dispatch({type: 'SET_ERROR', error: getAxiosErrorType(err)});
+            dispatch({type: 'SET_ERROR', error: error.kind as AxiosErrorKind});
```
rosvik, line 44: Is casting to `AxiosErrorKind` safe? `error.kind` can at least in theory have other values than the ones listen in `AxiosErrorKind`.

src/api/client.ts
adriansberg: Good question. Should we notify if there's a timeout or network error, for requests that shouldn't be skipped?

src/api/utils.ts
adriansberg: I like the idea of 1:1 as well, I can look at your suggestion. That would affect the comments below as well, the casting to `Http...`.

src/modules/geocoder/use-geocoder.ts
adriansberg: It is probably not safe, as if casting ever is 😅 Unsure how we should handle this though.

PR #5430, opened 2025-09-08
fix: Harbors overrides not working

strandlie:
This one took some time to figure out, but I will try to explain clearly here. 

@HanneLH added some harborConnection overrides in [this commit](https://github.com/AtB-AS/firestore-configuration/commit/2d0441a82072653c62ef36753254d7100fd026f3). I.e. Sula _(NSR:StopPlace:42924)_ --> Gjæsingen _(NSR:StopPlace:44074)_ was added there. But when we went to buy an express boat ticket in the app it still looked like this:

<details>
<summary>Screenshots before</summary>

| PurchaseSummaryScreen | PurchaseHarborSearchScreen |
|--------|-------|
|  |  |
</details>

The reason that they did not appear is that the `overrides` from `harborConnectionOverrides` in `firestore-configuration` were just IDs, and these IDs were then used to look up the actual harbor in the [applyOverrides](https://github.com/AtB-AS/mittatb-app/blob/e421a9a69cdd5c83b162c3c3302b79cbe4c2adb1/src/modules/harbors/use-harbors.tsx#L51) function. But when you have selected a `fromHarborId` the `{ data } = useHarborsQuery({fromHarborId})` gets refetched with only the harbors that are reachable from `fromHarborId`, and then the `applyOverrides` have no overlap between the overrides and the reachable harbors from `fromHarborId` which means that the [union](https://github.com/AtB-AS/mittatb-app/blob/e421a9a69cdd5c83b162c3c3302b79cbe4c2adb1/src/modules/harbors/use-harbors.tsx#L63) is empty. 

Long story short: we need to use `allHarbors` when we are mapping overrides, and not just the reachable harbors. Its not perfect using the `useHarborsQuery` twice, but React Query will cache the result so it will actually only be one result. 

<details>
<summary>Screenshots after</summary>

| PurchaseSummaryScreen | PurchaseHarborSearchScreen |
|--------|-------|
|  |  |
</details>

### DIFF
--- a/src/modules/harbors/use-harbors.tsx
+++ b/src/modules/harbors/use-harbors.tsx
@@ -10,9 +10,9 @@ import _ from 'lodash';
 import {StopPlaceFragmentWithIsFree} from './types';
 
 export const useHarbors = ({
-                             fromHarborId,
-                             transportModes,
-                           }: {
+  fromHarborId,
+  transportModes,
+}: {
   fromHarborId?: string;
   transportModes?: ProductTypeTransportModes[];
 } = {}) => {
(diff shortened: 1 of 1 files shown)


rosvik (approved):
Much easier to reason about now 🙌

PR #5432, opened 2025-09-09
feat: enable compact travel details map for screen reader users

rosvik:
Enables the compact travel details map for screen reader users. Already tested in https://github.com/AtB-AS/kundevendt/issues/21489#issuecomment-3269644306

closes https://github.com/AtB-AS/kundevendt/issues/21489


HanneLH (approved):
🌸

PR #5433, opened 2025-09-09
feat: update appearance settings page

rosvik:
As a suggestion from _markhor shaving hours_, it was proposed to change the two toggles on the appearance settings screen to three radio buttons instead. While working on this, I found the ThemeContext quite hard to understand, so I renamed some fields, and refactored it quite a bit to fit the new kind of selection 🤓

## Before / after 

<div>
Screenshot_1757418366
Screenshot_1757417856
</div>

## Acceptance criteria

- [ ] Color scheme selection works as you'd expect
- [ ] Android system font selection works like before
- [ ] Screen reader

### DIFF
--- a/src/translations/screens/subscreens/AppearanceSettings.ts
+++ b/src/translations/screens/subscreens/AppearanceSettings.ts
@@ -4,7 +4,7 @@ const AppearanceSettingsTexts = {
     title: _('Utseende', 'Appearance', 'Utsjånad'),
   },
   theme: {
-    heading: _('Fargetema', 'Color scheme', 'Fargetema'),
+    heading: _('Fargetema', 'Theme', 'Fargetema'),
     system: _('Automatisk', 'Automatic', 'Automatisk'),
     light: _('Lyst', 'Light', 'Lyst'),
     dark: _('Mørkt', 'Dark', 'Mørkt'),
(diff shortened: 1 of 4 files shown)


src/translations/screens/subscreens/AppearanceSettings.ts
```diff
@@ -3,14 +3,15 @@ const AppearanceSettingsTexts = {
...
-  actions: {
-    usePhoneTheme: _(
-      'Bruk fargetema fra telefonen',
-      "Use my phone's colour theme",
-      'Nytt fargetema frå mobilen min',
-    ),
-    darkMode: _('Mørk modus', 'Dark mode', 'Mørk modus'),
-    useSystemFont: _(
+  theme: {
+    heading: _('Fargetema', 'Color scheme', 'Fargetema'),
```
rosvik, line 7: @oyvindmikkelsenatb @Lovise-Git What's a good heading for this setting in English? Is "Color scheme" fine? I've also seen "Theme" and "Color mode" elsewhere.

adriansberg (approved):
I like this, very nice!

reidzeibel (approved):

src/translations/screens/subscreens/AppearanceSettings.ts
rosvik: I think I agree, and when there is no response from the designers, I believe the developer on vacation gets to decide! Theme it is 😄🚀

```suggestion
    heading: _('Fargetema', 'Theme', 'Fargetema'),
```

PR #5453, opened 2025-09-19
chore: Bump Mobile Token SDK, since Xcode 26 breaks the SDK

strandlie:
See Slack: https://mittatb.slack.com/archives/C03DTC05VRQ/p1758191551967599


rosvik (commented):
There's `@entur-private/abt-token-state-react-native-lib` in devDependencies, which should probably be bumped as well

strandlie:
> There's `@entur-private/abt-token-state-react-native-lib` in devDependencies, which should probably be bumped as well

Done!

tormoseng:
https://github.com/AtB-AS/mittatb-app/blob/f4e92f7b383c27a6ea17e24c5a604cd2e1803eb2/src/modules/mobile-token/hooks/use-validate-token.tsx#L18

PR #5454, opened 2025-09-22
chore: Upgrade minor and patch releases of deps

cbrevik:
To keep us up to date, early in the sprint, I took the liberty of bumping dependencies that should be low-risk to upgrade. I left out these dependencies, which require more work (basically any that bump a major version, or I know has strings attached):
image

`react-native-reanimated` cannot be upgraded to 4+, since it only works on new architecture

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,7 +193,7 @@
     "prettier": "^3.6.2",
     "react-dom": "19.1.0",
     "react-native-dotenv": "^3.4.11",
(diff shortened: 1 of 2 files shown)


rosvik (approved):
Sweet! I guess this is a part of https://github.com/AtB-AS/kundevendt/issues/21657 (mentioning it, to make it easier to find)

PR #5459, opened 2025-09-22
chore: upgrade kettle to v1.4.0

rosvik:
This version supports 16kb page size on android. It also bumps KettleKit to v1.5.1, which has some changes internally: https://developer.kogenta.com/docs/changelog/android

part of https://github.com/AtB-AS/kundevendt/issues/21713


adriansberg (approved):
After a little struggling with building the app through Android Studio, and having trouble with Kettle not initializing, I was able to build the app and check that stuff works as before.

PR #5462, opened 2025-09-23
refactor: use react-native-image-resizer for image compression

rosvik:
This library supports 16kb page size, and seems like a lighter library than react-native-compressor.

Build 1758637366 on iOS and 1758636235 on Android

part of https://github.com/AtB-AS/kundevendt/issues/21713

### DIFF
--- a/src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
+++ b/src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
@@ -48,13 +48,12 @@ export const Root_ParkingPhotoScreen = ({
   };
 
   const onConfirmImage = async (photo: PhotoFile) => {
-    const compressed = await compressImage(photo.path, 1024, 1024);
-    if (!compressed) return;
+    const compressedBlob = await compressImage(photo.path, 1024, 1024);
+    if (!compressedBlob) return;
 
     // Convert to Base64
-    const image = await fetch(compressed.path);
-    const imageBlob = await image.blob();
-    const base64Image = await blobToBase64(imageBlob);
+    const base64Image = await blobToBase64(compressedBlob);
+
     // Remove metadata
     const base64data = base64Image.split(',').pop();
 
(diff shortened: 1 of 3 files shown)


src/stacks-hierarchy/Root_ScooterHelp/Root_ParkingViolationsQrScreen.tsx
```diff
@@ -61,19 +61,19 @@ export const Root_ParkingViolationsQrScreen = ({
...
-    });
-    const image = await fetch(compressed);
-    const imageBlob = await image.blob();
-    const base64Image = await blobToBase64(imageBlob);
-    // Remove metadata, e.g. 'data:image/png;base64',
-    // and keep just the base64 encoded part of the image
-    // Nivel does not accept the metadata being a part of the image.
+    const compressedBlob = await compressImage(params.photo, 2048, 2048);
+    if (!compressedBlob) {
+      setIsError(true);
```
marius-at-atb, line 66: Just a note that we should use react query for state management instead of doing this "manually":
```ts
const [isLoading, setIsLoading] = useState(false);
  const [isError, setIsError] = useState(false);
```

src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
```diff
@@ -48,16 +48,12 @@ export const Root_ParkingPhotoScreen = ({
   };
 
   const onConfirmImage = async (photo: PhotoFile) => {
-    const compressed = await Image.compress(photo.path, {
-      maxHeight: 1024,
-      maxWidth: 1024,
-      quality: 0.7,
-    });
+    const compressedBlob = await compressImage(photo.path, 1024, 1024);
+    if (!compressedBlob) return;
```
marius-at-atb, line 52: Probably better to handle this with a default fallback image, so that it doesn't affect the user 🤔

marius-at-atb (approved):
Lgtm!

src/stacks-hierarchy/Root_ParkingPhotoScreen.tsx
marius-at-atb: Great!

PR #5464, opened 2025-09-24
feat: Add rule variables for zones in Announcements

cbrevik:
Partly fixes https://github.com/AtB-AS/kundevendt/issues/21627

Add support for car pooling zones to the configuration and refactor `useFindCityZoneInLocation` to `useFindZoneInLocation` so it can be reused for carpooling. I've also added a new hook for stable geolocation so we only recompute zones if location has a significant update.


src/modules/geolocation/GeolocationContext.tsx
```diff
@@ -371,6 +373,41 @@ export function useGeolocationContext() {
...
+ * NOTE: If the location is currently not available (undefined), the last used location is returned.
+ *
+ * @param {number} thresholdMeters - The threshold in meters to consider the location stable.
+ * @returns {GeoLocation | undefined} - The stable location.
+ */
+export function useStableLocation(
+  thresholdMeters: number = 150,
+): GeoLocation | undefined {
+  const {location} = useGeolocationContext();
+  const lastUsedLocationRef = useRef<GeoLocation | undefined>(undefined);
```
strandlie, line 389: This might be a stupid comment, but my mental model of refs are `use them for storing mutable values that are not used for rendering and do not need to cause re-renders when they change`, since React does not invalidate branches/trees if a ref changes. 

And I can imagine that a developer would expect a map to re-render whenever the location changes. So my question is: should this be a state instead of a ref? Or is this value never used for rendering?

strandlie: Or maybe that concern is handled some other way?

cbrevik: Oh, that is actually a good point. Yeah it should definitely be a state value instead, I didn't catch that in my testing but I haven't done much location changes in same session for that.

strandlie (approved):

PR #5470, opened 2025-09-24
Announcement warning info

marius-at-atb:
Added some more info to the warning message so that devs easily know where it comes from.


Dahly96 (approved):
lgtm

PR #5484, opened 2025-09-30
feat: use a11yRole radio for RadioSectionItem

rosvik:
- An issue we had in previous react native versions where `accessibilityRole="radio"` wasn't respected seems to have fixed itself in 0.81 🎉 
- Also noticed that the `hideSubtext` prop was unnecessary and caused issues with undefined in a11y labels, so factored that out.

done while looking at https://github.com/AtB-AS/kundevendt/issues/19430


HanneLH (approved):
🌸

PR #5488, opened 2025-10-01
chore: Upgrade Intercom to latest

cbrevik:
As per: https://github.com/AtB-AS/kundevendt/issues/21657

- [x] Testing in staging first before opening for review

Verified that it works with new version, can chat and get unread dot:
| Dot | Chat |
|--------|--------|
| image | image |


rosvik (approved):
LGTM 👍

PR #5491, opened 2025-10-01
feat: add RNBarcodeInspectionView mobile token aztec code

rosvik:
Draft for now, need to figure out if the beginInspection call is properly done.

closes https://github.com/AtB-AS/kundevendt/issues/21653

### DIFF
--- a/src/modules/fare-contracts/details/Barcode.tsx
+++ b/src/modules/fare-contracts/details/Barcode.tsx
@@ -20,6 +20,13 @@ import {SvgXml} from 'react-native-svg';
 import {GenericSectionItem} from '@atb/components/sections';
 import {useGetSignedTokenQuery} from '@atb/modules/mobile-token';
 import {useRemoteConfigContext} from '@atb/modules/remote-config';
+import {
+  beginInspection,
+  endInspection,
+  RNBarcodeInspectionView,
+} from '@entur-private/abt-token-state-react-native-lib';
+import {CONTEXT_ID} from '@atb/modules/mobile-token';
+import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
 
 type Props = {
   validityStatus: ValidityStatus;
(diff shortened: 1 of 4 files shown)


src/modules/fare-contracts/details/Barcode.tsx
```diff
@@ -80,9 +92,53 @@ function useScreenBrightnessIncrease() {
...
+        // Prepare data for RNBarcodeInspectionView
+        await beginInspection(CONTEXT_ID, {
+          visualInspectionNonces: undefined,
+          includeCertificate: false,
+          deviceDetails: {
+            application: false,
+            os: false,
+            network: false,
+            bluetooth: true,
+          },
```
mikaelbr, line 115: I guess it is easiest to just check what Entur does 🤷 Hard to speculate in the effect of each setting. But really nice to see the different options I think

src/modules/fare-contracts/details/Barcode.tsx
```diff
@@ -80,9 +92,53 @@ function useScreenBrightnessIncrease() {
...
+    aztec_code_size_in_cm,
+    aztec_code_padding,
+    enable_new_token_barcode_base64,
+  } = useRemoteConfigContext();
+
+  useEffect(() => {
+    (async () => {
+      try {
+        // Prepare data for RNBarcodeInspectionView
+        await beginInspection(CONTEXT_ID, {
```
mikaelbr, line 107: Somewhat nitpicky, but kind of waste doing SIAF when you really need to await. I guess try/catching, but I would argue that it is kind of easier to read `.catch()` from the promise it self without the syntax churn

src/modules/fare-contracts/details/Barcode.tsx
```diff
@@ -80,9 +92,53 @@ function useScreenBrightnessIncrease() {
...
+          visualInspectionNonces: undefined,
```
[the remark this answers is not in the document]
rosvik: I assume so! From the example app, there seems to be a bit of logic needed on the backend, but don't know how big of a task that is to implement. There's a `RNVisualInspectionView` that seems to work the exact same way as `RNBarcodeInspectionView`, so I assume it's as easy as that to add once we have the nonces.

strandlie (approved):

PR #5518, opened 2025-10-17
feat: improved map camera movement2

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21944

This issue is to fix so that the icon will center only on open, and then the user can move freely and adjust the bottomsheet without any other action happening. The only other time it will autocenter is in follomode on active trip. When a user leave follow mode when in an active trip and moves freely, there should be no centering when the bottomsheet is adjusted

### DIFF
--- a/src/modules/map/MapV2.tsx
+++ b/src/modules/map/MapV2.tsx
@@ -10,7 +10,7 @@ import {
   Camera,
   MapView,
 } from '@rnmapbox/maps';
-import {Feature, Point, GeoJsonProperties} from 'geojson';
+import {Feature} from 'geojson';
 import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {View} from 'react-native';
 import {MapCameraConfig, getSlightlyRaisedMapPadding} from './MapConfig';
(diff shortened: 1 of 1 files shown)


src/modules/map/MapV2.tsx
```diff
@@ -114,7 +111,33 @@ export const MapV2 = (props: MapProps) => {
...
+
+  useEffect(() => {
+    setPreviousFeature(previousMapState?.feature);
+  }, [previousMapState]);
+
+  useEffect(() => {
+    if (followUserLocation) {
+      setStalePaddingBottomMap(paddingBottomMap);
+    }
+  }, [followUserLocation, paddingBottomMap]);
```
strandlie: I see a lot of `previous`, `stale` `stablePrevious` etc in this file now. Can we document why we need that, and if there are gotcha's or something else to know about?

src/modules/map/MapV2.tsx
```diff
@@ -256,46 +330,77 @@ export const MapV2 = (props: MapProps) => {
...
-    [activeShmoBooking?.state, paddingBottomMap, dispatchMapState],
+    [
+      activeShmoBooking?.state,
+      paddingBottomMap,
+      mapState.bottomSheetType,
+      dispatchMapState,
+    ],
   );
 
+  const handleBreakFollow = useCallback(() => {
```
marius-at-atb: Could be a good idea to add a comment here about why it is like this

PR #5523, opened 2025-10-17
fix: fixed minimized height issue on different devices

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21885

This solves the issue that the minimized view of the bottomsheet could look different on different devices and text sizes

Før:

image

Etter:

image

### DIFF
--- a/src/components/bottom-sheet-map/MapBottomSheet.tsx
+++ b/src/components/bottom-sheet-map/MapBottomSheet.tsx
@@ -174,12 +174,14 @@ export const MapBottomSheet = ({
 
     const prevSnapPoints = snapPoints ? snapPoints : [];
 
-    const handleV =
+    const handleVerticalSize =
       toNum(styles.handleIndicatorStyle?.height) +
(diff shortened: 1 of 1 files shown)


src/components/bottom-sheet-map/MapBottomSheet.tsx
```diff
@@ -124,51 +127,78 @@ export const MapBottomSheet = ({
...
       )
     );
   };
 
+  const computedSnapPoints = useMemo(() => {
+    if (!canMinimize) return snapPoints;
+
+    const prevSnapPoints = snapPoints ? snapPoints : [];
+
+    const handleV =
```
marius-at-atb: What does the `V` stand for? Should be spelled out when it's hard to guess.

Dahly96: its vertical, and handle is the drag handle, so handleVertical, the size vertical for the drag handle

PR #5532, opened 2025-10-23
feat: create EventStreamContext and add event log to debug menu

rosvik:
As part of https://github.com/AtB-AS/kundevendt/issues/21648, and towards making it easier to debug the event stream, this adds a event stream log section to the debug menu. To achieve this, we needed an EventStreamContext to make state from the event stream available to the rest of the app. This context will likely be useful in the future, when events contain data that are used for other tasks than invalidating queries.

Simulator Screenshot - iPhone 16 Pro - 2025-10-23 at 15 38 00

## Acceptance criteria

- [x] Fare contract events are shown when purchasing a ticket
- [x] Open events are shown in the log on app open or after disconnects.
- [x] Errors shows in the log on disconnects


tormoseng:
✅

PR #5533, opened 2025-10-24
fix: use popTo when navigating backwards

rosvik:
This fixes issues where actions that should go backwards in the stack instead added the next screen on top

- Completing a ticket purchase added a Root screen on top
- Clicking fare contract benefits adds a Root screen on top
- Picking a fare zone when searching by text navigates to a new zone map
- The close button when adding favorites departures navigates forwards

Likely caused by the react-navigation v7 upgrade: https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-goes-back-use-popto-instead

Ref. slack discussion https://mittatb.slack.com/archives/C0116FMPX4Y/p1761217564887349

## Acceptance criteria

- [ ] Functionality mentioned above works as expected ☝️ 
- [ ] Should verify that navigation in obscure places in the app still works as expected.


hanne-at-atb:
Looks good ✅

PR #5545, opened 2025-10-28
feat: add icons to geofence snackbar

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/18949

Adds icons on snackbar for geofencing zones and updates texts


marius-at-atb (approved):
Looks good!

PR #5562, opened 2025-11-05
refactor: Remove MBXAccessToken and MAPBOX_PUBLIC_TOKEN

marius-at-atb:
Removing MAPBOX_PUBLIC_TOKEN since it is currently not needed and only used in the AtB app.
AFAICT it is only needed when also using OfflineManager from Mapbox.


tormoseng:
@marius-at-atb Any reason for not removing `MAPBOX_PUBLIC_TOKEN` from `env/atb/dev/.env` ?

tormoseng:
✅  By the way, the map works as before during testing in Staging (https://github.com/AtB-AS/mittatb-app/actions/runs/19096863133).

marius-at-atb:
@tormoseng Good catch! No good reason - updated.

rosvik (approved):

PR #5569, opened 2025-11-06
feat: support Announcement sort by index

marius-at-atb:
Adds control to the order of Announcements.

Updated docs: https://github.com/AtB-AS/docs-private/pull/85


tormoseng:
✅  Nice! Tested OK!

PR #5570, opened 2025-11-07
feat: Add bike ticket purchase flow

strandlie:
Closes https://github.com/AtB-AS/kundevendt/issues/4282

Adds a complete bike ticket purchase flow where you can select a bike on an equal footing with user profiles in the bottom sheet for travellers. Also introduces support for "baggage products" as a specific type of supplement product.

## Changes
* Support baggage products with counts and offer on an equal footing with user profiles in `use-offer-state.ts`. 
* Rewrite how to calculate the price total in `use-offer-state.ts`.
* :warning: Add own subtype `BaggageTicketOffer` Zod-schema, based on the `TicketOffer`-schema. This new schema assumes that there is **only one** supplement product in the `supplementProducts`-array, and the `fareProduct` (ref) is null/undefined, if it is a baggage-type offer. This is something we will have to change in the future when we want to support more complex offers. Backend are aware of this, and will ensure that this assumption/requirement is held until we decide to rewrite the ticket data structure. 
* Introduced a new `BaggageProduct` type (a subtype of `SupplementProduct` with specific properties) and corresponding runtime validation using zod in `src/modules/configuration/types.ts`. Use this type to `safeParse` baggage products in `getBaggageProducts` and `use-offer-state` and friends.
* Added  `baggageType` on `SupplementProduct` which is used in the OfferSearch-request.
* Add `supplementProducts`  on `TicketOffer`.
* Make `fareProduct` on `TicketOffer` nullable (it will always be `null` on baggage-type offers).

## Test changes
* Added comprehensive tests to ensure correct handling of baggage products in purchase selections, including validation against product limitations and allowing the bicycle ticket to be selected by itself.
* Had to replace `.toBe()` with `.toStrictEqual()` for deep object comparison in several tests.

Dependent on:
- https://github.com/AtB-AS/config-specs/pull/147 ✅ 
- https://github.com/AtB-AS/utils/pull/45 ✅ 
- https://github.com/AtB-AS/product/pull/28 ✅ 
- https://github.com/AtB-AS/firestore-configuration/pull/919 ✅

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
+++ b/src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
@@ -301,6 +301,7 @@ export function useOfferState(
           .map((sp) => ({
             id: sp.baggageType,
             // Must be a valid user profile ref, and we are unsure if Entur supports 'ANYONE' yet for baggage products.
+            // Doesn't actually have an affect on the purchased product.
             userType: 'ADULT',
             baggageTypes: [sp.baggageType],
             count: sp.count,


src/components/sections/items/CounterSectionItem.tsx
```diff
@@ -173,3 +178,12 @@ const useStyles = StyleSheet.createThemeHook((theme) => {
     },
   };
 });
+
+function mapIllustrationNameToSvg(name?: string) {
+  switch (name) {
+    case 'BicycleFill':
```
rosvik: I guess this one fell through the cracks while Håkon was away: https://github.com/AtB-AS/firestore-configuration/pull/908#discussion_r2494833104 

But now that we have
```
"baggageType": "BICYCLE",
"illustration": "BicycleFill"
```

Is the `illustration` field even needed, or can we map `baggageType` to an SVG instead?

tormoseng:
> Triggered a build yesterday, so build 1763997892 is testable with bike ticket purchase. Cc @tormoseng et al

Seems to be working fine! But, as we talked about @cbrevik , only the 1 adult part of the single ticket is shown under the "Repeat purchase" section in the app. And the single bike doesn't show at all. Maybe we need to talk to @Sebstorvik about how we like this to be presented (do we want bike tickets to appear or not).

src/stacks-hierarchy/Root_PurchaseOverviewScreen/use-offer-state.ts
rosvik: The pricing calculations are getting hard to reason about, and it's important to get correct. Is it time to extract it out of use-offer-state and add some tests?

src/modules/purchase-selection/use-selectable-baggage-products.ts
rosvik: Could this re-use `isSelectableSupplementProduct`? Also, since it's only used once, could it just as well be moved closer to where it's used?

src/modules/purchase-selection/purchase-selection-builder.ts
```diff
@@ -100,18 +106,11 @@ const createBuilder = (
...
-        currentSelection = {
-          ...currentSelection,
-          userProfilesWithCount: onlyWithActualCount,
-        };
-      }
+      buildState.userProfilesWithCount = userProfilesWithCount;
+      return builder;
+    },
+    baggageProducts: (baggageProductsWithCount) => {
+      buildState.baggageProductsWithCount = baggageProductsWithCount;
```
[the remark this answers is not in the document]
cbrevik: There are more things that needs to be fixed in the builder, but this was sort of a minimum change that at least does not introduce more inconsistent state 😅

src/stacks-hierarchy/Root_PurchaseOverviewScreen/components/Travellers/use-baggage-count-state.ts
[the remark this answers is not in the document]
rosvik: Now I want one of those robot lawnmowers, except it's a remote controlled upside down helicopter.

PR #5580, opened 2025-11-13
feat: add icons to geofencing zone in map

jorunnl:
Closes https://github.com/AtB-AS/kundevendt/issues/18745

Adds icons to geofencing zones for e-scooters in map

Had some discussion with @Aliaaen about size of icons and zoomlevel, landed on the current solution.

Uses the polylabel algorithm to calculate center of each polygon to place the icon.
It mostly works well, however it can look a bit off for very small zones.
Had to exclude very large zones (like Ryde's noEntry-zone covering the entire planet)  from getting icons to reduce processing time to an acceptable amount.

image
image

### DIFF
--- a/src/modules/map/components/mobility/GeofencingZones.tsx
+++ b/src/modules/map/components/mobility/GeofencingZones.tsx
@@ -43,7 +43,7 @@ const GeofencingZonesForVehicle = ({
 
   return (
     <>
-      {data?.iconFeatures?.map((iconFeatureCollection) => (
+      {data?.iconFeatureCollections?.map((iconFeatureCollection) => (
         <GeofencingZoneIcon
           iconFeatureCollection={iconFeatureCollection}
           key={`iconGeofencingZone_${iconFeatureCollection.renderKey}`}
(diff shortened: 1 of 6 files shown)


src/modules/map/components/mobility/GeofencingZones.tsx
```diff
@@ -112,3 +117,44 @@ const GeofencingZone = ({geofencingZone}: GeofencingZoneProps) => {
...
+          iconImage: [
+            'concat',
+            'geofencingzone_',
+            lowerCaseCode,
+            '_',
+            themeName,
+          ],
+          iconSize: 0.85,
+        }}
+        aboveLayerID={MapSlotLayerId.GeofencingZones}
```
marius-at-atb: I think we should either handle onPress or move the icons behind other zones too

marius-at-atb (approved):
Great!

A couple of questions for @Aliaaen and @lineamb (which are non-blocking for this PR):
- Should we add shadows to these icons as well?
- Isn't it strange that you can currently tap e.g. a noParking icon, but end up with a noEntry explanation?

PR #5586, opened 2025-11-17
feat: update header buttons

rosvik:
_Makes sense to review one commit at a time!_

- Moves all header "Cancel" and "Close" buttons to the right
- Adds an icon to every header button. No longer an optional setting.
- Makes some screens slide in from the right instead of bottom. Figma is updated with rules for this: https://www.figma.com/design/2QTjAdekdIPuLFovQhVrY3/Components?node-id=14245-7215&t=Kn1ijhrPjLlJ8NlB-0
- Updates the "sort favorites" screen (Profile > Favorites > Favorite locations > Re-arrange favorites) background color and removes the submit/save button.

closes https://github.com/AtB-AS/kundevendt/issues/20607

### DIFF
--- a/src/translations/screens/FareContract.ts
+++ b/src/translations/screens/FareContract.ts
@@ -70,7 +70,7 @@ const FareContractTexts = {
       infoButtonA11yHint: _(
         'Aktivér for å gå til billettinformasjon',
         'Activate to go to ticket information',
-        'Aktivér for å gå til billetinformasjon',
+        'Aktivér for å gå til billettinformasjon',
       ),
     },
     paymentMethod: _(


src/translations/screens/FareContract.ts
```diff
@@ -66,6 +66,12 @@ const FareContractTexts = {
   details: {
     header: {
       title: _('Billettdetaljer', 'Ticket details', 'Billettdetaljar'),
+      ticketInformation: _('Om', 'About', 'Om'),
+      infoButtonA11yHint: _(
+        'Aktivér for å gå til billettinformasjon',
+        'Activate to go to ticket information',
+        'Aktivér for å gå til billetinformasjon',
```
marius-at-atb:
```suggestion
        'Aktivér for å gå til billettinformasjon',
```

PR #5592, opened 2025-11-19
feat: add shmo booking stream event handling

adriansberg:
Implement handling for Shmo booking update events in the event stream, allowing the application to invalidate queries related to active Shmo bookings. Instead of setting a `refetchInterval` to poll the update of the price we now invalidate the active booking queries when an event of type `SHMO_BOOKING_UPDATED` appears in the event log.

related https://github.com/AtB-AS/kundevendt/issues/21932

### DIFF
--- a/src/modules/event-stream/handle-stream-event.ts
+++ b/src/modules/event-stream/handle-stream-event.ts
@@ -4,6 +4,7 @@ import {fareContractsQueryKey} from '../ticketing/use-fare-contracts';
 import {getBonusAmountEarnedQueryKey} from '../bonus';
 import {getActiveShmoBookingQueryKey} from '../mobility/queries/use-active-shmo-booking-query';
 import {getShmoBookingQueryKey} from '../mobility/queries/use-shmo-booking-query';
+import {languageGlobal} from '../locale';
 
 export const handleStreamEvent = (
   streamEvent: StreamEvent,
@@ -27,12 +28,12 @@ export const handleStreamEvent = (
         });
       }
       break;
-    case EventKind.ShmoBookingUpdate:
+    case EventKind.ShmoBookingUpdated:
       queryClient.invalidateQueries({
-        queryKey: getActiveShmoBookingQueryKey(),
+        queryKey: getActiveShmoBookingQueryKey(languageGlobal),
       });
       queryClient.invalidateQueries({
-        queryKey: getShmoBookingQueryKey(streamEvent.bookingId),
+        queryKey: getShmoBookingQueryKey(streamEvent.bookingId, languageGlobal),
       });
       break;
   }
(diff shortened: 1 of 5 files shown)


src/modules/event-stream/handle-stream-event.ts
```diff
@@ -25,5 +27,10 @@ export const handleStreamEvent = (
         });
       }
       break;
+    case EventKind.ShmoBookingUpdate:
+      queryClient.invalidateQueries({
+        queryKey: getActiveShmoBookingQueryKey(acceptLanguage),
+      });
```
rosvik: Do we really need to pass a language here? Invalidating just `"GET_ACTIVE_SHMO_BOOKING_QUERY_KEY"` should invalidate queries using both languages.

adriansberg: Ah, probably not? I saw that `acceptLanguage` is required in the query key, and assumed that it couldn't be undefined when used to invalidate either. Pushed a change that makes it optional now, and will test that some.

rosvik (approved):
Looks good to me 👍

src/modules/mobility/queries/use-active-shmo-booking-query.tsx
```diff
@@ -4,14 +4,12 @@ import {ONE_MINUTE_MS} from '@atb/utils/durations';
...
 import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
 
-export const getActiveShmoBookingQueryKey = (acceptLanguage: string) => [
+export const getActiveShmoBookingQueryKey = (acceptLanguage?: string) => [
   'GET_ACTIVE_SHMO_BOOKING_QUERY_KEY',
   acceptLanguage,
 ];
 
-export const useActiveShmoBookingQuery = (
-  refetchInterval: number | false = false,
```
marius-at-atb: We should still have refetch interval for when the stream svc feature flag is off.
It may even be a good idea to have a long refetchInterval when it's on, just in case we have issues with the reliability of the events.
So for now, I think we should set refetch interval same as before when enable_event_stream is false, and e.g. 90 sec when true.

marius-at-atb: Both ok IMO. Of course would be nice to define it only in one place.
I suppose the `refetchInterval` param could be replaced by e.g. a `shouldRefetchAtIntervals` boolean. Then you could set the refetchInterval value inside the query.

marius-at-atb:
@Dahly96 Excellent question! Yes, it could be that we should do that. Was thinking to talk about it with the Entur team first to be sure about the content of the events, hopefully in tomorrow's meeting.

marius-at-atb:
Btw that's not a blocker for this PR. In the first round of event handling, it's simplest and safest to start with just ids and invalidation. So if we go for full event data, it should be added later anyway.

adriansberg:
@marius-at-atb 100% agree, and like my previous comment we all should be aligned for how we proceed with a shift like that. 😀

PR #5599, opened 2025-11-21
feat: add travel aid onboarding screen

rosvik:
closes https://github.com/AtB-AS/kundevendt/issues/21984

### DIFF
--- a/src/stacks-hierarchy/Root_TravelAidOnboardingScreen.tsx
+++ b/src/stacks-hierarchy/Root_TravelAidOnboardingScreen.tsx
@@ -14,12 +14,12 @@ export const Root_TravelAidOnboardingScreen = () => {
   const {setPreference} = usePreferencesContext();
   const {continueFromOnboardingSection} = useOnboardingNavigation();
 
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TravelAidOnboardingScreen.tsx
```diff
@@ -0,0 +1,45 @@
...
+        rightIcon: {svg: Confirm},
+        expanded: true,
+      }}
+      secondaryFooterButton={{
+        onPress: onDismiss,
+        text: t(TravelAidTexts.onboarding.dismiss),
+        rightIcon: {svg: Close},
+        expanded: true,
+      }}
+      testID="locationWhenInUsePermission"
```
marius-at-atb: I guess testID should be called something with travelAid instead

rosvik:
> Btw we often have illustrations as well on onboarding screens, might want to add that if we have a suitable one

This screen is specifically for people with voiceover and large text sizes, so an illustration will either not be seen, or take up very limited screen space 😬

PR #5601, opened 2025-11-24
feat: Add GlobalMessage in Profile

marius-at-atb:
simulator_screenshot_F6B592D3-D8F6-43F6-8D86-29BEC36FADD7

Adding a GlobalMessage to Profile since it might be useful in the future.

E.g. for the new SPAR feature, it would have been nice to inform users on old app versions that they have to update the app to access it.

No rule variables added for now, but e.g. `isLoggedIn` could be relevant here? Let me know if you have any thoughts!

Tip for testing:
- Can use the existing profile-test GlobalMessage in Firestore in staging

Acceptance Criteria:
- [x] Normal GlobalMessage with context `app-profile` in the profile screen
- [x] The profile screen should look the same as before when no GlobalMessage is active

Test input
- [x] Light and dark mode
- [x] 200% zoom 
- [x] Screenreader
- [x] Language


rosvik:
@Sebstorvik @marius-at-atb `authenticationType` is available as a rule variable by default for all global messages. (Which is set to `'none' | 'anonymous' | 'phone'`)

rosvik (approved):

HanneLH:
@marius-at-atb 
Remember to updated the document regarding Global messages ▶️ https://github.com/AtB-AS/docs-private/blob/main/global-messages.md

PR #5608, opened 2025-11-27
refactor: use VehicleFragment type for all vehicles-v2 data

rosvik:
Related to https://github.com/AtB-AS/atb-bff/pull/420 and https://github.com/AtB-AS/atb-bff/pull/418.

- The types in `fragments/vehicles.ts` was actually mobility types, so I moved them there.
- For vehicles data (from Enturs vehicles-v2 service) we really only need the `VehicleFragment` type, so I cleand that up and also updated some naming of variables, since it was quite inconsistent.

### Acceptance criteria

- [ ] Bus in map should work as before


marius-at-atb (approved):
Lgtm!

Something we may want to address at some point, is that Entur sends us vehicles from both vehicles and mobility. These are named the same, but are different.
After this PR, they will be even less differentiated 🤔

PR #5612, opened 2025-11-27
refactor: use inapp browser for external links

reidzeibel:
closes: https://github.com/AtB-AS/kundevendt/issues/18945

Changed most of the `Linking.openURL()` calls to `openInAppBrowser()`. There are still some `Linking.openURL` which we use to:
- call phone number
- redirect to vipps
- redirect to deeplinks
- redirect to app store/play store

Otherwise, if it opens a web browser, it should use the `openInAppBrowser()` instead. 

I have listed most of the URLs in the app in the original issue.

### DIFF
--- a/src/modules/bonus/onboarding/BonusOnboarding_Download.tsx
+++ b/src/modules/bonus/onboarding/BonusOnboarding_Download.tsx
@@ -10,11 +10,10 @@ import {Confirm} from '@atb/assets/svg/mono-icons/actions';
 import {BrandingImage, findOperatorBrandImageUrl} from '@atb/modules/mobility';
 import {useFirestoreConfigurationContext} from '@atb/modules/configuration';
 import {ThemeText} from '@atb/components/text';
-import {Platform, View} from 'react-native';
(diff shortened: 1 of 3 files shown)


marius-at-atb:
Actually, I think this solves https://github.com/AtB-AS/kundevendt/issues/22097 as well!

reidzeibel:
Links documentation here: https://github.com/AtB-AS/docs-private/pull/88

src/screen-components/travel-details-screens/components/TripSection.tsx
```diff
@@ -532,7 +533,7 @@ const AuthorityRow = ({id, name, url}: AuthorityFragment) => {
             TripDetailsTexts.trip.leg.buyTicketFromA11yLabel(name),
           )}
           leftIcon={{svg: ExternalLink}}
-          onPress={() => url && Linking.openURL(url)}
+          onPress={() => url && openInAppBrowser(url, 'close')}
```
rosvik: Wasn't external providers supposed to be external browser?

PR #5614, opened 2025-11-28
feat: added new bottomsheet for RefundBottomSheet

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/21749

Added new bottomsheet for RefundBottomSheet

image


src/modules/fare-contracts/components/RefundSectionItem.tsx
rosvik: Sections in the app are set up to work like this:

```
<Section>
	<SectionItem>
	<SectionItem>
	<SectionItem>
</Section>
```

Where the `<Section>` component applies border radius and dividers between SectionItems. This change will break this system, because the new structure becomes 


```
<Section>
	<SectionItem>
	<SectionItem>
	<>
		<SectionItem>
		<BottomSheet>
	</>
</Section>
```

Not quite sure how to fix that though 🤔

PR #5618, opened 2025-12-01
fix: Show recent ticket with bikes (not standalone)

cbrevik:
Followup on https://github.com/AtB-AS/kundevendt/issues/4282#issuecomment-3588455967

Fixes so bicycles shows up on recent tickets if bought alongside a single ticket. Standalone bike-tickets are not shown, since they are not a "sellable" fare product in the app. We could hack it so it is hardcoded to use single ticket in that case, but we've decided to not support it for now. CC @Sebstorvik @tormoseng 

image


tormoseng:
✅ Tested OK!

PR #5623, opened 2025-12-02
chore: remove some unused code

rosvik:
I ran `yarn unimported`, and removed some of the files that isn't imported anywhere.

There's quite a bit of shared mobility code here: ShmoTesting, Bicycles, Scooters etc. Are there reasons to keep some of it?

### DIFF
--- a/src/modules/map/components/mobility/ShmoTesting.tsx
+++ b/src/modules/map/components/mobility/ShmoTesting.tsx
@@ -1,0 +1,334 @@
+import React, {useEffect} from 'react';
+import {
+  InitShmoOneStopBookingRequestBody,
+  ShmoBookingEvent,
+  ShmoBookingEventType,
+} from '@atb/api/types/mobility';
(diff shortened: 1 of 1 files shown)


src/modules/map/components/mobility/ShmoTesting.tsx
```diff
@@ -1,334 +0,0 @@
...
-import {useBottomSheetContext} from '@atb/components/bottom-sheet';
-import {ToggleSectionItem} from '@atb/components/sections';
-
-type ShmoTestingProps = {
-  selectedVehicleId?: string;
-  showSelectedFeature: boolean;
-  setShowSelectedFeature: (showSelectedFeature: boolean) => void;
-};
-
-export const ShmoTesting = ({
```
marius-at-atb: We should keep this file, as it's purely for testing purposes, and might be useful again for further integrations

marius-at-atb (approved):
Nice!

PR #5624, opened 2025-12-02
fix: Remove title on Traveller-selection when multiple

cbrevik:
Fixes: https://github.com/AtB-AS/kundevendt/issues/22327

Ref:
image

image


marius-at-atb (approved):
Lgtm!

PR #5627, opened 2025-12-02
refactor: Nearest stop places with tanstack query

marius-at-atb:
Resolves https://github.com/AtB-AS/kundevendt/issues/20116

This is a refactor of how we use the nearby stop places endpoint in the app, now using tanstack query instead, and deleting the old hook.

Acceptance criteria:
- [x] Calls to `/bff/v2/departures/stops-nearest` should be called when you see the NearbyStopPlacesScreenComponent (3 places -> Departures + Favourite edits in both Assistant and Profile, see screenshots)
   - [x] On entry
   - [x] If you move the selected location at least 75m
   - [x] If the data is >1h old
- [x] Also, ensure that navigation to other screens is correct from this screen regarding location
   
Places in the app:

### DIFF
--- a/src/api/types/generated/NearestStopPlacesQuery.ts
+++ b/src/api/types/generated/NearestStopPlacesQuery.ts
@@ -1,13 +1,14 @@
 import * as Types from '@atb/api/types/generated/journey_planner_v3_types';
 import type {SituationFragment} from '@atb/api/types/generated/fragments/situations';
 
-export type NearestStopPlacesQueryVariables = Types.Exact<{
-  count?: Types.InputMaybe<Types.Scalars['Int']['input']>;
-  distance: Types.Scalars['Float']['input'];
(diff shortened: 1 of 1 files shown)


src/api/types/generated/NearestStopPlacesQuery.ts
```diff
@@ -1,6 +1,14 @@
 import * as Types from '@atb/api/types/generated/journey_planner_v3_types';
 import type {SituationFragment} from '@atb/api/types/generated/fragments/situations';
 
+export type NearestStopPlacesQueryVariables = Types.Exact<{
```
marius-at-atb: This is copied from the BFF, replacing `StopsNearestQuery`

rosvik: Maybe use `number` and `string` here?

PR #5636, opened 2025-12-04
fix: don't persist nearest stops

marius-at-atb:
Resolves https://github.com/AtB-AS/kundevendt/issues/20116, follow up to https://github.com/AtB-AS/mittatb-app/pull/5627

Since the location will be unique for each app open, we skip storing the data.


rosvik (approved):

tormoseng:
✅

PR #5638, opened 2025-12-04
feat: shmo testing toggle

marius-at-atb:
This makes it possible to turn on `ShmoTesting` without editing code.
Also since `ShmoTesting` is now referenced, it is no longer marked as "dead code", ref https://github.com/AtB-AS/mittatb-app/pull/5623#discussion_r2588930329.


rosvik (approved):
Nice!

tormoseng:
✅

PR #5712, opened 2026-01-16
fix: ValidityLine without LinearGradient and using Reanimated instead

cbrevik:
Sort of a "response" to myself in this PR: https://github.com/AtB-AS/mittatb-app/pull/5710

We don't use any linear gradient colors here, actually. And `ValidityLine` is the only place where `react-native-linear-gradient` is in use. So I thought, how difficult is it to do this with just a `View`? Answer is: pretty easy to do a prototype of at least.

So interesting to see if this works well across devices, if so we could remove a library, which I look at as a 🎉 

Also refactoring it to use Reanimated instead, so a deadlock with navigation does not occur, ref: https://app.bugsnag.com/atb-as/mittatb/errors/6968cbe74ef4ae6af7de9452?filters%5Bapp.release_stage%5D=staging&filters%5Bevent.since%5D=1h

Also see https://github.com/facebook/react-native/issues/53128#issuecomment-3766913922

## Here is this PR (just a skewed View)
https://github.com/user-attachments/assets/8568ace5-5500-4986-9b48-0ce3514844d9

## Here is with LinearGradient (old way)
https://github.com/user-attachments/assets/f06dd9c9-398e-436d-a25f-7a0345b33849

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,6 @@
     "react-native-in-app-review": "4.4.2",
     "react-native-inappbrowser-reborn": "^3.7.0",
     "react-native-kettle-module": "2.0.0",
-    "react-native-linear-gradient": "^2.8.3",
     "react-native-localize": "^3.2.1",
     "react-native-pager-view": "^6.3.3",
     "react-native-permissions": "^5.4.2",
--- a/src/components/line-with-vertical-bars/LineWithVerticalBars.tsx
+++ b/src/components/line-with-vertical-bars/LineWithVerticalBars.tsx
@@ -8,10 +8,9 @@ import {
   Dimensions,
   View,
 } from 'react-native';
-import LinearGradient from 'react-native-linear-gradient';
 
 const SPACE_BETWEEN_VERTICAL_LINES = 72;
-const AnimatedLinearGradient = Animated.createAnimatedComponent(LinearGradient);
+const AnimatedVerticalLineSegment = Animated.createAnimatedComponent(View);
 
 export const LineWithVerticalBars = ({
   backgroundColor,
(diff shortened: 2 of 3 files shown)


reidzeibel (approved):
These are actually 🔥

rosvik (approved):
Nice!

rosvik (approved):

reidzeibel (approved):

michaelknoch:
hey @cbrevik, just found your PR here while investigating the animation deadlock. Did moving to reanimated reliably fix the crash?

cbrevik:
Hi there @michaelknoch! This PR did not fix it at all. 

We attempted a LOT of different things, and to make a pretty long story short, the real underlying issue was that we were using a dependency that did main queue/thread setup when it was initialised the first time.

Basically, when we "animated" from screen A to screen B, then screen B was showing a UI element which a dependency/library was rendering. And since this UI element had not been shown before, the native module that underpins it was being initialised for the first time. And therein lies the problem, because that caused a race condition with the nodes/screens being animated. 

Basically the same thing that is described in this issue from the Mapbox dependency: https://github.com/rnmapbox/maps/pull/3954

(Mapbox was not the issue for us, another internal SDK/dependency was).

So if I were going to suggest something, it is to go into `node_modules` and search for `requiresMainQueueSetup`. And if any of the dependencies return `YES` or `true` (depending on Objective-C vs Swift native modules), try to find out if any of them are used in the offending screens. 

If they are, try to find out if it ACTUALLY requires main queue setup (not all native modules actually need that). And if they don't, patch it to return `NO` / `false`. Or if they do, try to find ways of initialising them earlier where they might not race condition with heavy animations, or not use them at all.

This was a very specific and subtle bug for us, so what you experience might differ. But our error was the exact same one as noticed in RNMapbox.

cbrevik:
This was the PR that fixed it for us and has more info: https://github.com/AtB-AS/mittatb-app/pull/5897

Again this is an internal dependency, but another dependency might be giving you similar issues.

PR #5944, opened 2026-04-07
fix: qr code scan multiple requests bug fix

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23523

## Description
Ensures that no new requests are sent from qr-scanning while a previous qr-code is being processed.


reidzeibel (approved):
I think this works 👍🏼

marius-at-atb (approved):

PR #5947, opened 2026-04-07
fix: voiceover for 0 points says 0 instead of unknown

jorunnl:
## Description
Fixes voiceover for when the user has 0 points

## Acceptance Criteria

- [x] When user has 0 points, voiceover says 0 points


marius-at-atb (approved):

hanne-at-atb:
So when testing with android it still says unknown number of points, @jorunnl  😢

jorunnl:
Should be fixed now @hanne-at-atb

hanne-at-atb:
Yes, now it reads as expected! ✅

PR #5951, opened 2026-04-08
feat: log fare contract zod errors to bugsnag

rosvik:
## Issue Reference

https://mittatb.slack.com/archives/C02DL21RK7C/p1774022695138279

## Description

If the app receives any invalid fare contracts from the list endpoint, it's logged to Bugsnag


reidzeibel (approved):
Looks OK, but maybe we could also log some breadcrumbs there? like the farecontract ID or something for traceability?

reidzeibel (approved):
That works 👍🏼

PR #5956, opened 2026-04-10
fix: array index was used as `now` when filtering departures

rosvik:
Since `isValidDeparture` takes two params, calling `filter(isValidDeparture)` caused two params to be passed to `isValidDeparture`: departure (as expected) and an index. 

In practice, this meant we were running

```ts
.filter((departure, i) => isValidDeparture(departure, i))
```

When now was set to 0, 1, 2 etc., this was interpreted as seconds since Jan 1st 1970. So in practice, departures was never more than one minute in the past, because it seemed they were 56 years in the future.

This caused departures to stay as "now" for 5 minutes on the front page (until we call for a full refresh) instead of being filtered out.

### Acceptance criteria

- [ ] On the front page, departures are removed 1 minute after passing the stop

### DIFF
--- a/src/departure-list/utils.ts
+++ b/src/departure-list/utils.ts
@@ -49,7 +49,9 @@ export function updateStopsWithRealtime(
 
 function filterOutOutdatedDepartures(quayGroup: QuayGroup) {
   const newDepartureGroups = quayGroup.group.map(function (group) {
-    const newDepartures = group.departures.filter(isValidDeparture);
+    const newDepartures = group.departures.filter((departure) =>
+      isValidDeparture(departure),
+    );
     // Optimization to avoid having to sort list to often.
     // If after filtering it has the same length it means we could
     // just use the previous departure list.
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):
Nice catch!

mortennordseth (approved):
💯

tormoseng:
@rosvik What is deciding when to remove a departure. Is it the `/realtime` request?

I observed that a specific departure was listed in the `/realtime` response and it said "Now" in the app. When this departure was not longer a part of the `/realtime` response, the "Now" was visible until 2 min after it was gone from the  response (on 2 devices I followed concurrently). Repeated the with another departure, and then the departure was gone after approx 1 min on one of the devices, while it took 2.5 min on the other device.

Could it be the effect you're mentioning here: https://mittatb.slack.com/archives/C0116FMPX4Y/p1776066708923329?thread_ts=1775811535.167259&cid=C0116FMPX4Y

rosvik:
@tormoseng It should be decided by comparing the most recent estimated departure time and current time. 

So if the the /realtime gives estimated time 12:00, I'd expect it to be removed 12:01.

If /realtime stops sending updates for a departure, I believe the app keeps the previous estimate.

tormoseng:
Mhm. I checked again. With a departure where the last expected departure time was 08:55:40, was not part of the /realtime 08:56:02. But it disappeared around 08:58:00. So, there is still a lag in when a departure disappears. 🤔

tormoseng:
Observation handled by https://github.com/AtB-AS/mittatb-app/pull/5974, so moving this issue further.

PR #5959, opened 2026-04-10
refactor: short wait time logic

strandlie:
## Summary

- **Consolidates the "short transfer" warning logic** into a single `isShortWaitTime` function in `@atb/modules/trip-patterns`, replacing three scattered implementations (`timeIsShort` in `date.ts`, `significantWaitTime` in `travel-details-screens/utils.ts`, and an inline check in `TravelCardLegs`)
- **Fixes a bug in `TravelCardLegs`** where the short-transfer notification icon was shown for *all* wait times instead of only tight transfers (31–180 seconds)
- **Removes `timeIsShort`** from `date.ts` and **`significantWaitTime`** from `travel-details-screens/utils.ts` — both replaced by the shared functions in `trip-patterns/utils.ts`

### What is `isShortWaitTime`?

A wait time is "short" when it is **significant** (> 30s, worth displaying) **and** dangerously tight (≤ 3 min). This triggers the info notification warning the user about a tight transfer. The definition is now in one place and used by:

- `TravelCardLegs` — notification icon on search result legs
- `WaitSection` — info box in trip details
- `hasShortWaitTime` — used by trip analytics
- `hasShortWaitTimeAndNotGuaranteedCorrespondence` — used by the Trip component

## Test plan

- [x] `yarn tsc` passes
- [x] `yarn lint` passes
- [x] `yarn prettier` passes
- [x] `yarn test` — 51 suites, 637 tests, all passing
- [x] Verify short-transfer notification icon appears on search results only for transfers between 31s and 3min
- [x] Verify trip details WaitSection info box appears only for tight transfers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
+++ b/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
@@ -1,0 +1,41 @@
+import {isShortWaitTime, significantWaitTime} from '../utils';
+
+describe('significantWaitTime', () => {
+  it('should be false for 0 seconds', () => {
+    expect(significantWaitTime(0)).toBe(false);
+  });
+
+  it('should be false for 30 seconds (boundary)', () => {
+    expect(significantWaitTime(30)).toBe(false);
+  });
+
+  it('should be true for 31 seconds', () => {
+    expect(significantWaitTime(31)).toBe(true);
+  });
+});
+
+describe('isShortWaitTime', () => {
+  it('should be false for 0 seconds (not significant)', () => {
+    expect(isShortWaitTime(0)).toBe(false);
+  });
+
+  it('should be false for 30 seconds (boundary, not significant)', () => {
+    expect(isShortWaitTime(30)).toBe(false);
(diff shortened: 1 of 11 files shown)


rosvik (approved):
Nice!

reidzeibel:
@rosvik @strandlie just one question, what happens if the transfer is 30 seconds or below? 🤔

Are they being displayed on the trip suggestion or hidden because it is too short?

strandlie:
> @rosvik @strandlie just one question, what happens if the transfer is 30 seconds or below? 🤔
> 
> Are they being displayed on the trip suggestion or hidden because it is too short?

Good question! Same as before, this does not change that - so the trip will be displayed, but no warning about a short wait time. 

That is really a special case, for example when you have `WalkLeg` --> `BusLeg`. The theoretical "wait time" between those is 0 from Entur.

reidzeibel:
@strandlie got it, then everything is good!

tormoseng:
✅

PR #5960, opened 2026-04-10
feat: add points button to ticketing screen

jorunnl:
## Issue Reference

## Description
Adds button on ticket screen and map screen that leads to the Points screen

Some things worth noting:
- Moves compass on map when button is present, moves based on fontSize, since button size increases

image image


strandlie (approved):

src/modules/map/hooks/use-map-view-config.ts
```diff
@@ -43,12 +43,25 @@ export const useMapViewConfig = (
   );
   const {enable_surface_view_map} = useRemoteConfigContext();
 
+  // `mapbox` (v10) Adds compass offset `compassViewMargins` is still supported but generates issues:
+  // Mapbox error fireEvent failed: <rnmapbox_maps.RCTMGLEvent: 0x6000028a0fe0>
+  const compassPosition = useMemo(
+    () => ({
+      top: includeBonusOffset
+        ? Math.round(COMPASS_BASE_TOP + 15 * fontScale)
```
adriansberg, line 51: Is this 15 arbitrary or calculated from something?

adriansberg (approved):
Lgtm. Didn't verify the placement of stuff with Figma.

src/modules/map/hooks/use-map-view-config.ts
jorunnl: pretty arbitrary. I looked into more complex solutions, but think it is unnecessary

PR #5961, opened 2026-04-13
feat: Map tile preloader

marius-at-atb:
Preloading tiles at the next zoom level makes the map experience smoother, mostly due to improved in-transitions.
This is relevant for 3d buildings after https://github.com/AtB-AS/mittatb-app/pull/5230, and was originally made as part of https://github.com/AtB-AS/mittatb-app/pull/5321 to allow for better zoom transitions for vehicles and stations. Splitting it up into 2 PRs, starting with just this one for preloading (part 2: https://github.com/AtB-AS/mittatb-app/pull/5973/changes/091a371d25dfcc39b7e64f63340250b4204d7d49).

Added behind a new feature flag in remote config: `enable_map_tile_preloading`

Mapbox does not have a built in solution for tile preloading, so it has been solved with a second invisible map that zooms further than the main map, which triggers requests and populates the cache. The preloading map is not just invisible, but also just a couple of pixels large and without symbols, in order to minimize rendering work. Also, the Camera is only updated at strategic times, either when you have started zooming or move some distance. So the component rarely re-renders unless you are actually zooming and need it.

Preloading _slightly_ increases number of tile requests and may have a performance impact on some devices, which should be tested.

Staging builds that can be tested:
[iOS](https://github.com/AtB-AS/mittatb-app/actions/runs/24310680717)
[Android](https://github.com/AtB-AS/mittatb-app/actions/workflows/build-staging-android.yml)

**Acceptance Criteria:**
- [x] Overall map experience should be better, not worse.
- [ ] Smoother in-transitions for buildings. (Note -> when zooming fast, data may still not load early enough for the in-transition. This is expected.)
- [ ] Faster loading of vehicles and stations data on the next zoom level (this may be difficult to tell from just this PR, but is required for part 2 of https://github.com/AtB-AS/mittatb-app/pull/5321, where it will matter a lot).
- [x] Not too many tile requests.

### DIFF
--- a/src/modules/map/Map.tsx
+++ b/src/modules/map/Map.tsx
@@ -11,6 +11,7 @@ import {
   Viewport,
   Camera,
   MapView,
+  MapState,
 } from '@rnmapbox/maps';
 
 import {Feature} from 'geojson';
@@ -56,7 +57,11 @@ import {useMapContext} from '@atb/modules/map';
 import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
 import {NationalStopRegistryFeatures} from './components/national-stop-registry-features';
 import {OnPressEvent} from 'node_modules/@rnmapbox/maps/src/types/OnPressEvent';
-import {VehiclesAndStations} from './components/mobility/VehiclesAndStations';
+import {
+  StationsWithClusters,
+  VehiclesAndStations,
+  VehiclesWithClusters,
+} from './components/mobility/VehiclesAndStations';
 import {SelectedFeatureIcon} from './components/SelectedFeatureIcon';
 import {ShmoBookingState} from '@atb/api/types/mobility';
 import {useStablePreviousValue} from '@atb/utils/use-stable-previous-value';
@@ -68,8 +73,9 @@ import {ShmoTesting} from './components/mobility/ShmoTesting';
 import {usePreferencesContext} from '../preferences';
 import {useBottomSheetContext} from '@atb/components/bottom-sheet';
 import {GeofencingZonesAsTiles} from './components/mobility/GeofencingZonesAsTiles';
+import {MapTilePreloader} from './components/MapTilePreloader';
 
-const DEFAULT_ZOOM_LEVEL = 14.5;
+export const DEFAULT_ZOOM_LEVEL = 14.5;
 
 export const Map = (props: MapProps) => {
   const {
@@ -92,6 +98,7 @@ export const Map = (props: MapProps) => {
   const {getCurrentCoordinates} = useGeolocationContext();
   const mapCameraRef = useRef<Camera>(null);
   const mapViewRef = useRef<MapView>(null);
+  const mapStateRef = useRef<MapState | null>(null);
   const [initMapLoaded, setInitMapLoaded] = useState(false);
   const {bottomSheetMapRef} = useBottomSheetContext();
   const onPressStartRef = useRef<boolean>(false);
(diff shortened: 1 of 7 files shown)


src/modules/map/components/mobility/VehiclesAndStations.tsx
```diff
@@ -54,22 +58,26 @@ export const VehiclesWithClusters = ({
...
+        hideSymbols ? 'hidden' : 'visible'
+      }`}
       sourceID={vehiclesAndStationsVectorSourceId}
       sourceLayerID="combined_layer"
       minZoomLevel={minZoomLevel}
       aboveLayerID={MapSlotLayerId.Vehicles}
-      filter={filter}
-      style={style}
+      style={hideSymbols ? {} : style}
+      {...filter}
```
marius-at-atb, line 69: undefined filter crashes Android, so added as `{...filter} `instead

src/modules/map/Map.tsx
```diff
@@ -510,6 +520,27 @@ export const Map = (props: MapProps) => {
...
+              hideSymbols={true}
+            />
+          )}
+          {!!showStations && (
+            <StationsWithClusters
+              selectedFeatureId={undefined}
+              showNonVirtualStations={true}
+              hideSymbols={true}
+            />
+          )}
```
marius-at-atb, line 541: Note: We may consider adding geofencingZones here as well to preload gfz tiles

src/modules/map/components/MapTilePreloader.tsx
```diff
@@ -0,0 +1,128 @@
...
+  (
+    {startingCoordinates, tabBarHeight, children}: MapTilePreloaderProps,
+    /** Should be set like this: onCameraChanged={(state) => {mapStateRef.current = state;}} */
+    mapStateRef: ForwardedRef<MapState>,
+  ) => {
+    const mapCameraOffscreenRef = useRef<Camera>(null);
+
+    const mapViewConfig = useMapViewConfig({
+      includeVehiclesAndStationsVectorSource: true,
+      includeBasemapStyle: true,
```
marius-at-atb, line 41: Setting includeBasemapStyle=true means we preload 3d assets as well.
If performance issues are discovered, we can turn this off for those devices.

src/modules/map/components/MapTilePreloader.tsx
```diff
@@ -0,0 +1,128 @@
...
+          }
+        }
+      },
+      [mapStateRef],
+      100, // Refresh rate. Might be lower for slow devices. No need for very short interval, this is "just for preloading behind the scenes".
+    );
+
+    const {width, height} = Dimensions.get('window');
+    return (
+      <MapView
```
marius-at-atb, line 87: Note to self: we may want to also set preferredFramesPerSecond to e.g. 10 in case there is a performance difference.

src/modules/map/components/MapTilePreloader.tsx
```diff
@@ -0,0 +1,128 @@
...
+            previousCenterRef.current = center;
+            previousUpdateTimeRef.current = now;
+          }
+        }
+      },
+      [mapStateRef],
+      100, // Refresh rate. Might be lower for slow devices. No need for very short interval, this is "just for preloading behind the scenes".
+    );
+
+    const {width, height} = Dimensions.get('window');
```
reidzeibel: Can use `useWindowDimensions()` instead maybe?

reidzeibel (approved):
just one nitpick

src/modules/map/components/MapTilePreloader.tsx
marius-at-atb: True, updated :+1:

tormoseng:
✅  Didn't find any noticable effects on performance, and since we're not enabling 3d for older devices anyway, it should be fine.

PR #5968, opened 2026-04-13
fix: Use correct props for gfz message when using gfz as tiles

marius-at-atb:
## Issue Reference

With `enable_geofencing_zones_as_tiles` true, this message in the ActiveScooterSheet didn't work due to wrong prop lookups.

## Description
This PR uses the correct prop lookups and also respects the debug override option.

### Acceptance Criteria:
- [ ] When driving into a geofencing zone with code other than `allowed`, the message in the bottom sheet should behave the same when enable_geofencing_zones_as_tiles/isGeofencingZonesAsTilesEnabled is true as when it is false.

### DIFF
--- a/src/modules/map/hooks/use-shmo-warnings.tsx
+++ b/src/modules/map/hooks/use-shmo-warnings.tsx
@@ -93,7 +93,7 @@ export const useShmoWarnings = (
       if (code !== 'allowed') {
         const isStationParking: boolean =
           (isGeofencingZonesAsTilesEnabled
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):

tormoseng:
@marius-at-atb I did not see the message when driving outside with our Ryde. "Geofence zones as tiles" was set to `true` in debug.

tormoseng:
Works now after https://github.com/AtB-AS/mittatb-app/pull/6083

PR #5974, opened 2026-04-14
fix: improve handling of passed departure favorites

rosvik:
## Issue Reference

https://github.com/AtB-AS/mittatb-app/pull/5956

https://mittatb.slack.com/archives/C0116FMPX4Y/p1775811535167259

## Description

After https://github.com/AtB-AS/mittatb-app/pull/5956, there were another couple of issues with filtering out old departures from the DeparturesWidget, which caused departures to hang around longer than they should have. This seems to be for two reasons:

1. The `isNumberOfMinutesInThePast` was inaccurate, because it used minutes for calculation without taking rounding into account. I've now converted this to use seconds.
2. A `useMemo` in quay-section.tsx prevented the sorted array of departures from updating. I've added `now` to the dependency array, which updates every 10s.

## Acceptance criteria

- [x] Departures favorites are removed 1 minute after departure (max 70s, since it only filters every 10s)

### DIFF
--- a/src/departure-list/section-items/line.tsx
+++ b/src/departure-list/section-items/line.tsx
@@ -50,12 +50,14 @@ export type LineItemProps = SectionItemProps<{
   group: DepartureGroup;
   searchDate: string;
   onPressDeparture: QuaySectionProps['onPressDeparture'];
+  now: number;
 }>;
 export function LineItem({
   group,
   searchDate,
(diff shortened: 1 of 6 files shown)


reidzeibel (approved):

tormoseng:
✅  Looking good! 🐐

PR #5976, opened 2026-04-15
feat: use IsBonusEnabled feature toggle

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23671

## Description
Points/Bonus-related stuff on dashboard, Profile and tickets only visible when feature toggle is set to true
* Ensures that enrollment and bonus services are not called unnecessary when toggle is false


src/modules/bonus/use-is-bonus-active.ts
```diff
@@ -0,0 +1,7 @@
+import {KnownProgramId, useIsEnrolled} from '@atb/modules/enrollment';
+import {useFeatureTogglesContext} from '@atb/modules/feature-toggles';
+
+export function useIsBonusActiveForUser(): boolean {
+  const {isPointsEnabled} = useFeatureTogglesContext();
```
marius-at-atb: Since everything else is called bonus in the code, let's use it here as well for consistency.

src/modules/enrollment/use-program.ts
```diff
@@ -5,13 +5,17 @@ import {ONE_HOUR_MS} from '@atb/utils/durations';
 
 export const GET_PROGRAM_KEY = 'GET_PROGRAM';
 
-export function useProgram(programId: KnownProgramId): ProgramType | undefined {
+export function useProgram(
```
Dahly96: I do believe queries should follow the naming rules of useProgramQuery and use-program-query.ts

Dahly96 (approved):
Nicenice, just a small comment to fix before merge

PR #5977, opened 2026-04-15
chore: add more rules to allow usage of common tools

reidzeibel:
## Description
Adds a set of rules to allow running common bash commands without asking for permissions. Should be pretty straightforward, if anything should be included, feel free to suggest.


rosvik (approved):
🤖

PR #5980, opened 2026-04-16
chore: add feature-toggle skill to create feature toggles

reidzeibel:
## Description
Added a Claude Code skill to create feature toggles.

Can be invoked by `/create-feature-toggle {enable_feature_name} with default value {true/false}` or just a descriptive request to create one, e.g: "create a feature toggle called "enable_show_timer" with default value true

image

image


marius-at-atb (approved):
Lgtm!

rosvik (approved):

PR #5983, opened 2026-04-16
refactor: standardize Linking.openURL usage with openUrl utility

reidzeibel:
## Issue Reference
fixes https://github.com/AtB-AS/kundevendt/issues/22097

## Description
- Introduces a new `openUrl(url, bugsnagMessage, onError?)` utility in `src/utils/open-url.ts` that wraps `Linking.openURL` with consistent error handling — reports to Bugsnag and optionally calls an `onError` callback.
- All errors are grouped under the `linkingOpenUrl` Bugsnag grouping hash, with per-call-site context in metadata.

### DIFF
--- a/src/modules/fare-contracts/PurchaseReservation.tsx
+++ b/src/modules/fare-contracts/PurchaseReservation.tsx
@@ -29,7 +29,7 @@ export const PurchaseReservation: React.FC<Props> = ({reservation, now}) => {
   const {theme} = useThemeContext();
 
   async function openVippsUrl(vippsUrl: string) {
-    await openUrl(vippsUrl, 'Could not open Vipps URL');
+    await openUrl(vippsUrl);
   }
 
   const isSubAccountReservation = customerProfile?.subAccounts?.some(
--- a/src/modules/fare-contracts/PurchaseReservation.tsx
+++ b/src/modules/fare-contracts/PurchaseReservation.tsx
@@ -7,9 +7,9 @@ import {
   useTicketingContext,
 } from '@atb/modules/ticketing';
 import {TicketingTexts, useTranslation} from '@atb/translations';
-import Bugsnag from '@bugsnag/react-native';
+import {openUrl} from '@atb/utils/open-url';
 import React from 'react';
-import {Linking, View} from 'react-native';
+import {View} from 'react-native';
 import {formatToLongDateTime} from '@atb/utils/date';
 import {fromUnixTime} from 'date-fns';
 import {NativeBlockButton} from '@atb/components/native-button';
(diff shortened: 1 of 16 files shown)


src/utils/open-url.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Open a URL with consistent error handling.
+ * Always reports failures to Bugsnag with the provided message.
+ *
+ * @param url - The URL to open (http, https, tel, etc.)
+ * @param bugsnagMessage - Descriptive message for Bugsnag metadata
+ * @param onError - Optional callback invoked on failure (after Bugsnag report)
+ */
+export async function openUrl(
+  url: string,
+  bugsnagMessage: string,
```
rosvik: Is a custom Bugsnag message necessary here? I'm thinking the URL should be enough context

reidzeibel: True, we can put "cannot open ${url}" instead 🤦🏼

reidzeibel: nevermind, removed it since the error message should be the "error" from Linking.openUrl()

src/modules/fare-contracts/PurchaseReservation.tsx
```diff
@@ -29,11 +29,7 @@ export const PurchaseReservation: React.FC<Props> = ({reservation, now}) => {
   const {theme} = useThemeContext();
 
   async function openVippsUrl(vippsUrl: string) {
-    try {
-      await Linking.openURL(vippsUrl);
-    } catch (err: any) {
-      Bugsnag.notify(err);
-    }
+    await openUrl(vippsUrl);
   }
```
rosvik: Nit, but could just remove this function now, and call openUrl directly

src/utils/open-url.ts
```diff
@@ -0,0 +1,25 @@
+import {Linking} from 'react-native';
+import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
+
+/**
+ * Open a URL with consistent error handling.
+ * Always reports failures to Bugsnag with the provided message.
+ *
+ * @param url - The URL to open (http, https, tel, etc.)
+ * @param bugsnagMessage - Descriptive message for Bugsnag metadata
```
rosvik:
```suggestion
```

src/utils/open-url.ts
```diff
@@ -0,0 +1,25 @@
+import {Linking} from 'react-native';
+import {notifyBugsnag} from '@atb/utils/bugsnag-utils';
+
+/**
+ * Open a URL with consistent error handling.
+ * Always reports failures to Bugsnag with the provided message.
```
rosvik:
```suggestion
 * Always reports failures to Bugsnag.
```

rosvik (approved):

PR #5989, opened 2026-04-16
chore: update mobile token SDK version to 3.4.8

reidzeibel:
## Issue Reference
solves https://github.com/AtB-AS/kundevendt/issues/23447

## Description
Entur used Mobile Token SDK version 3.4.8 and seems to have no issues, so we should update our version since it has been quite some time from our last update.


ios/Podfile
rosvik, line 1: Could this cause the app size to increase? Probably not a big issue, but maybe keep an eye on it once we have a staging build

rosvik (approved):
🚀

ios/Podfile
reidzeibel: Might be, but we'll see 😬

rosvik: 11 MB more after I updated 👀 (Weird that "documents and data" also changed, but that's probably unrelated)

IMG_4722
IMG_F14D1B62B79D-1

reidzeibel: FYI @Sebstorvik

PR #5990, opened 2026-04-16
fix: do not display sent or received ticket info if self purchase

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/23683

## Description
Do not show the component if not sent/received fare contract (self purchase)


rosvik:
I'm afraid this issue will reappear if the component is reused somewhere else or the parent component is refactored and we forget to add the guard. I think we should rather disable the query or fall back to the users phone number to make handling of this bug more explicit.

reidzeibel:
@rosvik true, I'll rework it!

reidzeibel:
@rosvik updated 🚀

rosvik (approved):

PR #5993, opened 2026-04-17
fix: strictly parse fare contract dates

rosvik:
Turns out date-fns `parseISO` handles more of ISO 8601 than we took into account when using it in `convertIsoStringFieldsToDate`. It's quite liberal when parsing dates, and can ends up converting some order IDs to dates if they contains numbers in the right places.

This caused the following Zod error when fetching `ticket/v4/list`

```json
[
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "orderId"
    ],
    "message": "Invalid input: expected string, received Date"
  }
]
```

https://mittatb.slack.com/archives/C02DL21RK7C/p1776415955164209

### DIFF
--- a/src/modules/purchase-selection/purchase-selection-builder.ts
+++ b/src/modules/purchase-selection/purchase-selection-builder.ts
@@ -17,7 +17,7 @@ import {
   isValidSelection,
   isWithinUserProfileMaxCount,
 } from './utils';
-import {isValidDateString} from '@atb/utils/date';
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):
🚀

tormoseng:
✅  Worked like a charm. Tested earlier version vs this with orderId`41ZILZ5F`.

PR #6001, opened 2026-04-21
feat: start trip citybike

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23697
Closes https://github.com/AtB-AS/kundevendt/issues/22834
Closes https://github.com/AtB-AS/kundevendt/issues/22835

This PR covers the start trip functionlity for citybikes with a happypath flow, while preserving the existing scooter integration. This also covers starting a trip with stationId and with a vehicleId. 

The design is taken from: https://www.figma.com/design/YfLcr1BKOuaXk9HIUilXJK/City-bike-integration-V1?node-id=890-59052&p=f&t=oHbUVqATV4C5OseF-0

There will be another issue fixing the style details and colors, to not overwhelm this PR with changes, since there is alot to make this fully happen


src/api/types/mobility.ts
```diff
@@ -112,6 +112,8 @@ const PropulsionTypeSchema = z.enum([
   'COMBINED',
 ]);
 
+export type PropulsionTypeBooking = z.infer<typeof PropulsionTypeSchema>;
```
marius-at-atb: From before, `PropulsionTypeSchema` is used for map icons which are not connected to bookings, and only map items can have type `COMBINED`.

Perhaps we should change the existing `PropulsionTypeSchema` to `MapItemPropulsionTypeSchema`, and then also have a clean `PropulsionTypeSchema` without `COMBINED`. Then the `PropulsionType` can be used directly, no need to infer from the schema.

src/modules/mobility/utils.ts
```diff
@@ -301,21 +305,26 @@ export const findOperatorBrandImageUrl = (
   mobilityOperators?.find((op) => op.id === operatorId)?.brandAssets
     ?.brandImageUrl;
 
-export const getModeAndSubModeFromFormFactor = (
-  formFactor: FormFactor,
-): {
-  mode: AnyMode;
-  subMode?: AnySubMode;
-} => {
+export const getTrasportModeAndSubModeByFormFactorAndPropulsionType = (
```
marius-at-atb: I generally like verbose names, but here it makes sense to just call it `getTransportModeAndSubMode`, especially since `propulsionType` is optional.

src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
```diff
@@ -59,6 +67,19 @@ export const ActiveScooterSheet = ({
...
+  const {mode, subMode} =
+    getTrasportModeAndSubModeByFormFactorAndPropulsionType(
+      activeBooking?.asset?.formFactor,
+      activeBooking?.asset?.propulsionType,
+    );
+  const operator = useOperators().byId(activeBooking?.asset.operator.id);
+
+  const priceAdjustments = activeBooking?.asset?.formFactor
+    ? operator?.priceAdjustments?.[
+        activeBooking.asset.formFactor as keyof typeof operator.priceAdjustments
```
marius-at-atb: Rather than type casting, it can be looked up and found in the list of keys

src/modules/mobility/components/BikeStationIntegrationView.tsx
```diff
@@ -120,8 +118,36 @@ export const BikeStationIntegrationView = ({
...
+                        currency: 'NOK',
+                      },
+                      system: {
+                        id: station.system.id,
+                        name: station.system.name,
+                        operator: {
+                          id: station.system.operator.id,
+                          name: station.system.operator.name,
+                        },
+                      },
```
marius-at-atb: The system should always be the same, so this can be simplified with e.g.
`const {system} = station` and `...system`

src/modules/mobility/components/BikeStationIntegrationView.tsx
```diff
@@ -120,8 +118,36 @@ export const BikeStationIntegrationView = ({
...
+                      station: {id: station.id},
+                      vehicleType: {...e.vehicleType, name: {}},
+                    };
+
+                    queryClient.setQueryData(
+                      ['getVehicle', MOCK_VEHICLE_ID],
```
marius-at-atb: As usual when we need to reference the same queryKey several places, extract a function to get it and re-use..

src/modules/mobility/components/BikeStationIntegrationView.tsx
```diff
@@ -120,8 +118,36 @@ export const BikeStationIntegrationView = ({
...
+                      mockVehicle,
+                    );
+                    onPressVehicleType(MOCK_VEHICLE_ID);
```
marius-at-atb: Should use `mockVehicle.id` here, ref DRY/SSOT. (and as input to the [queryKey function](https://github.com/AtB-AS/mittatb-app/pull/6001/changes#r3123595108))

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -30,19 +31,40 @@ export const MobilityTexts = {
...
-    switch (propulsionType) {
-      case PropulsionType.ElectricAssist:
-      case PropulsionType.Electric:
-        return _(
-          'Elektrisk bysykkel',
-          'Electric city bike',
-          'Elektrisk bysykkel',
-        );
-      case PropulsionType.Human:
+  bikeNameByPropulsionType: (
```
marius-at-atb: With these changes, it looks like it should be called `vehicleNameByPropulsionType`.
Nit: Consider also standardising whether formFactor or propulsionType is the first param (ref [getTransportModeAndSubMode](https://github.com/AtB-AS/mittatb-app/pull/6001/changes#r3123367758))

src/modules/mobility/components/ShmoActionButton.tsx
```diff
@@ -42,6 +49,9 @@ export const ShmoActionButton = ({
   const coordinates = getCurrentCoordinatesGlobal();
   const {warningMessage} = useShmoWarnings(vehicleId);
   const {logEvent} = useAnalyticsContext();
+  const {vehicle} = useVehicle(vehicleId);
+
+  const isStationBasedBooking = vehicleId === MOCK_VEHICLE_ID;
```
marius-at-atb: While this works, I think we can make it less hacky by adding `isStationBasedBooking` as a param for `onVehicleTypeSelected`, and in the bicycle state in `mapStateReducer`.
Then in `useVehicleQuery`, also accept a param to determine the `refetchOnMount` logic.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_CityBikeStartTripScreen.tsx
```diff
@@ -0,0 +1,159 @@
...
+  } = useActiveShmoBookingQuery(isFocusedAndActive, ONE_SECOND_MS * 10);
+
+  useEffect(() => {
+    if (
+      activeBooking &&
+      activeBooking?.state !== ShmoBookingState.NOT_STARTED
+    ) {
+      navigation.goBack();
+    }
+  }, [activeBooking, navigation]);
```
marius-at-atb, line 51: Hmm, from a state management perspective, using a separate screen differs from how the rest of the booking state is handled. This also leads to the need for onNotStartedBookingState, with logic like this useEffect, and this other snippet, which isn't purely declarative:
```ts
    if (res.state === ShmoBookingState.NOT_STARTED) {
      onNotStartedBookingState?.();
    }
```
Instinctively I think this screen should be a view in Map (not a route). Potentially even talk to UI/UX about just having it in a bottom sheet 🤔

marius-at-atb: Having thought a bit more about it now:
Until the day we area ready to move the logic in MapBottomSheets to the navigator (with all bottom sheets as routes), I think it is best to not use routes for states.

src/modules/mobility/components/BikeStationIntegrationView.tsx
```diff
@@ -120,8 +118,36 @@ export const BikeStationIntegrationView = ({
                   })();
 
                   if (vehicle?.id) {
-                    // sanity check, already handled in the filter, but atleast prevents crashes if something unexpected happens
                     onPressVehicleType(vehicle.id);
+                  } else {
+                    const mockVehicle: Vehicle = {
```
marius-at-atb, line 130: Nice!
Btw `currentRangeMeters` should actually be an optional field in Entur's Mobility API. They will update it soon, and after that we can skip `currentRangeMeters` here.

marius-at-atb (changes requested):
Nice work!
See some comments.

src/modules/mobility/components/BikeStationIntegrationView.tsx
Dahly96: yes, now it should be the same after station is fetch with the same types from mobility. Before it was the fragment types from the BFF

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_CityBikeStartTripScreen.tsx
Dahly96: Its the same flow as parking picture, and a nice solution for the flow in the scetches. I dont think we should rework the flow now for this PR, then it should be a new issue to move from screen to something else

src/modules/mobility/components/BikeStationIntegrationView.tsx
Dahly96: nice

src/modules/mobility/utils.ts
```diff
@@ -347,3 +346,10 @@ export function getThemedIllustrationForFormFactor(formFactor: FormFactor) {
       return null;
   }
 }
+
+export function isValidKey<T extends object>(
+  obj: T,
+  key: PropertyKey,
+): key is keyof T {
+  return key in obj;
+}
```
marius-at-atb: Could be defined as a general util, not just for mobility

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_CityBikeStartTripScreen.tsx
marius-at-atb: I suppose we can come back to it :)
Make sure to test this with android go back options as well.

marius-at-atb (approved):
🚴

PR #6005, opened 2026-04-22
fix: Standarized fetching of situations

strandlie:
## Issue Reference
Closes https://github.com/AtB-AS/kundevendt/issues/23714

## Description
Previously, when in context of a trip-pattern leg we either looked at only `leg.situations`, or tried to compose situations from `leg.situations` and `leg.from/toPlace.quay.situations`. 

Neither of these approaches are correct, since `leg.situations` only displays situations regarding the actual transportation part of the leg, and not stop places in either end, and `leg.from/toPlace.quay.situations` can have situations not relevant for the leg. 

This PR tries to standardize the use of situations to only and always use `leg.situations` and `leg.estimatedFrom/toCall.situations` since these are "smart" and will filter out situations irrelevant for that leg, and _will_ capture both situations regarding the transportation part and for either stop place. 

Other uses of situations should not be affected. 

## Test input
@tormoseng Since we are changing to an Entur-curated source of situations, it is very hard to fabricate some response. But that means we can also shift the responsibility to them if something is wrong. So I think this is better, even if it is hard to verify that it solves the actual problem in the issue with synthetic data

### DIFF
--- a/src/api/types/generated/fragments/trips.ts
+++ b/src/api/types/generated/fragments/trips.ts
@@ -35,15 +35,17 @@ export type TripPatternFragment = {
       aimedDepartureTime: any;
       expectedDepartureTime: any;
       stopPositionInPattern: number;
+      cancellation: boolean;
       destinationDisplay?: {frontText?: string; via?: Array<string>};
       quay: {publicCode?: string; name: string};
       notices: Array<NoticeFragment>;
-      cancellation: boolean;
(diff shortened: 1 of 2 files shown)


reidzeibel (approved):
Looks ok, tested with departures/trips that passes through or starts from:
- Sluppenvegen
- Astronomvegen

PR #6006, opened 2026-04-22
feat: use the travel card component for booking screen

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/23646

## Summary
Refactors the `TravelCard` component to be more modular and reuses it for the booking trip selection screen, replacing the previous custom card implementation.

### Key changes

- **Modular TravelCard API** — Replaced the `TravelCardType` enum (`'trip-search' | 'saved-trip' | 'booking'`) with boolean feature flags (`includeDayInfo`, `includeFromToInfo`, `includeLegNotifications`, `includeSituationNotices`), so callers declare exactly which features they need.
- **Tag support** — Added `tagLabel` / `tagType` props to `TravelCard`, rendering a `Tag` inside the card with proper accessibility (reads tag type + label with a screenreader pause).
- **Accessibility prefix** — Replaced the internal type-based a11y prefix with a required `a11yPrefix` prop, giving callers full control over the screenreader prefix text.
- **Disabled state** — Added `isDisabled` prop that visually dims the card and disables interaction.
- **TravelCardNotices** — Extracted situation/notice display into a separate `TravelCardNotices` component shown below the journey legs.
- **Detailed a11y labels** — Added `getDetailedSituationOrNoticeA11yLabel` that reads out per-leg situation/notice text (e.g. "Bus 1: Bus stop has moved").
- **TripSelection cleanup** — Removed `BookingTrip` wrapper, `TripSelectionTag`, and related styles. Consolidated `getBookingTagLabel`/`getBookingTagType` into a single `getBookingTagInfo` helper.
- **Test coverage** — Added 52 tests for `situations/utils.ts` covering notice/situation collection, validity filtering, criticality ranking, and a11y label generation.

### Troms
image

### NFK
image

### DIFF
--- a/src/screen-components/travel-card/TravelCard.tsx
+++ b/src/screen-components/travel-card/TravelCard.tsx
@@ -1,5 +1,5 @@
 import React, {useState} from 'react';
-import {StyleSheet} from '@atb/theme';
+import {StyleSheet, Statuses} from '@atb/theme';
 import type {TripPattern} from '@atb/api/types/trips';
 import {TravelCardLegs} from './TravelCardLegs';
 import Animated, {FadeIn} from 'react-native-reanimated';
@@ -13,6 +13,7 @@ import {getTranslatedModeName} from '@atb/utils/transportation-names';
 import {CompositeAccessibilityProvider} from '@atb/modules/composite-accessibility';
 import {getDetailedSituationOrNoticeA11yLabel} from '@atb/modules/situations';
 import {TravelCardNotices} from './TravelCardNotices';
+import {Tag} from '@atb/components/tag';
 
 type TravelCardProps = {
   tripPattern: TripPattern;
@@ -25,8 +26,8 @@ type TravelCardProps = {
   includeLegNotifications?: boolean;
   includeSituationNotices?: boolean;
   isDisabled?: boolean;
-  extraA11yLabels?: Record<string, string | undefined>;
-  extraA11yOrder?: {before?: string[]; after?: string[]};
+  tagLabel?: string;
+  tagType?: Statuses;
 };
 
 export const TravelCard: React.FC<TravelCardProps> = ({
@@ -40,8 +41,8 @@ export const TravelCard: React.FC<TravelCardProps> = ({
   includeLegNotifications = false,
   includeSituationNotices = false,
   isDisabled = false,
-  extraA11yLabels,
-  extraA11yOrder,
+  tagLabel,
+  tagType = 'info',
 }) => {
   const styles = useThemeStyles();
   const [maxWidth, setMaxWidth] = useState(0);
@@ -59,22 +60,15 @@ export const TravelCard: React.FC<TravelCardProps> = ({
       ? getDetailedSituationOrNoticeA11yLabel(tripPattern, language, t)
       : undefined;
 
-  const baseOrder = ['cardPrefix', 'header', 'legs', 'situationOrNotice'];
-  const order = [
-    ...(extraA11yOrder?.before ?? []),
-    ...baseOrder,
-    ...(extraA11yOrder?.after ?? []),
-  ];
-
   return (
     <Animated.View entering={FadeIn}>
       <CompositeAccessibilityProvider
         parentLabels={{
           cardPrefix: prefixA11yLabel,
           situationOrNotice: situationOrNoticeA11yLabel,
-          ...extraA11yLabels,
+          tag: tagLabel,
         }}
-        order={order}
+        order={['cardPrefix', 'header', 'legs', 'situationOrNotice', 'tag']}
       >
         {(accessibilityProps) => (
           <NativeBlockButton
@@ -87,6 +81,11 @@ export const TravelCard: React.FC<TravelCardProps> = ({
             }
             {...accessibilityProps}
           >
+            {tagLabel && (
+              <View aria-hidden={true}>
+                <Tag labels={[tagLabel]} tagType={tagType} />
+              </View>
+            )}
             <TravelCardHeader
               tripPattern={tripPattern}
               includeDayInfo={includeDayInfo}
--- a/src/screen-components/travel-card/TravelCard.tsx
+++ b/src/screen-components/travel-card/TravelCard.tsx
@@ -14,15 +14,16 @@ import {CompositeAccessibilityProvider} from '@atb/modules/composite-accessibili
 import {getDetailedSituationOrNoticeA11yLabel} from '@atb/modules/situations';
 import {TravelCardNotices} from './TravelCardNotices';
 
-export type TravelCardType = 'trip-search' | 'saved-trip' | 'booking';
-
 type TravelCardProps = {
   tripPattern: TripPattern;
   onDetailsPressed(tripPattern: TripPattern, resultIndex?: number): void;
   cardIndex: number;
-  numberOfCards: number;
   testID?: string;
-  type: TravelCardType;
+  a11yPrefix: string;
+  includeDayInfo?: boolean;
+  includeFromToInfo?: boolean;
+  includeLegNotifications?: boolean;
+  includeSituationNotices?: boolean;
   isDisabled?: boolean;
   extraA11yLabels?: Record<string, string | undefined>;
   extraA11yOrder?: {before?: string[]; after?: string[]};
@@ -32,9 +33,12 @@ export const TravelCard: React.FC<TravelCardProps> = ({
   tripPattern,
   onDetailsPressed,
   cardIndex,
-  numberOfCards,
   testID,
-  type,
+  a11yPrefix,
+  includeDayInfo = false,
+  includeFromToInfo = false,
+  includeLegNotifications = false,
+  includeSituationNotices = false,
   isDisabled = false,
   extraA11yLabels,
   extraA11yOrder,
@@ -43,20 +47,15 @@ export const TravelCard: React.FC<TravelCardProps> = ({
   const [maxWidth, setMaxWidth] = useState(0);
   const {t, language} = useTranslation();
 
-  const includeDayInfo = type === 'saved-trip';
-  const includeFromToInfo = type === 'saved-trip';
-
   const uniqueModes = Array.from(
     new Set(tripPattern.legs.map((leg) => leg.mode)),
   );
   const modes = uniqueModes.map((mode) => t(getTranslatedModeName(mode)));
 
-  const prefixA11yLabel = `${t(
-    TravelCardTexts.card.typePrefix(type, cardIndex, numberOfCards),
-  )}. ${t(TravelCardTexts.card.modesPrefix(modes))}.`;
+  const prefixA11yLabel = `${a11yPrefix}. ${t(TravelCardTexts.card.modesPrefix(modes))}.`;
 
   const situationOrNoticeA11yLabel =
-    type === 'booking' && !isDisabled
+    includeSituationNotices && !isDisabled
       ? getDetailedSituationOrNoticeA11yLabel(tripPattern, language, t)
       : undefined;
 
@@ -103,7 +102,7 @@ export const TravelCard: React.FC<TravelCardProps> = ({
                 <TravelCardLegs
                   tripPattern={tripPattern}
                   maxWidth={maxWidth}
-                  type={type}
+                  includeLegNotifications={includeLegNotifications}
                 />
               </View>
               <View>
--- a/src/screen-components/travel-card/TravelCard.tsx
+++ b/src/screen-components/travel-card/TravelCard.tsx
@@ -17,10 +17,9 @@ import {Tag} from '@atb/components/tag';
 
 type TravelCardProps = {
   tripPattern: TripPattern;
-  onDetailsPressed(tripPattern: TripPattern, resultIndex?: number): void;
-  cardIndex: number;
+  onDetailsPressed(tripPattern: TripPattern): void;
   testID?: string;
-  a11yPrefix: string;
+  a11yLabelPrefix: string;
   includeDayInfo?: boolean;
   includeFromToInfo?: boolean;
   includeLegNotifications?: boolean;
@@ -33,9 +32,8 @@ type TravelCardProps = {
 export const TravelCard: React.FC<TravelCardProps> = ({
   tripPattern,
   onDetailsPressed,
-  cardIndex,
   testID,
-  a11yPrefix,
+  a11yLabelPrefix,
   includeDayInfo = false,
   includeFromToInfo = false,
   includeLegNotifications = false,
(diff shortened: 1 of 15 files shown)


src/stacks-hierarchy/Root_TripSelectionScreen/components/TripSelection.tsx
```diff
@@ -209,38 +177,52 @@ function TripSelectionTag({
...
       <Tag
         labels={[t(TicketingTexts.booking.numAvailableTickets(availableSeats))]}
         tagType="info"
       />
     );
+  }
+
+  if (!tag) return null;
+
+  return <View style={styles.tagContainer}>{tag}</View>;
```
rosvik: Maybe I'm missing something, but in general I think we should avoid giving padding to components, and add it when used in the parent instead. Here it also made the logic more messy, so I think it's better to wrap in `<View>` in the parent, and keep returning early like before.

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,88 @@ import {TravelCardHeader} from './TravelCardHeader';
...
   testID?: string;
-  type: TravelCardType;
+  a11yPrefix: string;
+  includeDayInfo?: boolean;
+  includeFromToInfo?: boolean;
+  includeLegNotifications?: boolean;
+  includeSituationNotices?: boolean;
+  isDisabled?: boolean;
+  tagLabel?: string;
+  tagType?: Statuses;
```
rosvik: Would make sense to combine this into one prop that share the type with getBookingTagInfo, since that also returns `{label: string; type: Statuses}`

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,88 @@ import {TravelCardHeader} from './TravelCardHeader';
...
-
-export type TravelCardType = 'trip-search' | 'saved-trip';
+import {getDetailedSituationOrNoticeA11yLabel} from '@atb/modules/situations';
+import {TravelCardNotices} from './TravelCardNotices';
+import {Tag} from '@atb/components/tag';
 
 type TravelCardProps = {
   tripPattern: TripPattern;
   onDetailsPressed(tripPattern: TripPattern, resultIndex?: number): void;
   cardIndex: number;
```
rosvik: Not introduced here, but I don't think `cardIndex` is in use

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,88 @@ import {TravelCardHeader} from './TravelCardHeader';
...
-  numberOfCards: number;
   testID?: string;
-  type: TravelCardType;
+  a11yPrefix: string;
```
rosvik: To be more clear here, maybe rename to `a11yLabelPrefix`?

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,88 @@ import {TravelCardHeader} from './TravelCardHeader';
...
             onPress={() => onDetailsPressed(tripPattern, cardIndex)}
             testID={testID}
-            style={styles.container}
-            accessibilityRole="button"
-            accessibilityHint={t(TravelCardTexts.card.a11yHint)}
+            style={[styles.container, isDisabled && styles.containerDisabled]}
+            accessibilityRole={isDisabled ? 'none' : 'button'}
+            accessibilityHint={
+              isDisabled ? undefined : t(TravelCardTexts.card.a11yHint)
+            }
```
rosvik, line 97: Can we use the `disabled` prop here?

src/screen-components/travel-card/TravelCardLegs.tsx
```diff
@@ -39,7 +41,7 @@ export const TravelCardLegs: React.FC<TravelCardContentProps> = ({
     return previousLeg && previousLeg.interchangeTo?.staySeated === true;
   };
 
-  const a11yLabel = useA11yLabel(filteredLegs);
+  const a11yLabel = useA11yLabel(filteredLegs, includeLegNotifications);
```
rosvik: Is `includeLegNotifications` a a11y flag only? Sounds strange to me that screen reader reads out different texts when there's no visual difference

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: combined 👍🏼

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: renamed

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: removed!

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: the `disabled` prop also sets opacity to 0.2, if we add opacity here, the displayed component will be different to the design, and I don't want to touch this opacity since it will affect many other components.

Ideally we should use `disabled`, what is the recommended action here?

src/screen-components/travel-card/TravelCardLegs.tsx
reidzeibel: on the saved trip or trip search, activating the Travel Card will bring the user to the trip details, so the leg notices aren't read out on the travel card, it is clearer to "see" it on the trip details. For booking purposes, the trip details aren't shown, instead the user is brought to the payment overview page, so the notices must be read on the Travel Card. 

Trip search: Can go to details for complete information.
Saved trip: Can go to details for complete information.
Booking: Cannot go to details for complete information, read out the leg notices.

reidzeibel: We could read everything out on all places instead though, but I didn't do it (yet) because this one only focuses on booking 🤷🏼

src/screen-components/travel-card/TravelCard.tsx
rosvik: Ok, that makes sense! Just wondering, do whatever makes it works as intended. I know there's been discussions on how to handle disabled states, so probably something we'll revisit soon.

src/screen-components/travel-card/TravelCardLegs.tsx
rosvik: My thought is that screen reader should read out whatever is visible on the screen. 

Is `<TravelCardNotices>` always rendered now? Should they? If so, I think we should also read out the notices.

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,85 @@ import {TravelCardHeader} from './TravelCardHeader';
...
-  onDetailsPressed(tripPattern: TripPattern, resultIndex?: number): void;
-  cardIndex: number;
-  numberOfCards: number;
+  onDetailsPressed(tripPattern: TripPattern): void;
   testID?: string;
-  type: TravelCardType;
+  a11yLabelPrefix: string;
+  includeDayInfo?: boolean;
+  includeFromToInfo?: boolean;
+  includeLegNotifications?: boolean;
```
strandlie, line 28: I'm  little unsure about this. Having a closed set of behaviors from some pre-determined types (`trip-search`, `saved-trip`, `booking`) is in my head much easier to manage over time than all the possible combinations of flags like this

src/screen-components/travel-card/TravelCard.tsx
strandlie: I see we are using the disabled prop also in line 78?

src/screen-components/travel-card/TravelCardLegs.tsx
strandlie: So this is a toggle for whether we include the "This trip has warnings. Go to details page for complete information part"" inn the a11y message?

reidzeibel: @rosvik `<TravelCardNotices>` is always rendered, the difference is that in the front page/dashboard for saved trips, we do what @strandlie mentioned above. Because we have details when activating the card, but not for the <TripSelection>. 

I was thinking that having the notices there would add extra unnecessary time for the user while waiting for the text to be read out. But on second thought, screenreader users are mostly used to hearing long information in faster speed, so adding this to the screenreader readout seems fine.

Thoughts?

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: It depends really, if we have selective rendering, having multiple types MAY add complexity in determining which part of the component gets rendered, because then we have to use those types as conditionals, and if the types bloat up, then what I do here should be better in the long run. What do you think? 

having predetermined types should be ok if we have scope on where and how this component is used.

rosvik: > Having a closed set of behaviors from some pre-determined types (trip-search, saved-trip, booking) is in my head much easier to manage over time than all the possible combinations of flags like this

Interesting! I have the exact opposite thought 😄 When I read through the code and read `type: 'saved-trip'` I don't know what that actually entails, and if I want to re-use the component another place, I need to implement a new `type` instead of just enabling the features I need.

I'm biased by previous experiences though: We had a similar setup for the departure screens before (`map`, `favorites` and `departures`), and it was a big source of bugs and headaches.

src/screen-components/travel-card/TravelCardLegs.tsx
rosvik: I'm a bit confused about the whole thing at this point 😅 I think we should do one of two things:

1. Remove the need for the flag by finding a solution that works for both cases. If it means the screen reader is a bit more verbose, I think that's ok 🤷 
2. Otherwise, rename `includeLegNotifications` to make it clear that it is a screen reader / a11y flag, and not visual.

reidzeibel: I'll just add it to the screen reader, more information seems to be the go-to anyways

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: @strandlie good catch, that should not be there, it was added accidentally during my last commit on friday 😬

src/screen-components/travel-card/TravelCard.tsx
strandlie: Well, there are pros and cons to both I guess. 

It was just a thought and I don't feel very strongly about it, so it's up to you :)

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: Will revisit the disabled states here https://github.com/AtB-AS/kundevendt/issues/23781

src/screen-components/travel-card/TravelCardLegs.tsx
reidzeibel: fixed on recent changes

src/screen-components/travel-card/TravelCard.tsx
reidzeibel: I'll leave this for now, based on our (me and @rosvik) recent experience with departure screens and predetermined types. But we'll see over the course of Planning and Purchase.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -108,10 +109,15 @@ export const Results: React.FC<Props> = ({
...
-                  numberOfCards={tripPatterns.length}
                   testID={'tripSearchSearchResult' + i}
-                  type="trip-search"
+                  a11yLabelPrefix={t(
+                    TravelCardTexts.card.a11yPrefix.tripSuggestion(
+                      i + 1,
+                      tripPatterns.length,
+                    ),
+                  )}
+                  includeDayInfo
```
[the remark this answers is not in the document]
reidzeibel: Nope, this is needed for clarity

src/screen-components/travel-card/TravelCard.tsx
```diff
@@ -8,60 +8,100 @@ import {TravelCardHeader} from './TravelCardHeader';
...
         {(accessibilityProps) => (
           <NativeBlockButton
-            onPress={() => onDetailsPressed(tripPattern, cardIndex)}
+            onPress={() => onDetailsPressed(tripPattern)}
             testID={testID}
-            style={styles.container}
-            accessibilityRole="button"
-            accessibilityHint={t(TravelCardTexts.card.a11yHint)}
+            style={[styles.container, isDisabled && styles.containerDisabled]}
+            accessibilityRole={isDisabled ? 'none' : 'button'}
```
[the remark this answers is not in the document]
reidzeibel: This should be revisited in https://github.com/AtB-AS/kundevendt/issues/23781

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -108,10 +109,15 @@ export const Results: React.FC<Props> = ({
                 <TravelCard
                   tripPattern={tripPattern}
                   onDetailsPressed={onDetailsPressed}
-                  cardIndex={i}
-                  numberOfCards={tripPatterns.length}
                   testID={'tripSearchSearchResult' + i}
-                  type="trip-search"
+                  a11yLabelPrefix={t(
+                    TravelCardTexts.card.a11yPrefix.tripSuggestion(
+                      i + 1,
```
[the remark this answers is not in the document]
reidzeibel: fixed

src/screen-components/travel-card/TravelCardNotices.tsx
```diff
@@ -0,0 +1,89 @@
...
+          message: t(TripDetailsTexts.messages.departureIsRailReplacementBus),
+        });
+      }
+      if (leg.bookingArrangements) {
+        legNotifications.push({
+          key: `booking-${leg.id}`,
+          message: t(TripSearchTexts.results.resultItem.footer.requiresBooking),
+        });
+      }
+    }
```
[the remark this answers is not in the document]
reidzeibel: Guarded by data from Entur, no way a rail replacement or booking without leg id

src/stacks-hierarchy/Root_TripSelectionScreen/components/TripSelection.tsx
```diff
@@ -272,11 +335,10 @@ const useBookingTripStyles = StyleSheet.createThemeHook((theme) => {
   };
 });
 
-const useBookingTripSelectionStyles = StyleSheet.createThemeHook((theme) => {
+const useBookingTripSelectionStyles = StyleSheet.createThemeHook(() => {
   return {
     container: {
       width: '100%',
```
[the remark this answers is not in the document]
reidzeibel: Not an issue, rowgap exists on the legacy component, the new component has its own padding

reidzeibel:
Ready for another round of review @rosvik @strandlie

strandlie:
Looking at the screenshots, I think there is too much padding in the Booking TravelCard now. I.e. compare these two

| Booking | Travel Search |
|--------|-------|
|  |  |

reidzeibel:
@strandlie I think the image you posted is broken? 🤔

strandlie (approved):
Other than the padding, I think this looks good now 🎉

PR #6009, opened 2026-04-22
feat: persist previous zone selection in storage

rosvik:
## Issue Reference

closes https://github.com/AtB-AS/kundevendt/issues/22913

## Description

If the user haven't given access to geolocation, we now store the user's previous zone selection in async storage, and fall back to this instead of the organization's default zone.


reidzeibel (approved):
👍🏼

PR #6010, opened 2026-04-22
refactor: mobility station handling and remove bike/car station queries

adriansberg:
Refactor mobility station handling by removing bike and car station queries and consolidating them into a single station query.

resolve https://github.com/AtB-AS/team-platform/issues/1045


src/modules/mobility/components/CarPreviews.tsx
```diff
@@ -103,12 +103,12 @@ const useSheetStyle = StyleSheet.createThemeHook((theme) => {
  *  - If the station capacity is 3 or more : show 1 car image and the plus text
  *
  *  @param stationCapacity: Station capacity
- *  @param vehicleTypesAvailable: Types of available cars, array of CarAvailabilityFragment
+ *  @param vehicleTypesAvailable: Types of available cars
```
jorunnl:
```suggestion
 *  @param vehicleTypesAvailable: Types of available vehicles
```

src/modules/mobility/components/CarImage.tsx
```diff
@@ -5,7 +5,7 @@ import {CarFill} from '@atb/assets/svg/mono-icons/transportation';
 import {StyleSheet} from '@atb/theme';
 
 type Props = {
-  uri: string | undefined;
+  uri: string | null | undefined;
```
jorunnl, line 8: Is this necessary? 👀

src/api/types/mobility.ts
```diff
@@ -225,6 +225,42 @@ export const VehicleSchema = z.object({
 
 export type Vehicle = z.infer<typeof VehicleSchema>;
 
+const StationVehicleTypeSchema = z.object({
```
jorunnl, line 228: I find it confusing that this is not just called `StationVehicleSchema`

jorunnl (commented):
tiny feedbacks

marius-at-atb (approved):
Looks great!

src/modules/mobility/components/CarImage.tsx
adriansberg: I'd rather type it like this than doing a check when passing the prop. Since it can be `null` it couldn't be passed directly, and had to be checked and pass `undefined` instead (`?? undefined`).

src/api/types/mobility.ts
adriansberg: The naming follows the gql schema from Entur, and IMO we should follow that.

PR #6019, opened 2026-04-24
refactor: move fare-zones-selector as its own module

reidzeibel:
## Description
Tidy up, fare-zones-selector can be a module 👍🏼


rosvik (approved):
🙌

PR #6024, opened 2026-04-27
feat: add info about ending trip for regular citybikes

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23754

Adding a infobox on the activeShmoSheet when there is a regular citybike trip, since there should not be possible to end the trip manually without the user placing the bike in a parkingslot. Also added the possibility to reset the removed info state from debug menu for testing. But in production, if the user closes the info, it should not appear again

image

### DIFF
--- a/src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
+++ b/src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
@@ -45,6 +45,10 @@ import {useOperators} from '../../use-operators';
 import {SupportButton} from '../SupportButton';
 import {BrandingImage} from '../BrandingImage';
 import {isValidKey} from '@atb/stacks-hierarchy';
+import {EndManualTripCard} from '../EndManualTripCard';
+import {ThemedCityBike} from '@atb/theme/ThemedAssets';
+import {usePersistedBoolState} from '@atb/utils/use-persisted-bool-state';
+import {storage, StorageModelKeysEnum} from '@atb/modules/storage';
 
 type Props = {
   navigateSupportCallback: () => void;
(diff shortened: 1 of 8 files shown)


src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
```diff
@@ -231,32 +240,42 @@ export const ActiveShmoSheet = ({
...
+                      type="error"
+                      message={formatFriendlyShmoErrorMessage(
+                        sendShmoBookingEventError,
+                        t,
+                      )}
+                    />
+                  )}
+                  {activeBooking.asset.formFactor === FormFactor.Bicycle &&
+                  activeBooking.asset.propulsionType ===
+                    PropulsionType.Human ? (
```
marius-at-atb, line 267: This is an interesting one. Vehicle type and geofencing zones together may define rules that apply for parking. E.g. in the future, there may be electric vehicles that are free floating in the outskirts of the city, and station based in the city centre.

We can come back to this later, but it means we should calculate on the server what the parking rules for a vehicle are. May impact design as well.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_CityBikeStartTripScreen.tsx
```diff
@@ -86,8 +86,7 @@ export const Map_CityBikeStartTripScreen = ({navigation}: Props) => {
           <View style={styles.title}>
             <ThemeIcon size="large" svg={Parking} />
             <ThemeText style={theme.typography.heading__xl}>
-              {t(MobilityTexts.cityBike.location(1))} // TODO: UPDATE WITH
-              LOCATION WHEN ITS ADDED TO ENTUR API
+              {t(MobilityTexts.cityBike.location(1))}
```
marius-at-atb, line 89: Still hard coded. Is this explicitly part of an issue to fix?

marius-at-atb (approved):
Lgtm

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_MapStack/Map_CityBikeStartTripScreen.tsx
Dahly96: yes, or until we acctually have the field. It does not exist yet from either Urban sharing, Entur or us

PR #6027, opened 2026-04-27
feat: add enrollmentIds to ruleVariables for announcements

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23670

## Description
Adds enrollmentIds to ruleVariables for announcements, so announcement can be showed only to enrolled users (or only to unenrolled users)

image


adriansberg (approved):
This looks good to me! :shipit:

PR #6031, opened 2026-04-28
fix: compass is below bonus button and visible with bottom sheet on Map

jorunnl:
## Issue Reference

## Description
- Uses actual button height to calculate compass position
- Button is also visible when bottom sheet is visible
image


adriansberg (approved):

adriansberg:
I did not test Android

PR #6033, opened 2026-04-28
fix: avoid gfz layers blinking on vehicle toggle

marius-at-atb:
This fixes a bug where when toggling between vehicles in dark mode, with 3d it would sometimes hide layers and then go to daylight very briefly, and then show layers again with night light back.

Also cleans up the code and uses rnmapbox as intended, which is possible after the [patch](https://github.com/AtB-AS/mittatb-app/pull/6013).

GeofencingZonesAsTiles refactored to use one layer for all zone codes, where the order is instead handled with a sortKey. (Reducing amount of layers is good for performance.)

Acceptance Criteria:
- [x] Toggling vehicles to show and hide vehicles and geofencing zones should be smooth, also in dark mode.
- [x] With the gfz as tiles feature flag false, the zones should now show below the buildings in 3d (same with true and false)


Dahly96 (approved):
lgtm

PR #6036, opened 2026-04-29
fix: use useSafeAreaInsets to handle android compass bonusButton overlap

jorunnl:
## Issue Reference

## Description
Now the button does not overlap on android either:

image

### DIFF
--- a/src/modules/map/Map.tsx
+++ b/src/modules/map/Map.tsx
@@ -17,6 +17,7 @@ import {
 import {Feature} from 'geojson';
 import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {Platform, View} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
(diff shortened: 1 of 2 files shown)


strandlie:
This is a little nitpicky, but if we were to implement `SafeAreaInsets` properly they should be used for everything "touching" the edge of the screen, and not just the Points-button. 🤓

jorunnl:
> This is a little nitpicky, but if we were to implement `SafeAreaInsets` properly they should be used for everything "touching" the edge of the screen, and not just the Points-button. 🤓

The `safeAreaTop` here is actually added to the compass position, not the Points-button. The Points-button (and other MapButtons) already gets `SafeAreaInsets` from `useControlPositionsStyle`.
The problem on Android was that Mapbox does not automatically account for `SafeAreaInsets` when positioning its ornaments. On iOS on the other hand it always adds `SafeAreaInsets` when positioning ornaments.

I will apply `safeAreaTop` to the compass also when the button is not visible, too keep the look consistent between Android and iOS.

strandlie (approved):

PR #6040, opened 2026-04-30
feat: Added rounded transport mode icons for TravelCard

strandlie:
## Issue Reference
Part of https://github.com/AtB-AS/kundevendt/issues/23634

## Description
Adds rounded icons to the TravelCard, according to Figma sketches. 

| Before | After |
|--------|-------|
|  |  |

### DIFF
--- a/src/components/icon-box/CounterIconBox.tsx
+++ b/src/components/icon-box/CounterIconBox.tsx
@@ -16,6 +16,7 @@ export const CounterIconBox = ({
   size = 'normal',
   spacing = 'compact',
   style,
+  rounded,
   notification,
 }: {
   count: number;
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):
LGTM, yet more props to think about 🤷🏼 (no this is not related to your changes 😆)

rosvik (approved):
🚀

strandlie:
> LGTM, yet more props to think about 🤷🏼 (no this is not related to your changes 😆)

It's a valid point! But in this case I think it makes sense, since we are adding the `rounded` prop to `CounterIconBox`, and the `roundedProp` already exists on `TransportationIconBox`

PR #6041, opened 2026-04-30
feat: end trip citybike from station

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23778

Added handling for when Entur ends a trip when a citybike is placed into a station.

Added State value to the streamevent, enabling the possibillity to show the tripsummary when we receive a finished event.

image


src/index.tsx
```diff
@@ -96,19 +96,19 @@ export const App = () => {
...
+                                <FiltersContextProvider>
+                                  <SearchHistoryContextProvider>
+                                    <FirestoreConfigurationContextProvider>
+                                      <TicketingContextProvider>
+                                        <StoredTripPatternsContextProvider>
+                                          <MobileTokenContextProvider>
+                                            <AppLanguageContextProvider>
+                                              <GeolocationContextProvider>
+                                                <MapContextProvider>
+                                                  <EventStreamContextProvider>
```
adriansberg, line 111: Why was this moved?

src/modules/event-stream/use-setup-event-stream.ts
```diff
@@ -49,12 +56,36 @@ export const useSetupEventStream = () => {
       Bugsnag.leaveBreadcrumb('Received event from stream', {
         data: event.data,
       });
+
+      const handleFinishedBookingEvent = async (bookingId: string) => {
```
adriansberg: This now adds logic related to an implementation somewhere else in the app. I haven't put any thought into how the event stream should integrate with other parts of the app, but my initial thought was that having implementation specific stuff in here was not optimal. Maybe passing the `dispatchMapState` to the handler as a util for the handler to use for integrating with other parts of the app might be a more preferable solution.

src/index.tsx
Dahly96: To be able to use useMapContext inside the EventStreamContext. Had a discussion about it in the office. Seems like the eventLog in the EventStreamContext is only used in debugging, so there was nothing stopping it from being lower in the three

src/modules/event-stream/use-setup-event-stream.ts
```diff
@@ -49,12 +56,36 @@ export const useSetupEventStream = () => {
...
+        userId,
+        {
+          isEventStreamFareContractsEnabled,
+        },
+        handleFinishedBookingEvent,
+      );
+
+      if (streamEvent.event === EventKind.ShmoBookingUpdated) {
+        streamEvent.bookingId;
+      }
```
marius-at-atb:
```suggestion
```

src/modules/event-stream/use-setup-event-stream.ts
marius-at-atb: Interesting case of simplicity vs separation of concerns.

From before, we have one global event handler for all stream events.
As long as all use cases only rely on invalidating queries, that works fairly well.

However in this case it would also require a useEffect with more complexity than a direct handler, which likely isn't worth it.

This may be the case for many upcoming events as well. E.g. we have plans to support in-app notifications by showing a global snackbar.

The problem could be solved with a new hook from EventStreamContext, such as:
```ts
useStreamEventListener(eventKind: EventKind, handler: (streamEvent: StreamEvent) => void)
```

However that sounds like a separate PR to me, and maybe something @rosvik wants to have a say in?

src/modules/event-stream/use-setup-event-stream.ts
Dahly96: yeah, this should not be there

src/modules/event-stream/use-setup-event-stream.ts
rosvik: Kind of expected that situation would appear where a simple query invalidation wouldn't be enough when handling events, so you're very welcome to change it 😄 I like the `useStreamEventListener()` suggestion. 

Seems to me like we have two kinds of events / event handling:
1. Global events (query invalidation) - events that can happen at any time, in any context, and affects global state.
2. Local events (finishedBookingEvent, maybe bus in map in the future) - events that have an affect on a certain screen or as a response to a user action. Affects local state.

Would be nice if a new solution handles both of those cases.

But yes, that's another PR. This looks like a fine solution to me for the time being.

src/modules/event-stream/handle-stream-event.ts
```diff
@@ -24,6 +26,17 @@ export const handleStreamEvent = (
...
     case EventKind.ShmoBookingUpdated:
+      const existingBooking: ShmoBooking | undefined = queryClient.getQueryData(
+        getShmoBookingQueryKey(streamEvent.bookingId, languageGlobal),
+      );
+
+      if (
+        existingBooking?.state !== ShmoBookingState.FINISHED &&
+        streamEvent.state === ShmoBookingState.FINISHED
+      ) {
+        await handleFinishedBookingEvent(streamEvent.bookingId);
```
rosvik: Do we need to make this (and `handleStreamEvent`) async? Could you try to remove it?

```suggestion
        handleFinishedBookingEvent(streamEvent.bookingId);
```

src/modules/event-stream/use-setup-event-stream.ts
marius-at-atb: Indeed. Created a [new issue](https://github.com/AtB-AS/kundevendt/issues/23848) for it, which we can prioritise as  part of the normal flow.
Feel free to add more details :+1:

marius-at-atb (approved):
LGTM

rosvik (approved):

PR #6043, opened 2026-05-05
feat: add price info on bottomsheets

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23813

Added price info to BottomSheets with a priceInfo screen

[Figma](https://www.figma.com/design/YfLcr1BKOuaXk9HIUilXJK/City-bike-integration-V1?node-id=1109-29319&p=f&t=fzRWkx2aWjGAaklV-0)

image

image

image

image


strandlie:
Having a separate "Kampanje" information object like this feels like we are breaking some pattern in the app, and just passing the data model straight through instead of communicating the relevant information to the user. 

What about something like this:

prisinfo

gorandalum:
> Having a separate "Kampanje" information object like this feels like we are breaking some pattern in the app, and just passing the data model straight through instead of communicating the relevant information to the user.

I wondered about this when looking at the screenshots. How do I know what will be my price, the standard price or the campaign price?

Dahly96:
> Having a separate "Kampanje" information object like this feels like we are breaking some pattern in the app, and just passing the data model straight through instead of communicating the relevant information to the user.
> 
> What about something like this:
> 
> prisinfo

after talking with Design we decided to only show campaign prices if there is a campaign, not showing the original price above aswell. I have updated the screenshots

src/stacks-hierarchy/Root_ShmoPricingDetailsScreen.tsx
```diff
@@ -0,0 +1,187 @@
...
+          effectiveStartMinute,
+          perMinPricingSegment.end,
+        ),
+      );
+    }
+    if (hasMultiplePerMinPricingPlans) {
+      return t(
+        MobilityTexts.pricingDetails.minutePriceFrom(effectiveStartMinute),
+      );
+    }
```
marius-at-atb: ` if (hasMultiplePerMinPricingPlans)` repeated unnecessarily. 
Can just do 
 ```ts
if (hasMultiplePerMinPricingPlans) {
   if (perMinPricingSegment.end != null) {
   ...
   } else {
   ...
   }
}
```

File: src/stacks-hierarchy/Root_ShmoPricingDetailsScreen.tsx [hunk 2]
```diff
@@ -0,0 +1,187 @@
...
+      }}
+      headerContent={(ref) => (
+        <ScreenHeading
+          ref={ref}
+          text={t(MobilityTexts.pricingDetails.priceInfo)}
+        />
+      )}
+      contentColor={contentColor}
+    >
+      <ScrollView contentContainerStyle={styles.container}>
```
marius-at-atb, line 112: The content inside ScrollView looks bloated.
Both hasCampaign and !hasCampaign has a section where all rows have text on the left and right.
In both cases, it can be just one component/mapping that takes in an array of rows to render.
Try to instead put the logic for which data to show in the array.

The pricingPlan.perMinPricing mapping is also mostly duplicated, see if you can be more concise.

src/stacks-hierarchy/Root_ShmoPricingDetailsScreen.tsx
```diff
@@ -0,0 +1,187 @@
...
+  );
+
+  const firstPerMinPricingPlanRate = pricingPlan.perMinPricing?.[0]?.rate ?? 0;
+  const freeMinCount =
+    freeMinutesPriceAdjustment && firstPerMinPricingPlanRate > 0
+      ? computeFreeMinuteCount(
+          freeMinutesPriceAdjustment,
+          firstPerMinPricingPlanRate,
+        )
+      : 0;
```
marius-at-atb: Looking at it now, it's not that complicated to support correct handling even if the priceAdjustments exceeds the max cost of the first rate, so let's just ensure correctness even with surprising data.

marius-at-atb (changes requested):

src/stacks-hierarchy/Root_ShmoPricingDetailsScreen.tsx
Dahly96: Sure

src/modules/mobility/utils.ts
```diff
@@ -264,8 +265,21 @@ export const getFreeMinutes = (
...
   freeMinutes: PriceAdjustmentType,
-  rate: number,
-): number => Math.floor(Math.abs(freeMinutes.amount) / rate);
+  perMinPricing: ShmoPricingSegment[],
+): number => {
+  //Calculating free minutes based on the per min price segments, until the budget runs out or there are no more segments with a cost
+  let budget = Math.abs(freeMinutes.amount);
+  let total = 0;
+  for (const segment of perMinPricing) {
+    if (budget <= 0 || segment.rate <= 0) break;
```
marius-at-atb: I don't think we should break on `segment.rate <= 0`, since there might be a segment with rate > 0 later. To avoid dividing by 0, just use the `segLength` as minutes instead in that case.
Then when returning total after the loop, probably just cap it to some common sense max number (e.g. 3h = 180min).

Otherwise lgtm!

marius-at-atb (approved):
Lgtm!

PR #6044, opened 2026-05-05
feat: replace Firestore bonusProducts with GET /bonus/v2/products

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23798

## Description
Frontend now gets data about bonus products from backend instead of firestore. This is used both on Bonus Screen under Profile and relevant bottom sheets.

Other minor changes:
- HowPointsWork is in separate file because it got very long (also might be reused for bottom sheets about Bonus in the Future)
- Created useRelevantBonusProduct hook to simplify logic in bottom sheet
- Created useIsBonusEnrollable hook since logic is now used two places (and easier to understand with a readable name)

### Screenshots
image
image
image


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/BonusHowPointsWork.tsx
```diff
@@ -0,0 +1,214 @@
...
+  return (
+    <>
+      {[
+        ...new Set(
+          activeBonusProducts
+            ?.filter(
+              (product) =>
+                product.productType === BonusProductTypeEnum.enum.VOUCHER,
+            )
+            .flatMap((product) => product.operatorIds) ?? [],
```
adriansberg: Maybe this could be extracted to a helper function to help readibility?

src/modules/bonus/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  'VOUCHER',
+  'TICKET',
+  'SHARED_MOBILITY',
+]);
+
+export const BonusProductSchema = z.object({
+  id: z.string(),
+  isActive: z.boolean(),
+  operatorIds: z.array(z.string()),
+  formFactors: z.array(
```
adriansberg: Could use the `FormFactorSchema` directly here?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/BonusHowPointsWork.tsx
jorunnl: I can move it outside the JSX into a named variable at least 🙆‍♀️

adriansberg (approved):
LGTM! Fix that unnecessary call to `useIsBonusActiveForUser` and I think it's good.

PR #6046, opened 2026-05-06
feat:  use new endpoint to claim bonus product vouchers

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23799

## Description
Updates OperatorActionButton to use new endpoint in bonus service to claim bonus products with type VOUCHER.

Also:
- removes bonus-checkbox from integrated bicycle bottom sheet, since bonus products for SHARED_MOBILITY are not supported yet

image image image


src/modules/bonus/use-relevant-bonus-product.ts
```diff
@@ -1,9 +1,11 @@
 import {FormFactor} from '@atb/api/types/generated/mobility-types_v2';
 import {useActiveBonusProductsQuery} from './queries';
+import {BonusProductType} from './types';
 
 export const useRelevantBonusProduct = (
   operatorId: string | undefined,
   formFactor: FormFactor,
+  productType?: BonusProductType['productType'],
```
adriansberg: Is there no separate type for the product type?

src/modules/mobility/components/BicycleSheetNotIntegratedView.tsx
```diff
@@ -51,7 +52,11 @@ export const BikeStationNotIntegratedView = ({
 
   const {operatorBenefit} = useOperatorBenefit(operatorId);
   const isBonusActiveForUser = useIsBonusActiveForUser();
-  const bonusProduct = useRelevantBonusProduct(operatorId, FormFactor.Bicycle);
+  const bonusProduct = useRelevantBonusProduct(
+    operatorId,
+    FormFactor.Bicycle,
+    BonusProductTypeEnum.enum.VOUCHER,
```
adriansberg: Should an actual enum be used instead of the Zod type enum?

adriansberg (approved):
Looks good, a couple comments

src/modules/bonus/use-relevant-bonus-product.ts
jorunnl: fixed now

src/modules/mobility/components/BicycleSheetNotIntegratedView.tsx
jorunnl: Agreed, fixed now

adriansberg (approved):
Great work!

PR #6048, opened 2026-05-06
feat: only show refresh control when it is user initialized

rosvik:
## Issue Reference

fixes https://github.com/AtB-AS/kundevendt/issues/23323

## Description

Adds a new `useManualRefreshControlProps` hook that return refresh control props that only shows as refreshing when it is initiated by the user. This means that when navigating to a page that reloads on focus, the spinner won't trigger and get 'stuck' as described in https://github.com/AtB-AS/kundevendt/issues/23323.

Since the pull to refresh indicator isn't visible on open anymore, I added loading indicators on SPAR and payment method screens that shows on initial load.


marius-at-atb (approved):
Lgtm!

PR #5941, opened 2026-03-30
feat: Experimentally enable new TravelCard in search and details

cbrevik:

PR #6050, opened 2026-05-06
feat: Experimentally enable new travel card in details

strandlie:
## Issue Reference
Duplicate of [this PR](https://github.com/AtB-AS/mittatb-app/pull/5941). Created a new one because of conflicts. 

## Description
This PR adds the new TravelCard to trip-search behind a feature toggle

simulator_screenshot_BC3C241E-0B11-4D3C-9886-26C3A0681B10

### Other small changes
- Expands the experimentalFeature function to take a featureToggle name as a parameter. In practice that means that we OR the previous experimentalFeature logic and the featureToggle, so if either is true we display the logic. 
- Also tweaks some parameters to reduce accidental saving of trips. Also made this configurable through RemoteConfig, if we want to.


reidzeibel (approved):
LGTM

tormoseng:
👍  Notified @Sebstorvik and @helenekulsetas about the swipe sensitivity on Slack.

PR #6053, opened 2026-05-07
fix: Keep isStationBasedBooking value

marius-at-atb:
First commit: Fixes a bug that `isStationBasedBooking` was immediately lost due to dispatch of `BICYCLE`.
Second commit: Simplify isStationBasedBooking management. Fixes a bug where if useVehicle was used anywhere without isStationBasedBooking=true, it would overwrite the mock data.


adriansberg (approved):

tormoseng:
👍

PR #6057, opened 2026-05-08
feat: add bonus checkbox to scooter sheet

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/23800

## Description
Adds checkbox to use bonus on scooter sheet.
Functionality:
- If there exists a BonusProduct for current operator and FormFactor, checkbox shows
- If checkbox is checked, price updates are highlightet
- When button is pressed BonusProductId (if checkbox is active) is sent in initShmoOneStopBooking call to MobilityService, priceadjustments are applied there

Worth noting:
- Price info is removed from activeShmoSheet, as this will not be there in updated future design anyways

image  image


adriansberg (approved):
Clean!

PR #6059, opened 2026-05-11
Add BonusProductGroup support and update BonusProduct types

adriansberg:
## Summary

- Add `BonusProductGroupSchema`/`BonusProductGroupType` with `id`, `description`, `price: {amount, currency}`
- Add `getActiveBonusProductGroups` API call and `useActiveBonusProductGroupsQuery` hook
- Update `BonusProductSchema`: `operatorId` (singular string), `bonusProductGroupId`, `price` as `{amount, currency}` object
- Update `BonusProductList` to render one row per group, aggregating form factors and operator names from member products
- Fix all `operatorIds` → `operatorId` references across components and hooks

## Test plan

- [ ] Bonus screen shows one row per product group
- [ ] Map button navigates with correct form factors
- [ ] Pay with points checkbox shows correct point amount
- [ ] Voucher operator links render correctly in "How points work" screen

Resolves https://github.com/AtB-AS/kundevendt/issues/23862

### DIFF
--- a/src/modules/bonus/components/BonusProductList.tsx
+++ b/src/modules/bonus/components/BonusProductList.tsx
@@ -4,7 +4,7 @@ import {GenericSectionItem, Section} from '@atb/components/sections';
 import {ThemeText} from '@atb/components/text';
 import {BonusStarFill} from './BonusStarFill';
 import {useFirestoreConfigurationContext} from '@atb/modules/configuration';
-import {BonusProductType} from '@atb/modules/bonus';
+import {BonusProductGroupType, BonusProductType} from '@atb/modules/bonus';
 import {
   BonusProgramTexts,
   getTextForLanguage,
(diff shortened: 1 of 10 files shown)


src/modules/bonus/components/BonusProductList.tsx
```diff
@@ -57,100 +57,109 @@ export const BonusProductList = ({bonusProducts, onNavigateToMap}: Props) => {
...
-                      )
-                      .join(', ')}
-                  </ThemeText>
+      {bonusProductGroups.map((group) => {
+        const memberProducts = bonusProducts.filter(
+          (bp) => bp.bonusProductGroupId === group.id,
+        );
+        const formFactors = [
+          ...new Set(memberProducts.flatMap((bp) => bp.formFactors)),
+        ] as FormFactor[];
```
marius-at-atb, line 66: `as FormFactor[]` can be replaced by `Set<FormFactor>` I think

marius-at-atb (approved):
Nice!

src/modules/bonus/components/BonusProductList.tsx
adriansberg: The set is spread into an array, so the type looks to be correct. Could probably refactor to use Set methods instead of map methods, if that's what you meant.

PR #6063, opened 2026-05-12
fix: prevent android overflow on ChooseTicketRecipientScreen

rosvik:
fixes https://github.com/AtB-AS/kundevendt/issues/23892

`ChooseTicketRecipientScreen` didn't use `FullScreenView` which caused a bug with safe area on android.

### DIFF
--- a/src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/Root_ChooseTicketRecipientScreen.tsx
+++ b/src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/Root_ChooseTicketRecipientScreen.tsx
@@ -1,10 +1,8 @@
-import {FullScreenHeader} from '@atb/components/screen-header';
 import {RootStackScreenProps} from '@atb/stacks-hierarchy/navigation-types';
 import {StyleSheet, useThemeContext} from '@atb/theme';
 import {OnBehalfOfTexts, useTranslation} from '@atb/translations';
 import {useCallback, useRef} from 'react';
-import {KeyboardAvoidingView, RefreshControl, View} from 'react-native';
-import {ScrollView} from 'react-native-gesture-handler';
+import {KeyboardAvoidingView, View} from 'react-native';
 import {animateNextChange} from '@atb/utils/animation';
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/Root_ChooseTicketRecipientScreen.tsx
```diff
@@ -47,19 +46,16 @@ export const Root_ChooseTicketRecipientScreen = ({
...
-        >
+    <FullScreenView
+      headerProps={{
+        leftButton: {type: 'back'},
+        title: t(OnBehalfOfTexts.chooseReceiver.header),
+      }}
+      refreshControlProps={refreshControlProps}
+      focusRef={focusRef}
+    >
+      <KeyboardAvoidingView behavior="position">
```
gorandalum, line 57: We have normally used behavior "padding". Is there a reason this instance needs "position". And if "position" is better than "padding", should some (or all) other places be changed as well?

gorandalum (approved):
Looks good!

src/stacks-hierarchy/Root_ChooseTicketRecipientScreen/Root_ChooseTicketRecipientScreen.tsx
rosvik: Right! "padding" was used before as well, but that didn't work when I put it inside the FullScreenView. Looked into it, and the react native documentation is [lacking](https://reactnative.dev/docs/keyboardavoidingview#behavior). [This stack overflow answer](https://stackoverflow.com/a/58385319) gives a bit more detail, but not enough for me to understand if any of them are "better".

Added some test input to the issue to get this tested on android after the change.

reidzeibel: The link to the stackOverflow answer is wrong I think

rosvik: Sorry! https://stackoverflow.com/a/58385319

PR #6069, opened 2026-05-15
feat: add EditActionSectionItem section component

gorandalum:
Resolves [#23866](https://github.com/AtB-AS/kundevendt/issues/23886)

A section item with optional left icon, primary text, optional
inline subtext, and a right icon defaulting to the Edit icon.
Text and subtext wrap inline when they exceed available width.

Adds a dedicated Storybook story and includes the component in
the AllSectionItemsList story.

<div>

</div>

### DIFF
--- a/src/components/sections/items/EditActionSectionItem.tsx
+++ b/src/components/sections/items/EditActionSectionItem.tsx
@@ -1,0 +1,70 @@
+import {Edit} from '@atb/assets/svg/mono-icons/actions';
+import {NativeBlockButton} from '@atb/components/native-button';
+import {ThemeText} from '@atb/components/text';
+import {ThemeIcon, ThemeIconProps} from '@atb/components/theme-icon';
+import {StyleSheet, useThemeContext} from '@atb/theme';
+import React, {forwardRef} from 'react';
+import {AccessibilityProps, View} from 'react-native';
+import {SectionItemProps} from '../types';
+import {useSectionItem} from '../use-section-item';
+import {useSectionStyle} from '../use-section-style';
+
+type Props = SectionItemProps<
+  {
+    leftIcon?: ThemeIconProps['svg'];
+    rightIcon?: ThemeIconProps['svg'];
+    text: string;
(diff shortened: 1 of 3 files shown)


src/components/sections/items/EditActionSectionItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+          />
+        </View>
+      </NativeBlockButton>
+    );
+  },
+);
+
+const useStyles = StyleSheet.createThemeHook((theme) => ({
+  button: {flex: 1},
+  gap: {gap: theme.spacing.small, alignItems: 'flex-start'},
```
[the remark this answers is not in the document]
gorandalum: Did a slight reorganisation of the styles now.

src/components/sections/items/EditActionSectionItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <NativeBlockButton
+        ref={ref}
+        accessible
+        accessibilityRole="button"
+        onPress={onPress}
+        style={[topContainer, styles.button]}
+        testID={testID}
+        {...props}
+      >
+        {leftIcon && <ThemeIcon svg={leftIcon} size="normal" />}
```
reidzeibel, line 38: The right icon has `color` props defined, is there any reason why this should not have any color? 

Would it make sense to pass a `{children}` (a `<ThemeIcon>` component) instead of just SVG here?

gorandalum: In the design the component has no color on the left icon and blue color (or interactive.0.default.background) on the right icon. The blue color is ment to signal that it is an edit action. Some screenshots can be seen in this issue:
https://github.com/AtB-AS/kundevendt/issues/20169

Since that as of now are the limitations of the color choices in the design I think the component can have the same limitations. If we later see that the component needs to support a wide variety of colors then we can add more flexibility by loosening up the parameters :)

reidzeibel (approved):

PR #6070, opened 2026-05-15
feat: added terms for citybike

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23078

Added termspage for citybike onboarding. I made the terms component reusable between scooter and bicycle and separated the agreement logic based on formfactor. So you have to agree to both individually.

image

image


src/modules/configuration/converters.ts
```diff
@@ -211,6 +211,24 @@ export function mapToScooterConsentLines(scooterConsentLines?: any) {
     .filter(isDefined);
 }
 
+export function mapToBicycleConsentLines(bicycleConsentLines?: any) {
```
marius-at-atb: Can just rename `mapToScooterConsentLines` to `mapToConsentLines` and re-use that instead.

src/modules/mobility/components/onboarding/RulesScreenComponent.tsx
```diff
@@ -52,36 +84,46 @@ export const RulesScreenComponent = ({
...
-        backgroundColor: theme.color.background.neutral[0],
-        rightIcon: {svg: ExternalLink},
-      }}
-      contentNode={<ListElement scooterConsentLines={scooterConsentLines} />}
+      contentNode={
+        <ListElement
+          consentLines={consentLines}
+          privacyLinkText={t(MobilityTexts.shmoRequirements.terms.privacyLink)}
+          onPrivacyLinkPress={onOpenTerms}
+          showPrivacyLink={!!mobilityTermsUrl}
```
marius-at-atb, line 96: naming: privacy link or terms?

marius-at-atb (approved):
Nice work!

src/modules/mobility/components/onboarding/RulesScreenComponent.tsx
Dahly96: the text says "Personvern", so i would say privacy

PR #6071, opened 2026-05-15
refactor: improve how colors are passed to ThemeText

strandlie:
## Summary

`ThemeText` has two separate color-related props: `color` (which `ContrastColor` to derive a foreground from) and `type` (which what "mode" the text should be in (`primary`, `secondary`, `disabled`)). Many call sites were confusing these by passing `"primary"` or `"secondary"` to `color`, which is wrong and unnecessary when we already have the type prop to do the exact same thing. This PR cleans that up.

**Core changes:**
- Added `resolveColorValue()` to `src/theme/colors.ts` — a tested, pure helper that resolves any color input (`ContrastColor`, status name, text-color name, or raw hex) to a final `string`. Centralises logic that was previously duplicated in both ThemeText and ThemeIcon.
- Added unit tests in `src/theme/__tests__/resolve-color-value.test.ts` covering all branches.
- Narrowed the public `ThemeText` `color` prop type to `ContrastColor | ColorValue` — string semantic aliases (`"primary"`, `"info"`, etc.) are no longer part of the documented API. 
- Removed the unused `color` prop from `ScreenHeading` (no caller ever supplied a non-default value).

**Call-site migrations (~70 files, no visual change):**
- `color="primary"` / `color="secondary"` / `color="disabled"` → `type="primary"` etc.
- `color={theme.color.foreground.dynamic.primary}` → `type="primary"` (removes manual theme lookup for a value `ThemeText` already derives internally)

## Test input
- This has changed colors in a lot of files, but the main thing to check is that no colors are wrong or different in text or icons. In my testing I could not find anything wrong

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -194,6 +194,9 @@ module.exports = {
 
     // Error on React hooks being called inside StyleSheet.createThemeHook
     'rulesdir/no-hooks-in-create-theme-hook': 'error',
+
+    // Error on passing TextColor string literals as the color prop of ThemeText
+    'rulesdir/no-text-color-in-theme-text': 'error',
   },
   overrides: [
     // Allow 'export default' from these paths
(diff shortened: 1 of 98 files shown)


gorandalum (approved):
Looks like a good change! 👍

reidzeibel (approved):

PR #6081, opened 2026-05-18
fix: Prevent undefined values to propagate

strandlie:
## Description
Small fix of a bug. This is just a fallback, but is properly fixed in the BFF here: https://github.com/AtB-AS/atb-bff/pull/452


reidzeibel (approved):

rosvik:
Once we get our types in order, I guess this shouldn't be necessary.

strandlie:
Not necessary.

PR #6082, opened 2026-05-19
feat: show tariff zones in the regular map view

gorandalum:
Add tariff zone boundary lines and labels to the main map, with a
toggle in the map filter sheet. Extract shared
TariffZoneLinesAndLabels component and mapZonesToPolygonCollection
utility to reuse across the fare zone selector map and the regular
map.

- The new design on the zone lines is added to both general map and the existing zone selector map
- The new design on zone labels was not easy to implement, so the labels are kept as is in both places

### DIFF
--- a/src/modules/map/components/TariffZoneLinesAndLabels.tsx
+++ b/src/modules/map/components/TariffZoneLinesAndLabels.tsx
@@ -1,0 +1,75 @@
+import MapboxGL from '@rnmapbox/maps';
+import {Expression} from 'node_modules/@rnmapbox/maps/src/utils/MapboxStyles';
+import turfCentroid from '@turf/centroid';
+import {FeatureCollection, Point, Polygon} from 'geojson';
+import React from 'react';
+import {useThemeContext} from '@atb/theme';
+
+const labelZoomOpacity: Expression = [
+  'interpolate',
+  ['linear'],
+  ['zoom'],
+  10,
+  1,
+  10.5,
+  0,
+];
+
+type TariffZonesProps = {
+  polygonCollection: FeatureCollection<Polygon>;
+  showLabelsAtAllZoom?: boolean;
+};
+
+export const TariffZoneLinesAndLabels = ({
+  polygonCollection,
+  showLabelsAtAllZoom,
+}: TariffZonesProps) => {
+  const {theme} = useThemeContext();
+
+  const labelPointsCollection =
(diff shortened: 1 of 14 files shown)


marius-at-atb:
In order to clearly separate GeofencingZones and TariffZones, could you update the zone naming to be prefixed with tariff?

src/modules/map/components/TariffZoneLinesAndLabels.tsx
```diff
@@ -0,0 +1,75 @@
...
+  polygonCollection,
+  showLabelsAtAllZoom,
+}: TariffZonesProps) => {
+  const {theme} = useThemeContext();
+
+  const labelPointsCollection =
+    mapPolygonsToLabelPointsCollection(polygonCollection);
+
+  return (
+    <>
```
strandlie, line 33: Is this fragment necessary? Maybe the Mapbox component needs _something_? Or?

src/modules/map/components/TariffZoneLinesAndLabels.tsx
```diff
@@ -0,0 +1,75 @@
...
+  10.5,
+  0,
+];
+
+type TariffZonesProps = {
+  polygonCollection: FeatureCollection<Polygon>;
+  showLabelsAtAllZoom?: boolean;
+};
+
+export const TariffZoneLinesAndLabels = ({
```
strandlie, line 23: This code is very similar to the code for showing the [geofencing zones for scooters](https://github.com/AtB-AS/mittatb-app/blob/master/src/modules/map/components/mobility/GeofencingZones.tsx). Do we really need a separate implementation for this? 

It is at least tempting to propose a refactor of that code instead, such that it can be used for this as well,

gorandalum: I think it is a good suggestion, and I gave it a quick try. However I found that reusing a shared component was better in theory than in practice. The two implementations are quite different with regards to styling (Geofencing zones use dynamic MapboxGL style expressions from custom props), onPress logic, label implementation (text vs icon), and some more. All in all based on my quick testing I think it is better to keep them separate for now.

strandlie: Sounds like that component is ripe for a refactor at some point then :)

strandlie (approved):

PR #6083, opened 2026-05-19
fix: map onPress handling with geofencingZonesAsTiles

marius-at-atb:
After https://github.com/AtB-AS/mittatb-app/pull/6033, the geofencingZones would always capture the onPress events when shown. This meant you couldn't go directly from one scooter to another.

Fixed by handling it the same way as for GeofencingZones (geojson) -> using the onFeatureClick handler instead.

Also
- fixed an incorrectly named prop in GeofencingZonePropsSchema. This likely also fixes the missing shmo warnings in the bottom sheet during an active trip.
- added a debug util to clear locally cached tiles.

**Acceptance Criteria:**
- [x] You can click on map items also when a vehicle with geofencingZones as tiles is selected
- [x] Clicks on geofencing zones should show the snackbar correctly
- [x] Warning messages in the active scooter sheet should show correctly
- [x] Still working with the old geojson gfz as well

### DIFF
--- a/src/modules/map/Map.tsx
+++ b/src/modules/map/Map.tsx
@@ -214,21 +214,6 @@ export const Map = (props: MapProps) => {
     [showSnackbar, getGeofencingZoneContent],
   );
 
-  const geofencingZoneOnPress = useCallback(
-    (e: OnPressEvent) => {
-      const featuresAtClick = e.features;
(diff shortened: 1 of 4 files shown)


Dahly96 (approved):
tested, lgtm

PR #6085, opened 2026-05-19
test: adjust search filter tests

tormoseng:
Adjusted the e2e tests for the travel search filters after the changes in https://github.com/AtB-AS/kundevendt/issues/23886.


reidzeibel (approved):
💡

PR #6086, opened 2026-05-19
fix: various booking travel card fixes

reidzeibel:
## Issue Reference
related to https://github.com/AtB-AS/kundevendt/issues/23646

## Description
- Reorder a11y text screenreader for `Travel Card`, now the trip status inside "Tag" will be read earlier, rather than later, so the user gets the availability information earlier before the trip detail information.
- Fixes margin on the old booking layout
- Improves a11y text hint, `Travel Card` is used for:
  - Saved Trips -> activate to Trip Details
  - Trip Search ->  activate to Trip Details
  - Booking Option ->
    - activate to Ticket Summary if purchasing for self
    - activate to Choose Recipient if purchasing for others


tormoseng:
👏  Very good! Tested OK.

marius-at-atb (approved):
Lgtm

PR #6089, opened 2026-05-20
fix: Fix bug with originalTime in TravelCard Header

strandlie:
## Issue Reference
Fixes a bug related to https://github.com/AtB-AS/kundevendt/issues/23634

## Description
Currently, we show the "Opprinnelig xx.xx - yy.yy" text for all tripPatterns that has started. This PR changes so that we only show Opprinnelige if they actually are different.

### DIFF
--- a/src/screen-components/travel-card/TravelCardHeader.tsx
+++ b/src/screen-components/travel-card/TravelCardHeader.tsx
@@ -45,6 +45,14 @@ export const TravelCardHeader: React.FC<
     endTimeLabel: expectedEndTimeLabel,
   } = useTimeLabels(expectedStartTime, expectedEndTime, includeDayInfo);
 
+  /**
+   * If the start time of a trip is in the past (hasStarted), and the
+   * includeDayInfo flag is true, we want to include day info in the
+   * time label.
+   *
+   * See documentation in useTimeLabels for details about special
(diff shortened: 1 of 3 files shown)


src/screen-components/travel-card/TravelCardHeader.tsx
```diff
@@ -45,14 +45,22 @@ export const TravelCardHeader: React.FC<
...
   } = useTimeLabels(expectedStartTime, expectedEndTime, includeDayInfo);
 
+  /**
+   * If the start time of a trip is in the past (hasStarted), and the
+   * includeDayInfo flag is true, we want to include day info in the
+   * time label.
+   *
+   * See documentation in useTimeLabels for details about special
+   * handling of the current day.
+   */
```
gorandalum, line 51: I don't think this comment is really necessary, this is just a more verbose explanation of the code `hasStarted && includeDayInfo`.

gorandalum (approved):

src/screen-components/travel-card/TravelCardHeader.tsx
strandlie: 🫡 I'll remove the top part at least. The bottom part is a slight gotcha, so I think there is value to that.

PR #6094, opened 2026-05-21
feat: added selection of vehicles in station from vehicleTypeId

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23737

Since propulsionType could be the same for different vehicleTypes, iv changed the queries to be dependant on vehicleTypeId, so the correct vehicles is returned for the different queries


src/modules/mobility/queries/use-vehicles-by-propulsion-types-queries.tsx
```diff
@@ -1,34 +1,35 @@
...
       gcTime: ONE_MINUTE_MS,
       staleTime: 5000,
       retry: 3,
     })),
   });
 
-  return {humanQuery, electricQuery, electricAssistQuery};
+  return Object.fromEntries(
+    vehicleTypeIds.map((id, index) => [id, results[index]]),
+  ) as Record<string, (typeof results)[number]>;
```
marius-at-atb: Let's see if we can return the useQueries result directly instead, both avoiding the type cast, and stadardize useQueries usage the same way as for useQuery.

adriansberg (approved):
lgtm

PR #6095, opened 2026-05-21
feat(debug): show dot indicator when server override is active

strandlie:
When a server override is active in the debug menu, there was no visible indication of this anywhere in the app. Easy to forget between sessions.

This adds a red dot at every level of the navigation path leading to the server overrides screen — so the signal is visible regardless of where you are in the app.

| Profile Screen | Debug Menu |
|--------|-------|
|  |  |

## How

- Added a subscriber system to `debug-server-overrides-cache.ts` so React components can react to changes
- New `useDebugServerOverrides` hook that re-renders when overrides change
- Added `standalone` prop to `NotificationIndicator` for inline (non-overlaying) use
- Red dot appears on the **Profile tab icon** (bottom tab bar), the **Debug button** in the profile menu, and the **Server overrides** expandable section in the debug screen


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_RootScreen/Dashboard_RootScreen.tsx
```diff
@@ -444,7 +446,7 @@ const useStyle = StyleSheet.createThemeHook((theme) => ({
     margin: 0,
   },
   contentSection: {
-    marginTop: theme.spacing.medium,
+    marginVertical: theme.spacing.medium,
```
gorandalum: I haven't actually looked in the app, but on first glance this seems a little "risky" with regards to other components having their margin changed.

rosvik:
Nice idea! Could we instead do something like add a notification "dot" on the profile tab icon to show that _some_ server override is active? Would take up less space, and also be visible on other screens. (Search for `getProfileNotification`)

strandlie:
> Nice idea! Could we instead do something like add a notification "dot" on the profile tab icon to show that _some_ server override is active? Would take up less space, and also be visible on other screens. (Search for `getProfileNotification`)

Sure! I'll change to that. That also solves @gorandalum point from above.

strandlie:
@gorandalum @rosvik Changed now.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_RootScreen.tsx
```diff
@@ -297,6 +300,15 @@ export const Profile_RootScreen = ({navigation}: ProfileProps) => {
...
                   testID="debugButton"
+                  leftElement={
+                    serverOverrides.length > 0 ? (
+                      <NotificationIndicator
+                        color={theme.color.status.error.primary}
+                        iconSize="normal"
+                        standalone
+                      />
+                    ) : undefined
+                  }
```
rosvik: You could instead set a `rightIcon` with `ChevronRight` and `notificationColor`. That way we don't need a new "standalone" prop

strandlie: | Screen1 | Screen2 | 
|--------|-------|
|  |  |

I didn't think of that I could override it like that. Great idea. The ServerOverrides ExpandableSectionItem looks a little funky if I use `suffixNode` and remove standalone (shown in the screenshot), but I think thats a price I'm willing to pay for some debug code.

rosvik (approved):
🚀

PR #6098, opened 2026-05-22
fix: paging bug when arriveBy: true

strandlie:
## Issue Reference
Not an issue, but a bug report in Slack: https://mittatb.slack.com/archives/C0116FMPX4Y/p1779427611851349

## Description
When `arriveBy: true` we should not use `nextPageCursor` to fetch more result. Instead we should use ´previousPageCursor`

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-infinite-query.ts
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-infinite-query.ts
@@ -49,7 +49,10 @@ export const useTripsInfiniteQuery = (
     staleTime: 30 * ONE_MINUTE_MS,
     gcTime: 30 * ONE_MINUTE_MS,
     initialPageParam: undefined,
-    getNextPageParam: (lastPage) => lastPage?.trip?.nextPageCursor,
+    getNextPageParam: (lastPage) =>
+      tripsInfiniteQueryProps.arriveBy
+        ? lastPage.trip.previousPageCursor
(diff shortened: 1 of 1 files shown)


src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-infinite-query.ts
```diff
@@ -49,7 +49,10 @@ export const useTripsInfiniteQuery = (
     staleTime: 30 * ONE_MINUTE_MS,
     gcTime: 30 * ONE_MINUTE_MS,
     initialPageParam: undefined,
-    getNextPageParam: (lastPage) => lastPage?.trip?.nextPageCursor,
+    getNextPageParam: (lastPage) =>
+      tripsInfiniteQueryProps.arriveBy
+        ? lastPage.trip.previousPageCursor
+        : lastPage?.trip?.nextPageCursor,
```
gorandalum: Is it actually so that if `tripsInfiniteQueryProps.arriveBy` is `true`, then `lastPage` and `lastPage.trip` will never be undefined. But if `false`, they can be undefined 🤔

gorandalum (approved):

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/use-trips-infinite-query.ts
strandlie: That is a very good point. According to the types yes, but I don't trust it. So I added some ?'s

PR #6099, opened 2026-05-22
feat: improved a11y in happypath vehicles

Dahly96:
There was a issue with textsize in vehicleCards during vehicletrips, that is now fixed. Also gave more space between counter in active trip.

image
image

image


adriansberg (approved):
Looks good!

PR #6113, opened 2026-05-27
feat: add geocoder v3 BFF API support

gorandalum:
Integrates the new /bff/v2/geocoder endpoints backed by Entur's
geocoder v3. A feature toggle (use_geocoder_v3) switches between
the existing v1 and new v3 endpoints at runtime.

Also renames tariff_zones to fare_zones to align with the updated
API response shape.

### DIFF
--- a/src/modules/favorites/types.ts
+++ b/src/modules/favorites/types.ts
@@ -23,7 +23,7 @@ export type SearchLocation = {
   category: FeatureCategory[];
   label?: string;
   postalcode?: string;
-  tariff_zones?: string[];
+  fare_zones?: string[];
   housenumber?: string;
 } & (
   | {resultType: 'search'}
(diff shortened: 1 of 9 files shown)


src/modules/favorites/types.ts
```diff
@@ -23,7 +23,7 @@ export type SearchLocation = {
   category: FeatureCategory[];
   label?: string;
   postalcode?: string;
-  tariff_zones?: string[];
+  fare_zones?: string[];
```
rosvik, line 26: Can this cause backwards compatibility issues with stored favorite locations? It's maybe not in use for favorites?

rosvik (approved):

src/modules/favorites/types.ts
gorandalum: Nope, should be okay. Only used when searching for stop places in the zone selector.

PR #6117, opened 2026-05-27
refactor: replace individual margins with gap in trip search

gorandalum:
Use gap and marginTop on the content container instead of
individual marginTop/paddingTop on child components
(GlobalMessage, CityZoneMessage, DayLabel, NonTransitResults).

Remove the inner wrapper View for from/to results — children
are now direct siblings in the gap layout. GlobalMessage is
rendered without a wrapper so it takes no space when empty.

**Before/After**
<div>

</div>

### Acceptance criteria
- [x] Should no longer be 36px margin above non transit results, but 24px instead.
- [x] Margins should look ok with other components on the trip search screen, like global message, empty results, city zone message.


rosvik (approved):

tormoseng:
@gorandalum Question. Is the margin correct between time component and global message / city box?

<details>
<summary>Margins</summary>

File (30)
File (31)


</details>

Also, not this issue, but the buttons in the city box is a bit off (it's also in Prod so can create a separate issue if you would like)

<details>
<summary>City box buttons</summary>

File (30)

</details>

gorandalum:
**Margin below trip components**
This is correct from my understanding. The trip components is part of the header, which has a bottom margin of 12. The body below it has a top margin of 12, making the total gap 24.

**Misplaced buttons in city box**
https://github.com/AtB-AS/mittatb-app/pull/6123

PR #6118, opened 2026-05-28
refactor: move spacing responsibility from TravelCard to parent containers

gorandalum:
The gap between travel cards on TripSearchScreen should be 8px
(small) instead of 16px (custom). This gave some necessary changes
as the margin was hardcoded on the cards.

Components should generally not specify their own margins. That
should be parent's responsibility, as the parent knows the rendering
context. For lists in particular, using 'gap' on the container is
preferable to margins on items, which would add unwanted space at the
start or end of the list.

**Before/After**
<div>

</div>

<div>

</div>

<div>

</div>

### Acceptance criteria
- [x] Margins should be according to Figma


src/stacks-hierarchy/Root_TripSelectionScreen/components/TripSelection.tsx
```diff
@@ -100,15 +100,16 @@ export function BookingTripSelection({
...
+              includeLegNotifications
+              includeSituationNotices
               isDisabled={
                 !(
                   tp.booking.availability === 'available' &&
                   !tp.booking.disabledReason
                 )
               }
               tag={getBookingTagInfo(t, tp.booking, tp.booking.disabledReason)}
+              style={styles.travelCard}
```
rosvik: I've tried to reduce the number of places we have `style` props like this, since it often creates issues inside the component when mixing internal vs external styles.

Could this instead be solved with adding a `<View>` wrapper with gap?

gorandalum: I removed it. Added screenshots from the trip selection screen.

strandlie (approved):
Great work!

src/stacks-hierarchy/Root_TripSelectionScreen/components/TripSelection.tsx
rosvik: Nice ✨

rosvik (approved):

PR #6122, opened 2026-05-28
refactor: add MarginStyle for stricter style props

gorandalum:
Introduce MarginStyle which only allows margin-related ViewStyle
keys, and apply it to MessageInfoBox, MessageInfoText, and
GlobalMessage so callers can only pass margins via style.

Update createThemeHook to preserve narrow inferred types, so
styles created via useStyles flow through without casts. Clean
up callers that previously passed non-margin properties.

This is done kinda as a PoC because of [this comment](https://github.com/AtB-AS/mittatb-app/pull/6118#discussion_r3317384262), which correctly pinpoints that the ability to send in styles from outside to components is a risk factor. But trying to remove the style on MessageBox is too large a change, so making a more narrow MarginStyle was more manageable.

### Acceptance criteria
Verify some message boxes/texts in the app that they look as intended:
- [x] Message-boxes on the purchase overview screen (just check some of them, not necessary with all)
- [x] The anonymous warning message info text on select payment method bottom sheet
- [x] Global message on fare contract details

### DIFF
--- a/src/theme/StyleSheet.ts
+++ b/src/theme/StyleSheet.ts
@@ -13,6 +13,9 @@ export type NamedStyles<T> = {
 };
 export type ThemedStyles<T> = (theme: Theme, insets: EdgeInsets) => T;
 
+// MarginStyle is a subset of ViewStyle containing only margin-related
+// fields, e.g. 'margin', 'marginTop', 'marginHorizontal', etc. All other
+// ViewStyle fields are typed as `never`, so they cannot be set.
 type MarginKeys = Extract<keyof ViewStyle, `margin${string}`>;
 export type MarginStyle = Pick<ViewStyle, MarginKeys> & {
   [K in Exclude<keyof ViewStyle, MarginKeys>]?: never;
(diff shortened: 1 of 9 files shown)


strandlie:
Its kindof hard to understand the intention and consequences of this by itself. Maybe you can show me sometime today?

src/theme/StyleSheet.ts
```diff
@@ -13,9 +13,14 @@ export type NamedStyles<T> = {
 };
 export type ThemedStyles<T> = (theme: Theme, insets: EdgeInsets) => T;
 
+type MarginKeys = Extract<keyof ViewStyle, `margin${string}`>;
+export type MarginStyle = Pick<ViewStyle, MarginKeys> & {
+  [K in Exclude<keyof ViewStyle, MarginKeys>]?: never;
+};
```
rosvik, line 22: A comment wouldn't hurt here 😅

rosvik (approved):

strandlie (approved):

PR #6124, opened 2026-06-01
fix: walk leg can use expected time

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/22871#issuecomment-4543787406

## Description
In a trip pattern which involves a transit, the walk leg has `expected` time and not just `aimed` time, even though it is a `no-realtime` part of the trip.

Currently, when the time is updated due to a delayed bus trip, the start of walk leg is not updated according to the schedule, because it kept on using the `aimed` time, which is the original time.

This PR attempts to fix it by using `expected` time whenever available, so the trip timing is consistent across legs.

### DIFF
--- a/src/screen-components/travel-details-screens/components/Time.tsx
+++ b/src/screen-components/travel-details-screens/components/Time.tsx
@@ -64,7 +64,7 @@ export const Time: React.FC<{
     case 'no-realtime': {
       return (
         <ThemeText typography="body__m__strong" testID="schCaTime">
-          {scheduled}
+          {expected || scheduled}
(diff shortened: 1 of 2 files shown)


strandlie:
Do we really need the `no-realtime` state anymore, now that walk legs has both `expected` and `aimedTime`?

reidzeibel:
@strandlie I can check quickly with bikes

reidzeibel:
@strandlie no-realtime state removed!

strandlie (approved):

PR #6126, opened 2026-06-02
chore(security): Added dependabot file to limit dependabot PRs

strandlie:
## Issue Reference
Closes https://github.com/AtB-AS/kundevendt/issues/23814

## Description
Adds a dependabot file to our repo, to configure automatic dependabot PRs. For now we aim to reduce noise, so we only enable PRs for security updates. 

Unfortunately Cocoapods is not supported. Another argument for moving to SPM when we can!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+# Dependabot configuration
+#
+# Policy: security updates only.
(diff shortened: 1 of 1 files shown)


rosvik (approved):
🚀

reidzeibel (approved):

PR #6129, opened 2026-06-03
fix: clarify impossible trip status text with transfer cause

strandlie:
## Description
When the journey planner marks a trip pattern as `impossible` — meaning a transfer can no longer be made, e.g. due to delays — the TravelCard status text just said *"Ikke lenger mulig"* without explaining why. This change updates the text to *"Overgang ikke lenger mulig"* (with matching English and Nynorsk variants) so it's clear that the transfer is the cause.

### DIFF
--- a/src/translations/components/TravelCard.ts
+++ b/src/translations/components/TravelCard.ts
@@ -69,9 +69,9 @@ const TravelCardTexts = {
     originalTime: _('Opprinnelig', 'Original', 'Opprinnelig'),
     pastTime: _('Avreise passert', 'Departure passed', 'Avreise passert'),
     notPossible: _(
-      'Ikke lenger mulig',
-      'No longer possible',
-      'Ikkje lenger mogleg',
+      'Ikke lenger mulig pga. overgang',
+      'No longer possible due to transfer',
+      'Ikkje lenger mogleg pga. overgang',
     ),
     cancelled: _('Innstilt', 'Cancelled', 'Innstilt'),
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):

rosvik (approved):

gorandalum (approved):
I think the change is an improvement, but has it been considered to make the phrasing less deterministic? Currently, it states the trip is 'no longer possible' as a fact, but in reality, a passenger might still make the connection because:
- The bus could make up time later in the route making the trip possible again.
- The passenger could walk faster than estimated during a walking leg between departures.
- In rare cases it might even be a corresponding connection, meaning the next vehicle will wait even if the scheduled times no longer align.

If we instead would state something like "Trip might not be possible anymore", then it would be more up to the user to evaluate the trip details up against their walk/jog speed and risk tolerance.

strandlie:
@gorandalum Yep, there are some nuances that needs to be better adressed so we are having meeting. 

Converting this to draft for now.

strandlie:
@reidzeibel @rosvik @gorandalum We decided that this is in fact too final, especially considering the uncertainty in the data. But changing from `Ikke lenger mulig` to `Overgang ikke lenger mulig` is an improvement, and good enough for now, so we will add that to 1.86. 

Then, in 1.87 we will add a more nuanced information on the problematic interchange, something like this (not final yet):



Figma reference for later: https://www.figma.com/design/YPLCPDkCV7aZxH2KVI0QdG/Planning-and-purchase?node-id=6682-18980&p=f&t=LbFnLd7D7JBKHjBJ-0

PR #6133, opened 2026-06-04
feat: add apple pay info on PaymentMethodScreen

rosvik:
## Issue Reference

https://mittatb.slack.com/archives/C02EEG7D8EL/p1780496470920329

## Description

Simulator Screenshot - iPhone 16 Pro - 2026-06-04 at 10 06 48


reidzeibel (approved):
Just wondering if we should write "Når du kjøper billetter kan du også betale med Vipps eller Apple Pay" / "When you buy a ticket you can also pay with Vipps or Apple Pay" instead.

Aliaaen:
I really like @reidzeibel 's suggestion! Good clarification

PR #6134, opened 2026-06-04
fix: trigger ThemeText re-renders on fontScale changes

rosvik:
## Description

After the new architecture update, changing font size updates text size without updating its container. This causes clipping when it is increased, and excessive padding when it is decreased. Re-rendering text nodes on changes fixes this.

## Acceptance criteria

- [ ] Increasing and decreasing font size doesn't cause clipping or broken padding around text.

### DIFF
--- a/src/components/text/ThemeText.tsx
+++ b/src/components/text/ThemeText.tsx
@@ -81,8 +81,17 @@ export const ThemeText: React.FC<ThemeTextProps> = ({
   // If markdown is enabled, we need to wrap the content in a <View></View> to properly align the <Text></Text> elements,
   // by doing this we also avoid to wrap the list elements inside the markdown render in a Text component, which is not allowed.
   if (isMarkdown) {
-    return <View>{content}</View>;
+    return <View key={fontScale}>{content}</View>;
   }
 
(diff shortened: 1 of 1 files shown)


reidzeibel (approved):

strandlie:
Won't this lead to multiple nodes with duplicate keys, if we have several ThemeText in one component? 

Or are we sure fontScale is unique for each ThemeText instance?

rosvik:
> Won't this lead to multiple nodes with duplicate keys, if we have several ThemeText in one component?
> Or are we sure fontScale is unique for each ThemeText instance?

Good question! From what I understand, that's not how it works. (Font scale definitely won't be unique, rather the same for all Text nodes) 

Key props are 'special' in react, and don't propagate in the same way as regular props:

1. Setting `key` inside ThemeText won't affect keys in the parent context. So returning `<Text key=x>` is different from setting `<ThemeText key=x>` directly. 
2. Setting `key` on ThemeText also doesn't affect the child context (`props.key` is undefined). We use `<ThemeText key={x}>` a few places, but I've verified that this isn't affected by this change.

tormoseng:
Magical 🎉

PR #6141, opened 2026-06-05
fix: improve coupon codes screen accessibility and zoom layout

adriansberg:
## Issue Reference

closes https://github.com/AtB-AS/kundevendt/issues/23925

## Description
Accessibility and layout fixes for the AtB Bonus **My coupon codes** screen (`Profile_BonusCouponCodesScreen`):

- **Spell out the code** in the copy button's accessibility label (`spellOut`), so screen readers read it character-by-character instead of mispronouncing it as a word.
- **Add an accessibility hint** on the copy button ("Press to copy the coupon code"), mirroring the existing `installId` copy pattern.
- **Add `accessibilityRole="button"`** so the copy control is exposed as a button (surfaces label + hint correctly in the Accessibility Inspector / VoiceOver).
- **Hide the duplicate visual code** `ThemeText` from the screen reader (`accessibilityElementsHidden` / `importantForAccessibility`), so the code is announced once (spelled out) instead of twice.
- **Reflow the voucher card on zoom**: with large font scale the code + Copy button now wrap below the operator name (`flexWrap` + `minWidth` floor on the info column) instead of crushing the name text to one character per line.

New translation keys added under `BonusProgramTexts.myCouponCodes`: `copyA11yHint`, `codeA11yLabel` (nb/en/nn).

### DIFF
--- a/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_BonusCouponCodesScreen.tsx
+++ b/src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_BonusCouponCodesScreen.tsx
@@ -194,10 +194,22 @@ const VoucherCard = ({voucher}: {voucher: BonusVoucher}) => {
             </View>
           </View>
           <View style={styles.cardCode}>
-            <ThemeText typography="body__s__strong">{voucher.code}</ThemeText>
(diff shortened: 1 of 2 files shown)


tormoseng (approved):

tormoseng:
✅  Tested through https://github.com/AtB-AS/kundevendt/issues/23925

PR #6147, opened 2026-06-11
fix: only buyer sees earned bonus on ticket sent to others

jorunnl:
## Description
When a ticket is sent to others:
Only the user who has bought the ticket should be able to see how many points was earned for the purchase, as they are the one who received the point(s)

image image


Dahly96 (approved):

marius-at-atb (approved):

tormoseng:
✅

PR #6152, opened 2026-06-11
added check for showing bonus points earned

Dahly96:
Removed bonus view on refunded tickets

image

image

### DIFF
--- a/src/modules/fare-contracts/details/DetailsContent.tsx
+++ b/src/modules/fare-contracts/details/DetailsContent.tsx
@@ -56,7 +56,10 @@ import {
   useBonusAmountEarnedQuery,
   useIsBonusActiveForUser,
 } from '@atb/modules/bonus';
-import {useFareContractLegs} from '@atb/modules/fare-contracts';
+import {
+  isRelativeValidityStatus,
+  useFareContractLegs,
+} from '@atb/modules/fare-contracts';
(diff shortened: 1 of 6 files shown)


src/modules/fare-contracts/details/DetailsContent.tsx
```diff
@@ -154,6 +154,12 @@ export const DetailsContent: React.FC<Props> = ({
     benefits && benefits.length > 0 && validityStatus === 'valid';
 
   const isBonusActiveForUser = useIsBonusActiveForUser();
+  const shouldShowBonusAmountEarned =
+    (validityStatus === 'valid' ||
```
adriansberg: Could maybe use `RelativeValidityStatus` here, instead of checking the three?

adriansberg: I'm not familiar with the statuses. Is `sent` the status for on-behalf tickets? I.e. the one who sends should see that they earned a point.

adriansberg: And `cancelled` should not show earnt?

Dahly96: Talked with Tor and we figured we would try with the 3 at first, but i can check if i can just check for the type instead of typing it out

adriansberg (approved):

PR #6158, opened 2026-06-12
feat: use citybike illustrations

Dahly96:
Added missing citybike illustrations

Closes https://github.com/AtB-AS/kundevendt/issues/23807

The start trip functionallity was down at entur, so showing the in-use sheet with the new illustration was not possible. So the last image with end trip was just forcefully put into a empty sheet just to show what the section would look like with the illustration.

image

image

image

image


adriansberg (approved):
LGTM 👍

PR #6160, opened 2026-06-15
fix: travel card and trip details UI tweaks after demo

strandlie:
## Description

Three small UI fixes to the new travel card / trip details design:

- **Status text without icon** — `StatusText` now supports rendering without an icon, and the icons currently used on the status configs are removed. The text renders where the icon would otherwise sit.
- **Saved-trip icon overlap** — the bookmark icon on a saved trip in search results no longer overlaps the duration. The saved state is passed through `SaveableTripSearchResultRow`, and the duration drops below the icon when a trip is saved.
- **Timestamp alignment** — trip-leg timestamps in trip details are left-aligned so they line up with the time in the travel card header above them, instead of being indented by the right-aligned label column.

## Notes for reviewers

- The trip-leg label column (`NEW_TRIP_DIMENSIONS.labelWidth` in `TripRow.tsx`) is a fixed, reserved gutter — it has to be predetermined because rows without a timestamp (transport, walk, wait, "X mellomstopp") still need their content to align, and the timeline line is absolutely positioned at `labelWidth + decorationContainerWidth/2`. This is the existing pattern (`theme.tripLegDetail.labelWidth`); the value was just retuned for the left-aligned layout. Trade-off: it's sized for `HH:mm` at default font, so very large accessibility font sizes / the `ca.` approximate-time prefix have less headroom than before. Can follow up with a font-scale-aware version if we want.

### Status text without icon

| Before | After |
|--------|-------|
|  |  |

### Saved-trip icon overlap

| Before | After |
|--------|-------|
|  |  |

### Trip details timestamp alignment

| Before | After |
|--------|-------|
|  |  |

### DIFF
--- a/src/modules/experimental-store-trip-patterns/SaveableTripSearchResultRow.tsx
+++ b/src/modules/experimental-store-trip-patterns/SaveableTripSearchResultRow.tsx
@@ -11,11 +11,10 @@ import {getTripPatternAnalytics} from '@atb/screen-components/travel-details-scr
 import {useFirestoreConfigurationContext} from '@atb/modules/configuration';
 import {useTimeContext} from '../time';
 
-export const SaveableTripSearchResultRow: React.FC<
-  React.PropsWithChildren<{
-    tripPattern: TripPattern;
-  }>
-> = ({children, tripPattern}) => {
+export const SaveableTripSearchResultRow: React.FC<{
+  tripPattern: TripPattern;
+  children: (isStored: boolean) => React.ReactNode;
+}> = ({children, tripPattern}) => {
   const {isSaveTripsEnabled} = useFeatureTogglesContext();
   const {tripPatterns, addTripPattern, removeTripPattern, canAddTripPattern} =
     useStoredTripPatterns();
@@ -69,7 +68,7 @@ export const SaveableTripSearchResultRow: React.FC<
   );
 
   if (!isSaveTripsEnabled || !canAdd) {
-    return children;
+    return children(false);
   }
 
   return (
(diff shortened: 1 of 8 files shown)


reidzeibel:
I think you missed the screenshots part on the description @strandlie

strandlie:
> I think you missed the screenshots part on the description @strandlie

@reidzeibel  Sorry about that! I promise I added them yesterday, but maybe I forgot to press Save 😞

reidzeibel (approved):

tormoseng:
Looking good - except for the "loading icons bug" reported in [Slack](https://mittatb.slack.com/archives/C09CJU3JUH5/p1781855527836439?thread_ts=1781185740.199769&cid=C09CJU3JUH5).

@strandlie Is it on purpose that the travel card is expanded when you save it? If the travel card doesn't have an expected time, the travel card is expanded when the save icon is added. See screenshots.

<details>
<summary>Screenshot</summary>

Skjermbilde 2026–06–19 kl  10 20 36
Skjermbilde 2026–06–19 kl  10 21 10 2


</details>

strandlie:
> Looking good - except for the "loading icons bug" reported in [Slack](https://mittatb.slack.com/archives/C09CJU3JUH5/p1781855527836439?thread_ts=1781185740.199769&cid=C09CJU3JUH5).
> 
> @strandlie Is it on purpose that the travel card is expanded when you save it? If the travel card doesn't have an expected time, the travel card is expanded when the save icon is added. See screenshots.
> 
> Screenshot
> Skjermbilde 2026–06–19 kl  10 20 36 Skjermbilde 2026–06–19 kl  10 21 10 2

Yes it is on purpose, since I just expanded on how it already worked. But it does look a little wonky. Maybe @RHatlem could look if we should solve it differently?

RHatlem:
@strandlie Is it possible to make them expanded only when the card has expected time? The space looked a bit weird

tormoseng:
Loading error handled in https://github.com/AtB-AS/mittatb-app/pull/6193

Fix related to https://github.com/AtB-AS/mittatb-app/pull/6160#issuecomment-4777165903 will also be handled in a separate issue.

PR #6161, opened 2026-06-15
fix: Require login to use transfer code

cbrevik:
Related to: https://github.com/AtB-AS/kundevendt/issues/23977

Must be verified as a requirement first, but seems like a sensible one so pupils do not accidentally add school tickets to an anonymous account and lose them.

CC @Aliaaen & @Sebstorvik 

## Changes
- Re-use `Root_LoginRequiredForFareProductScreen` to gate usage of transfer code
- Made title/text on `Root_LoginRequiredForFareProductScreen` route params so they can be overridden

image

### DIFF
--- a/src/translations/screens/Ticketing.ts
+++ b/src/translations/screens/Ticketing.ts
@@ -90,6 +90,14 @@ const TicketingTexts = {
         'Ein hentekode kan brukast til å hente ein billett i appen. Vanlegvis er det nokon andre som har sendt deg koden, til dømes skulen eller arbeidsgivaren din.',
       ),
     },
+    loginRequired: {
+      title: _('Du må logge inn', 'You must log in', 'Du må logge inn'),
+      text: _(
+        'Du må logge inn for å bruke en hentekode.',
+        'You must log in to use a transfer code.',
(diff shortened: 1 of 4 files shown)


src/translations/screens/Ticketing.ts
```diff
@@ -90,6 +90,14 @@ const TicketingTexts = {
         'Ein hentekode kan brukast til å hente ein billett i appen. Vanlegvis er det nokon andre som har sendt deg koden, til dømes skulen eller arbeidsgivaren din.',
       ),
     },
+    loginRequired: {
```
rosvik, line 93: Nit: Could rename this to make it clear that it's transfer code specific texts, and not generic "login required" texts.

rosvik (approved):

src/translations/screens/Ticketing.ts
cbrevik: It is nested under `transferCode` in the texts so maybe that is clear enough?

rosvik: Oh right, agreed!

PR #6164, opened 2026-06-15
feat: Add keyboard avoiding behaviour to transfer code sheet

cbrevik:
Related to: https://github.com/AtB-AS/kundevendt/issues/23977#issuecomment-4715634548

Handle so bottom sheet avoids keyboard. The most stable way to do this is to use `BottomSheetTextInput` from `@gorhom/bottom-sheet`: https://gorhom.dev/react-native-bottom-sheet/components/bottomsheettextinput

https://github.com/user-attachments/assets/d2d47672-8d5c-41c8-8fbd-99ef4f280ce4


cbrevik:
CC @Dahly96 since it also adds something to `BottomSheetModal` props

rosvik (approved):

tormoseng:
@cbrevik Works well on iOS, but not on my Android 12. I don't see any difference - the keyboard is still hiding the input field.

tormoseng:
✅  After https://github.com/AtB-AS/mittatb-app/pull/6166

PR #6165, opened 2026-06-16
fix: prevent intermittent blank travel card legs

strandlie:
## Description
This is a difficult one to explain and reproduce, and I have struggled with understanding exactly what happens myself, but I will give it a try. I think that the render-invisible-to-measure-then-render-properly approach in `OverflowContainer` has a bug, where the faulty assumption is that the following holds: 

**Invariant:** The `tryCompute` function will always be called after `setFitResult(null)` is called in the `useEffect`

Counterexample where that doesn't hold: 

1. Overflow container renders one `TripPattern` with 1 foot leg, and two bus legs. `maxWidth` is stable, because the `View` wrapping `TravelCardLegs` has a settled size.
2. The `TripPattern` is updated, but only times changes such that the foot leg is no longer significant according to [isSignificantFootLegOrWalkTime](https://github.com/AtB-AS/mittatb-app/blob/5236123d158c9f8326b1bea7d61a683d3c76fa92/src/screen-components/travel-details-screens/utils.ts#L202) and thus gets filtered out of [TravelCardLegs](https://github.com/AtB-AS/mittatb-app/blob/master/src/screen-components/travel-card/TravelCardLegs.tsx#L31)
3. That means the number of `children` to [OverflowContainer](https://github.com/AtB-AS/mittatb-app/blob/master/src/screen-components/travel-card/TravelCardLegs.tsx#L72) has changed, which in turn triggers the [useEffect of OverflowContainer](https://github.com/AtB-AS/mittatb-app/blob/master/src/components/overflow-container/OverflowContainer.tsx#L33-L39), which sets `fitResult` to `null` since `maxWidth` is not 0. 
4. But since the [View wrapping children in OverflowContainer](https://github.com/AtB-AS/mittatb-app/blob/master/src/components/overflow-container/OverflowContainer.tsx#L94-L98) is keyed only off of `maxWidth` (which is stable) the `onLayout` function is never called --> `tryCompute` is never called on its children (or called less times than the number of children, which means `fitResult` is [never set](https://github.com/AtB-AS/mittatb-app/blob/master/src/components/overflow-container/OverflowContainer.tsx#L45))
5. I.e. the invariant does not hold. QED.

This PR tries to fix that, by keeping the useEffect and the `View wrapping children in OverflowContainer` in sync. It also adds a test that tries to communicate how `OverflowContainer` works, the nuance of the issue, and tries to catch the issue if it ever happens again. 

## Example 

Bug report from [Slack](https://mittatb.slack.com/archives/C09CJU3JUH5/p1781185740199769): 

```
I'we experienced the bug where the transport icon is missing/invisible in the 
trip search a couple of times now. Think we should prioritize to look into it. 
See pictures in :thread:.

Can't replicate it consistently, but i have seen it a couple of times when i keep 
loading in more results by continuing to press "load more results" at the bottom
```

### DIFF
--- a/src/components/overflow-container/OverflowContainer.tsx
+++ b/src/components/overflow-container/OverflowContainer.tsx
@@ -91,8 +91,11 @@ export const OverflowContainer: React.FC<Props> = ({
   return (
     <View style={[styles.row, {gap}]}>
       {!fitResult && (
+        // Key includes the child count so a count change remounts this pass and
+        // every child re-fires onLayout — reused views don't re-measure, which
+        // would otherwise leave the row blank.
         <View
-          key={maxWidth}
+          key={`${maxWidth}-${children.length}`}
           style={[styles.measureWrapper, {width: maxWidth}]}
           pointerEvents="none"
         >
--- a/src/components/overflow-container/__tests__/OverflowContainer.test.tsx
+++ b/src/components/overflow-container/__tests__/OverflowContainer.test.tsx
@@ -1,0 +1,123 @@
+import React, {useEffect} from 'react';
+import {View} from 'react-native';
+import {
+  fireEvent,
+  render,
+  type RenderResult,
+} from '@testing-library/react-native';
+import {OverflowContainer} from '../OverflowContainer';
+
+// Minimal mock for the reveal animation OverflowContainer imports from reanimated.
+jest.mock('react-native-reanimated', () => {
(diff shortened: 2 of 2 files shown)


rosvik (approved):
Well done, and thanks for the deep-dive here 🙌 Makes a lot more sense when I read it before 16:00. Also can't argue with a PR description that ends on "QED" 😅

PR #6169, opened 2026-06-17
fix: use correct screen transition after zone selection

rosvik:
There was a small bug with screen transitions after zone selection was done during ticket purchase. If the user entered the zone selection screen and navigated back, the transition setting wasn't remembered, and it would navigate with the `slide-from-bottom` animation instead of `slide-from-right`.

https://github.com/user-attachments/assets/37148b07-2657-4028-bf71-b78ac79bab37

## Acceptance criteria

- [ ] Exiting the ticket purchase flow has the same animation as going in, also after changing zones

### DIFF
--- a/src/stacks-hierarchy/Root_PurchaseFareZonesSearchByMapScreen.tsx
+++ b/src/stacks-hierarchy/Root_PurchaseFareZonesSearchByMapScreen.tsx
@@ -41,11 +41,15 @@ export const Root_PurchaseFareZonesSearchByMapScreen = ({
         to: selection.zones.to.id,
       });
     }
-    navigation.popTo('Root_PurchaseOverviewScreen', {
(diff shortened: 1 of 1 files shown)


gorandalum (approved):

hanne-at-atb:
I see that when changing zones in map and then clicking back, the state of zones are saved so there is no difference between the buttons "back" and "save selected zone". Might be outside scope for this issue, but do you know if this is by design @rosvik? :D 

Other than that the transition looks good! ✅

rosvik:
@hanne-at-atb I feel like that's something we did on purpose at some point, but not quite sure. I think it's been like that for a while at least, and not touched by this PR 😊

hanne-at-atb:
Alrighty. Then all is good 🎉

PR #6171, opened 2026-06-17
Add app Feide-verification for Firebase-user in Profile

cbrevik:
Related to: https://github.com/AtB-AS/team-platform/issues/1080

Dependent on: https://github.com/AtB-AS/identity/pull/78 ✅ 
Dependent on: https://github.com/AtB-AS/identity/pull/79

## What
Adds the ability for a logged-in (phone-authenticated) pupil to connect their Firebase account to a Feide identity, so they can retrieve their school ticket (skoleskyss) in the app. This PR just introduces that verification-step towards Feide, and does not do anything with school ticket yet. This is coming in a later PR when work with `farecontract-admin` is done!

## Future work
- If the pupil has a ticket, we might want to go to the ticket screens to refresh after moving tickets to current ABT-account
- If they have a travelcard, do we want to initiate switching to mobile token?
- Lots of unknowns!

## Flow
The flow mirrors the existing Vipps "connect-while-logged-in" age verification:

1. The user starts the connection, which fetches a Feide authorization URL from the identity service. The app appends the `state`, `nonce` and PKCE-`challenge` to the authorization URL, and opens it in an in-app browser.
2. After login, Feide redirects back via the `atb://auth/feide` deep link. The screen validates the returned `state` against the locally stored value, then exchanges the authorization `code`, `nonce` and PKCE-`verifier` server-side to link the Feide user to the current Firebase user.
3. The screen reflects connection status (connected / error) to the user.

## Changes 
- A new **Feide og skoleskyss** entry appears under Profile → Settings.
- New `Profile_FeideConnectionScreen` handling the deep-link callback, state validation, connection status, and error display.
- Add API-endpoints in `src/api/identity.ts`:
    - `initFeideConnect` — fetches authorization URL and opens the in-app browser, generating/storing `state`, `nonce` and a PKCE-pair with `verifier` and a `challenge`.
    - `connectFeide` — trades the authorization `code`, `nonce` and `verifier` server-side and links Feide to the Firebase user.
    - `getFeideConnection` — returns whether the user is connected (and the sub).
- React Query hooks towards endpoints:
    - `useConnectFeideMutation` (invalidates the connection query on success),
    - `useGetFeideConnectionQuery` (gated by feature toggle + auth status).
- Remote flag / feature toggle `enable_feide_connection` & `isFeideConnectionEnabled` (defaults to `false`).

https://github.com/user-attachments/assets/85036185-342f-4a01-9143-7e582767b293

### DIFF
--- a/src/api/identity.ts
+++ b/src/api/identity.ts
@@ -19,6 +19,11 @@ export type FeideConnectResponse = {
   hasNin: boolean;
 };
 
+export type FeideConnectionStatus = {
+  connected: boolean;
+  feideSub?: string;
+};
+
 export const getServerTime = async () => {
   const response = await client.get('/identity/v1/time', {
     authWithIdToken: false,
@@ -211,6 +216,21 @@ export const connectFeide = async (
         ...opts,
       },
     )
+    .then((res) => res.data);
+};
+
+/**
+ * Whether the logged-in customer is connected to Feide (and with which sub).
+ * Mirrors getAgeVerification.
+ */
+export const getFeideConnection = (
+  opts?: AxiosRequestConfig,
+): Promise<FeideConnectionStatus> => {
+  return client
+    .get<FeideConnectionStatus>('/identity/v1/feide/connection', {
+      ...opts,
+      authWithIdToken: true,
+    })
     .then((res) => res.data);
 };
 
--- a/src/api/identity.ts
+++ b/src/api/identity.ts
@@ -175,31 +175,26 @@ async function generateNonce() {
 }
 
 /**
- * Start Feide-connection: Get authorization-URL from identity og open it in
- * in-app browser. Redirect catched via deep-link in screen which calls this.
- * Mirrors initAgeVerification (Vipps connect-while-logged-in).
+ * Start Feide-connection: Get authorization-URL from identity and return the
+ * full URL (with state, nonce and PKCE challenge) for the caller to open in an
+ * in-app browser. Redirect catched via deep-link in screen.
+ * Mirrors authorizeVippsUser (Vipps login).
  */
 export const initFeideConnect = async (
   opts?: AxiosRequestConfig,
-): Promise<void> => {
+): Promise<string> => {
   const state = await generateFeideState();
   const nonce = await generateFeideNonce();
   const {verifier, challenge} = generatePkcePair();
   await storage.set('feide_code_verifier', verifier);
-  return client
-    .get(
-      `/identity/v1/feide/authorization-url?callbackUrl=${FEIDE_CALLBACK_URL}`,
-      {
-        ...opts,
-      },
-    )
-    .then(async (response) => {
-      const authorisationUrl = response.data;
-      openInAppBrowser(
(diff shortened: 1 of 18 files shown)


src/api/identity.ts
```diff
@@ -161,3 +172,76 @@ async function generateNonce() {
...
+  return client
+    .get(
+      `/identity/v1/feide/authorization-url?callbackUrl=${FEIDE_CALLBACK_URL}`,
+      {
+        ...opts,
+      },
+    )
+    .then(async (response) => {
+      const authorisationUrl = response.data;
+      openInAppBrowser(
```
strandlie: Little bit nitpicky, but it feels weird to me to have API logic mixed in with "behaviour"-logic, so I think `openInAppBrowser` belongs in `useInitFeideConnectMutation`

strandlie: Or maybe even in `Profile_FeideConnectionScreen`?

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_FeideConnectionScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+          await storage.set('feide_nonce', '');
+        } else if (parsed.searchParams.get('error')) {
+          setLocalError(true);
+        }
+      }
+    };
+
+    const subscription = Linking.addEventListener('url', handleUrl);
+    return () => subscription.remove();
+  }, [appStatus, connectFeide]);
```
strandlie: Why is `appStatus` included in the dependency array? I understand the setup and teardown of the listener, but the `appStatus` is a little unclear to me

src/api/identity.ts
cbrevik: Unsure, its the same exact pattern as how `initAgeVerification` is implemented. 😅 The question is more, how do I ensure it is only done once. But I see your point

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_FeideConnectionScreen.tsx
cbrevik: Copy-pasted from Vipps/age-verification again 😅 So I assumed it had something to do with webviews/in-app-browser-window switches. iOS and Android behaves a bit janky with this. But to be honest I haven't followed up why they do it in Vipps/age-verification, just assumed thats why

src/api/identity.ts
cbrevik: I'll try to move it out. Looks like `useAuthenticateVipps` does something similar which has a good pattern

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_ProfileStack/Profile_FeideConnectionScreen.tsx
cbrevik: I can try to remove it, and rather re-add if there are issues

strandlie (approved):

tormoseng:
✅

PR #6172, opened 2026-06-17
feat: redesign interchange section in trip details

gorandalum:
Replace the MessageInfoBox-based interchange notice with a dedicated
component using an icon and structured title/body text. Use the new
`interchangeFrom` field on legs so the section renders at the start
of the next leg rather than at the end of the previous one.

**Before/After**
<div>


strandlie (approved):
Nice refactor!

PR #6173, opened 2026-06-17
feat: show correspondence icon between legs on travel card

gorandalum:
Render a small correspondence icon between transport legs whenever
the next leg has a guaranteed (non stay-seated) interchange. When
a foot leg sits between two transit legs, the icon lands after the
walk. Also announce the interchange via the travel card's a11y
label.

**Before/After**
<div>


strandlie (approved):
Looks good! Some type issues?

PR #6177, opened 2026-06-19
feat: new icon to stationSheet

Dahly96:
Added a new blue parking icon (P) to the station bottomsheet, replacing the old mono parking icon

image

From [figma](https://www.figma.com/design/YfLcr1BKOuaXk9HIUilXJK/City-bike-integration-V1?node-id=1119-93908&t=dNfx9Q544enciAeP-0)

image

### DIFF
--- a/src/modules/mobility/components/BikeStationIntegrationView.tsx
+++ b/src/modules/mobility/components/BikeStationIntegrationView.tsx
@@ -30,8 +30,7 @@ import {
   MOCK_VEHICLE_ID,
 } from '../queries/use-vehicle-query';
 import {SupportButton} from './SupportButton';
-import {ThemedCityBikeStation} from '@atb/theme/ThemedAssets';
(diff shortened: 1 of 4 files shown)


src/modules/mobility/components/BikeStationIntegrationView.tsx
```diff
@@ -32,6 +31,7 @@ import {
 } from '../queries/use-vehicle-query';
 import {SupportButton} from './SupportButton';
 import {ThemedCityBikeStation} from '@atb/theme/ThemedAssets';
+import Parking from '@atb/assets/svg/color/icons/vehicles/light/Parking';
```
jorunnl: even though the icon now is the same for light and dark I would still facilitate for having different ones for light and dark in case of future changes.

Dahly96: agreed

reidzeibel (approved):

hanne-at-atb:
Tested ok ✅

PR #6181, opened 2026-06-19
fix: fix for reanimated warning

Dahly96:
There was a issue where  useWindowDimensions(); was mutated, causing warnings in the terminal


cbrevik (approved):
As I said, I find it weird that Reanimated is complaining here, but change is a good one anway 👍🏻

PR #6183, opened 2026-06-22
feat: Enforce unique Feide-connections

cbrevik:
Allow only a single connection to Feide at a time, so multiple app-users cannot be connected to the same Feide-user.

Will route users to customer support if there is a conflict, since this most likely requires manual clean up if they have a new phone number.

Fixes: https://github.com/AtB-AS/identity/pull/78#issuecomment-4751887689

Dependent on: https://github.com/AtB-AS/identity/pull/81

image


marius-at-atb (approved):
Lgtm

tormoseng:
✅

PR #6186, opened 2026-06-22
feat(mobility): consume shmo vehicle contract v2 (benefit, bonus offer, action button)

adriansberg:
## What

Consume the shared-mobility **vehicle contract v2** in the app: a single benefit and an
optional bonus offer resolved server-side by mobility-service, plus an action button that
drives the CTA. History cleaned up into 4 reviewable commits and rebased on `master`.

## Commits

1. **adopt shmo vehicle contract v2 types** — slim `BenefitSchema` to the display payload
   (`title`, `description`, `illustrationName`, `priceAdjustments`); `kind`, `eligibility`
   and vehicle/system filtering all live server-side now. `Vehicle` gains optional nullable
   `bonusOffer` and `actionButton` (discriminated union `START_TRIP | APP_SWITCH{url,label}`).
   `BonusOffer` and `Benefit` share one `PriceAdjustmentSchema`. Adds `getAppSwitchUrl` /
   `getMockVehicle` client calls + schema-parse tests (with/without the new fields).
2. **consume vehicle benefit, price details and app switch** — `PriceDetailsCard` renders the
   server benefit text + price adjustments; `VehicleSheet` prefers `vehicle.bonusOffer`
   (client lookup only as fallback), never stacks benefit + bonus, and drives the app switch
   via `useAppSwitchMutation`.
3. **resolve benefit illustration by name** — `BenefitIllustration` maps `illustrationName`
   to a themed asset with a safe fallback for unknown values.
4. **wire pricing-details screen and map bottom sheets** — the shmo pricing-details route +
   threading the selected benefit through the map bottom sheets and navigation types.

## Notes

- No DB changes.
- `preReqs` on the action button is intentionally **not** consumed (deferred per the contract).
- `getMockVehicle` is added for completeness; the station-based-vehicle rename is tracked
   separately (see #6191).
- App-side deep-integration stays gated by the `isShmoDeepIntegration*` remote-config toggles
   with an app-switch fallback, so the server's `actionButton` is safe across app versions.
- Ships in sync with mobility#113, benefit#16, bonus#41.

## Verification

- `pnpm tsc` — no new errors from this change. (Pre-existing `ThemedAssets.tsx` /
   `icon-box/utils.ts` asset-codegen errors exist on `master` too — `TicketValid` /
   `ParkingThick` assets — unrelated to this PR.)
- `pnpm lint`, `pnpm prettier` — clean on all changed files.
- `pnpm test` (mobility/benefit) — 11 passing, incl. the schema tests.

Resolves AtB-AS/kundevendt#24093

---
📋 **Change overview & deploy runbook:** https://claude.ai/code/artifact/59f5bed8-d0bd-43d3-810d-e6cf899b09c5

### DIFF
--- a/src/api/mobility.ts
+++ b/src/api/mobility.ts
@@ -2,6 +2,7 @@ import {client} from '@atb/api/index';
 import {stringifyUrl} from '@atb/api/utils';
 import qs from 'query-string';
 import {AxiosRequestConfig} from 'axios';
+import {z} from 'zod';
 import {
   AssetSchema,
   AssetFromQrCodeQuery,
@@ -28,6 +29,10 @@ import {
   ViolationsReportQueryResult,
   ViolationsReportQueryResultSchema,
 } from './types/mobility';
+
+const AppSwitchUrlSchema = z.object({url: z.string()});
+
+export type AppSwitchUrl = z.infer<typeof AppSwitchUrlSchema>;
 
 export const getActiveShmoBooking = (
   acceptLanguage: string,
@@ -175,6 +180,38 @@ export const getStation = (
     .catch((error) => {
       console.error('Error in StationSchema parsing: ', error);
       return null;
+    });
+};
+
+export const getAppSwitchUrl = (
+  vehicleTypeId: string,
+  bonusProductId?: string,
+): Promise<AppSwitchUrl> => {
+  const url = `/mobility/v1/app-switch/vehicle-type/${vehicleTypeId}`;
+  const query = qs.stringify({bonusProductId});
+  return client
+    .post<AppSwitchUrl>(stringifyUrl(url, query), undefined, {
+      authWithIdToken: true,
+    })
+    .then((response) => AppSwitchUrlSchema.parse(response.data));
+};
+
+export const getMockVehicle = (
+  stationId: string,
+  vehicleTypeId: string,
+  opts?: VehicleRequestOpts,
+): Promise<Vehicle> => {
+  return client
+    .get(`/mobility/v1/stations/${stationId}/mock-vehicles/${vehicleTypeId}`, {
+      ...opts,
+    })
+    .then((response) => {
+      const result = VehicleSchema.safeParse(response.data);
+      if (!result.success) {
+        console.error(result.error.format());
+        throw result.error;
+      }
+      return result.data;
     });
 };
 
--- a/src/api/types/benefit.ts
+++ b/src/api/types/benefit.ts
@@ -1,7 +1,8 @@
 import {z} from 'zod';
 import {PriceAdjustmentEnum} from '@atb-as/config-specs/lib/mobility';
+import {LanguageAndTextTypeArray} from '@atb-as/config-specs/lib/common';
 
-export const MobilityPriceAdjustmentSchema = z
+export const PriceAdjustmentSchema = z
   .object({
     amount: z.number(),
     adjustmentType: PriceAdjustmentEnum,
@@ -13,16 +14,15 @@ export const MobilityPriceAdjustmentSchema = z
     description: x.description,
   }));
 
-export const MobilityPriceAdjustmentBenefitSchema = z.object({
-  kind: z.literal('MOBILITY_PRICE_ADJUSTMENT'),
-  vehicleTypeIds: z.array(z.string()).default([]),
-  systemIds: z.array(z.string()).default([]),
-  priceAdjustments: z.array(MobilityPriceAdjustmentSchema),
+// The backend resolves a single benefit for the vehicle (kind, eligibility,
+// vehicle/system filtering all happen server-side), so the app only models the
+// display payload.
+export const BenefitSchema = z.object({
+  title: LanguageAndTextTypeArray.default([]),
+  description: LanguageAndTextTypeArray.default([]),
+  illustrationName: z.string().optional(),
+  priceAdjustments: z.array(PriceAdjustmentSchema),
 });
 
-export type MobilityPriceAdjustmentType = z.infer<
-  typeof MobilityPriceAdjustmentSchema
->;
-export type MobilityPriceAdjustmentBenefitType = z.infer<
-  typeof MobilityPriceAdjustmentBenefitSchema
->;
+export type PriceAdjustmentType = z.infer<typeof PriceAdjustmentSchema>;
+export type BenefitType = z.infer<typeof BenefitSchema>;
--- a/src/api/types/mobility.ts
+++ b/src/api/types/mobility.ts
@@ -7,7 +7,8 @@ import {isValidPhoneNumber} from 'libphonenumber-js';
 import {isValidEmail} from '@atb/utils/validation';
 import {Feature, MultiPolygon, Point, Polygon} from 'geojson';
 import {Base64ImageSchema} from '@atb/utils/image';
-import {MobilityPriceAdjustmentBenefitSchema} from '@atb/api/types/benefit';
+import {BenefitSchema, PriceAdjustmentSchema} from '@atb/api/types/benefit';
+import {LanguageAndTextTypeArray} from '@atb-as/config-specs/lib/common';
 
 export const ViolationsReportingInitQuerySchema = z.object({
   lng: z.string(),
@@ -236,6 +237,33 @@ const RentalUrisSchema = z
 
 export type RentalUris = z.infer<typeof RentalUrisSchema>;
 
+export const BonusOfferSchema = z.object({
+  bonusProductId: z.string(),
+  bonusProductPrice: z.object({
+    amount: z.number(),
+    currencyCode: z.string(),
+  }),
+  priceAdjustments: z.array(PriceAdjustmentSchema),
+});
+
+export type BonusOffer = z.infer<typeof BonusOfferSchema>;
+
+export enum ActionButtonType {
+  START_TRIP = 'START_TRIP',
+  APP_SWITCH = 'APP_SWITCH',
+}
+
+export const ActionButtonSchema = z.discriminatedUnion('type', [
+  z.object({type: z.literal(ActionButtonType.START_TRIP)}),
+  z.object({
+    type: z.literal(ActionButtonType.APP_SWITCH),
+    url: z.string(),
+    label: LanguageAndTextTypeArray.default([]),
+  }),
+]);
+
+export type ActionButton = z.infer<typeof ActionButtonSchema>;
+
 export const VehicleSchema = z.object({
   id: z.string(),
   lat: z.number(),
@@ -250,7 +278,11 @@ export const VehicleSchema = z.object({
   station: z.object({id: z.string()}).nullable().optional(),
   rentalUris: RentalUrisSchema.nullable().optional(),
   vehicleType: VehicleTypeSchema,
-  benefit: MobilityPriceAdjustmentBenefitSchema.nullable().optional(),
+  // Tolerant: a malformed/unknown benefit shape degrades to undefined rather
+  // than failing the whole vehicle parse.
+  benefit: BenefitSchema.nullable().optional().catch(undefined),
+  bonusOffer: BonusOfferSchema.nullable().optional(),
+  actionButton: ActionButtonSchema.nullable().optional(),
 });
 
 export type Vehicle = z.infer<typeof VehicleSchema>;
--- a/src/modules/mobility/components/BenefitIllustration.tsx
+++ b/src/modules/mobility/components/BenefitIllustration.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import {View, ViewStyle} from 'react-native';
+import {SvgProps} from 'react-native-svg';
+// eslint-disable-next-line no-restricted-syntax -- resolve a themed asset dynamically by its server-driven name
+import * as ThemedAssets from '@atb/theme/ThemedAssets';
+
+type ThemedAsset = (props: SvgProps) => React.JSX.Element;
+
+const themedAssets = ThemedAssets as unknown as Record<
+  string,
+  ThemedAsset | undefined
+>;
+
+/**
+ * Resolves the server-driven `illustration` string to a themed asset by name,
+ * e.g. "BonusStar" -> ThemedBonusStar. Any existing themed asset can be used
+ * without maintaining a registry; unknown values render nothing.
+ */
+const resolveIllustration = (
+  illustration: string | undefined,
+): ThemedAsset | undefined => {
+  if (!illustration) return undefined;
+  const asset = themedAssets[`Themed${illustration}`];
+  return typeof asset === 'function' ? asset : undefined;
+};
+
+type BenefitIllustrationProps = {
+  illustrationName: string | undefined;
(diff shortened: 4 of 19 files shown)


src/modules/mobility/components/BenefitIllustration.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import {View, ViewStyle} from 'react-native';
+import {SvgProps} from 'react-native-svg';
+// eslint-disable-next-line no-restricted-syntax -- resolve a themed asset dynamically by its server-driven name
```
adriansberg, line 4: **`eslint-disable no-restricted-syntax` here — worth a team decision.** 👀

This resolves the server-driven `illustration` string to a themed asset dynamically (`Themed{name}`), so any existing illustration in `ThemedAssets` can be used without maintaining a per-asset map. Indexing the assets by name needs a namespace import (`import * as ThemedAssets`), which trips the repo-wide `no-restricted-syntax` "No wildcard imports" rule — hence the disable.

Why I think it's acceptable here: `ThemedAssets` already eagerly imports every asset at module top, so a namespace import doesn't change what's bundled (no tree-shaking loss).

Alternatives if we'd rather not disable the rule:
1. **Explicit record in `ThemedAssets.tsx`** — e.g. `export const themedIllustrations: Record<string, ThemedAsset> = { Ticket: ThemedTicket, … }`, imported normally. Convention-clean, but needs an entry per asset (a registry, just centralized).
2. **Component-local registry** (the original approach) — same maintenance + duplication, living in the mobility module.
3. **Keep the namespace import + disable** (current) — zero maintenance, any themed asset usable by name.

No strong opinion — flagging for discussion. Happy to switch to (1) if we'd rather avoid the disable.

src/api/types/benefit.ts
marius-at-atb: I think this should just be called BenefitSchema, and `kind`, `source`, `vehicleTypeIds` and `systemIds` should be removed.

src/api/types/benefit.ts
```diff
@@ -13,8 +14,19 @@ export const MobilityPriceAdjustmentSchema = z
...
+]);
+
+export type MobilityBenefitSource = z.infer<typeof MobilityBenefitSourceSchema>;
+
 export const MobilityPriceAdjustmentBenefitSchema = z.object({
   kind: z.literal('MOBILITY_PRICE_ADJUSTMENT'),
+  source: MobilityBenefitSourceSchema.optional(),
+  title: LanguageAndTextTypeArray.default([]),
+  description: LanguageAndTextTypeArray.default([]),
+  illustration: z.string().optional(),
```
Dahly96: discussed renaming this to illustrationName, to match the pattern used for ticket in TicketingTileIllustration.

src/api/types/mobility.ts
```diff
@@ -236,6 +238,40 @@ const RentalUrisSchema = z
 
 export type RentalUris = z.infer<typeof RentalUrisSchema>;
 
+const BonusOfferPriceAdjustmentSchema = z.object({
```
marius-at-atb: The PriceAdjustment should be the same for BonusOffer and Benefit. Also I don't think we need systemIds on it anymore, since that logic is instead applied server side.

src/api/types/benefit.ts
```diff
@@ -13,8 +14,19 @@ export const MobilityPriceAdjustmentSchema = z
     description: x.description,
   }));
 
+export const MobilityBenefitSourceSchema = z.enum([
```
Dahly96: Is this to separate a regular benefit (open for all campaign) vs bundling? (And other kinds in the future if necessary). In what case is mobility-benefit used vs campaign?

src/api/types/benefit.ts
```diff
@@ -13,8 +14,19 @@ export const MobilityPriceAdjustmentSchema = z
...
+  'campaign',
+  'mobility-benefit',
+]);
+
+export type MobilityBenefitSource = z.infer<typeof MobilityBenefitSourceSchema>;
+
 export const MobilityPriceAdjustmentBenefitSchema = z.object({
   kind: z.literal('MOBILITY_PRICE_ADJUSTMENT'),
+  source: MobilityBenefitSourceSchema.optional(),
```
Dahly96: Same as [this comment](https://github.com/AtB-AS/mittatb-app/pull/6186/changes#r3458151258)

src/api/mobility.ts
```diff
@@ -178,6 +183,38 @@ export const getStation = (
...
+  const url = `/mobility/v1/app-switch/vehicle-type/${vehicleTypeId}`;
+  const query = qs.stringify({bonusProductId});
+  return client
+    .post<AppSwitchUrl>(stringifyUrl(url, query), undefined, {
+      authWithIdToken: true,
+    })
+    .then((response) => AppSwitchUrlSchema.parse(response.data));
+};
+
+export const getMockVehicle = (
```
marius-at-atb, line 199: `getStationBasedVehicle` and just `vehicles` instead of `mock-vehicles` sounds best to me. However if you find that confusing, using mock is also fine.

src/modules/mobility/components/sheets/VehicleSheet.tsx
```diff
@@ -127,18 +135,67 @@ export const VehicleSheet = ({
   } = useFeatureTogglesContext();
 
   const isBonusActiveForUser = useIsBonusActiveForUser();
-  const bonusProduct = useRelevantSharedMobilityBonusProduct(vehicleTypeId);
+  // Prefer the server-provided bonus offer; fall back to the client-side
+  // lookup only when the vehicle response omits it.
+  const clientBonusProduct =
+    useRelevantSharedMobilityBonusProduct(vehicleTypeId);
```
marius-at-atb, line 141: Finding the bonusProduct should be based off bonusProductId from bonusOffer. Let's not keep the old method around.

src/modules/mobility/components/sheets/VehicleSheet.tsx
```diff
@@ -127,18 +135,67 @@ export const VehicleSheet = ({
...
-    !isBicycle && isBonusActiveForUser && !!bonusProduct;
+    !isBicycle && isBonusActiveForUser && hasBonusOffer && !!clientBonusProduct;
+
+  // Drive the CTA from the server `actionButton` when present, otherwise fall
+  // back to the existing deep-integration / operator branching.
+  const showStartTrip = actionButton
+    ? actionButton.type === ActionButtonType.START_TRIP
+    : isDeepIntegrationEnabled &&
+      !!operatorId &&
+      !!operatorIsIntegrationEnabled;
```
marius-at-atb, line 196: Here as well, remove the legacy option.

src/modules/mobility/components/sheets/VehicleSheet.tsx
```diff
@@ -292,6 +361,54 @@ export const VehicleSheet = ({
...
+  serverBonusOffer,
+  clientBonusProduct,
+}: {
+  payWithBonusPoints: boolean;
+  vehicle: Vehicle;
+  vehicleTypeId: string | undefined;
+  serverBonusOffer: BonusOffer | undefined;
+  clientBonusProduct: BonusProductType | undefined;
+}): MobilityPriceAdjustmentBenefitType | undefined => {
+  if (!payWithBonusPoints) return vehicle.benefit ?? undefined;
```
marius-at-atb, line 370: By updating the MobilityPriceAdjustmentBenefitType as commented above, and removing the legacy solution, this function can be simplified a lot.

src/modules/mobility/components/sheets/VehicleSheet.tsx
```diff
@@ -292,6 +361,54 @@ export const VehicleSheet = ({
...
+
+  const systemId = vehicle.system.id;
+  const bonusPriceAdjustments = serverBonusOffer
+    ? serverBonusOffer.priceAdjustments
+        .filter((pa) => pa.systemIds.includes(systemId))
```
marius-at-atb: This filtering should already have been applied server side.

src/modules/mobility/components/BenefitIllustration.tsx
Dahly96: @marius-at-atb talked about reusing the TicketingTileIllustration logic, of mapping a name to an icon. As a separate new file for mobility. But im happy with either

marius-at-atb: Interesting discussion.
I'm fine with both this, or the same as was done for TicketingTileIllustration.
Maybe this can even be generic and reused by TicketingTileIllustration.
E.g. just `ThemedIllustration`. Thoughts?
@rosvik

src/modules/mobility/components/PriceDetailsCard.tsx
```diff
@@ -69,8 +71,34 @@ export const PriceDetailsCard = ({
       )
     : t(ScooterTexts.per.unit(ratePrUnit?.perUnit ?? ''));
 
+  const benefitTitle = getTextForLanguage(benefit?.title, language);
+  const benefitDescription = getTextForLanguage(benefit?.description, language);
+  const hasBenefitInfo = !!benefitTitle || !!benefitDescription;
+
   return (
     <Section>
+      {hasBenefitInfo && (
```
marius-at-atb, line 78: Could you check with @lineamb if this matches the most recent Figma sketches?

marius-at-atb (commented):
Nice!
The comments are mostly regarding cleanup of the legacy stuff, which you can choose if you want to include in this PR or in a new one.

src/api/types/benefit.ts
Dahly96: I dont think this should drive decisions in the frontend. Is it used for anything special as a frontend prop?
I think the pattern should be that the backend benefit type has some sort of source/decision on what kind of benefit it is. So when a call comes to getVehicle, and we fetch all benefits related to that vehicleTypeId, we should know there if there is some different rules (bundling) that should be verified before choosing what benefit to go for, like if there is multiple benefits for a vehicleTypeId. And the app just contains the simple benefit to show for that specific vehicle. Might not even need vehicleTypeIds and systemIds anymore, since its locked to a vehicle

marius-at-atb: Related: https://github.com/AtB-AS/mittatb-app/pull/6186#discussion_r3458102867. `systemId` also not needed going forward.

src/api/mobility.ts
marius-at-atb: Todo in new PR - use this instead of local mocking.

marius-at-atb: Update: planned changes [here](https://github.com/AtB-AS/mittatb-app/pull/6191/changes/49f3a27e0a0427203886cf3a76ce236fa1626f56..257e5fc686e9298637af1ba3f34eb71540da81d5) (also includes a bugfix for the selected map item icon after qr scanning).

src/modules/mobility/components/sheets/VehicleSheet.tsx
adriansberg: Agreed. Finding the bonus product by id (from `bonusOffer.bonusProductId`) is a rewire rather than a deletion, so I split it into a follow-up: AtB-AS/kundevendt#24106 (item 1). Kept it out of this PR to stay focused on the data-shape changes.

src/modules/mobility/components/sheets/VehicleSheet.tsx
adriansberg: Simplified here: the type is slimmed and the server-offer branch passes adjustments through. The remaining legacy `clientBonusProduct` branch removal depends on the by-id lookup, so it is tracked in AtB-AS/kundevendt#24106 (item 1).

src/modules/mobility/components/sheets/VehicleSheet.tsx
adriansberg: This one cascades into the `isShmoDeepIntegration*` toggles and `operatorIsIntegrationEnabled`, so I split it into its own change: AtB-AS/kundevendt#24106 (item 2).

src/api/types/benefit.ts
adriansberg: Tied into the BenefitIllustration component approach, so tracking the rename together with that discussion in AtB-AS/kundevendt#24106 (item 3).

src/modules/mobility/components/BenefitIllustration.tsx
adriansberg: Captured the open ThemedIllustration direction in AtB-AS/kundevendt#24106 (item 3) to resolve separately. Kept the current approach for now.

src/modules/mobility/components/sheets/VehicleSheet.tsx
adriansberg: Done. The bonus offer is now system-filtered server-side (bonus-service takes a `systemId` query param and returns only that system's adjustments), so the client-side filter is removed.

src/api/types/benefit.ts
adriansberg: Done — renamed to `BenefitSchema` and dropped `kind`/`vehicleTypeIds`/`systemIds`. The app models only the display payload now; kind/eligibility/vehicle+system filtering all live on the Vehicle response server-side.

src/api/types/mobility.ts
adriansberg: Done — `BonusOffer` and `Benefit` now share the same `PriceAdjustmentSchema` (mobility.ts imports it from benefit.ts).

src/api/types/benefit.ts
adriansberg: The field is now `illustrationName` (done). The broader ThemedIllustration component direction is tracked in AtB-AS/kundevendt#24106.

src/modules/mobility/components/sheets/VehicleSheet.tsx
```diff
@@ -196,30 +255,21 @@ export const VehicleSheet = ({
...
             />
           </View>
 
-          {isDeepIntegrationEnabled &&
-          operatorId &&
-          operatorIsIntegrationEnabled ? (
+          {showStartTrip && operatorId ? (
             <>
-              {showBonusCheckbox && (
+              {showBonusCheckbox && clientBonusProduct && (
```
marius-at-atb, line 270: Showing PayWithBonusPointsCheckbox should be based on bonusOffer. However, it can be addressed in a follow-up PR since we need that anyway after this one.

marius-at-atb (approved):
Approving as is so we can move on to the [follow up PRs](https://github.com/AtB-AS/kundevendt/issues/24106), which are required before new release.

src/api/mobility.ts
marius-at-atb: New update: the previous PR has been replaced by [this standalone one](https://github.com/AtB-AS/mittatb-app/pull/6218), and the `getMockVehicle` function here should be removed.

src/modules/mobility/components/sheets/VehicleSheet.tsx
adriansberg: Yes, covered in follow-up issue: https://github.com/AtB-AS/kundevendt/issues/24106

PR #6189, opened 2026-06-23
refactor: replace LegsSummary with BookingSummary

gorandalum:
The app typed the BFF DatedServiceJourney response as the broad
DatedServiceJourney from journey_planner_v3_types, but the BFF GraphQL
query actually returns a much smaller subset. The previous mapping in
useFareContractLegs cast that partial object back to Leg, leaving many
mandatory-typed fields actually undefined at runtime — a real footgun
for any caller relying on those type guarantees.

This commit tightens DatedServiceJourneyQuery to mirror the BFF query
and derives a narrower DatedServiceJourney in bff/types.ts. Mapping
this narrower DatedServiceJourney to Leg isn't viable anymore since
many fields mandatory on Leg are simply missing from the DSJ payload.
Instead, introduce a specialized BookingJourneySegment type carrying
only the data needed to render the legs summary (now renamed to
BookingSummary), and rename useFareContractLegs to
useFareContractBookingSegments accordingly.

### Acceptance criteria

- [x] Leg info (what departure the booking is for) is shown in the purchase process
- [x] Leg info (what departure the booking is for) is shown in the purchased fare contract


strandlie (approved):
Good cleanup! Makes sense.

PR #6190, opened 2026-06-23
feat: show error message on fare contract loading errors

rosvik:
## Background

This all started with reports from users that had just upgraded their app, that said their period tickets were missing. They then bought a new ticket, and ended up with two. This caused refund requests to customer service.

The likely technical explanation for this is that we discard the react query cache on app update (`buster: APP_VERSION`  in ReactQueryProvider), and if the app is then opened without a network connection, no tickets are shown from cache, and no information is given that things failed.

I wanted to add an error message for this, but not one that would appear on every failed request to `/ticket/v4/list`, since we have a 7 day cache that is fine to fall back to in most cases.

## Solution

Simulator Screenshot - iPhone 16 Pro - 2026-06-25 at 13 00 21

This PR adds an error message that appears

1. When the app has just been opened, and the first request to `/ticket/v4/list` fails.
2. When we know the current data is outdated, e.g. after ticket purchase and the fare contract stream event is received.
3. On pull to refresh

I didn't find an very elegant solution to this (e.g. using the built-in  `stale` state) in react query, so I opted to create a zustand store with a `needsRefresh` field for this instead. `needsRefresh` is true on app start, and whenever the fare contracts query is invalidated.

## Note

I feel like this was a lot harder to work with than it should have been, mostly because fare contract state is still connected to firestore loading logic in TicketingContext in ways that is hard to wrap your head around. We should migrate away from this as soon as we're sure the stream service solution is working well enough to not need a feature toggle.

## Acceptance criteria

- [x] When the app is launched without a network connection, an error message is shown in the "My tickets" list.
- [x] If requests to `/ticket/v4/list` after ticket purchase, the message is shown.
- [x] On pull to refresh, the message is shown
- [x] When requests to `/ticket/v4/list` otherwise fails without user interaction (e.g. after app focus), the message is not shown.
- [x] Clicking "Try again" should reload the list

### DIFF
--- a/src/modules/ticketing/use-fare-contracts.ts
+++ b/src/modules/ticketing/use-fare-contracts.ts
@@ -8,6 +8,8 @@ import {getAvailabilityStatus, AvailabilityStatus} from '@atb-as/utils';
 import {isDefined} from '@atb/utils/presence';
 import {ONE_WEEK_MS} from '@atb/utils/durations';
 import {getServerNowGlobal} from '../time';
+import {useFeatureTogglesContext} from '../feature-toggles';
+import {useRemoteConfigContext} from '../remote-config';
 
 type AvailabilityStatusInput = {
   availability: Exclude<AvailabilityStatus['availability'], 'invalid'>;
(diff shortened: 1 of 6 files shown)


src/modules/ticketing/use-fare-contracts.ts
```diff
@@ -28,14 +31,38 @@ export const useFareContracts = (
...
+    isFetching,
+    isError,
+    status,
+    isFetchedAfterMount,
+    dataUpdatedAt,
+  } = useGetFareContractsQuery({
+    enabled:
+      enable_ticketing &&
+      isEventStreamEnabled &&
+      isEventStreamFareContractsEnabled,
```
rosvik, line 52: The query needs to be enabled for props like `isError` to work. When we previously only used `refetch`, it was fine for it to be disabled. 

Enabling it also fixes a minor bug where the `isRefetching` state isn't working in production.

reidzeibel (approved):
Looks correct by looking at the code.

tormoseng:
✅  Will also test missing ticket responses when upgrading to a new version in Prod

PR #6198, opened 2026-06-24
feat: notify on forced selection changes when switching product

gorandalum:
When changing fare product forces a swap of traveller or zone, because
the previous selection isn't available for the new product, the
builder now reports the forced changes and the purchase overview shows
a snackbar so the user understands why their selection changed.


strandlie (approved):
It makes sense to have data about the build as part of a `PurchaseSelectionBuildResult`

PR #6193, opened 2026-06-24
refactor: Fix OverflowContainer rendering race condition

strandlie:
## Issue Reference
Closes https://github.com/AtB-AS/kundevendt/issues/24085

## Description
Refactors `OverflowContainer` to remove the fragile two-pass, `useEffect`-driven measurement that caused intermittent blank legs (getting stuck in the measure pass).

- The visible fit is now **derived during render** from cached child widths, instead of being set from a layout handler in a second pass. A child whose width is measured before `maxWidth` arrives no longer gets stuck, and no reset/remount is needed.
- The fit math is extracted into a pure `computeFit` helper with unit tests.

No behavior change for consumers. A follow-up PR adds travel-card skeleton loading on top of this (it relies on the reliable measurement this refactor provides).

PR #6204, opened 2026-06-25
feat: skeleton loading for travel cards

strandlie:
## Issue Reference
Follow up to https://github.com/AtB-AS/mittatb-app/pull/6193 and https://github.com/AtB-AS/kundevendt/issues/24085

## Screenshots
| Trip Search | Trip Search (Load more) | Booking (Trip Selection) |
|--------|-------|-------|
|  |  |  |

## Description
Adds skeleton-loading placeholders for travel cards in the new trip search and booking trip selection, both during load and measure. 

- New `WithTravelCardSkeleton` wraps each result slot: it shows a `TravelCardSkeleton` immediately, mounts the real card hidden to measure its legs, then reveals it once measured. The same slot instance persists from loading through measure, so the shimmer stays continuous.
- `OverflowContainer` gains an `onReady` callback (fires once the legs are measured and revealed); `TravelCard` → `TravelCardLegs` forward it.
- Consumers: trip-search results, booking trip selection, and the walk/bike (non-transit) row + day labels, all showing skeletons sized to the real content (font-scaled).


src/screen-components/travel-card/skeleton/TravelCardSkeleton.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <View style={styles.legsArea}>
+          <LegsSkeleton />
+        </View>
+        <ThemeIcon svg={ChevronRight} color="secondary" />
+      </View>
+    </View>
+  );
+};
+
+const useThemeStyles = StyleSheet.createThemeHook((theme) => ({
```
strandlie, line 35: A lot of hard coded values below here. Hard to know how else to do that, since a skeleton is mocking something that is sized depending on data.

src/components/overflow-container/OverflowContainer.tsx
```diff
@@ -37,9 +38,12 @@ export const OverflowContainer: React.FC<Props> = ({
...
 
-  const entering = FadeIn.duration(revealDuration).easing(
-    Easing.out(Easing.ease),
-  );
+  const onReadyMemo = useCallback(() => onReady?.(), [onReady]);
+  useEffect(() => {
+    if (fitResult != null) {
+      onReadyMemo();
+    }
+  }, [fitResult, onReadyMemo]);
```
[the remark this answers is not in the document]
strandlie: No. This cost is marginal, and checking also has its cost.

src/stacks-hierarchy/Root_TabNavigatorStack/TabNav_DashboardStack/Dashboard_TripSearchScreen/components/Results.tsx
```diff
@@ -90,45 +97,85 @@ export const Results: React.FC<Props> = ({
...
+    const slotCount =
+      filteredTripPatterns.length + (isSearching ? SKELETON_COUNT : 0);
+
+    return (
+      <View style={styles.container} testID="tripSearchContentView">
+        {Array.from({length: slotCount}, (_, i) => {
+          const tripPattern = filteredTripPatterns[i];
+          return (
+            <Fragment key={i}>
+              {tripPattern ? (
```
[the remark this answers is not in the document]
strandlie: Fixed.

src/screen-components/travel-card/skeleton/LegsSkeleton.tsx
```diff
@@ -0,0 +1,36 @@
...
+    </View>
+  );
+};
+
+const useStyles = StyleSheet.createThemeHook((theme, _, {fontScale}) => ({
+  row: {
+    flexDirection: 'row',
+    gap: theme.spacing.xSmall,
+  },
+  // Stadium shape and height matching the transport-mode boxes (icon + padding).
```
rosvik, line 30: What's a stadium?

rosvik (approved):
Nice work 🙌 LGTM :shipit:

src/screen-components/travel-card/skeleton/LegsSkeleton.tsx
strandlie: It came from claude, but I found it pretty accurate 🤷🏼‍♂️

rosvik: TIL 💡

It sounds more grand than 'pill' at least 😆

hanne-at-atb:
When testing I noticed that while loading we now show the arrow indicating it can be clicked, but in reality you can not click while loading and have to wait until actual content is shown. Have we made an conscious choice to show it like this @strandlie / @RHatlem ? 
Not a big problem, but people with slow internet connection will see this skeleton loader for several seconds.

hanne-at-atb:
I also notice that the screen reader skips the loading trip search, so it is not in the focus order and the user does not get info that new search results are loading in the list of results. If you click "load more results" the screen reader notifies that more results are loaded. But yet you can not reach the visual element that is loading with screen reader.

RHatlem:
@hanne-at-atb @strandlie If you can't click on the arrow while loading, i think it is better to not show the arrow befor you can click on it

strandlie:
@hanne-at-atb @RHatlem These are fixed in latest staging now!

hanne-at-atb:
Looks and sounds good to me! ✅

PR #6205, opened 2026-06-25
feat: show affected stop places in situation bottom sheet

gorandalum:
Adds an "Affects" line to the situation bottom sheet that lists the
stop places (or quays) impacted by a situation. Up to 10 names are
shown in full; beyond that the list is truncated to "first 8 + N other
stops".


rosvik (approved):
LGTM 🚀

PR #6206, opened 2026-06-26
fix: refresh inspection on screen focus

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/24116

## Description
Attempt to fix the issue where the inspection is invalid when screen is just unlocked/focused, by restarting the inspection process.

Hopefully this solves the FRAM token issue.

The old fix https://github.com/AtB-AS/mittatb-app/pull/5406, is only used for the old renderer/MobileTokenAztec, and doesn't affect the new RNBarcodeView


strandlie (approved):

tormoseng:
✅  Looks good for both FRAM and AtB

PR #6211, opened 2026-06-30
fix: add bottom safe area inset to refund reason screen

rosvik:
## Issue Reference

https://mittatb.slack.com/archives/C0116FMPX4Y/p1782818808543369

## Description

Same solution as e.g. Root_PurchaseConfirmationScreen


reidzeibel (approved):
Looks good, but does `SafeAreaView` work here? or `useSafeAreaInsets`?

rosvik:
> Looks good, but does `SafeAreaView` work here? or `useSafeAreaInsets`?

I actually have an alternative solution ready, but doing the simple fix for the release ✌️

PR #6212, opened 2026-06-30
refactor: add useStreamEventListener hook

rosvik:
## Issue Reference

closes https://github.com/AtB-AS/kundevendt/issues/23848

## Description

Adds a `useEventStreamListener` hook to make it more ergonomic to act on events from the event stream. When adding a new listener, supply event kind and a callback. This works in the following way:

- `useSetupEventStream` returns `subscribe` and `unsubscribe` methods. These stores `StreamEventListener` objects, that contain the callback and eventKind, in addition to an id that's used for unsubscribing.
- On events, any callbacks that are saved with the recieved eventKind, will be called.
- An `useStreamEventListener` hook is returned from the context, which makes it easy to add more event listeners from anywhere in the app.
- Global event listeners are added to `src/modules/event-stream/use-global-event-stream-listeners.ts`, which is called from RootStack.


marius-at-atb (approved):
Looks great!

PR #6214, opened 2026-07-01
fix: getFeatureFromScan

marius-at-atb:
Bugfix: select the correct map item after a qr scan. This fixes the map item icon of a vehicle/station after scanning with a qr code.

| Before | After |
|--------|-------|
|  |  |

Acceptance Criteria:
- [x] Scanning a vehicle works
   - [ ] Correct item in bottom sheet
   - [ ] Correct icon/item in map
- [x] Scanning a station or vehicle that returns a station works (when only `stationId` is set in the qr response)
   - [ ] Correct item in bottom sheet
   - [ ] Correct icon/item in map

### DIFF
--- a/src/modules/map/utils.ts
+++ b/src/modules/map/utils.ts
@@ -36,7 +36,6 @@ import {
   isVehicleCluster,
   isStationCluster,
 } from '@atb/modules/mobility';
-import {MapBottomSheetType} from './MapContext';
 import {
   FormFactor,
   PropulsionType,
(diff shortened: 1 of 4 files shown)


src/modules/map/utils.ts
```diff
@@ -133,38 +135,49 @@ export const getFeaturesAtPoint = async (
...
-  mapBottomSheetType?: MapBottomSheetType,
-): FormFactor | undefined {
-  switch (mapBottomSheetType) {
-    case MapBottomSheetType.BikeStation:
-      return FormFactor.Bicycle;
-    case MapBottomSheetType.CarStation:
-      return FormFactor.Car;
-    default:
-      return undefined;
+export const getFeatureFromScan = ({
```
adriansberg, line 145: Could possibly add a comment that either `vehicle` or `station` can be passed as an argument.

adriansberg (approved):
Looks good!

tormoseng:
👍  Will need to test city bike scanning in Prod

PR #6215, opened 2026-07-01
fix: sort departure list by expected departure time

strandlie:
## Issue Reference
Slack (with video and description): https://mittatb.slack.com/archives/C0116FMPX4Y/p1780299944608929

## Description
The departure list ("avganger") could show departures out of order relative to their displayed times — e.g. "5 min" listed above "3 min". 

The full-fetch path returned departures in BFF order and was never re-sorted client-side, and that is a problem since the BFF / Entur seems to return in `aimedDepartureTime` order, while we display `expectedDepartureTime`. The refresh-fetch path already sorted by `expectedDepartureTime`, so this does that for full-fetch also.

### DIFF
--- a/src/departure-list/utils.test.ts
+++ b/src/departure-list/utils.test.ts
@@ -65,6 +65,14 @@ describe('sortEstimatedCallsByExpectedTime', () => {
       '2024-01-01T07:03:00Z',
     ]);
   });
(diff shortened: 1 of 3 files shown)


strandlie:
Closing this. Didn't get time to look into it further.

PR #6218, opened 2026-07-03
feat: get station based vehicles from new endpoint

marius-at-atb:
## Issue Reference

Part of https://github.com/AtB-AS/kundevendt/issues/24096
Depends on https://github.com/AtB-AS/mobility/pull/113

## Description

Rather than create a station based vehicle in the app, it is now fetched from the mobility svc based on vehicleTypeId and stationId. With the new API for deep integration and benefits from mobility svc, this means we can align station based vehicles with normal vehicles. Otherwise we'd have to add new props to vehicleTypesAvailable on `station` as well. However that would not play well with potential refresh mechanisms, while a separate endpoint does.

This also fixes a bug for getting the correct state after opening a station based vehicle, and then logging in or out:

| Before | After |
|--------|-------|
|  |  |

**Acceptance Criteria:**
- [x] Correct station based vehicle data received (matching the station and vehicleType selected just before)
- [x] Existing vehicles not affected
- [x] No unnecessary api calls / guaranteed errors

### DIFF
--- a/src/api/mobility.ts
+++ b/src/api/mobility.ts
@@ -166,7 +166,6 @@ export const getVehicle = (
         : `/mobility/v1/stations/${stationId}/mock-vehicles/${vehicleTypeId}`,
       {
         ...opts,
-        ...{baseURL: 'http://localhost:8080'},
       },
     )
     .then((response) => {
--- a/src/api/mobility.ts
+++ b/src/api/mobility.ts
@@ -152,14 +152,23 @@ export const getVehicles = (
 };
 
 export const getVehicle = (
-  id?: string,
+  vehicleId?: string,
+  vehicleTypeId?: string,
+  stationId?: string,
   opts?: VehicleRequestOpts,
 ): Promise<Vehicle | null> => {
-  if (!id || id === '') return Promise.resolve(null);
+  if (!vehicleId && (!vehicleTypeId || !stationId))
+    return Promise.resolve(null);
   return client
-    .get(`/mobility/v1/vehicles/${id}`, {
-      ...opts,
-    })
+    .get(
+      !!vehicleId
+        ? `/mobility/v1/vehicles/${vehicleId}`
+        : `/mobility/v1/stations/${stationId}/mock-vehicles/${vehicleTypeId}`,
+      {
+        ...opts,
+        ...{baseURL: 'http://localhost:8080'},
+      },
+    )
     .then((response) => {
       const result = VehicleSchema.safeParse(response.data);
       if (!result.success) {
--- a/src/modules/mobility/components/ShmoActionButton.tsx
+++ b/src/modules/mobility/components/ShmoActionButton.tsx
@@ -18,7 +18,7 @@ import {useMapContext, useShmoWarnings} from '@atb/modules/map';
 import {MessageInfoText} from '@atb/components/message-info-text';
 import {AgeVerificationEnum} from '../queries/use-get-age-verification-query';
 import {useAnalyticsContext} from '@atb/modules/analytics';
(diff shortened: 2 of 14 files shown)


src/modules/mobility/components/ShmoActionButton.tsx
```diff
@@ -46,10 +46,13 @@ export const ShmoActionButton = ({
   const {theme} = useThemeContext();
   const styles = useStyles();
   const coordinates = getCurrentCoordinatesGlobal();
-  const {warningMessage} = useShmoWarnings(vehicleId);
   const {logEvent} = useAnalyticsContext();
-  const {vehicle} = useVehicle(vehicleId);
   const {mapState} = useMapContext();
+  const {vehicle} = useMapVehicle();
+  const {warningMessage} = useShmoWarnings(
+    // Shmo warnings not yet supported for station based vehicles.
```
marius-at-atb, line 53: Todo: https://github.com/AtB-AS/kundevendt/issues/24153

src/api/mobility.ts
```diff
@@ -147,14 +147,22 @@ export const getVehicles = (
...
+  if (!vehicleId && (!vehicleTypeId || !stationId))
+    return Promise.resolve(null);
   return client
-    .get(`/mobility/v1/vehicles/${id}`, {
-      ...opts,
-    })
+    .get(
+      !!vehicleId
+        ? `/mobility/v1/vehicles/${vehicleId}`
+        : `/mobility/v1/stations/${stationId}/mock-vehicles/${vehicleTypeId}`,
```
marius-at-atb, line 166: We could have created completely separate functions and hooks for the new endpoint, but the whole point of station-based vehicles is to keep them exactly in line with normal vehicles, so opted for this instead.

Dahly96 (approved):

hanne-at-atb:
Now doing a call to `/mobility/v1/vehicles/56583` that fails with 404 and retries many times after starting a trip with city bike. The call is done just after the `bookings/active` call. 

Fixed by: https://github.com/AtB-AS/mittatb-app/pull/6225

Retested ok ✅

hanne-at-atb:
When starting a city bike trip for electric bike we should not have to push the button to undock the ebike.
 
Before this PR (testing yesterday 6.jul) when starting a trip the ebike was undocked automatically after 10-15 sec and you could make a trip. 

Now testing this merged PR I start a trip (`/one-stop` call made) and the ebike does not undock automatically. I wait for about 1 min and the app shows screen with dock number this whole time, suddenly the dock number disappears (`/active `call done). And then the screen in the app after approx. 2 min changes back to show the bottom sheet with the city bike overview and option to start trip. (`/active `call followed by a call to `/stations/YTR:Station:10907/mock-vehicles/YTR:VehicleType:ebike`)
So the behaviour has changed after this PR and requires you to manually undock 🤷🏼  <-- should look into this!

It might be affected by pricePlan changes done by Urban sharing yesterday evening inbetween testing.

hanne-at-atb:
For a brief moment between swapping bottom sheets, the sheet has a loading state or in-between state where it shows a UFO icon and loading spinner in the sheet. 
Edit: Tested previous builds and have confirmed that the UFO was introduced by this PR, so I think we should look at it. 
 
https://github.com/user-attachments/assets/042b58f3-f346-44ca-aec6-fa4509773b61

Fixed in https://github.com/AtB-AS/mittatb-app/pull/6239
Retested ok ✅

hanne-at-atb:
2 of 3 findings fixed. The last one I'm not sure is caused by this PR. Should look more into it and retest once we have everything in place for start/stop of ebikes and normal bikes. Possibly testing in prod. 

This PR is good: ✅

PR #6223, opened 2026-07-08
test: add e2e exploring and creation with mcp

tormoseng:
Adding the possibility to enable Claude Code to interact with the app through Appium and the WebdriverIO MCP server (https://github.com/webdriverio/mcp). This allows to specify commands in Claude that are translated to requests that Appium understands so that Claude can interact with the app.

Tested that Claude can explore and create a test based on e.g. "Open the app and buy  a single ticket, but stop before the payment confirmation" that also utilize the exiting test-ids and page objects.


rosvik (approved):
Fancy ✨

PR #6224, opened 2026-07-08
revert: incomplete vehicle 404 fix from #6222

adriansberg:
Reverts #6222.

The `useMapVehicle` guard added in #6222 only covered the `Map.tsx` fetch path. The active-trip 404 on `/mobility/v1/vehicles/:id` also originates from `ActiveShmoSheet` → `useShmoWarnings` → `useVehicle`, which #6222 did not touch — so testers still hit repeated 404s during an active booking (confirmed for bicycle rides).

Reverting here so the fix can be replaced with a single centralized guard in `useVehicleQuery` (separate PR).


jorunnl (approved):

hanne-at-atb:
Revert seems ok ✅

PR #6213, opened 2026-07-01
refactor: auto handle safe area padding of FullScreenView

rosvik:
## Issue Reference

https://github.com/AtB-AS/mittatb-app/pull/6211#issuecomment-4851548341

## Description

Checks if there is a bottom tab bar on the current screen, and adds bottom safe area inset if needed.

## Acceptance criteria

- [x] Screens without the bottom tab bar has the appropriate amount of bottom safe area padding
- [x] Screens with the bottom tab bar doesn't have excessive bottom padding.

PR #6227, opened 2026-07-09
fix: improved android keyboard avoiding full screen views

rosvik:
## Issue Reference

Fixes feedback from @hanne-at-atb on https://github.com/AtB-AS/mittatb-app/pull/6213#issuecomment-4891826117

## Description

- Sets keyboard avoiding behavior for FullScreenFooter to be `undefined` on Android, while keeping `'padding'` on iOS. I've found this works the best on on both platforms.
- Removed a bit of padding at the bottom of the payment method bottom sheet
- Updated some screens to use FullScreenView: 
	- FareContractDetailsScreen
	- ContactShmoOperatorScreen
	- EnrollmentScreen

### DIFF
--- a/src/components/screen-footer/FullScreenFooter.tsx
+++ b/src/components/screen-footer/FullScreenFooter.tsx
@@ -1,6 +1,6 @@
 import {StyleSheet, Theme} from '@atb/theme';
 import React from 'react';
-import {KeyboardAvoidingView, View} from 'react-native';
+import {KeyboardAvoidingView, Platform, View} from 'react-native';
 
 type ScreenFooterProps = {
   children: React.ReactNode;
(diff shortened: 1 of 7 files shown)


src/components/screen-footer/FullScreenFooter.tsx
```diff
@@ -17,7 +17,10 @@ export function FullScreenFooter({
   const containerStyle = {...styles.container, backgroundColor: footerColor};
 
   return avoidKeyboard ? (
-    <KeyboardAvoidingView behavior="padding" style={containerStyle}>
+    <KeyboardAvoidingView
+      behavior={Platform.OS === 'android' ? undefined : 'padding'}
```
rosvik, line 21: We don't actually use the `avoidKeyboard` prop for FullScreenFooter anywhere, but this should make it work better on Android once we do.

gorandalum (approved):

hanne-at-atb:
Looks good in test ✅

PR #6230, opened 2026-07-09
chore: update header borders

rosvik:
Two very small details with header borders:

- On ticketing root screen `textOpacity={0}` was used instead of `showBorder={false}` (No functional difference)
- Removed a border on the location search by map screen that I don't think should be there 👇

IMG_F931AC0DBACF-1

### DIFF
--- a/src/components/screen-header/ScreenHeader.tsx
+++ b/src/components/screen-header/ScreenHeader.tsx
@@ -1,9 +1,4 @@
-import {
-  LayoutChangeEvent,
-  LayoutRectangle,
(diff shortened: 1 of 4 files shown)


gorandalum (approved):
Nice. I also came to think of that the `showBorder` and `textOpacity` stuff should probably not be configurable when using `FullScreenView`? Maybe take a look at omitting them from the `headerProps` field type.

rosvik:
@gorandalum Omitted those now. Also noticed a `style` prop that was unused

PR #6232, opened 2026-07-10
feat: prevent QR code screenshots on iOS

rosvik:
## Issue Reference

First part of https://github.com/AtB-AS/kundevendt/issues/23927

## Description

Adds a TurboModule component `<SecureView>` that hides its children from screenshots and recordings on iOS. It's a bit of a hack to make this work, since iOS doesn't provide an official API for this. Instead we put react content inside a special canvas layer that iOS creates as part of a UITextField with `isSecureTextEntry=true`. See the comment in SecureViewImpl.swift for a proper explanation.

Android should be unaffected for now, and work like before.

The resulting screenshots look like this:

IMG_4943

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -216,6 +216,7 @@ module.exports = {
         'types/*',
         'src/translations/**',
         'src/modules/storybook/stories/**',
+        'src/modules/native/*NativeComponent.ts',
       ],
       rules: {
         'no-restricted-exports': [
--- a/ios/TurboModules/RCTSecureView.h
+++ b/ios/TurboModules/RCTSecureView.h
@@ -2,12 +2,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/**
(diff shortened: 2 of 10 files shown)


strandlie:
Nice solution. Could we add some info to that view? 

`Det er ikke tillatt å dele billetter med andre. Prøver du å kjøpe til andre? Trykk "Kjøp til andre og send til deres mobilnummer"`

rosvik:
@strandlie Not sure. We might be able to hide something behind the QR code that becomes visible when the QR is hidden. Next task here is to show an alert after screenshot is taken, so I'm thinking that's good enough 🤷 https://www.figma.com/design/hLy6zqtbKs0QI2XlXxd4A6/%F0%9F%8F%9C%EF%B8%8F-Sandkasse-for-alt-mulig?node-id=13304-55671&t=UkBJl90HS9DYZoKc-0

gorandalum (approved):

PR #6233, opened 2026-07-10
refactor: extract map styles to @atb-as/mapbox-shared

gorandalum:
Move mapbox style JSON, pin types, and NSR layer/util definitions out
of the app and consume them from the newly published
@atb-as/mapbox-shared package (^1.0.3), shared with the web planner.


rosvik (approved):
Clean! 🧼

PR #6243, opened 2026-07-14
fix: refresh geolocation when app returns from background

strandlie:
## Issue Reference
Fixes AtB-AS/kundevendt#23978

## Description
Location stayed stale after returning from background. Adds `maximumAge: 60s` (library default: infinite), a one-shot fresh fix on background→active, and a staleness guard on the `getCurrentCoordinates` cache.

### DIFF
--- a/src/modules/geolocation/GeolocationContext.tsx
+++ b/src/modules/geolocation/GeolocationContext.tsx
@@ -4,6 +4,7 @@ import React, {
   useContext,
   useEffect,
   useReducer,
+  useRef,
   useState,
 } from 'react';
 import {Alert, Linking, Platform} from 'react-native';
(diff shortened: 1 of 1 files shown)


src/modules/geolocation/GeolocationContext.tsx
```diff
@@ -230,6 +236,21 @@ export const GeolocationContextProvider = ({children}: Props) => {
     }
   }, [currentLocationError, appStatus, geoLocationName]);
 
+  const previousAppStatusRef = useRef(appStatus);
+
+  useEffect(() => {
```
strandlie: The point of this is to force a new location fetch when the app comes to the foreground again. Might not be necessary,

strandlie:
Started a staging build [here](https://github.com/AtB-AS/mittatb-app/actions/runs/29332069386) to test

rosvik (approved):

PR #6247, opened 2026-07-21
fix: overflowing trip details row labels

gorandalum:
Needed to go back to right aligned row label to make room for the
"ca." prefix. Also fixed wrapping when the row label content is
too big.

**Background**
https://mittatb.slack.com/archives/C0116FMPX4Y/p1784496657121129

**Before**
<div>

</div>

**After**
<div>

</div>

### Acceptance criteria
- [x] There is room for the "ca." prefix on same line when standard font size on normal sized phones
- [x] The text wraps if not enough room for "ca." prefix on same line


Dahly96 (approved):
LGTM

tormoseng:
✅

PR #6253, opened 2026-07-28
feat: added bundlingView in the app for deepintegration vehicles

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/22839

Added title, description and illustrationName if it exisits on a benefit, it should show under the price info as shown in the sketches.
It uses the newly added benefit attribute on a vehicle fetched from mobility. It does not need to be bundling. But for now we do have a title and description for bundling, so it will show. But we could add info on other benefits if we like aswell, and it will show in the same way

Seed data PR for the benefits needed in the database in order for it to show in the app https://github.com/AtB-AS/benefit/pull/17

image


marius-at-atb (approved):
Lgtm

PR #6254, opened 2026-07-30
feat: show parking spaces for citybikes in active trip

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/23712

Made it possible to see stations in the map in an active citybike trip, displaying the number of slots available for parking in the map.

You can press the parking places and you get a snackbar with the same info, displaying the amount of free parkingplaces.

image

image


src/modules/map/Map.tsx
```diff
@@ -223,6 +236,35 @@ export const Map = (props: MapProps) => {
...
+      // Always refetch (staleTime 0): dock availability changes constantly - not least
+      // from the user's own rental, so a value cached before the trip started is wrong.
+      // Going through the shared cache still keeps useStationQuery consumers in sync.
+      const station = await queryClient.fetchQuery({
+        queryKey: getStationQueryKey(stationId),
+        queryFn: ({signal}) => getStation(stationId, {signal}),
+        staleTime: 0,
+      });
+      const parkingSpacesAvailable =
+        station?.numDocksAvailable ?? fallbackParkingSpacesAvailable;
```
marius-at-atb: Rather than sending a station query, we should just use the data already in the map item. That is not outdated, since it is refreshed by the vector source.

marius-at-atb: Update: I see that `fallbackParkingSpacesAvailable` is already assigned that data, so we can just rename it to `parkingSpacesAvailable` and remove the query above.

src/modules/mobility/queries/use-station-query.tsx
```diff
@@ -2,9 +2,11 @@ import {useQuery} from '@tanstack/react-query';
 import {getStation} from '@atb/api/mobility';
 import {ONE_MINUTE_MS} from '@atb/utils/durations';
 
+export const getStationQueryKey = (id: string) => ['getStation', id];
```
marius-at-atb, line 5: No longer needed to export this after https://github.com/AtB-AS/mittatb-app/pull/6254/changes#r3703383513

src/modules/map/hooks/use-map-symbol-styles.ts
```diff
@@ -166,9 +186,7 @@ export const useMapSymbolStyles = ({
         : [
             'case',
             ['==', iconCode, 'citybike'],
-            'bikes',
-            ['==', iconCode, 'ebike'],
-            'bikes',
```
marius-at-atb, line 171: Not safe to remove these two lines I think

marius-at-atb (changes requested):
Nice work!
See a couple of comments.

Note: Something we can come back to later is to make an implementation that works for any formFactor/vehicle with regards to parking (e.g. cars as well).

src/modules/map/Map.tsx
Dahly96: Agreed, done

Dahly96: Renamed, and the count no longer touches the query. Kept the query itself for the **station name** though — that is not part of the tile data (`StationFeatureProperties` has no name, nor does the `stations` table in the map service), and the issue asks for the station name as the snackbar heading. Happy to drop the query and the title together if you would rather not fetch here.

src/modules/mobility/queries/use-station-query.tsx
Dahly96: Still needed, since the query stays for the station name — see my reply above. Will remove it if we decide to drop the title.

src/modules/map/hooks/use-map-symbol-styles.ts
Dahly96: Restored. They are actually unreachable now — I made station pins always resolve to `citybike`, since there is no `stationpin_ebike_*` sprite in the sheet (only `citybike` and `sharedcar`) and e-bikes share stations with regular bikes, so the pin should not vary by propulsion. But relying on that from a different hunk is fragile, so keeping them is the safer call.

src/modules/map/Map.tsx
```diff
@@ -318,19 +357,26 @@ export const Map = (props: MapProps) => {
...
+          // instead of opening a bottom sheet. Other features stay non-interactive.
+          if (isActiveBikeTrip && isBikeStation(featureToSelect)) {
+            const {id, capacity, num_vehicles_available} =
+              featureToSelect.properties;
+            showBikeStationParkingSnackbar(
+              id,
+              Math.max(0, capacity - num_vehicles_available),
+            );
+          }
+          return;
```
marius-at-atb: If the feature is a cluster of stations, it should zoom in rather than do nothing.

src/modules/map/Map.tsx
marius-at-atb: Interesting. We can either
- skip the specific name
- keep the request as is
- add the name on map items and get it from there instead

I'm tempted to suggest the latter, but as long as performance is good, I guess any of the 3 options are ok.

Dahly96: agreed, we should have name on the mapItem, but i created a separate issue for that:

https://github.com/AtB-AS/kundevendt/issues/24310

marius-at-atb (approved):
🎉

PR #6255, opened 2026-07-31
feat: added fix for big text in startTripOverlay

Dahly96:
Fixes this comment https://github.com/AtB-AS/kundevendt/issues/22834#issuecomment-4924739759

 With big text size the stationSlotName was pushed up behind the camera. This makes it stop before it goes out of frame
 
 
image


gorandalum (approved):

hanne-at-atb:
Tested and ok ✅

PR #6256, opened 2026-07-31
fix: fixed issue with too many payment cards causing scrolling to bug out

Dahly96:
when selecting a paymentcard from a bottomsheet sheet, like vehiclesheet, if you had to many the scroll would bugout and not make the confirmbutton unreacheble. 

I changed it to a confirm bottomsheet instead, matching other places in the app that has autosave, and not a confirm button.  Also addressed the scroll-issue with leaving it up to the buttomsheet itself to handle scroll instead of nesting a scrollview.

Fixes https://github.com/AtB-AS/kundevendt/issues/23943

### DIFF
--- a/src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
+++ b/src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
@@ -1,11 +1,8 @@
 import React, {useEffect, useState} from 'react';
 import {View} from 'react-native';
-import {StyleSheet, useThemeContext} from '@atb/theme';
-import {Button} from '@atb/components/button';
+import {StyleSheet} from '@atb/theme';
 import {useTranslation} from '@atb/translations';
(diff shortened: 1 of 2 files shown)


src/modules/mobility/components/sheets/SelectShmoPaymentMethodsSheet.tsx
```diff
@@ -79,54 +69,29 @@ export const SelectShmoPaymentMethodSheet = ({
...
-            ) {
+        {recurringPaymentMethods?.map((method, index) => (
+          <SinglePaymentMethod
+            key={method.recurringPayment?.id}
+            paymentMethod={method}
+            selected={
+              selectedPaymentMethod?.recurringPayment?.id ===
+              method.recurringPayment?.id
+            }
+            onSelect={(val: PaymentMethod) => {
```
marius-at-atb, line 80: Just my preference - call it paymentMethod instead of val

marius-at-atb (approved):
Lgtm

PR #6257, opened 2026-08-04
fix: hide refund action when refund window has passed

gorandalum:
Uses the new lastRefundTime field from the sales API to hide the
refund link on ticket details when the refund window has passed,
and to disable the confirm button on the refund confirmation screen
with an accompanying message when the window closes while the sheet
is open.

Fixes AtB-AS/kundevendt#23751


marius-at-atb (approved):
Lgtm

PR #6259, opened 2026-08-05
fix: use vehicle based endpoint for app switch

marius-at-atb:
Follow-up to https://github.com/AtB-AS/mittatb-app/pull/6244 after an API change.
Requires update to https://github.com/AtB-AS/mobility/pull/116/.

## Description

After realising that the app switch endpoint is vehicle specific due to rentalAppUri, we are changing
`POST /mobility/v1/app-switch/vehicle-type/${vehicleTypeId}?bonusProductId={bonusProductId}&rentalAppUri={rentalAppUri}`;
to
`POST /mobility/v1/vehicles/${vehicleId}/app-switch?bonusProductId={bonusProductId}`;

This also has the benefit of not using the rentalAppUri input directly in the output (avoids "shit in, shit out").

Also added support for the station based vehicle version

PR #6264, opened 2026-08-06
fix: pick real vs. mock-vehicle app-switch endpoint by booking type, not fetched id

Dahly96:
Follow-up to #6259.

## Description

`vehicle?.id` is truthy for a station-based (mock) vehicle too — it's the server's synthetic `"mock:..."` id, not a real vehicle id — so `getVehicleAppSwitch`'s `!!vehicleId` check in `getVehiclePath` always picked the real-vehicle endpoint (`/vehicles/:vehicleId/app-switch`), even for docked vehicles that have no real vehicle id at all.

This broke the app-switch/voucher flow for docked, non-deep-integrated vehicle types (e.g. Trondheim Bysykkel): the request would carry a fake vehicle id to the real-vehicle endpoint, which now (mobility PR [#116](https://github.com/AtB-AS/mobility/pull/116)) only resolves real Entur vehicle ids and 404s on anything else.

## Fix

Gate `vehicleId` on `mapState.isStationBasedBooking` before deriving it from the fetched vehicle, matching how `useMapVehicle` already decides this before fetching the vehicle in the first place:

```diff
- const vehicleId = vehicle?.id;
+ const vehicleId = mapState.isStationBasedBooking ? undefined : vehicle?.id;
```

With `vehicleId` correctly `undefined` for a mock vehicle, `getVehiclePath` falls through to the dedicated `/stations/:stationId/mock-vehicles/:vehicleTypeId/app-switch` route.

## Tests

`tsc --noEmit`, `eslint`, `prettier -c src`, and the full Jest suite (62 suites / 806 tests) all pass.

### DIFF
--- a/src/modules/mobility/components/sheets/VehicleSheet.tsx
+++ b/src/modules/mobility/components/sheets/VehicleSheet.tsx
@@ -100,7 +100,7 @@ export const VehicleSheet = ({
     appStoreUri,
   } = useMapVehicle();
 
-  const vehicleId = vehicle?.id;
+  const vehicleId = mapState.isStationBasedBooking ? undefined : vehicle?.id;
 
   const formFactor = vehicle?.vehicleType.formFactor ?? FormFactor.Other;
   const propulsionType = vehicle?.vehicleType.propulsionType;


marius-at-atb (approved):
:+1:

PR #6276, opened 2026-08-12
fix: update get apk script

tormoseng:
There has been an update in the Firebase appBinaryDownload URI, where a short lived token now is included in the URI. Therefore, the script doesn't need an Authority header anymore - simplifying the script.


reidzeibel (approved):

hanne-at-atb (approved):
👍🏼

PR #6277, opened 2026-08-13
feat: made end-trip card not dismissable

Dahly96:
Made EndTripCard not dismissable

image


marius-at-atb (approved):
Nice

hanne-at-atb:
Tested and you can not dismiss it now. 
Did you also remove the button to reset the dismissed notice from debug menu @Dahly96 ? 😄 Didn't find it which I guess is as expected.

PR #6280, opened 2026-08-14
fix: wrap price information rows on large font scale

adriansberg:
## Issue Reference

https://github.com/AtB-AS/kundevendt/issues/23813

## Description

The rows in the SHMO price information screen were a flex row with two non-shrinkable `ThemeText`s. Since `Text` defaults to `flexShrink: 0`, the label kept its full single-line width at increased phone zoom/font scale, overflowed the card, and pushed the value outside the clipped bounds so it looked like it disappeared.

- Label and value can now shrink, so the text wraps instead of overflowing.
- Value stays right-aligned when it breaks onto two lines.
- Added a small gap between label and value.


hanne-at-atb (approved):
👍🏼

hanne-at-atb:
Works well from what I can see! ✅

PR #6281, opened 2026-08-14
fix: hide shared car stations during active city bike trip

adriansberg:
## Issue Reference

https://github.com/AtB-AS/kundevendt/issues/23712#issuecomment-5291466306 (note 1)

## Description

Shared car stations (Hyre) were shown in the map during an active city bike trip. They are not relevant during a bike trip, and they are not even interactive while a trip is in use (`Map.tsx` returns early on tap for everything but bike stations), so they only add noise.

Station visibility was previously derived from the map filter alone, in two places: `showStations` in `Map.tsx` and the form factor filter expression inside `StationsWithClusters` (which read `mapFilter` from context on its own). This PR moves that decision to a single place in `Map.tsx` and passes `showCityBikeStations` / `showSharedCarStations` down as props, with shared car stations turned off while a city bike trip is active.

City bike stations, scooters and everything else are unaffected, and nothing changes when there is no active bike trip.

### DIFF
--- a/src/modules/map/Map.tsx
+++ b/src/modules/map/Map.tsx
@@ -134,13 +134,7 @@ export const Map = (props: MapProps) => {
   const startingCoordinates = getCurrentCoordinatesGlobal() || FOCUS_ORIGIN;
 
   const showVehicles = mapFilter?.mobility.SCOOTER?.showAll ?? false;
-  const showStations =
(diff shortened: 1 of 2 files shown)


reidzeibel (approved):
Looks ok!

PR #6283, opened 2026-08-17
fix: place Mapbox information button next to the logo

adriansberg:
## Issue Reference

Closes https://github.com/AtB-AS/kundevendt/issues/24327

## Screenshots 

<details><summary>↓</summary>
<p>
iOS:

image

Android:

image

</p>
</details> 

## Description

The Mapbox attribution ("i") button was positioned too far right of the logo on both platforms, and on iOS it also sat above the logo instead of beside it.

**iOS.** `attributionPosition: {bottom, left}` is mapped to the `bottomLeading` ornament position by rnmapbox, which is where MapboxMaps places the Mapbox logo by default. MapboxMaps stacks ornaments that share a position vertically and only honours the margins of the first one, so the button was pinned above the logo and its `bottom` value was discarded. This PR patches rnmapbox to map `{bottom, left}` to `bottomLeft` instead, so the button gets its own position and keeps its margins, and adjusts the margins to line it up with the logo. Note that the patched position does not flip in RTL layouts, unlike `bottomLeading`.

**Android.** The logo and the attribution button are separate plugins with independent margins, so it was only a matter of replacing the hardcoded margins with values matching the logo (which happen to be the Mapbox defaults).

Verified in the iOS and Android simulators.


reidzeibel (approved):
Looks ok!

PR #6285, opened 2026-08-18
feat: add transfer risk messages to trip detail screen

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/24007

## Description

Adds two new states to the trip details screen for transfers between vehicle A and vehicle B, based on the gap between A's **expected arrival** and B's **expected departure**. Both are real-time values, so the state changes during an ongoing trip as delay data arrives (the screen already refreshes every 20 s).

| Gap | State | |
|---|---|---|
| ≥ +2:00 | **Vent i x minutter** | existing |
| +0:01 … +1:59 | **Kort byttetid** / Under x minutt(er) ventetid | updated |
| 0:00 … −2:00 | **Usikker overgang** | 🆕 |
| ≤ −2:01 | **Overgang ikke lenger mulig** | 🆕 |

A negative gap means B is expected to leave before A arrives. Note the boundaries are mirror images: 2 minutes late is still *uncertain*, and 2 minutes early is already *Vent i*.

The two new messages are deliberately **not** shown when:

- the interchange is **guaranteed** — B waits for A, so there is no risk
- the next leg is a **foot leg** — the message attaches to the last leg before a vehicle, so a transfer with a walk in the middle reports once rather than twice, measured on the time left *after* the walk

Transfers involving walking are otherwise included, not just same-stop transfers. Applies to all modes, not only bus.

## Screenshots
|||
|---|---|
| **Kort byttetid**<br>kort byttetid | **Vent i x minutter**<br>vent i x minutter |
| **Usikker overgang**<br>usikker overgang | **Overgang ikke lenger mulig**<br>overgang ikke lenger mulig |

## Also in this PR

**Short-transfer range changed from 31–180 s to 1–119 s.** `isShortWaitTime` is the single definition of a short transfer and is shared, so this changes four surfaces beyond the new messages:

- trip details top banner (`hasShortWaitTime`)
- correspondence banner (`hasShortWaitTimeAndNotGuaranteedCorrespondence`)
- travel card notification icon (`getMsgTypeForTravelCardLeg`)
- legacy trip details wait box

Net effect: **1–30 s gaps now warn** (previously silent), and **2–3 minute transfers no longer warn** (previously did). Consequently the new trip details screen now shows a wait row for gaps of 1–30 s, which previously rendered nothing at all.

Short waits are stated as "Under x minutt(er) ventetid" rounded up to whole minutes, so there will never be a "Under 0 minutter ventetid" showing.

### Why `significantWaitTime` is unchanged

The app now has two wait thresholds that look like they should agree:

- `significantWaitTime` (> 30 s) — decides which legs render at all (`getFilteredLegsByWalkOrWaitTime`)
- `isShortWaitTime` (> 0 s, < 120 s) — decides the wording of the wait row

Lowering `significantWaitTime` to `> 0` was considered, so that nothing falls between the two, but it is blocked: three surfaces pass the raw seconds straight to `secondsToDuration`, which renders anything under 30 s as **"0 minutter"**.

- `legacy/TripSection.tsx:404` — a visible row reading "Vent i 0 minutter"
- `TravelCardLegs.tsx:168` — screen reader: "vent 0 minutter"
- `ResultItem.tsx:445` — screen reader: "gå 1 min. vent 0 min"

The new trip details screen is immune because short waits are rounded up and phrased as "Under x minutt(er)" rather than stated exactly. Bringing the other three in line is a prerequisite for changing the threshold, and is left out of this PR.

### DIFF
--- a/src/modules/trip-patterns/__tests__/interchange-risk.test.ts
+++ b/src/modules/trip-patterns/__tests__/interchange-risk.test.ts
@@ -1,0 +1,19 @@
+import {getInterchangeRisk} from '../utils';
+
+describe('Interchange risk evaluator', () => {
+  it('passes when there is time to spare', () => {
+    expect(getInterchangeRisk(1)).toBeUndefined();
+    expect(getInterchangeRisk(600)).toBeUndefined();
+  });
+  it('catches an uncertain interchange', () => {
+    expect(getInterchangeRisk(0)).toBe('uncertain');
+    expect(getInterchangeRisk(-60)).toBe('uncertain');
+  });
+  it('treats -120 seconds as the last uncertain second', () => {
+    expect(getInterchangeRisk(-120)).toBe('uncertain');
+  });
+  it('catches an impossible interchange from -121 seconds', () => {
+    expect(getInterchangeRisk(-121)).toBe('impossible');
+    expect(getInterchangeRisk(-600)).toBe('impossible');
+  });
+});
--- a/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
+++ b/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
@@ -15,24 +15,28 @@ describe('significantWaitTime', () => {
 });
 
 describe('isShortWaitTime', () => {
-  it('should be false for 0 seconds (not significant)', () => {
+  it('should be false for 0 seconds (an interchange risk, not a short wait)', () => {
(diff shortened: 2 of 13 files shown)


rosvik (approved):
Nice! Some very verbose comments for some of the functions here, is it worth shortening some of them?

reidzeibel:
@rosvik I can do that ✅

PR #6286, opened 2026-08-19
fix: priceInfo interval display

Dahly96:
Fixed bug where interval showing wierd timeranges, like from this images:

image

After update:

image


marius-at-atb (approved):

hanne-at-atb:
Looks good! Should just verify this in production where we have more realistic data.

PR #6287, opened 2026-08-19
fix: fixed issue with filter showing in active trip

Dahly96:
There was possible to toggle the filter in the map when in an active trip if the app was reloaded while in an active trip, the filter then appeared again


gorandalum (approved):

hanne-at-atb:
Looks good ✅

PR #6288, opened 2026-08-19
feat: auto-save trips on ticket purchases from trip search

reidzeibel:
## Issue Reference
https://github.com/AtB-AS/kundevendt/issues/24289

## Description
Adds an autosave trip feature when user purchases a ticket through the trip-search. The feature is gated behind experimental feature

Trip will only saved if:
- User press the "kjøp billett" button on the trip details screen
- User does not edit anything on the purchase overview screen
- User successfully purchases a ticket

If the user purchases ticket through the ticket screen, edits the purchase overview (adding traveler, editing departure time, changing the stop place), or does not purchase the ticket, the trip will not be saved.

The trip pattern shown in the trip details is threaded along to the purchase confirmation screen where the actual save happens.

### DIFF
--- a/src/modules/remote-config/remote-config.ts
+++ b/src/modules/remote-config/remote-config.ts
@@ -20,6 +20,11 @@ export type RemoteConfig = {
    * transaction.
    */
   enable_auto_sale: boolean;
+  /**
+   * Configuration whether a trip should be saved automatically when a ticket
+   * for it is purchased from travel search. Requires `enable_save_trips`.
+   */
+  enable_auto_save_trips: boolean;
   enable_bonus: boolean;
   enable_car_sharing_in_map: boolean;
   enable_city_bikes_in_map: boolean;
@@ -105,6 +110,7 @@ export const defaultRemoteConfig: RemoteConfig = {
   enable_activate_ticket_now: false,
   enable_apple_pay: false,
   enable_auto_sale: false,
+  enable_auto_save_trips: false,
   enable_bonus: false,
   enable_car_sharing_in_map: false,
   enable_city_bikes_in_map: false,
@@ -201,6 +207,9 @@ export function getConfig(): RemoteConfig {
   const enable_auto_sale =
     values['enable_auto_sale']?.asBoolean() ??
     defaultRemoteConfig.enable_auto_sale;
+  const enable_auto_save_trips =
+    values['enable_auto_save_trips']?.asBoolean() ??
+    defaultRemoteConfig.enable_auto_save_trips;
   const enable_bonus =
     values['enable_bonus']?.asBoolean() ?? defaultRemoteConfig.enable_bonus;
   const enable_car_sharing_in_map =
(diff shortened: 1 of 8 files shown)


src/modules/remote-config/remote-config.ts
```diff
@@ -20,6 +20,11 @@ export type RemoteConfig = {
    * transaction.
    */
   enable_auto_sale: boolean;
+  /**
+   * Configuration whether a trip should be saved automatically when a ticket
+   * for it is purchased from travel search. Requires `enable_save_trips`.
+   */
+  enable_auto_save_trips: boolean;
```
rosvik: Do we actually need a feature toggle for this? Fine if someone has asked for it, but I think it would be nice to avoid if we can 🤞

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
```diff
@@ -56,6 +57,9 @@ export const Root_PurchaseOverviewScreen: React.FC<Props> = ({
   const [selection, setSelection] = useParamAsState(params.selection);
   const {snackbarProps, showSnackbar, hideSnackbar} = useSnackbar();
 
+  const initialSelectionRef = useRef(params.selection);
+  const isSelectionEdited = !isEqual(selection, initialSelectionRef.current);
```
rosvik: Does this mean that we won't save the trip if the user e.g. adds a traveler? Is that the behavior we want?

src/stacks-hierarchy/Root_PurchaseConfirmationScreen/Root_PurchaseConfirmationScreen.tsx
```diff
@@ -197,6 +209,23 @@ export const Root_PurchaseConfirmationScreen: React.FC<Props> = ({
...
+    // Save the trip the purchase originated from, so it shows up on the
+    // dashboard. `params.tripPattern` is only set when coming from travel
+    // search with an unedited selection.
+    if (
+      !hasSavedTripRef.current &&
+      isSaveTripsEnabled &&
+      isAutoSaveTripsEnabled &&
+      !recipient &&
+      params.tripPattern &&
+      canAddTripPattern(params.tripPattern)
```
rosvik: I think it would be cleaner if this was computed outside the `useCallback`, so that we add less variables to the dependency array.

src/stacks-hierarchy/Root_PurchaseOverviewScreen/Root_PurchaseOverviewScreen.tsx
reidzeibel: This one is debatable, not sure how many edits we allow, but I believe time and route should not be editable (they should be set during the trip search)

src/modules/remote-config/remote-config.ts
reidzeibel: Can ask if this can be removed

rosvik (approved):
Nice!

PR #6290, opened 2026-08-20
feat: added numDocksAvailable

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/24361

Added numDocksAvailable in map-service to get a more accurate number when showing parking spaces in active trip with citybike. Also added name to get rid of the extra servercall to fetch name in the app.

This in Map need to be merged first: https://github.com/AtB-AS/map/pull/26


adriansberg (approved):
This also lgtm. Map PR is also approved.

marius-at-atb (approved):

PR #6294, opened 2026-08-24
feat: renamed texts to be more generic for all vehicles

Dahly96:
Renamed scooter specific texts in the app to be more generic since its reused for both scooters, citybikes and cars


marius-at-atb (approved):
🛵

hanne-at-atb:
I see it's been updated for age verification screen now. Can you point me in the direction of other screens that was also updated @Dahly96 ?

hanne-at-atb:
Tested quite a few of the textual changes and it loos good! ✅

PR #6297, opened 2026-08-25
fix: bonus product operator list include no duplicates

jorunnl:
## Description
Bonus product operator list can no longer include duplicates

Before:
image

After:


adriansberg (approved):
LGTM

hanne-at-atb:
LGTM too ✅

PR #6298, opened 2026-08-25
fix: update age verification text to include city bikes

adriansberg:
Updates the age verification screen description to mention city bikes in addition to e-scooters, since the age limit now also applies to city bikes.


adriansberg:
Solved in https://github.com/AtB-AS/mittatb-app/pull/6294

PR #6305, opened 2026-08-31
feat: added shared cars to the mapFilter

Dahly96:
Closes https://github.com/AtB-AS/team-platform/issues/1124

Added shared cars (free floating cars) in the same filter as carStations

image

image

image


marius-at-atb (approved):
Lgtm. Wouldn't mind some before and after screenshots in the description ;)

PR #6307, opened 2026-09-04
fix: handle time interval in shared mobility pricing

jorunnl:
closes https://github.com/AtB-AS/kundevendt/issues/24365

## Summary

image
image
image
image
image

- Price segments now account for `interval` when rendering `perMinPricing`/`perKmPricing` rates, so a segment with `interval: 30, rate: 5` shows as "5 kr per 30 min" instead of "5 kr/min"
- Segments with `interval: 0` (charged once, not per unit) now render as "(engangsbeløp)"/"(one-time fee)" instead of being mislabeled as per-minute/per-km
- Minute ranges/boundaries ≥ 60 min are now formatted as hours + minutes (e.g. "0-1 time 30 min"), capped at hours
- `computeFreeMinuteCount` now accounts for `interval` when converting a free-minutes campaign budget into minutes, instead of treating `rate` as a flat per-minute cost
- The campaign free-minutes row on the pricing details screen now goes through the same interval-aware label formatting as regular segments

### DIFF
--- a/src/modules/mobility/components/PriceDetailsCard.tsx
+++ b/src/modules/mobility/components/PriceDetailsCard.tsx
@@ -17,6 +17,7 @@ import {
 } from '@atb/translations/screens/subscreens/MobilityTexts';
 import {
   computeFreeMinuteCount,
+  formatMinuteBoundary,
   formatRatePerUnit,
   getFreeMinutes,
   getFreeUnlock,
@@ -56,6 +57,10 @@ export const PriceDetailsCard = ({
 
   const hasFreeMinutes = !!(freeMinutes && pricingPlan.perMinPricing?.length);
 
+  const freeMinuteCount = hasFreeMinutes
+    ? computeFreeMinuteCount(freeMinutes!, pricingPlan.perMinPricing!)
+    : 0;
+
   const minutePriceStat = hasFreeMinutes
     ? zeroAmount
     : ratePrUnit?.formattedRate;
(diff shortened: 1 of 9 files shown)


marius-at-atb:
At this point – let's add a test for each case specified in Figma.

src/modules/mobility/components/PriceDetailsCard.tsx
```diff
@@ -56,17 +54,27 @@ export const PriceDetailsCard = ({
...
   const minutePriceStat = hasFreeMinutes
     ? zeroAmount
     : ratePrUnit?.formattedRate;
 
   const minutePriceDescription = hasFreeMinutes
     ? t(
-        ScooterTexts.freeMinutesDescription(
-          computeFreeMinuteCount(freeMinutes!, pricingPlan.perMinPricing!),
+        MobilityTexts.pricingDetails.freeMinutesDescription(
+          `${formatMinuteBoundary(freeMinuteCount, t)}${freeMinuteCount < 60 ? ' min' : ''}`,
```
Dahly96: Should this 'min' be added to the function itself, or extracted somehow? It feels like the util is partially doing the job of returning hour and minutes, but you need to add min manually in the cases where its just minutes

src/modules/mobility/utils.ts
```diff
@@ -214,6 +206,19 @@ export const hasMultiplePricingPlans = (plan: ShmoPricingPlan) =>
   (plan.perKmPricing && plan.perKmPricing.length > 1) ||
   (plan.perMinPricing && plan.perMinPricing.length > 1);
 
+export const formatMinuteBoundary = (
```
Dahly96: Follow up on [this](https://github.com/AtB-AS/mittatb-app/pull/6307/changes#r3955187943) comment. This function return either a number of minutes or a formatted string with the amount of hours and the text hour/hours. Should maybe be the same handling for minutes.

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -83,24 +81,83 @@ export const MobilityTexts = {
...
+      }
+      if (interval % 60 === 0) {
+        const hours = interval / 60;
+        return hours === 1
+          ? _('per time', 'per hour', 'per time')
+          : _(`per ${hours} timer`, `per ${hours} hours`, `per ${hours} timar`);
+      }
+      return _(
+        `per ${interval} minutter`,
+        `per ${interval} min`,
```
Dahly96: Could maybe be the same format as the rest, with minutes? or just 'min' for all 3

marius-at-atb (approved):
Well done!
Easier to follow after the extracted utils.

PR #6308, opened 2026-09-07
fix: follow mode citybike

Dahly96:
Fixed that the citybike now autofollows the user on starting a trip. 
What was wrong: From scooters we went from no booking to IN_USE, making the useffect work for first time startup. But for citybike we go from Preparing -> IN_USE, so it did not register. So we alter to see if it went from any state to IN_USE.

Acceptance criteria:

The map autofollows the user on starting a trip


marius-at-atb (approved):

hanne-at-atb:
Test ok ✅ 
Works for both city bikes and e-scooter!

PR #6310, opened 2026-09-07
refactor: change lib for image compressing

gorandalum:
The react-native-image-resizer lib was not working with newer
react-native versions, and hasn't had any new releases the last 4
years.

Instead using the already present skia lib to compress the images.

This is an image after compression, taken with my iPhone 17, and it ended up at 89kb.
skia_image

### Acceptance criteria

Some specific things because of lib change:
- [x] Taking picture of parked el scooter should work as before on iOS*
- [x] Taking picture of parked el scooter should work as before on Android*
- [ ] Report wrongly parked scooter should work as before (iOS)**
- [ ] Report wrongly parked scooter should work as before (Android)**

\* Confirm with shmo-crowd whether verification of image in third party is necessary
** If the report wrongly parked functionality should still live on in the app


reidzeibel (approved):
🚀

rosvik (approved):
Nice!

Dahly96 (approved):

tormoseng:
✅ 

Wrongly parked escooters is disabled due to a missing agreement with Nivel. Will test if that is up and running again later.

PR #6311, opened 2026-09-07
feat: deeplink map

Dahly96:
Closes https://github.com/AtB-AS/kundevendt/issues/24431

Added deeplink to the map, making it possible to add announcements with button to go to the map for the future


marius-at-atb (approved):

src/stacks-hierarchy/RootStack.tsx
```diff
@@ -293,6 +293,25 @@ export const RootStack = () => {
                   ],
                 } as ResultState;
               }
+              if (path.includes('map')) {
```
jorunnl, line 298: Is it an idea to also handle formfactor queryparam in the url to turn on form factor in map filter on the navigation to the map? 🤔

for example /map/CAR would turn on car stations in the filter. We use this type of functionality on navigation to the map from Bonus products

Dahly96: Hei Jorunn! Jeg støtter dine tanker. Jeg ser på det!

src/modules/bonus/components/BonusProductList.tsx
```diff
@@ -35,24 +35,17 @@ export const BonusProductList = ({
...
-        updatedMobility[ff] = {
-          operators: updatedMobility[ff]?.operators ?? [],
-          showAll: true,
-        };
-      });
-      const updatedFilter = {...mapFilter, mobility: updatedMobility};
-      setMapFilter(updatedFilter);
+      if (!onNavigateToMap) return;
+      const updatedFilter = enableFormFactorsInMapFilter(formFactors);
+      if (!updatedFilter) return;
```
jorunnl: Since enableFormFactorsInMapFilter can (according to your comment) return undefined if the app is cold startet from deep link, I think this is unwanted behavior.
I think the function should navigate to map without updated filter

rosvik (commented):
FYI: I'm working on a refactor of deep link logic in RootStack, but I think it should work nicely with this change 👌

https://github.com/AtB-AS/mittatb-app/compare/master...rosvik/deep-link-refactor

rosvik (approved):

reidzeibel:
@rosvik maybe we could just merge this so you can merge your PR?

PR #6312, opened 2026-09-07
feat: handle nullable benefit title and description

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/24392

## Description
App can now handle nullable title and description from benefit in the APIs from benefit and mobility.
These values were set to nullable in the db here: https://github.com/AtB-AS/benefit/pull/20
To actually return null values from backend will be done in the future, in this issue: https://github.com/AtB-AS/team-platform/issues/1154


Dahly96 (approved):
LGTM!

PR #6316, opened 2026-09-08
refactor: add deep-links module

rosvik:
## Issue Reference

Initial part of https://github.com/AtB-AS/kundevendt/issues/24379

## Description

- Adds a `src/deep-links/` module that handles deep link logic that previously cluttered RootStack quite a bit.
- Fixes a bug that caused routing of deep links to fail when the app was launched from a cold start (ea6a39ced4b50afcc2d8cd5077d6cb037421b643)
- Makes deep link routing stricter. Previously, we checked for the url to contain "privacy" to route to the privacy screen. Now we require the path to be strictly equal to "privacy", but allow trailing slashes, query params and so on. (See `parseDeepLink()`.
- Adds test for all known deep links, to avoid breaking existing usage. (use-deep-link-options.test.ts)

Expecting to fix some conflicts with https://github.com/AtB-AS/mittatb-app/pull/6311 before I merge this 😬

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,8 +14,8 @@ module.exports = {
       statements: 100,
     },
   },
-  // Added polylabel & tinyqueue to RN allowlist so Jest can transpile ESM packages.
+  // Added polylabel, tinyqueue & @react-navigation to RN allowlist so Jest can transpile ESM packages.
   transformIgnorePatterns: [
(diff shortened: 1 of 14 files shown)


reidzeibel (approved):
Tested in my Android device and doesn't seem like anything is breaking. 🚀

gorandalum (approved):
Nice!

PR #6317, opened 2026-09-08
fix: maestro token inspection safeguard

tormoseng:
Fix for https://github.com/AtB-AS/kundevendt/issues/24341#issuecomment-5580667381

If, for some reason, the inspection is on another device, the tests will now remove remote tokens and reload so that inspection will be on the device under test. Somewhat messy handling of logic in Maestro, but it is what it is.

✅  All tests OK: https://github.com/AtB-AS/mittatb-app/actions/runs/34237887721

However, there were some unexpected closure of the emulator, etc. So to fasten the tests, I added a simplified `smoke.yaml` that cover
- log in
- find departures
- travel search
- purchase a ticket
- log out

✅  Smoke tests OK: https://github.com/AtB-AS/mittatb-app/actions/runs/34328623216

Updated the Github action yaml to run the `smoke.yaml` nightly, and a possibility to run all tests ad-hoc. Also updated the versions of the action usages within the yaml.


reidzeibel (approved):
🚀

PR #6321, opened 2026-09-09
feat: add departures deep links

rosvik:
## Issue Reference

closes https://github.com/AtB-AS/kundevendt/issues/2936

part of https://github.com/AtB-AS/kundevendt/issues/24379

## Description

Adds the `atb://departures` deep link, that works similarly to the widget routes, but without `showOnlyFavoritesByDefault` enabled.

Also removes the `stopName` query param from the widget, since this is fetched by the app anyway, and we don't want to rely on the url to have the correct name.

### DIFF
--- a/ios/departureWidget/WidgetViewModel.swift
+++ b/ios/departureWidget/WidgetViewModel.swift
@@ -58,7 +58,6 @@ struct WidgetViewModel {
 
         var queryItems = [
             URLQueryItem(name: "stopId", value: stopPlace.id),
-            URLQueryItem(name: "stopName", value: stopPlace.name),
             URLQueryItem(name: "quayId", value: quay.id),
             URLQueryItem(name: "latitude", value: String(stopPlace.latitude)),
             URLQueryItem(name: "longitude", value: String(stopPlace.longitude)),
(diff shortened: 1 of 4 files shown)


reidzeibel (approved):
This is very nice! if we can put this deeplink somewhere in the planner-web, then we can see the departures from planner-web inside our app, provided the planner-web is opened from a mobile device. 

Edit: Also if we have stop place name taken from the deeplink, we can actually pass a random name on that stop place name 🤷🏼  I did try with the deeplink module test, I can put a stop place with name "100%Valid" on the deeplink and it still loads.

rosvik:
> if we can put this deeplink somewhere in the planner-web, then we can see the departures from planner-web inside our app, provided the planner-web is opened from a mobile device.

I like that idea!

rosvik:
> Edit: Also if we have stop place name taken from the deeplink, we can actually pass a random name on that stop place name 🤷🏼 I did try with the deeplink module test, I can put a stop place with name "100%Valid" on the deeplink and it still loads.

Yup, I realized the same thing 😅 Removed the option now, since it didn't really add anything even when it was correct.

PR #6324, opened 2026-09-11
feat: add paused state with resume button + test pause button

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/24311

## Description

Adds handling of paused state in active booking bottom sheet. The bottom sheet now shows an error message (because the state should currently be unreachable) and a resume button.
image image

Also adds a pause button that is behind a feature toggle
image

### DIFF
--- a/src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
+++ b/src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
@@ -1,7 +1,10 @@
 import React, {RefObject, useCallback, useEffect} from 'react';
 import {useTranslation} from '@atb/translations';
 import {StyleSheet, useThemeContext} from '@atb/theme';
-import {MobilityTexts} from '@atb/translations/screens/subscreens/MobilityTexts';
+import {
+  MobilityTexts,
+  ShmoWarnings,
+} from '@atb/translations/screens/subscreens/MobilityTexts';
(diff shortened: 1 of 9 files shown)


Dahly96 (approved):
lgtm

src/modules/mobility/components/sheets/ActiveShmoSheet.tsx
```diff
@@ -225,6 +293,16 @@ export const ActiveShmoSheet = ({
                       </View>
                     </View>
                   )}
+                  {isPaused && (
+                    <MessageInfoBox
+                      type="error"
```
marius-at-atb, line 298: Note: in the future we can't assume it's an error.

src/translations/screens/subscreens/MobilityTexts.ts
```diff
@@ -691,6 +691,29 @@ export const GeofencingZoneExtraExplanations = {
...
+          'Sykkelen har låst seg',
+          'The bike has locked itself',
+          'Sykkelen har låst seg',
+        );
+      default:
+        return _(
+          'Kjøretøyet har låst seg',
+          'The vehicle has locked itself',
+          'Køyretøyet har låst seg',
+        );
```
marius-at-atb, line 714: We are not sure what's happened, so we should not be specific about the vehicle is locked.

PR #6326, opened 2026-09-11
feat: adjust the short wait time from 120s to 180s

reidzeibel:
## Issue Reference
Part of https://github.com/AtB-AS/kundevendt/issues/24416

## Description
Adjusts the short wait time threshold from 0-120 to 0-180, as discussed with @Sebstorvik and @helenekulsetas

### DIFF
--- a/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
+++ b/src/modules/trip-patterns/__tests__/is-short-wait-time.test.ts
@@ -32,12 +32,12 @@ describe('isShortWaitTime', () => {
     expect(isShortWaitTime(31)).toBe(true);
   });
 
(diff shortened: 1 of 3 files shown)


Sebstorvik:
Sorry, me and @helenekulsetas forgot to mention, but some of the OMS also want to try this change @reidzeibel 😄

Can you also change so that FRAM and NFK as well use 180 seconds? Don't change anything in Svipper

reidzeibel:
Right now the change is hardcoded in the app @Sebstorvik , so everyone will have 180s, I was planning to add the env adjustments after the core tasks for this alignment is done.

jorunnl (approved):
LGTM 💯

PR #6328, opened 2026-09-11
fix: show trip receipt when a booking ends while the app is backgrounded

Dahly96:
## Issue Reference

Closes https://github.com/AtB-AS/kundevendt/issues/24493

## Description

When a trip ended while the phone was locked in the user's pocket, the receipt sheet never appeared.

The receipt was triggered by a `SHMO_BOOKING_UPDATED` event from the event stream. That stream is a WebSocket with no replay — when the OS suspends the app the socket dies, and on reconnect the server only sends events from that point on. Any booking finished while backgrounded produced an event that was simply never delivered, so the listener never ran.

This replaces the event trigger with a reconciliation against server state. The active booking only exists while a booking is in a non-terminal state, so its disappearance *is* the signal that the trip ended. The id of the booking we've seen as active is kept in AsyncStorage, which lets us recognise it after the fact — including across an app restart.

On unlock, `refetchOnWindowFocus` refetches the active booking, it comes back `null`, and the stored id tells us which receipt to open.

### Notable details

- **Only a successful fetch counts.** A failed refetch also leaves us without booking data, and must not be mistaken for the booking having ended — otherwise going offline mid-trip would pop a receipt for a trip still running.
- **The stored id carries a `userId`**, so a booking is never shown to another user after a logout or account switch. Comparing on read avoids clearing the id on every cold start, which watching `userId` for changes would have done.
- **The listener in `MapContext` is removed** rather than kept as a fast path. `useGlobalEventStreamListeners` already invalidates the active booking query on the same event, so the foreground case now flows through the same single path at the same latency.
- **Cancelled bookings show the receipt too**, so it's clear no money was charged. The sheet already degrades to `00:00` / `0 kr`; only the heading needed a new string.
- `Root_ParkingPhotoScreen` shows the receipt immediately after the photo upload, so it clears the stored id to avoid the sheet reopening once the query catches up.

### Testing

7 unit tests for the detection hook. Manually verified on device.

Note: `tsc` and 4 test suites fail on `master` already (`@react-native-firebase` typings, missing `@atb-as/utils` exports). Unrelated to this change and unaffected by it.

### DIFF
--- a/src/modules/map/hooks/use-finished-booking-detection.ts
+++ b/src/modules/map/hooks/use-finished-booking-detection.ts
@@ -1,0 +1,56 @@
+import React, {useEffect} from 'react';
+import {useAuthContext} from '@atb/modules/auth';
+import {useActiveShmoBookingQuery} from '@atb/modules/mobility';
+import {MapStateActionType, ReducerMapStateAction} from '../mapStateReducer';
+import {
+  clearPendingFinishedBooking,
+  getPendingFinishedBooking,
+  setPendingFinishedBooking,
+} from '../pending-finished-booking';
+
+/**
+ * Opens the receipt for a booking which has ended while we weren't looking.
+ *
+ * The active booking only exists while the booking is in a non-terminal state,
+ * so when it disappears the booking has been finished or cancelled. We can't
+ * rely on the booking event for this, as the event stream is only connected
+ * while the app is in the foreground, and delivers no events for the time it
+ * was disconnected.
+ */
+export const useFinishedBookingDetection = (
+  isFocusedAndActive: boolean,
+  dispatchMapState: React.Dispatch<ReducerMapStateAction>,
+) => {
+  const {userId} = useAuthContext();
+  const {data: activeBooking, isSuccess} =
+    useActiveShmoBookingQuery(isFocusedAndActive);
+  const activeBookingId = activeBooking?.bookingId;
+
+  useEffect(() => {
+    // Only a successful fetch tells us whether there is an active booking. A
+    // failed one also leaves us without booking data, and must not be mistaken
+    // for the booking having ended.
+    if (!isSuccess || !userId) return;
+
+    if (activeBookingId) {
+      setPendingFinishedBooking({bookingId: activeBookingId, userId});
+      return;
+    }
+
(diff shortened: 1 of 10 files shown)


src/modules/map/hooks/use-finished-booking-detection.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    let isCancelled = false;
+    getPendingFinishedBooking().then((pendingFinishedBooking) => {
+      if (isCancelled || !pendingFinishedBooking) return;
+      clearPendingFinishedBooking();
+      // Belongs to a previous user, after a logout or an account switch.
+      if (pendingFinishedBooking.userId !== userId) return;
+      console.log('dispatching end booking');
```
jorunnl: Remoooooooove this line ⚔️

src/modules/map/pending-finished-booking.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * The booking we have seen as active, but not yet shown a receipt for.
+ *
+ * The event stream only delivers events while the app is in the foreground, so
+ * a booking which is finished or cancelled while the phone is locked never
+ * reaches us. Keeping the id in storage lets us detect it afterwards, by seeing
+ * that the active booking has disappeared. The user id is stored alongside it
+ * so a booking is never shown to another user after a logout or account switch.
+ */
+const PendingFinishedBookingSchema = z.object({
```
jorunnl: I think 'pending finished booking' is a strange name. I understand it is a booking that we are waiting to be finished, but it is also just an 'active booking'. Could it be renamed?

src/modules/map/hooks/use-finished-booking-detection.ts
Dahly96: donee

src/modules/map/pending-finished-booking.ts
Dahly96: renamed to LastActiveBooking

jorunnl: nice, I like it better

marius-at-atb (approved):
Lgtm!

PR #6331, opened 2026-09-14
fix: update order and text in ShmoContactSection

jorunnl:
## Description

Updates the order of section items in ShmoContactSection. Now 'Chat' is above 'Kontaktskjem' -- since Chat will likely give faster answers. Also updates Chat text to give user better understanding of what pressing it leads to.

image


Dahly96 (approved):

hanne-at-atb:
FYI: we did these updates @lineamb - consider updating the sketches and look into further changes if needed in next release

hanne-at-atb:
Tested and looks and behaves as expected ✅

PR #6332, opened 2026-09-14
fix: update paused state active shmo trip

jorunnl:
## Issue Reference
closes https://github.com/AtB-AS/kundevendt/issues/24311

## Description

Updates texts on error message and resume button on paused active shmo bottom sheet
image


marius-at-atb (approved):
Looks good. Maybe write a comment in https://github.com/AtB-AS/kundevendt/issues/24312 about not treating it as an error anymore after that.

hanne-at-atb:
Tested and text has been updated ✅

PR #6333, opened 2026-09-15
fix: show shared cars in map when navigating from ticket benefit tile

Dahly96:
Closes AtB-AS/kundevendt#24476 — Hyre filtered out in map on redirect from ticket details

Tapping "Vis i kart" on the Hyre benefit tile in ticket details did not just fail to show shared cars — it actively turned the car filter **off**.

### Fix

Use `useEnableFormFactorsInMapFilter`, the hook the bonus screen and the map deeplinks already use. It sets `showAll: true` per form factor, persists the filter, and returns it.

The tile now turns on "shared cars" rather than "Hyre specifically" — which matches what the filter sheet already does, since operator-level filtering was removed with the map v2 filter.

### Testing

1. Map tab → filter → turn the car toggle **on**, confirm cars are visible
2. Ticketing → valid periodic ticket → ticket details
3. Tap "Vis i kart" on the Hyre tile
4. Cars should be visible, and the car toggle should still be **on**

Before this change, step 4 showed no cars and the toggle had flipped off. Repeat with the toggle starting **off** — it should now flip on. Same for the bike tile.


src/modules/mobility/components/BenefitTiles.tsx
```diff
@@ -37,30 +31,16 @@ export const BenefitTile = ({
...
-      };
-    });
-
-    // Update stored filters (for persistence, and the filters bottom sheet)
-    setMapFilter({...mapFilter, mobility: mobilityFilters});
-    // Provide the same filters as intital filter state for the map screen
-    onNavigateToMap({...mapFilter, mobility: mobilityFilters});
+  const onPress = () => {
+    // Turn on the filter for the benefit's form factors, so the vehicles are
+    // visible in the map even if the user has previously turned the filter off.
```
jorunnl:
```suggestion

```

jorunnl (changes requested):
I think the comment is unnecessary, the code is understandable by it self

Otherwise looks good 🙆‍♀️

jorunnl (approved):
whooooops 💯

